PR #3, opened 2020-12-31
Created a bit nicer playground

KATT:
Just cloned the project to get a feel of the DX

Here's a PR with a bit nicer playground that can be started with `yarn playground`.

Will loop back later on with ideas on how to improve the lib.

But curious - what were your intentions with this lib? What are you going to use it for yourself?


examples/playground/client.ts
```diff
@@ -0,0 +1,20 @@
+import { trpc } from '../../src';
+
+import type { RootRouter } from './server';
```
KATT, line 3: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export
> import type only imports declarations to be used for type annotations and declarations. It always gets fully erased, so there’s no remnant of it at runtime. Similarly, export type only provides an export that can be used for type contexts, and is also erased from TypeScript’s output.

KATT: Should be safe for clients

PR #58, opened 2021-02-03
normi proof-of-concept

KATT:
- uses [normi](https://github.com/colinhacks/normi) for a centralized global cache 
- easy a/f to update local state


examples/next-ssg-chat/pages/index.tsx
```diff
@@ -23,6 +24,22 @@ const getLatestTimestamp = (m: Message[]) => {
...
 
+const MessageListItem = observer(({ m }: { m: Message }) => {
+  return (
+    <li
+      onClick={async () => {
+        const text = prompt('text');
+        await trpc.client.mutation('messages.edit', {
+          id: m.id,
+          text,
+        });
```
KATT, line 35: ☝️ 
All you need to do to update the state of MessageListItem`

cyrus-za:
I'd love to start using normi with trpc without needing to manually write any normi code. trpc should just do it based off looking for the id. Kind of like apollo client does with normalization. The developer doesnt need to know how apollo stores it, just that it is optimally normalized and that it requires a typename and id returned from the server. 

Or in trpc case only a id, because the type is explicitly specified by the router with zod schema etc.

KATT:
Hey @cyrus-za - You don't have to write any "normi code", apart for the code to set it up, or what are you referring to? 


I like the trpc+normi combo as well but I don't want to ship it as a core part because it has some quirks to it that people might not be prepared on.

Some gotchas

- `id` has to be globally unique
- _any_ property that belongs to an ID is tied to that without any params
  -  i.e. if you want to paginate a related object it shouldn't be part of any parent with an `id`

PR #74, opened 2021-02-09
client subscription support

KATT:
- client subscription support
- fix backpressure logic
- fix `example:playground`


packages/server/src/http.ts
```diff
@@ -256,7 +256,7 @@ export async function requestHandler<
             cleanup();
             resolve(buffer);
           };
-          if (requestTimeLeft >= backpressureMs) {
+          if (requestTimeLeft <= backpressureMs) {
```
KATT, line 259: this is why you test things

PR #87, opened 2021-02-18
Add middleware support

KATT:
- allows you to create a middleware-esque thing on a router that cascades down on all routes added after
- I don't love the name `preHook` - would love some input.
- (started on a proper "middleware"-thing in #88, but it's a bit complicated as it might require async iterables, so might do that later or never)
- see the tests for usage, here's an example of an admin "sub-router"   https://github.com/trpc/trpc/blob/5f96526432bd8d5454e145e62e3df9eba3f9785c/packages/server/test/preHook.test.ts#L67-L84

### DIFF
--- a/packages/server/src/router.ts
+++ b/packages/server/src/router.ts
@@ -283,34 +283,42 @@ export class Router<
       throw new Error(`Duplicate endpoint(s): ${duplicates.join(', ')}`);
     }
 
-    const addPreHooks = <TRoutes extends RouteRecord>(routes: TRoutes) => {
-      const newRoutes = {} as TRoutes;
-      for (const key in routes) {
-        const route = routes[key];
-        newRoutes[key] = {
-          ...route,
-          _preHooks: [...(route._preHooks ?? []), ...this._def.preHooks],
-        };
-      }
-      return newRoutes;
-    };
     return new Router<TContext, any, any, any>({
       queries: {
         ...this._def.queries,
-        ...addPreHooks(Router.prefixRoutes(router._def.queries, prefix)),
+        ...this.inhertPreHooks(
+          Router.prefixRoutes(router._def.queries, prefix),
+        ),
       },
       mutations: {
         ...this._def.mutations,
-        ...addPreHooks(Router.prefixRoutes(router._def.mutations, prefix)),
+        ...this.inhertPreHooks(
+          Router.prefixRoutes(router._def.mutations, prefix),
+        ),
       },
       subscriptions: {
         ...this._def.subscriptions,
-        ...addPreHooks(Router.prefixRoutes(router._def.subscriptions, prefix)),
+        ...this.inhertPreHooks(
+          Router.prefixRoutes(router._def.subscriptions, prefix),
+        ),
       },
       preHooks: this._def.preHooks,
     });
   }
 
+  private inhertPreHooks<TRoutes extends RouteRecord<TCtx>, TCtx>(
+    routes: TRoutes,
+  ): TRoutes {
+    const newRoutes = {} as TRoutes;
+    for (const key in routes) {
+      const route = routes[key];
+      newRoutes[key] = {
+        ...route,
+        _preHooks: [...(route._preHooks ?? []), ...this._def.preHooks],
+      };
+    }
+    return newRoutes;
+  }
   private static getInput<TRoute extends Route<any, any, any>>(
     route: TRoute,
     rawInput: unknown,
@@ -363,7 +371,6 @@ export class Router<
     return !!this._def[what][path];
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   /**
    * Function to be called before any route is invoked
    * Can be async or not async
--- a/packages/server/src/router.ts
+++ b/packages/server/src/router.ts
@@ -36,7 +36,6 @@ export type RouteWithInput<
 > = {
   input: RouteInputParser<TInput>;
   resolve: RouteResolver<TContext, TInput, TOutput>;
-  _preHooks?: PreHookFunction<TContext>[];
 };
 
 export type RouteWithoutInput<
@@ -47,11 +46,12 @@ export type RouteWithoutInput<
   input?: undefined | null;
   resolve: RouteResolver<TContext, TInput, TOutput>;
 };
-export type Route<TContext = unknown, TInput = unknown, TOutput = unknown> =
+export type Route<TContext = unknown, TInput = unknown, TOutput = unknown> = (
   | RouteWithInput<TContext, TInput, TOutput>
-  | (RouteWithoutInput<TContext, TInput, TOutput> & {
-      _preHooks?: PreHookFunction<TContext>[];
-    });
+  | RouteWithoutInput<TContext, TInput, TOutput>
+) & {
+  _preHooks?: PreHookFunction<TContext>[];
+};
 
 export type RouteRecord<
   TContext = unknown,
--- a/packages/server/src/router.ts
+++ b/packages/server/src/router.ts
@@ -23,11 +23,15 @@ export type RouteInputParser<TInput = unknown> =
   | RouteInputParserYupEsque<TInput>
   | RouteInputParserCustomValidatorEsque<TInput>;
 
+type RouteResolverOpts<TContext = unknown, TInput = unknown> = {
+  ctx: TContext;
+  input: TInput;
+};
 export type RouteResolver<
   TContext = unknown,
   TInput = unknown,
   TOutput = unknown
-> = (opts: { ctx: TContext; input: TInput }) => Promise<TOutput> | TOutput;
+> = (opts: RouteResolverOpts<TContext, TInput>) => Promise<TOutput> | TOutput;
 
 export type RouteWithInput<
   TContext = unknown,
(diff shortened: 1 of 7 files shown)


packages/server/src/router.ts
```diff
@@ -326,16 +362,27 @@ export class Router<
...
     return route.resolve({ ctx, input });
   }
 
   public has(what: 'subscriptions' | 'mutations' | 'queries', path: string) {
     return !!this._def[what][path];
   }
+
+  /**
+   * Function to be called before any route is invoked
+   * Can be async or not async
```
cyrus-za, line 380:
```suggestion
   * Can be async or sync
```

packages/server/test/preHook.test.ts
```diff
@@ -0,0 +1,167 @@
...
+  type Context = {
+    user?: {
+      id: number;
+      name: string;
+      isAdmin: boolean;
+    };
+  };
+  const resolverMock = jest.fn();
+
+  const headers: Record<string, string | undefined> = {};
```
cyrus-za, line 57: doesnt @types/node come with some Headers typing you can import?

KATT: i don't care enough about test code to care about changing this :)

there's some reason why i don't use that elsewhere, but i don't remember why

packages/server/test/preHook.test.ts
```diff
@@ -0,0 +1,167 @@
...
+        trpc
+          .router()
+          .preHook(preHookInChild)
+          .query('name', {
+            resolve() {
+              return 'Parent';
+            },
+          })
+          .merge(
+            'child.',
```
cyrus-za, line 136: I dont get it. Can we merge different routers and specify the same name (`child.`) twice?

cyrus-za: nevermind. I see lower down. so calling it `.child` actually appends to the previous route. That is actually good to know. It means I cannot do 

`mainRouter.merge('products.',productsRouter).merge('users.', usersRouter) as it will mean to query on users I'd need to go `products.users.list` which is fundamentally wrong as users is not a subroute to products, right?

KATT: you don't _have to_ use a prefix but you can if you want

packages/server/test/preHook.test.ts
```diff
@@ -0,0 +1,167 @@
...
+                },
+              }),
+          ),
+      ),
+  );
+
+  expect(await client.query('child.child.name')).toBe('GrandChild');
+  expect(preHookInParent).toHaveBeenCalledTimes(1);
+  expect(preHookInChild).toHaveBeenCalledTimes(1);
+  expect(preHookInGrandChild).toHaveBeenCalledTimes(1);
```
cyrus-za, line 152: nice. so it will call each of those "middleware" hooks when going to a subrouter. That's perfect.

packages/server/test/preHook.test.ts
KATT: these two examples are equivalent:

```ts

  type Context = {
    user?: {
      id: number;
      name: string;
      isAdmin: boolean;
    };
  };
  // **using prefix**
  trpc
    .router()
    .query('foo', {
      resolve() {
        return 'bar';
      },
    })
    .merge(
      'admin.',
      trpc
        .router<Context>()
        .preHook(({ ctx }) => {
          if (!ctx.user?.isAdmin) {
            throw httpError.unauthorized();
          }
        })
        .query('secretPlace', {
          resolve() {
            return 'a key';
          },
        }),
    );

  // **not** using prefix
  trpc
    .router()
    .query('foo', {
      resolve() {
        return 'bar';
      },
    })
    .merge(
      trpc
        .router<Context>()
        .preHook(({ ctx }) => {
          if (!ctx.user?.isAdmin) {
            throw httpError.unauthorized();
          }
        })
        .query('admin.secretPlace', {
          resolve() {
            return 'a key';
          },
        }),
    );
```

packages/server/test/preHook.test.ts
```diff
@@ -0,0 +1,167 @@
...
+        },
+      })
+      .merge(
+        'child.',
+        trpc
+          .router()
+          .preHook(preHookInChild)
+          .query('name', {
+            resolve() {
+              return 'Parent';
```
cyrus-za, line 132:
```suggestion
          .preHook(preHookInParent)
          .query('name', {
            resolve() {
              return 'Parent';
```
Makes the test more readable/digestable lower down if you use the same names. 

PS. same with GrandPa vs Parent above ☝🏻

cyrus-za (approved):
This is exactly what I requested. Thanks so much

packages/server/test/preHook.test.ts
KATT: In my app I've split up my routers in different files

<details><summary>api/trpc/__shared__.tsx</summary>

```ts
import { CreateExpressContextOptions } from '@trpc/server/dist/adapters/express';
import * as trpc from '@trpc/server';
import { inferAsyncReturnType } from '@trpc/server';
import { createUserSdk, decodeAuthToken } from '../__shared__';

export const createContext = async ({ req }: CreateExpressContextOptions) => {
  const auth0 = await decodeAuthToken(req);
  const roles = auth0['https://hasura.io/jwt/claims']['x-hasura-allowed-roles'];
  const sdk = createUserSdk(auth0);

  return {
    sdk,
    user: {
      auth0,
      roles,
    },
  };
};
export type Context = inferAsyncReturnType<typeof createContext>;
export function createRouter() {
  return trpc.router<Context>();
}
```
</details>

<details><summary>api/trpc/[...trpc].ts</summary>

```ts
import * as z from 'zod';
import { adminRouter } from './admin';
import { createRouter, createContext } from './__shared__';
import { createExpressMiddleware } from '@trpc/server/dist/adapters/express';
export type { Context } from './__shared__';

const appRouter = createRouter()
  .query('hello', {
    input: z
      .object({
        text: z.string().optional(),
      })
      .optional(),
    resolve({ input }) {
      return {
        greeting: `hello ${input?.text ?? 'world'}`,
      };
    },
  })
  .merge('admin.', adminRouter);

// Exporting type _type_ AppRouter only exposes types that can be used for inference
// https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

export const trpcMiddleware = createExpressMiddleware({
  router: appRouter,
  createContext,
});

```
</details>


<details><summary>api/trpc/admin.ts</summary>

```ts
import { createRouter } from './__shared__';
import * as z from 'zod';
import { httpError } from '@trpc/server';
import gql from 'graphql-tag';
import { v4 } from 'uuid';
import { UserRole_Enum } from '../__generated__/sdk';
import { env } from '../__shared__';

gql`
  mutation admin__addUser(
    $user: User_insert_input!
    $organization: Organization_insert_input!
    $userId: uuid!
    $organizationId: uuid!
  ) {
    insert_Organization_one(object: $organization) {
      id
    }
    insert_User_one(object: $user) {
      id
    }
    insert_OrganizationUser_one(
      object: { userId: $userId, organizationId: $organizationId }
    ) {
      userId
    }
  }
`;
function randomString(len: number) {
  const charSet =
    'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
  var randomString = '';
  for (let i = 0; i < len; i++) {
    const randomPoz = Math.floor(Math.random() * charSet.length);
    randomString += charSet.substring(randomPoz, randomPoz + 1);
  }
  return randomString;
}

export const adminRouter = createRouter().mutation('createUser', {
  input: z.object({
    user: z.object({
      name: z.string().min(1),
      email: z.string().email(),
    }),
    organization: z.object({
      name: z.string(),
      slug: z.string().regex(/^[a-z0-9-]+$/),
      url: z.string().optional(),
      businessType: z.string(),
      country: z.string(),
    }),
  }),
  async resolve({ ctx, input }) {
    if (!ctx.user.roles.includes('admin')) {
      throw httpError.unauthorized();
    }
    const userId = v4();
    const organizationId = v4();

    await ctx.sdk.admin__addUser({
      userId,
      organizationId,
      organization: {
        id: organizationId,
        ...input.organization,
      },
      user: {
        id: userId,
        ...input.user,
        role: UserRole_Enum.User,
      },
    });
    const headers = await (
      await fetch(`https://${env.AUTH0_MGMT_DOMAIN}/oauth/token`, {
        method: 'POST',
        headers: {
          'content-type': 'application/json',
        },
        body: JSON.stringify({
          client_id: env.AUTH0_MGMT_CLIENT_ID,
          client_secret: env.AUTH0_MGMT_CLIENT_SECRET,
          audience: `https://${env.AUTH0_MGMT_DOMAIN}/api/v2/`,
          grant_type: 'client_credentials',
        }),
      })
    ).json();
    const token = headers.access_token;
    if (typeof token !== 'string') {
      throw new Error('Coudl not get token');
    }
    const createUserResponse = await (
      await fetch(`https://${env.AUTH0_MGMT_DOMAIN}/api/v2/users`, {
        method: 'POST',
        headers: {
          'content-type': 'application/json',
          authorization: `Bearer ${token}`,
        },
        body: JSON.stringify({
          email: input.user.email,
          name: input.user.name,
          connection: 'Username-Password-Authentication',
          password: randomString(32),
        }),
      })
    ).json();

    if (
      !createUserResponse.created_at ||
      createUserResponse.email !== input.user.email
    ) {
      console.error('Auth0 response', createUserResponse);
      throw new Error(
        'User not created - auth0 said something unexpected - does the user already exist maybe?',
      );
    }

    return null;
  },
});

```
</details>

(kinda using hasura as an sdk on the backend which isn't amazing but it gives me typesafety db access and I'm in process of moving to prisma)

packages/server/test/preHook.test.ts
cyrus-za: ye I am the same with tests. Doesnt need to be as readable. I just figured while you writing its easier to import heders than to type the thing manually but its already done, no need to change it now

packages/server/test/preHook.test.ts
KATT: Fixed!

packages/server/test/preHook.test.ts
cyrus-za: yep makes sense. Most of our stuff is on prisma finally. Took months to get the team to adopt it

KATT:
@all-contributors add @cyrus-za for idea

PR #88, opened 2021-02-18
wip - proper middlewares

KATT:


packages/server/src/router.ts
```diff
@@ -370,7 +378,7 @@ export class Router<
     const target = this._def[opts.target];
     const route: Route<TContext> = target[opts.path as any];
     const { ctx } = opts;
-    for (const fn of route._preHooks ?? []) {
+    for (const fn of route._middlewares ?? []) {
```
KATT, line 381: here we'd need to do something like

```tsx
for await (const middleware of route._middlewares) {

}
```

PR #91, opened 2021-02-18
fix peer dep warning

KATT:


KATT:
build probably breaks b/c of https://github.com/babel/babel/issues/12816

PR #96, opened 2021-02-19
flatten router generics

KATT:


KATT:
@colinhacks could you do some magic here? it doesn't let me use the `flatten()`-thing everywhere

PR #107, opened 2021-02-24
svelte/sapper support

KATT:
the thing missing is that I don't know how to do a "isomorphic" fetch request in svelte


benbender:
@KATT I'm thinking about trpc+sveltekit (basically official next.js in svelteland). There is also an included, isomorphic fetch() and builtin data-loading mechanism (@see https://kit.svelte.dev/docs#loading). And, on top, there is https://github.com/SvelteStack/svelte-query (Which seems a bit dead at the moment and should be directly based on react-query/core - but hey. at least there is a reference-implementation...)

KATT:
Hey @benbender! 💯  

I really want SvelteKit-support!! I've had a stab at it a few times, it does require a bit refactoring as SvelteKit's API functions are a quite different from express/next/etc. (In sveltekit you don't deal with `res.send(x)`/`res.setHeaders(...)` return an object like `{ body: 'x', headers: ...}` instead)

There's work in progress on #731 as well.

benbender:
Hey,

I just noticed the PR after I posted this. Sorry for the noise ;)

I'll post some thoughts over at #731.

PR #174, opened 2021-03-15
replace tsdx with preconstruct

KATT:
- replace tsdx with https://preconstruct.tools/ 
- allows multiple entrypoints etc

general cleanup

- move babel, linting, etc to root
- do some lint cleanup

--- 

## Will be a major version bump

e.g. 

```ts
import * as trpcExpress from '@trpc/server/dist/adapters/express';
import * as trpcNext from '@trpc/server/dist/adapters/next';
```

changes to

```ts
import * as trpcExpress from '@trpc/server/adapters/express';
import * as trpcNext from '@trpc/server/adapters/next';
```


examples/playground/src/server.ts
```diff
@@ -3,7 +3,7 @@ import { EventEmitter } from 'events';
 import express from 'express';
 import * as trpc from '@trpc/server';
 import * as z from 'zod';
-import * as trpcExpress from '@trpc/server/dist/adapters/express';
+import * as trpcExpress from '@trpc/server/adapters/express';
```
KATT, line 6: this will be a minor breaking change

babel.config.js
```diff
@@ -0,0 +1,25 @@
...
+  ],
+  plugins: [
+    "babel-plugin-annotate-pure-calls",
+    "babel-plugin-dev-expression",
+    ["@babel/plugin-proposal-class-properties", { loose: true }],
+  ],
+  babelrcRoots: [
+    "packages/*"
+  ]
+};
```
KATT, line 25: This babel conf will be used for client and react - not sure if this is the right appraoch

KATT (commented):
🤷‍♂️  👇

babel.config.js
KATT: cc @sibelius :)

sibelius: I would rather go with multiple babel configs that you could share some common configs

it depends on what you wanna achieve with them

KATT: I have the root one which is for two browser packages and then the exception is node where I have node 12 target.

The question is that I don't know what browser support is kosher to target nowadays

PR #209, opened 2021-03-26
fix:added test directory to tsconfig files

colinhacks:
Fixes several type errors in test files. 

@KATT since you aren't seeing these, you might have somehow overridden the default tsconfig values globally on your computer? 

I also ran `yarn build` and this didn't result in any undesirable artefacts in `dist` (I was worried this change would cause `test` to be compiled in addition to `src` but that's not the case.

### DIFF
--- a/packages/server/tsconfig.json
+++ b/packages/server/tsconfig.json
@@ -1,4 +1,7 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["src"]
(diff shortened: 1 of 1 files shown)


KATT (approved):

KATT:
This was the same for me so I'm unsure what/when I broke it. Thanks!

PR #210, opened 2021-03-26
Added trpc export. Aliased createTRPCClient to createClient.

colinhacks:
`import * as` imports aren't a great developer experience. There's no way to get autocompleting imports that way. This makes a simple change that allows for importing from @trpc/server like this:

```ts
import { trpc } from "@trpc/server";
```

I also aliased `createTRPCClient` to `createClient`. Since the user is importing from `@trpc/client` I think having "TRPC" in the function name is redundant (and doesn't look very good imo).

### DIFF
--- a/packages/client/src/createTRPCClient.ts
+++ b/packages/client/src/createTRPCClient.ts
@@ -317,3 +317,5 @@ export function createTRPCClient<TRouter extends AnyRouter>(
 ) {
   return new TRPCClient<TRouter>(opts);
 }
+
+export { createTRPCClient as createClient };
--- a/packages/server/src/exports.ts
+++ b/packages/server/src/exports.ts
@@ -1,0 +1,6 @@
+export * from './http';
+export * from './router';
+export * from './types';
+export * from './subscription';
+export * from './transformer';
+export * from './adapters/standalone';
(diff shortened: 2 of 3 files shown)


KATT:
It's both things that you have to do exactly once in the application, have a look at https://github.com/trpc/trpc/tree/main/examples/next-prisma-todomvc/pages/api/trpc

In the react frontend I have as a convention to export `trpc` in utils so that default export would collide in your autocomoplete.

KATT (changes requested):
- It's both things that you have to do exactly once in the application, have a look at https://github.com/trpc/trpc/tree/main/examples/next-prisma-todomvc/pages/api/trpc
- In the react frontend I have as a convention to export `trpc` in utils so that default export would collide in your autocomplete, so because of that I don't want to merge it
  - if you have an other idea of a good convention I would be willing to change it
  - the reason I export a `trpc` object on the frontend is really just because react-query occupies the `useQuery`/`useMutation` namespace
- The `exports.ts` is an unnecessary abstraction, better to just leave it in `index` which causes collisions
- I don't mind the `createClient` but I also feel it's unnecessary and I get a bit annoyed when lib authors take generic names like that as it's likely that you have several packages installed that use `createClient`

colinhacks:
This all seems very reasonable 👍

Closing.

PR #220, opened 2021-03-28
Added support for useQuery(path, input, opts) and useQuery(path, opts)

colinhacks:
- Added support for useQuery(path, input, opts)
- Added new tests to react.test.ts to test new syntax

Important note: implementing this overload requires a runtime-only way of distinguishing between the `useQuery(path, input, opts)` usage and `useQuery(path, opts)` usage. It is assumed that `useQuery(path, opts)` usage is intended if:

1) the third argument is undefined
2) the second argument is an object
3) the second argument exclusively contains keys that exist on react-query's `UseQueryOptions`

If by chance the procedures `input` type extends `UseQueryOptions` then this may mistakenly result in the function treating the second argument as `opts` instead of `input`. I think this is a worthwhile tradeoff since it results in a much cleaner API and there's a low probability of this occurrence, since the keys are fairly obscure:

```
enabled
staleTime
refetchInterval
refetchIntervalInBackground
refetchOnWindowFocus
refetchOnReconnect
refetchOnMount
retryOnMount
notifyOnChangeProps
notifyOnChangePropsExclusions
onSuccess
onError
onSettled
useErrorBoundary
select
suspense
keepPreviousData
placeholderData
optimisticResults
retry
retryDelay
cacheTime
isDataEqual
queryFn
queryHash
queryKey
queryKeyHashFn
initialData
initialDataUpdatedAt
behavior
structuralSharing
getPreviousPageParam
getNextPageParam
_defaulted
```

One way to avoid this is to eliminate the `useQuery(path, opts)` usage. Like so:

```ts
trpc.useQuery('allPosts');
trpc.useQuery('allPosts', undefined, { retry: false });
```

The user must explicitly pass `undefined` as the input if they wish to pass `UseQueryOptions` to react-query. I think this is the best solution.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -30,7 +30,8 @@
   "dependencies": {
     "@babel/runtime": "^7.9.0",
     "@trpc/client": "^4.0.0",
-    "@trpc/server": "^4.0.0"
+    "@trpc/server": "^4.0.0",
+    "tslib": "^2.1.0"
   },
   "devDependencies": {
     "@types/express": "^4.17.11",
--- a/packages/react/src/createReactQueryHooks.ts
+++ b/packages/react/src/createReactQueryHooks.ts
@@ -26,6 +26,53 @@ import {
   DehydrateOptions,
 } from 'react-query/hydration';
 
+const useQueryOptionKeys = {
+  enabled: true,
+  staleTime: true,
+  refetchInterval: true,
+  refetchIntervalInBackground: true,
+  refetchOnWindowFocus: true,
+  refetchOnReconnect: true,
+  refetchOnMount: true,
+  retryOnMount: true,
+  notifyOnChangeProps: true,
+  notifyOnChangePropsExclusions: true,
+  onSuccess: true,
+  onError: true,
+  onSettled: true,
+  useErrorBoundary: true,
+  select: true,
+  suspense: true,
+  keepPreviousData: true,
+  placeholderData: true,
+  optimisticResults: true,
+  retry: true,
+  retryDelay: true,
+  cacheTime: true,
+  isDataEqual: true,
+  queryFn: true,
+  queryHash: true,
+  queryKey: true,
+  queryKeyHashFn: true,
+  initialData: true,
+  initialDataUpdatedAt: true,
+  behavior: true,
+  structuralSharing: true,
+  getPreviousPageParam: true,
+  getNextPageParam: true,
+  _defaulted: true,
+};
+
+// const useMutationOptionKeys = {
+//   mutationKey: true,
+//   onMutate: true,
+//   onSuccess: true,
+//   onError: true,
+//   onSettled: true,
+//   retry: true,
+//   retryDelay: true,
+//   useErrorBoundary: true,
+// };
 export type OutputWithCursor<TData, TCursor extends any = any> = {
   cursor: TCursor | null;
   data: TData;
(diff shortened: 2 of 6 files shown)


KATT:
Great work! Thanks for taking the time and digging in.

I think it'd be even cleaner to extend the `UseQueryOptions` with an `input` option and make opts mandatory in case it's mandatory input & optional if input is optional. Would be familiar to Apollo-users to where you have `variables:`.

I generally think it's a bit off when functions have several args who's meaning change depending on what you pass, that's why I settled with the tuple style first arg.

Usage would be like this:

```tsx
useQuery(path) // if `input` is nullable/undefineable
useQuery(path, { input }) // if input is mandatory
```

Sketch of type signature:

```tsx
// Targets queries without inputs / queries with `Maybe` inputs (nullable/undefinable)
function _useQuery<
  // `TQueriesWithoutInputs` is a bit tricky to get but I've done
  // that struct before when I was playing with this
  // (should even be in the history somewhere, can dig it out)
  TPath extends keyof TQueriesWithoutInputs & string,
  TProcedure extends TQueries[TPath],
  TOutput extends inferProcedureOutput<TProcedure>
>(
  path: TPath,
  // note - `opts` is optional
  opts?: UseQueryOptions<
    inferProcedureInput<TQueries[TPath]>,
    TError,
    TOutput
  > & {
    // input is optional & inferred
    input?: inferProcedureInput<TProcedure>;
  },
): UseQueryResult<TOutput, TError>;
// Targets queries with mandatory inputs
function _useQuery<
  TPath extends keyof TQueriesWithInputs & string,
  TProcedure extends TQueries[TPath],
  TOutput extends inferProcedureOutput<TProcedure>
>(
  path: TPath,
  // note - `opts` is mandatory
  opts: UseQueryOptions<
    inferProcedureInput<TQueries[TPath]>,
    TError,
    TOutput
  > & {
    // input is mandatory
    input: inferProcedureInput<TProcedure>;
  },
): UseQueryResult<TOutput, TError>;
```

KATT:
Other reasons not to change it from how it is now 

- it's currently consistent with react-query (I had it fully consistent before but that broke autocompletion in vscode in the tuple) 
- react users should be aware of tuples already considering use Effect

KATT:
Bottom line is that

- it's not broken as-is
- it follows convention of react-query already 
- this PR introduces new complexity with that hairy massive object that needs to me manually kept in sync with react-query features
- this PR breaks autocompletion on existing feature (at least on my machine)

So with that said, let's hold off jumping the gun on this one. I initially hated the tuple first, but after using it I quickly stopped caring as well.

colinhacks:
Okay you'r right, scratch that approach. This is too confusing.

But I think this:

```ts
useQuery(path, { input }) // if input is mandatory
```

is clearly worse than this:

```ts
trpc.useQuery('allPosts');
trpc.useQuery('allPosts', undefined, { retry: false });
```

If you need `opts`, passing in `undefined` as the second argument is simple and understandable. Everyone knows `trpc.useQuery('allPosts');` is equivalent to `trpc.useQuery('allPosts', undefined);`. This is nice, readable, understandable, there's no nested objects or tuples. 

Consistency with react-query is not a good goal. By limiting the number of inputs to 1, we're able to provide a cleaner API, so that's exactly what we should do.

KATT:
> is clearly worse than this:

That's your opinion, you have to understand and respect that this isn't a absolute objective truth & everyone might not agree with you on this. 😊

For _me_, `input` is a clear part of the options object as it's 

- not always required
- option objects are easier to read as they're named
- i generally refrain from ever having more than 2 args
- if an arg is optional, i always make it part of an options object
- i really dislike both both of these
  - when you s_sometimes_ pass `undefined` as a second arg (accepting `null`/`undefined` would be good if we decide on this)
  - when arguments changing order (as in your suggestion)

> Consistency with react-query is not a good goal. 

It's a good start for now - it's a very well-made library and a lot of time and care has been taken into it's design. tRPC is built around it so it's easy to just copy design decisions as the wrapper gets super simple. 

---

Hey, I'm not saying no, but I want more people saying yes before doing this as it's a breaking change - please try using it on your project for a week and let me know if still pokes your eye as much after doing so. 

Apollo uses `variables` as part of the options object and Relay uses it as a second arg so both can be argued for.


---

Previously I properly copied react-query and supported `useQuery(path, opts)` & `useQuery([path, input], opts)` but that broke my autocompletion in the same way that this PR does.

KATT:
Closing this for now due to inactivity. Added the label "icebox" so it's easier to find in the history if we want to pick it back up at a later point.

PR #244, opened 2021-04-03
fix - no global QueryClient

### LINKED ISSUE #242: @trpc/react query cache is shared between Next.js pages
Hey there I noticed potentially a big problem with the instructions on how to use library with security implications.

I noticed this issue https://github.com/Sendouc/sendou.ink/issues/354 and started investigating it. Essentially the query cache is shared between pages. In my case it just added useless bloat to the `getStaticProps` JSONs since it had queries from various pages but of course if you use say `getServerSideProps` with sensitive information then cache being shared like this between requests is dangerous.

Here is how I fixed it: https://github.com/Sendouc/sendou.ink/commit/e42bc3c690c76d36071bb06938633395c7776626

I noticed a discrepancy between how trpc docs instruct to do it: https://trpc.io/docs/nextjs#2-create-a-trpc-client and how React Query docs do: https://react-query.tanstack.com/guides/ssr#using-hydration (see the bolded part especially).

KATT:
closes #242. potential security issue so jumped straight on it.

- make sure QueryClient isn't created globally
- moved `trpc.dehydrate()` to `trpc.ssr().dehydrate()`
- moved utilities for queries to under hook `useQueryUtils` (`prefetch`, `invalidate`, `cancelQuery`, `setQueryData`) 
- update docs: https://www-git-feature-sec-take-2.tmp.trpc.io/

---

I want to do a bigger overhaul of `@trpc/react` later:

- it's a bit annoying that we have this big factory function only because we want the generic to be inferred on the hooks & functions - is there a better way? cc @colinhacks 
- bootrapping is a bit annoying for people that are starting a greenfield project - we could abstract away a lot of junk like setting up your own react-query provider, etc
- `dehydratedState`-prop could be something inferred/forced from trpc to avoid naming conflicts/ typos


examples/next-hello-world/pages/_app.tsx
```diff
@@ -1,11 +1,13 @@
 import type { AppProps /*, AppContext */ } from 'next/app';
-import { QueryClientProvider } from 'react-query';
+import { useState } from 'react';
+import { QueryClient, QueryClientProvider } from 'react-query';
 import { Hydrate } from 'react-query/hydration';
 import { trpc } from '../utils/trpc';
 
 function MyApp({ Component, pageProps }: AppProps) {
+  const [queryClient] = useState(() => new QueryClient());
```
cyrus-za, line 8: nice 👍🏻

cyrus-za (approved):
Good work. Good call on the security issue @Sendouc

I recall seeing a similar issue logged with nextjs on the with-apollo example a few years back where apollo state was global at first.

PR #306, opened 2021-04-24
add automatic SSR support

KATT:
# Major Version!

V6 with **automatic SSR** by simply wrapping your app in `_app.tsx` with `withTRPC()`. 

## Reference project
- have a look at https://hello-world-git-feature-sec.tmp.trpc.io/  <-- completely server rendered without any extract work 
- https://github.com/trpc/trpc/tree/feature/sec/examples/next-hello-world <-- code
- https://githubbox.com/trpc/trpc/tree/feature/sec/examples/next-hello-world <-- play with it in codesandbox

## Docs

https://www-git-feature-sec.tmp.trpc.io/docs/nextjs

## Breaking changes

- SSG-helpers are now separated and has to be included through `import { createSSGHelpers } from '@trpc/react/ssg';`
- Changed how you set up trpc hooks - it only accepts the interface of the router now

---

some details

- automatic SSR built-in 
  - ability to use _app `getInitialProps()` to fetch all queries in tree
  - uses http rather than invoking router directly - might add some milliseconds because of the http overhead, but increases sec as you never need to import backend fns on the frontend that risk ending up in the build

---

todos

- [x] cleanup `getDataFromTree` (uses an ugly `while(true)`-loop)
- [x] add support for ssr on infinite queries
- [x] remove old creator
- [x] fix tests
- [x] cleanup
- [x] update all examples
- [x] fix test coverage


Sendouc:
So essentially the change is to move the data fetching logic to _app.js? I wonder about this approach since:

```
Adding a custom getInitialProps in your App will disable Automatic Static Optimization in pages without Static Generation.
```
Source: https://nextjs.org/docs/advanced-features/custom-app#caveats

I think there might be the possibility of adding `getStaticProps` for _app.js some day but it's not a feature of Next.js yet. I think because of that the current way of doing it in this library is the best practice and it matches how it's done for other libraries in the Next.js examples. See for example the [with-apollo example](https://github.com/vercel/next.js/tree/canary/examples/with-apollo).

examples/next-hello-world/pages/api/trpc/[trpc].ts
```diff
@@ -44,6 +45,15 @@ export const appRouter = createRouter()
...
     };
   })
+  .middleware(({ ctx }) => {
+    console.log('received request', {
+      url: ctx.req.url,
+      'x-ssr': ctx.req.headers['x-ssr'],
+    });
+    // if (ctx.req.headers['x-ssr']) {
+    //   throw new Error('Random failure');
+    // }
```
Comment (cyrus-za) [inline examples/next-hello-world/pages/api/trpc/[trpc].ts:55]: reminder about uncommented code

cyrus-za (approved):
this is awesome! Code looks good. I DM'd you some questions which you answered to my satisfaction so approving

PR #343, opened 2021-05-06
Add crosswalk to related projects

danvk:
Disclaimer: it's my project. From what I can tell, the projects have similar goals but take pretty different approaches towards achieving them. (Crosswalk makes you add a build step to get runtime type validation, but in exchange you get to use TypeScript types to define your API.)


KATT:
Seem to require actual schemas, it's similar to tRPC in the same way that https://github.com/acacode/swagger-typescript-api is, but it doesn't automatically infer things like blitzjs/trpc does

PR #385, opened 2021-05-18
Draft: Using multiple providers

molebox:
**This is just a test**

This is not the recommended way to do things. But if you keep the migrations in sync then you should be good to go. I tried opening this fork in codesandebox and you have both the sqlite and postgres migrations. Im not sure what your use case is for this project though.


KATT:
@all-contributors add @molebox for examples

KATT:
Note to self when testing

https://github.com/molebox/trpc/tree/multiple-providers/examples/next-ssg-chat

PR #393, opened 2021-05-18
add combined transformers

### LINKED ISSUE #378: Use different transformers for client<->server
First of all I want to say thank you very much for making this library! I was already looking for a library like this for quite some time. 

There is one thing I would like to ask/suggest:

Superjson is great for preserving type information, but has some performance drops for example when using with lots of nested dates. There is another library, devalue, which has a much better performance, but it is not usably for client->server for security reasons. 

Thats why I currently use superjson for client->server and devalue for server->client in some of my apps.

Is it currently possible to use different transformers for each direction? If not, I would like to suggest adding this.

simonedelmann:
As discussed [here](https://github.com/trpc/trpc/issues/378) this adds a new interface `CombinedDataTransformer` which allows to specify different transformers for server->client and client->server communication. `CombinedDataTransformer`can be used interchangeable with `DataTransformer`.

```ts
type DataTransformer = {
  serialize(object: any): any;
  deserialize(object: any): any;
};

type CombinedDataTransformer = {
  input: DataTransformer;
  output: DataTransformer;
};
```

### Example usage with superjson and devalue

```ts
import superjson from 'superjson';
import devalue from 'devalue';

export const transformer = {
  input: superjson,
  output: {
    serialize: (object) => devalue(object),
    deserialize: (object) => eval(`(${object})`),
  },
};

// Add to `createTRPCCLient()`
export const client = createTRPCClient<AppRouter>({
  // [...]
  transformer: transformer,
});

// Add to API handler
  // [...]
  transformer: transformer,
});
```

### TODO

- [x] Code
- [x] Documentation
- [x] Tests
- [ ] Add to example

---

Closes #378

### DIFF
--- a/www/docs/server/data-transformers.md
+++ b/www/docs/server/data-transformers.md
@@ -56,7 +56,7 @@ If a transformer should only be used for one directon or different transformers 
 
 ### How to
 
-Here [superjson](https://github.com/blitz-js/superjson) is be used for uploading and [devalue](https://github.com/Rich-Harris/devalue) for downloading data, because devalue is a lot faster but insecure to use on the server. 
+Here [superjson](https://github.com/blitz-js/superjson) is be used for uploading and [devalue](https://github.com/Rich-Harris/devalue) for downloading data, because devalue is a lot faster but insecure to use on the server.
 
 #### 0. Install
 
@@ -73,8 +73,8 @@ import devalue from 'devalue';
 // [...]
 
 export const transformer = {
-  up: superjson,
-  down: {
+  input: superjson,
+  output: {
     serialize: (object) => devalue(object),
     deserialize: (object) => eval(`(${object})`),
   },
@@ -116,7 +116,7 @@ type DataTransformer = {
 };
 
 type CombinedDataTransformer = {
-  up: DataTransformer;
-  down: DataTransformer;
+  input: DataTransformer;
+  output: DataTransformer;
 };
 ```
--- a/www/docs/server/data-transformers.md
+++ b/www/docs/server/data-transformers.md
@@ -18,7 +18,6 @@ SuperJSON allows us to able to transparently use e.g. standard `Date`/`Map`/`Set
 
 ### How to
 
-
 #### 0. Install
 
 ```bash
@@ -51,11 +50,73 @@ export default trpcNext.createNextApiHandler({
 });
 ```
 
+## Different transformers for upload and download
+
+If a transformer should only be used for one directon or different transformers should be used for upload and download (e.g. for performance reasons), you can provide individual transformers for upload and download. Make sure you use the same combined transformer everywhere.
+
+### Howto
+
+For demonstration purpose [superjson](https://github.com/blitz-js/superjson) should be used for uploading and faster but insecure [devalue](https://github.com/Rich-Harris/devalue) for downloading data.
+
+#### 0. Install
+
+```bash
+yarn add superjson devalue
+```
+
+#### 1. Add to `utils/trpc.ts`
+
+```ts
+import superjson from 'superjson';
+import devalue from 'devalue';
+
+// [...]
+
+export const transformer = {
+  up: superjson,
+  down: {
+    serialize: (object) => devalue(object),
+    deserialize: (object) => eval(`(${object})`),
+  },
+};
+```
+
+#### 2. Add to `createTRPCCLient()`
+
+```ts
+import { transformer } from 'utils/trpc.ts';
+
+// [...]
+
+export const client = createTRPCClient<AppRouter>({
+  // [...]
+  transformer: transformer,
+});
+```
+
+#### 3. Add to API handler
+
+```ts
+import { transformer } from 'utils/trpc.ts';
+
+// [...]
+
+export default trpcNext.createNextApiHandler({
+  // [...]
+  transformer: transformer,
+});
+```
+
 ## `DataTransformer` interface
 
 ```ts
 type DataTransformer = {
   serialize(object: any): any;
   deserialize(object: any): any;
+};
+
+type CombinedDataTransformer = {
+  up: DataTransformer;
+  down: DataTransformer;
 };
 ```
--- a/www/docs/server/data-transformers.md
+++ b/www/docs/server/data-transformers.md
@@ -54,9 +54,9 @@ export default trpcNext.createNextApiHandler({
 
 If a transformer should only be used for one directon or different transformers should be used for upload and download (e.g. for performance reasons), you can provide individual transformers for upload and download. Make sure you use the same combined transformer everywhere.
 
-### Howto
+### How to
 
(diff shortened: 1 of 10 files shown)


simonedelmann:
@KATT I've added code and documentation but I am struggling with adding tests for that. Would be great if you could have a look! Thanks!

KATT:
great! @simonedelmann I added a basic superjson test now (that already should've been there).

when doing your tests for this upstream/downstream, try to assert that the right fn have been called depending on up/down so no one breaks it in the future

www/docs/server/data-transformers.md
```diff
@@ -51,11 +50,73 @@ export default trpcNext.createNextApiHandler({
...
+    serialize: (object) => devalue(object),
+    deserialize: (object) => eval(`(${object})`),
+  },
+};
+```
+
+#### 2. Add to `createTRPCCLient()`
+
+```ts
+import { transformer } from 'utils/trpc.ts';
```
KATT, line 87:
```suggestion
import { transformer } from '../../utils/trpc';
```

www/docs/server/data-transformers.md
```diff
@@ -51,11 +50,73 @@ export default trpcNext.createNextApiHandler({
...
+export const client = createTRPCClient<AppRouter>({
+  // [...]
+  transformer: transformer,
+});
+```
+
+#### 3. Add to API handler
+
+```ts
+import { transformer } from 'utils/trpc.ts';
```
KATT, line 100:
```suggestion
import { transformer } from '../../utils/trpc';
```

www/docs/server/data-transformers.md
KATT, line 87:
```suggestion
import { transformer } from '../utils/trpc';
```

www/docs/server/data-transformers.md
```diff
@@ -51,11 +50,73 @@ export default trpcNext.createNextApiHandler({
 });
 ```
 
+## Different transformers for upload and download
+
+If a transformer should only be used for one directon or different transformers should be used for upload and download (e.g. for performance reasons), you can provide individual transformers for upload and download. Make sure you use the same combined transformer everywhere.
+
+### Howto
+
+For demonstration purpose [superjson](https://github.com/blitz-js/superjson) should be used for uploading and faster but insecure [devalue](https://github.com/Rich-Harris/devalue) for downloading data.
```
KATT, line 59:
```suggestion
Here [superjson](https://github.com/blitz-js/superjson) is be used for uploading and [devalue](https://github.com/Rich-Harris/devalue) for downloading data, because devalue is a lot faster but insecure to use on the server. 
```

www/docs/server/data-transformers.md
```diff
@@ -51,11 +50,73 @@ export default trpcNext.createNextApiHandler({
 });
 ```
 
+## Different transformers for upload and download
+
+If a transformer should only be used for one directon or different transformers should be used for upload and download (e.g. for performance reasons), you can provide individual transformers for upload and download. Make sure you use the same combined transformer everywhere.
+
+### Howto
```
KATT, line 57:
```suggestion
### How-to
```

KATT:
@all-contributors add @simonedelmann for code and idea and tests

KATT:
Closes #378

simonedelmann:
Thanks for the test and your help with English documentation. It's not my native language unfortunately ;)

One question about naming: 

```ts
type CombinedDataTransformer = {
  up: DataTransformer;
  down: DataTransformer;
};
```

Are you happy with `up` and `down`? I don't really like them but I don't have a better idea..

KATT:
> Are you happy with `up` and `down`? I don't really like them but I don't have a better idea..

Maybe `input` and `output`? Or `upstream`/`downstream`?

Also should add jsdoc to them to clarify any confusion

KATT:
I quite like `input` and `output` I think

www/docs/server/data-transformers.md
```diff
@@ -51,11 +50,73 @@ export default trpcNext.createNextApiHandler({
...
+// [...]
+
+export const transformer = {
+  input: superjson,
+  output: {
+    serialize: (object) => devalue(object),
+    deserialize: (object) => eval(`(${object})`),
+  },
+};
+```
```
KATT, line 82: The only thing I don't like with this is that `devalue` isn't actually needed at all on the client and becomes an unnecessary import - got any idea to work around this?

KATT: I guess on the client you only need `input.serialize` & `output.deserialize` and the inverse for the server.... 🤔

simonedelmann: At least when using devalue this is no problem. Devalue has no dependencies and should be very small.

KATT:
```tsx
export type CombinedDataTransformerClient = {
  input: Pick<DataTransformer, 'serialize'>;
  output: Pick<DataTransformer, 'deserialize'>;
};
```

KATT: The types (on the trpcclient) should still reflect that you can omit it if you want so it's up to the dev to optimize to the level they want

simonedelmann: I've added `CombinedDataTransformerClient` type and used in client packages, but I am still not sure if this will prevent devalue to be imported on client.

simonedelmann:
I think, it should be done mainly now. 

Shall we add an example with combined transformers? Or shall I modify the todomvc example to make use of combined transformers?

www/docs/server/data-transformers.md
simonedelmann: > The types (on the trpcclient) should still reflect that you can omit it if you want so it's up to the dev to optimize to the level they want

Ah, I see. The type is added in https://github.com/trpc/trpc/pull/393/commits/9845d012a7e7bef1f1f77b591a72f61ea85b20ea

simonedelmann:
Hm.. thought about this now for quite some time.

As long as we make sure that we only eval a string which comes encoded from the server, it should be fine. 

Even if an user enters harmful text and the server responds that text back, that text will be encoded as string and is not going to be evaluated by eval anyway, right? And if somebody is able to let the server respond an unencoded response, the server is already compromised anyway, at least from what I unterstand.

simonedelmann:
I have just checked devalues readme:

https://github.com/Rich-Harris/devalue#other-security-considerations

Maybe we should also wrap eval in a isolated block?
From my understanding this should not be necessary, but there is no reason not to do it.

KATT:
I _think_ it's alright, but I'm not an XSS-injection expert

KATT:
You know / follow anyone on twitter you could maybe ping?

simonedelmann:
> You know / follow anyone on twitter you could maybe ping?

Unfortunately no. 

What about not using devalue in docs and tests? We can use something else, just for demonstration purpose. CombinedTransformers don't need devalue for being a helpful feature (and I can still use devalue if I want).

KATT (approved):

KATT:
published in new minor bump

simonedelmann:
Great, thank you!

PR #415, opened 2021-05-24
add guards for `Object.prototype`-props

KATT:
## issue

just realised that we don't safe-guard from querying on built-in object properties. 

i.e. 

- `client.query('toString')` for example does _not_ return `NOT_FOUND` as it's part of the built-in stuff
- also, you can't use `toString` as a procedure key

## fix

- refactor router to create all objects with `Object.create(null)` instead of `null`
- refactor spread operators to `Object.assign(Object.create(null), x, y)` (via `safeObject`-fn)

## notes
did it TDD-style - both of these tests were failing as you can see in the commit history

### DIFF
--- a/packages/server/src/router.ts
+++ b/packages/server/src/router.ts
@@ -89,6 +89,27 @@ export type DefaultErrorShape = {
   stack?: string;
 };
 
+/**
+ * Create an empty object with `Object.create(null)`.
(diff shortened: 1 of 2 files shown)


KATT:
I just want one 👍 from someone... 👀

simonedelmann:
I will have a look tomorrow!

kripod (approved):

PR #475, opened 2021-06-09
enable batching in ssr

KATT:
- enables batching when doing ssr
- skip using `suspense` in ssr and doing a little hack to fetch the queries
- calling `react-ssr-prepass` several times until there's no pending queries

### DIFF
--- a/packages/react/src/createReactQueryHooks.tsx
+++ b/packages/react/src/createReactQueryHooks.tsx
@@ -44,8 +44,6 @@ export type OutputWithCursor<TData, TCursor extends any = any> = {
   data: TData;
 };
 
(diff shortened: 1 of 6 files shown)


packages/react/src/createReactQueryHooks.tsx
```diff
@@ -191,22 +191,27 @@ export function createReactQueryHooks<TRouter extends AnyRouter>() {
...
+    const { client, isPrepass, queryClient, fetchQuery } = useContext();
 
+    if (
+      typeof window === 'undefined' &&
+      isPrepass &&
+      opts?.ssr !== false &&
+      !queryClient.getQueryCache().find(cacheKey)
+    ) {
+      fetchQuery(pathAndArgs);
+    }
```
KATT, line 210: this really doesn't feel kosher

PR #508, opened 2021-06-10
add websocket support

KATT:
# WebSocket support 🎉 

- 😻 add support for websockets transport using  the `ws`-package and a new `wsLink()`
- 🤔 planning to release this a bit earlier than it's ready but with hidden docs and some warnings if you use it

## How to review

- 👉  **Try it on CodeSandbox** - https://githubbox.com/trpc/trpc/tree/feature/websockets/examples/standalone-server
- 👀  Read the `test('$subscription()', ..` in the tests to understand usage both on the backend and frontend https://github.com/trpc/trpc/blob/feature/websockets/packages/server/test/websockets.test.ts

## Left to do

- [ ] someone to review / critisise the API
- [x] add support for queries
- [x] add support for subscriptions
- [x] add support for mutations
- [ ] graceful reconnect when server shutting down - broadcast some event to clients and they can have a grace period where they reconnect (and ideally not all at once), in k8s this is 30sec?
- [ ] cleanup api - need some input on both the backend and the frontend
- [x] more rigid testing 
- [ ] ~maybe creating a new `Subscription` instance on the backend for each isn't that performant?~ 🤷 
- [x] allow backend subscriptions to control when they stop themselves and notify client when they do
- [x] should `webSocketLink` be called `wsLink`?
- [x] either make server not depend on `ws` or make separate package?
- [x] observable logic is quite convoluted
- [x] errors and dege cases
  - [x] backend: what to do if a `Subscription` emits an error? auto-destroy the subscription? just pass to client?
  - [x] client: what to do if a subscription encounters an error? up to the subscriber i guess?
  - [x] some errors can probably terminate the subscription - for instance if `createContext` fails or if the ws server goes down - how to handle that?


packages/server/test/websockets.test.ts
```diff
@@ -0,0 +1,171 @@
...
+          "data": Object {
+            "id": "1",
+          },
+          "ok": true,
+        },
+      ],
+    ]
+  `);
+
+  $sub.destroy();
```
KATT, line 164: - you call `client.$subscription('x')` to return an observable
- then you subscribe to the observable
- then you call `$sub.destroy()`

i feel it's a bit convoluted, what you would think would be ideal, @gmencz?

packages/server/src/websockets/websocketHandler.ts
```diff
@@ -0,0 +1,218 @@
...
+      params: {
+        input: TInput;
+        path: string;
+      };
+    }
+  | {
+      id: number;
+      jsonrpc: '2.0';
+      method: 'stop';
+    };
```
KATT, line 39: is this a good way to stop a subscription?

packages/server/test/websockets.test.ts
gmencz: What is `onMessage`? Is it some topic you subscribe to and other clients can publish to it?

gmencz: Each of them would call the `client.subscribe(topic)` method and whenever a client publishes something on that topic, the server broadcasts to all subscribers of the topic, that is how https://github.com/FungiRealtime does it and it's been working pretty well

cyrus-za (commented):
This look amazing. Thanks for putting in the time

packages/server/test/websockets.test.ts
```diff
@@ -0,0 +1,169 @@
...
+      });
+      ee.emit('server:msg', {
+        id: '1',
+      });
+    });
+  });
+  const callback = jest.fn();
+  const $sub = client.$subscription('onMessage');
+
+  $sub.subscribe({ onNext: callback });
```
KATT, line 141: The subscriptions is in this format and you have to call `$sub.done()` to end the subscription. This allows multiple subscribers, but maybe that's over-doing it.

Could be nicer with this format:

```ts
const unsubscribe = client.$subscription('onMessage', {
  onNext(data: TOutput) {},
  onError(err: TRPCClientError) {},
});

unsubscribe();
```

Thoughts? 

---


Secondly - error handling - there are some errors that would be "fatal" and some that at "transient". I.e. one pull of the subscription in the backend could return an error which wouldn't cause the link to broken, but then there's other type of errors that could force-quit the subscription... unsure how to deal with the different cases. 


---

cc @cyrus-za @anthonyshort

gmencz: I really like that format. Not sure if I prefer `onData` over `onNext`

---

For error handling, inversion of control is what seems most "appropiate" for me, so the developer should handle the different types of errors instead of trpc handling them. What I'm thinking is that whenever any error occurs, that should trigger the `onError` callback, maybe this callback's error could also have a `code`? For example if the connection closes abnormally, the code would be `1006`, all the websocket error codes are listed here: https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent#properties. 

This way, the developer using trpc should handle the different type of errors based on whether they have a code or not:
```ts
const unsubscribe = client.$subscription('onMessage', {
  onNext(data: TOutput) {},
  onError(err: TRPCClientError, code?: number) {
    if (code) {
      // Do something...
    }
    
    // Do something else based on `err`
  },
});

unsubscribe();
```
Thoughts?

KATT: I have changed to your suggestion, @gmencz. What do you think now?

Still unsure how to handle errors nicely...

KATT: variants:

- intermittent error - one pull messed up but next `onNext()` could work
- fatal error for the specific subscription - an error led to subscription being cancelled
- error because server shut down
  - potentially because of a new version on the server - would be nice to handle gracefully
  - fails but potentially reconnect
- ..

PR #513, opened 2021-06-12
fix vercel url

boredland:
just stumbled upon

1. wrong example url for vercel (in other places throughout the docs it is correct)
~2. imo usually unnecessarily complex url for render.com~

### DIFF
--- a/examples/next-prisma-starter/src/pages/_app.tsx
+++ b/examples/next-prisma-starter/src/pages/_app.tsx
@@ -17,14 +17,14 @@ function getBaseUrl() {
   if (process.browser) {
     return '';
   }
-  // reference for vercel.com
+  // // reference for vercel.com
   // if (process.env.VERCEL_URL) {
   //   return `https://${process.env.VERCEL_URL}`;
   // }
 
   // // reference for render.com
-  // if (process.env.RENDER_EXTERNAL_URL) {
-  //   return process.env.RENDER_EXTERNAL_URL;
+  // if (process.env.RENDER_INTERNAL_HOSTNAME) {
+  //   return `http://${process.env.RENDER_INTERNAL_HOSTNAME}:${process.env.PORT}`;
   // }
 
   // assume localhost
--- a/examples/next-prisma-starter/src/pages/_app.tsx
+++ b/examples/next-prisma-starter/src/pages/_app.tsx
@@ -17,14 +17,14 @@ function getBaseUrl() {
   if (process.browser) {
     return '';
   }
-  // // reference for vercel.com
+  // reference for vercel.com
   // if (process.env.VERCEL_URL) {
-  //   return process.env.VERCEL_URL;
+  //   return `https://${process.env.VERCEL_URL}`;
   // }
 
   // // reference for render.com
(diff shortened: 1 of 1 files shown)


examples/next-prisma-starter/src/pages/_app.tsx
```diff
@@ -17,14 +17,14 @@ function getBaseUrl() {
   if (process.browser) {
     return '';
   }
-  // // reference for vercel.com
+  // reference for vercel.com
   // if (process.env.VERCEL_URL) {
-  //   return process.env.VERCEL_URL;
+  //   return `https://${process.env.VERCEL_URL}`;
```
KATT, line 22: This is good 👍

examples/next-prisma-starter/src/pages/_app.tsx
```diff
@@ -17,14 +17,14 @@ function getBaseUrl() {
...
   // }
 
   // // reference for render.com
-  // if (process.env.RENDER_INTERNAL_HOSTNAME) {
-  //   return `http://${process.env.RENDER_INTERNAL_HOSTNAME}:${process.env.PORT}`;
+  // if (process.env.RENDER_EXTERNAL_URL) {
+  //   return process.env.RENDER_EXTERNAL_URL;
```
KATT, line 27: This forces the server-side request to go through the interwebz in order to make a call to a server that sits in the same machine

KATT (changes requested):

examples/next-prisma-starter/src/pages/_app.tsx
KATT: Also might not work for PR previews, but not sure

boredland: I did not understand that. Will add a comment, ok?

KATT: Just revert the render bit, I'm happy with the vercel change

KATT: I use the render internal env var in my prod app

KATT (approved):

examples/next-prisma-starter/src/pages/_app.tsx
```diff
@@ -19,7 +19,7 @@ function getBaseUrl() {
   }
   // // reference for vercel.com
   // if (process.env.VERCEL_URL) {
-  //   return process.env.VERCEL_URL;
+  //   return `https://${process.env.VERCEL_URL}`;
```
KATT, line 22: Have you verified that vercel doesn't already include https?

boredland: ya, had an "only allows absolute urls" error on vercel and headed over to their docs: https://vercel.com/docs/environment-variables#system-environment-variables

```
VERCEL_URL | The URL of the deployment. Example: my-site-7q03y4pi5.vercel.app.
```

boredland: if we want this to be nice we could use sth like url.parse

KATT: Awesome, will double check but will merge this for now, thank you so much! 🙏

PR #529, opened 2021-06-15
support `null`/`undefined` procedure results

### LINKED ISSUE #527: Mutations with no return value stall forever
When a `resolve` callback’s return type is either:

- `Promise<void>`
- `Promise<null>`
- `Promise<undefined`

tRPC doesn’t ever close the response stream, thus, stalling the client indefinitely.

The following code **doesn’t work:**

```ts
trpc.router<Context>()
  .mutation("update", {
    input: z.object({ id: z.string() }),
    resolve: async ({ ctx, input }): Promise<void> => {
      await updateSection(input);
      // `return undefined` or `return null` wouldn’t solve the case
    },
  });
```

But the following return value change **does the trick:**

```ts
trpc.router<Context>()
  .mutation("update", {
    input: z.object({ id: z.string() }),
    resolve: async ({ ctx, input }): Promise<{ [key: string]: never }> => {
      await updateSection(input);
      return {};
    },
  });
```

KATT:
closes #527


packages/client/src/internals/executeChain.ts
```diff
@@ -17,7 +17,12 @@ export function executeChain<
   op: Operation<TInput>;
 }) {
   type TError = TRPCClientError<TRouter>;
-  const $result = observableSubject<TOutput | null, TError>(null);
+  const $result = observableSubject<
+    { type: 'init' } | { type: 'data'; data: TOutput },
+    TError
+  >({
+    type: 'init',
+  });
```
KATT, line 25: This was the culprit - when we initialize as call we started with an `observableSubject` of `null` and we also used this to distinguish if a request was pending or had results.

kripod (approved):

packages/server/test/websockets.test.ts
```diff
@@ -221,7 +221,7 @@ test('$subscription()', async () => {
   wsClient.close();
 });
 
-test('$subscription() - server randomly stop and restart (this test might be flaky, try re-running)', async () => {
+test.skip('$subscription() - server randomly stop and restart (this test might be flaky, try re-running)', async () => {
```
simonedelmann, line 224: What's the reason for this change?

packages/server/test/index.test.tsx
```diff
@@ -533,3 +536,36 @@ describe('createCaller()', () => {
     sub.start();
   });
 });
+
+// regression https://github.com/trpc/trpc/issues/527
+test('void mutation response', async () => {
+  const { client, close, wssPort, router } = routerToServerAndClient(
+    trpc
+      .router()
+      .mutation('undefined', {
```
simonedelmann, line 545: I wonder if we should have a test for `return undefined` as well? Should be the same, but I am not sure if `void` always equals `undefined`.

packages/client/src/internals/executeChain.ts
simonedelmann: I get why using `null` is not possible and I think you have found a working solution. I wonder if it's possible to make it a little more intuitive. We cannot use `undefined` instead of `null`, right?

simonedelmann (commented):
I left some comments, would be curious to hear what both of you think. 

Sorry for brevity and less code  snippets. Smartphone/Train...

packages/server/test/index.test.tsx
KATT: `void` === `undefined`

Issue #545, opened 2021-06-16
Transformers at server level


delaneyj:
https://trpc.io/docs/data-transformers show transformer at the client and in next handlers.  Would it be possible to match the client interface ala

```ts
const appRouter = router<Context>({
  transformer:superjson
})
```
so it working without being tied to Next.js

KATT:
There was a reason for this design that I have now forgotten.... wanna take a stab at it? I'd be happy to do the change.

Indeed this would be nicer... could be done as part of next major, #532 also has breaking changes.

KATT:
_If_ you do, start off with a fork of `feature/refactor-rpc`

PR #550, opened 2021-06-17
refactor router initialization

KATT:
- refactor router initialization in order to simplify work with #545

(sorry for all the force pushes to the branch, was fighting with CI)

### DIFF
--- a/packages/server/src/router.ts
+++ b/packages/server/src/router.ts
@@ -112,6 +112,13 @@ function safeObject<TObj1, TObj2>(
 function safeObject(...args: unknown[]) {
   return Object.assign(Object.create(null), ...args);
 }
+const defaultFormatter: ErrorFormatter<unknown, DefaultErrorShape> = ({
+  defaultShape,
+}: {
+  defaultShape: DefaultErrorShape;
+}) => {
+  return defaultShape;
+};
 export type MiddlewareFunction<TContext> = (opts: {
   ctx: TContext;
   type: ProcedureType;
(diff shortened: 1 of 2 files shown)


packages/server/src/router.ts
```diff
@@ -138,24 +144,18 @@ export class Router<
   }>;
```
KATT, line 144: Add `transformer:` up here

packages/server/src/router.ts
```diff
@@ -138,24 +144,18 @@ export class Router<
...
-        defaultShape: DefaultErrorShape;
-      }) => {
-        return defaultShape;
-      }) as any,
+    this._def = {
+      queries: (def?.queries ?? safeObject()) as TQueries,
+      mutations: (def?.mutations ?? safeObject()) as TMutations,
+      subscriptions: (def?.subscriptions ?? safeObject()) as TSubscriptions,
+      middlewares: def?.middlewares ?? [],
+      errorFormatter: def?.errorFormatter ?? (defaultErrorFormatter as any),
```
KATT, line 158: and here

packages/server/src/router.ts
```diff
@@ -490,16 +487,15 @@ export class Router<
...
       TContext,
       TQueries,
       TMutations,
       TSubscriptions,
       ReturnType<TErrorFormatter>
-    >({
-      ...this._def,
-      errorFormatter,
-    });
+    >;
```
KATT, line 498: and a `transformer()` function similar to this

KATT (commented):
the only tricky part with data transformers is how to deal with output emitted from subscriptions

cc @delaneyj

delaneyj (approved):

PR #560, opened 2021-06-18
fix request cancelling

### LINKED ISSUE #559: Cancelling mutations should throw an `AbortError` instead of `new Error('Done')`
Since upgrading to v7, I’ve noticed that the exception handling mechanism in use by our company doesn’t work anymore:

```js
const { client } = trpc.useContext();
let cancelablePromise;
try {
  cancelablePromise = client.mutation("something", { param: "example" });
  setTimeout(() => {
    cancelablePromise.cancel()
  }, 0);
  return await cancelablePromise;
} catch (error) {
  const originalError =
    error instanceof TRPCClientError ? error.originalError : error;
  if (
    originalError instanceof DOMException &&
    originalError.name === "AbortError"
  ) {
    // We should reach this branch
    throw new CancelledError();
  }
  // But instead, `new Error('Done')` is called via `.cancel()`
  throw error;
}
```

I think I found the guilty code snippet:

https://github.com/trpc/trpc/blob/4a5654b36ca3a520d7e369cd45f6eab4c95c6e57/packages/client/src/createTRPCClient.ts#L251-L256

It should be corrected to throw an `AbortError` instead, just like DOM APIs (e.g. `fetch`) do.

My use-case would be distinguishing regular and cancellation errors (the latter are caused by a debouncing mechanism).

KATT:
closes #559

### DIFF
--- a/packages/server/test/index.test.tsx
+++ b/packages/server/test/index.test.tsx
@@ -11,6 +11,7 @@ import { CreateHttpContextOptions } from '../src';
 import { routerToServerAndClient } from './_testHelpers';
 import WebSocket from 'ws';
 import { waitFor } from '@testing-library/react';
+import { httpBatchLink } from '../../client/src/links/httpBatchLink';
 test('mix query and mutation', async () => {
   type Context = {};
   const r = trpc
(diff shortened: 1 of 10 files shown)


packages/server/test/index.test.tsx
```diff
@@ -569,3 +571,85 @@ test('void mutation response', async () => {
...
+
+    expect(err.name).toBe('TRPCClientError');
+    expect(err.originalError?.name).toBe('AbortError');
+
+    close();
+  });
+
+  test('cancelling batch request should throw TRPCAbortError', async () => {
+    // aborting _one_ batch request doesn't necessarily mean we cancel the reqs part of that batch
+    // therefore, we throw TRPCAbortError instead
```
KATT, line 607: cc @kripod 

if you do a batch request it'll be `err.name === 'TRPCAbortError`

KATT: do you think i should throw the same error for non-batch http requests too to ensure it's consistent?

kripod: Yep, that sounds like a good idea – although I’m not sure why a `TRPCAbortError` is required at all. I think a `TRPCClientError` may be used over here too, wrapping `new DOMException("The operation was aborted.", "AbortError")` as its `originalError`.

KATT: Great idea!

kripod:
This is fantastic – thank you for the quick fix, @KATT! 😊

PR #568, opened 2021-06-20
remove http subscriptions

### LINKED ISSUE #367: Add `trpc.useContext().getQueryData()`
In order to [execute updates optimistically with React Query](https://react-query.tanstack.com/guides/optimistic-updates#updating-a-list-of-todos-when-adding-a-new-todo), data has to be read from the `queryClient`, as seen in the linked example:

```ts
// Snapshot the previous value
const previousTodos = queryClient.getQueryData('todos')
```

Unfortunately, the situation isn’t so easy with tRPC, as it doesn’t provide a `getQueryData` method, but only `setQueryData`. After digging into the source code, I figured out an escape hatch to perform query data reads with – but it [relies upon](https://github.com/trpc/trpc/blob/b0ad15ef95b8d7d0b7ac8747df54d5aa08154550/packages/react/src/createReactQueryHooks.tsx#L163-L176) the [`CACHE_KEY_QUERY` internal constant](https://github.com/trpc/trpc/blob/b0ad15ef95b8d7d0b7ac8747df54d5aa08154550/packages/react/src/internals/constants.tsx#L3). The issue is showcased below without the additional clutter of handling optimistic updates via `onMutate` and `onError`:

```ts
export function useCreatePage() {
  const { setQueryData, queryClient } = trpc.useContext();

  return trpc.useMutation("Page.create", {
    onSuccess: (data) => {
      setQueryData(["Page.find", data.id], data);

      const prevPagesBySiteId =
        queryClient.getQueryData<Page[]>([
          "Page.findBySiteId",
          data.siteId,
          "TRPC_QUERY", // CACHE_KEY_QUERY
        ]) ?? [];
      setQueryData(
        ["Page.findAllBySiteId", data.siteId],
        [...prevPagesBySiteId, data],
      );
    },
  });
}
```

The code above should be replaceable with:

```ts
export function useCreatePage() {
  const { getQueryData, setQueryData } = trpc.useContext();

  return trpc.useMutation("Page.create", {
    onSuccess: (data) => {
      setQueryData(["Page.find", data.id], data);

      const prevPagesBySiteId =
        getQueryData(["Page.findBySiteId", data.siteId]) ?? [];
      setQueryData(
        ["Page.findAllBySiteId", data.siteId],
        [...prevPagesBySiteId, data],
      );
    },
  });
}
```

KATT:
- remove support of http subscriptions
- removing `useLiveQuery` - might be re-added later in another form
- changing so we do an `TRPCAbortError` instead of a DOMException as DOMException does not work in node (cc @kripod )
- closes #367


kripod:
Oh I see the reason behind using `TRPCAbortError`. That’s a good solution for supporting Node, thank you for noticing 😊

Issue #474, opened 2021-06-09
next major


KATT:
- [x] cleanup `@deprecated`
- [x] change loggerLink to accept a custom logger with proper opts
- [x] maybe change ending link to return `Error` downstream if (`!result.ok`)
- [ ] ~maybe force you start trpc client with links`
- [ ] stricter context

PR #571, opened 2021-06-21
removing http handlers' `transform:`-property

KATT:
- change will be part of next major (#474)
- already deprecated since a few versions ago
- use `router().transform(superjson)` instead

---

fyi @simonedelmann @cyrus-za  @kripod


simonedelmann (commented):
Too late to make a review, but looks good. I am already using the new syntax without any problems.

PR #572, opened 2021-06-21
refactor `withTRPC()` to use 1 argument

KATT:
- refactor `withTRPC` to only take one argument that is an options object with `config:() => ..` and `ssr: boolean`


cyrus-za (approved):
So much better

PR #595, opened 2021-06-24
improve websockets support 

KATT:
- make both dev & prod env for ws example
- subscriptions now auto-resubscribe after disconnect and reconnect signal
- deployed at https://trpc-websockets-example.herokuapp.com
- code is a bit convuluted in the wslink/client, could do with a refactor


examples/.wip/next-prisma-starter-websockets/src/pages/_app.tsx
```diff
@@ -14,34 +16,18 @@ const MyApp: AppType = ({ Component, pageProps }) => {
...
+  const { APP_URL } = publicRuntimeConfig;
+  if (!process.browser) {
+    return httpBatchLink({
+      url: `${APP_URL}/api/trpc`,
     });
   }
-  return httpBatchLink({
-    url: `${getHostname()}/api/trpc`,
+  const client = createWSClient({
+    url: APP_URL.replace(/^http/, 'ws'),
```
marcusradell, line 27: This works, but I'd still be surprised to find out that the protocol got replaced.
I'd consider using a separate base URL for ws or exclude the protocol, even if those also have some drawbacks.

PR #620, opened 2021-07-03
BUG: Combined transformers not transforming errors on server

simonedelmann:
https://github.com/trpc/trpc/blob/8e6ea59f07de33addc486ec6cc956deb40789109/packages/server/src/http/internals/getHTTPStatusCode.ts#L34

When using combined transformers, this line is calling `transformer.input.deserialize`, but it is parsing an error which is serialized with `transformer.output.serialize`.

### DIFF
--- a/packages/server/src/http/internals/getHTTPStatusCode.ts
+++ b/packages/server/src/http/internals/getHTTPStatusCode.ts
@@ -1,5 +1,5 @@
 import { TRPCResponse, TRPC_ERROR_CODES_BY_KEY } from '../../rpc';
-import { DataTransformer } from '../../transformer';
+import { CombinedDataTransformer } from '../../transformer';
 import { invert } from './invert';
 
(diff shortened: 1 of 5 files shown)


simonedelmann:
Before the router was only storing a collapsed DataTransformer. Now router stores the full CombinedDataTransformer object.

If an error is thrown in a query/mutation, this error can now safely be deserialized on the server. (This is happening [here](https://github.com/trpc/trpc/blob/8e6ea59f07de33addc486ec6cc956deb40789109/packages/server/src/http/internals/getHTTPStatusCode.ts#L34))

The test which validates this behavior is a little ugly. I combined the "basic" test from `/server/test/errors.test.ts` with the combined-transformer test in `/server/test/transformers.test.ts`. Probably there is some unnecessary overhead.

KATT (approved):
❤️

PR #622, opened 2021-07-04
Fix disabled query SSR and add tests

### LINKED ISSUE #621: @trpc/react trpc.useQuery ignores enabled: false during SSR
The issue seems to be that it doesn't check for `opts?.enabled !== false` here:
https://github.com/trpc/trpc/blob/438c06d90a1a5c358468424860b4384675d5154e/packages/react/src/createReactQueryHooks.tsx#L211-L218

hypnodron:
Closes #621 

When writing tests I've encountered two different behavior depending on whether the `queryKey` changes or not and documented both of them in tests.

### DIFF
--- a/packages/react/src/createReactQueryHooks.tsx
+++ b/packages/react/src/createReactQueryHooks.tsx
@@ -212,6 +212,7 @@ export function createReactQueryHooks<TRouter extends AnyRouter>() {
       typeof window === 'undefined' &&
       isPrepass &&
       opts?.ssr !== false &&
+      opts?.enabled !== false &&
       !queryClient.getQueryCache().find(cacheKey)
     ) {
       fetchQuery(pathAndArgs);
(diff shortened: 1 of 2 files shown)


KATT:
Awesome, thanks!! Could you add it to the infinite query too (if they have that prop; I'm on my phone reviewing this)

hypnodron:
Fixed the issue for infinite query and added tests. 

Also I added options forwarding from `useQuery` to `fetchQuery`. I could not figure out the types so I casted the options to any for now. Even though `react-qeury`'s `fetchQuery` accepts only subset of options of `useQuery`, forwarding them directly seems fine and `react-qeury` just ignores excess options.

KATT:
Out hiking 🗻 will merge ASAP. Thanks so much for being so thorough.

KATT (approved):

KATT:
Thank you so much for this @hypnodron!!

KATT:
@all-contributors please add @hypnodron for tests and code

Issue #160, opened 2021-03-08
[RFC] Procedure call batching


KATT:
Would be cool if one could automatically batch all parallel queries into the same request on the client. https://github.com/graphql/dataloader for inspo.

Shouldn't be that hard to do. Could enable some cool stuff like making a db transaction across multiple parallel mutations.

---

considerations:

- should query/mutation/subscriptions be able to batched together or should we batch per type?
- query-specific: how to avoid caching of a mix of "public"/"private" procedures if one uses caching
- subscriptions: tricky
- opt-in or opt-out?

PR #669, opened 2021-07-18
fix: batching with zod `.optional()` input

KATT:
- `JSON.stringify([undefined])` === `[null]`
- this causes issues with `.optional()` zod schemas as it expects an object _or_ `undefined`, not `null`
- ~fixed by ugly hack that runs validator twice when it fails if the raw input was `=== null`.~ 
  fixed by @simonedelmann's [idea](https://github.com/trpc/trpc/pull/669#issuecomment-882703759)


simonedelmann:
I've just checked the PR. I am not sure if I like the hack, but it should do the job. But...

I am not sure if we should merge this. If we expect the library to always parse values correctly, we should set a transformer (SuperJSON or something similar) by default and not just handle one specific edge case. But if the library by default uses `JSON.stringify` and `JSON.parse` for encoding/decoding values and we expect our users to deal with its issues anyway (sending Dates for example), they should also be aware of this edge case and manage it on their own.

KATT:
> I am not sure if we should merge this. If we expect the library to always parse values correctly, we should set a transformer (SuperJSON or something similar) by default and not just handle one specific edge case, [...] , they should also be aware of this edge case and manage it on their own.

I understand your point of view, I just think this is a very common edge-case, especially since I want to enable batching per default (#160). 

I don't wanna add transformers as default as they add a perf hit and everyone don't benefit from having them. At least not superjson, as it's slow.

I wish `JSON.parse()` would throw on dates/map/set as default - I might actually add a specific callback to `JSON.parse()` to make this happen with a note that they should add a transformer as a follow-up.

simonedelmann:
> I understand your point of view, I just think this is a very common edge-case, especially since I want to enable batching per default (#160).

I see. So it is probably worth handling this specific case. Then I have only one more concern: Are `null` and `undefined` the only two values which are represented by `"null"` in JSON? So we make sure not to parse anything as `undefined` which actually didn't was `undefined` before.

> I wish `JSON.parse()` would throw on dates/map/set as default - I might actually add a specific callback to `JSON.parse()` to make this happen with a note that they should add a transformer as a follow-up.

I like the idea of throwing on Dates/Maps/Sets/... by default. The only thing I would suggest then is to do a little research, what kind of problems JSON.stringify/parse has, so we are not only addressing a few but all (or at least many) of them.

simonedelmann:
One more question: Couldn't we solve this client-side? So if the input value is `null`, we send `JSON.stringify(null)`. But if input value is not existing at all, we just send an empty string `""`, which is no valid JSON then but is clearly identifiable as `undefined`?

KATT:
> One more question: Couldn't we solve this client-side? So if the input value is `null`, we send `JSON.stringify(null)`. But if input value is not existing at all, we just send an empty string `""`, which is no valid JSON then but is clearly identifiable as `undefined`?

Good idea, but it would break this code wh

```tsx
export const appRouter = createRouter()
  .query('hello', {
    input: z.string().min(1).optional(),
    resolve: ({ input, ctx }) => {
    },
  })

await client.query('hello'); // will throw 
```

simonedelmann:
But why? `.optional()` does allow a value of undefined. So if we have no input client-side, why can't we call the parser with `parser(undefined)`?

KATT:
> But why? `.optional()` does allow a value of undefined. So if we have no input client-side, why can't we call the parser with `parser(undefined)`?

In a batched query it would have an array, so `[undefined]` would become `['']` according to your suggestion which would fail the validator on `.min(1)`.

simonedelmann:
Hm, okay... maybe I haven't understood batched queries. 🙈
For a batched query we are sending an array of inputs, for example `['world', 'foo', 'I don't know']`, right? But shouldn't we validate each of these strings against `z.string().min(1).optional()` and not the whole array?
 
Still the idea of masking `undefined` as an empty string `''` does not work for a batched query. We would not know if it is an empty string or undefined.

Just to make sure again, because I think I have misunderstood the issue: The problem is not sending `null` or `undefined` in a normal query but only sending `[undefined]`, right?

KATT:
> Just to make sure again, because I think I have misunderstood the issue: The problem is not sending `null` or `undefined` in a normal query but only sending `[undefined]`, right?

Correct! It's only an issue when batching - when we're not batching we're omitting `?input`

simonedelmann:
So the problem actually is that when batching is activated, there is no more difference between `null` and `undefined`. I am not sure if it is a good idea to enable batching by default if this could change the behavior of an app. 

Could we maybe change the batched query payload to an object? 
```
// Current behaviour
const inputs = ['world', null, undefined]
JSON.stringify(inputs)
// ['world',null,null]

// Better?
const inputs = {
  0: 'world', 
  1: null, 
  2: undefined, 
  length: 3
}
JSON.stringify(inputs)
// {"0":"world","1":null,"length":3}
```

With keys similar to an arrays index and an extra `length` parameter, we could fill all missing indexes with `undefined`.

Could that be an option?

KATT:
That's a great idea actually.

KATT:
I love the outside-the-box thinking.

I think this would be an acceptable approach even if it deviates even further from "standards" of REST-esque comms, but since tRPC isn't primarily for public APIs, I think this would be great!

KATT:
@simonedelmann this is now updated according to your idea!

KATT:
@all-contributors add @simonedelmann for reviews

PR #670, opened 2021-07-19
Batching as default behavior

KATT:
This PR changes so the default behavior of tRPC is to batch requests.

- On the server, batching is enabled unless explicitly 
- On the client, the default "ending link" will be `httpBatchLink`

---

Thoughts? Should this be regarded as a "power feature" or something that should be enabled by default?

---

**Update - docs! 📚**

  - https://www-git-feature-batching-as-default.tmp.trpc.io/docs/links
  - Only gotten a resounding "yes" from people who have responded so far


peterp:
I think it makes a ton of sense to make this the default.

Modern apps are very chatty and individual React components are often "thrown together on a page" without the consideration about what kind of network requests each component is doing, this is because of the way React component are architected where each piece of functionality is colocated within the component.

KATT:
Thanks @peterp 

Opting out can be done by a split link, example: 

```tsx
import { createTRPCClient } from '@trpc/client';
import { httpBatchLink } from '@trpc/client/links/httpBatchLink';
import { httpLink } from '@trpc/client/links/httpLink';
import { splitLink } from '@trpc/client/links/splitLink';
import AbortController from 'abort-controller';
import ws from 'ws';
import type { AppRouter } from './server';

// polyfill fetch & websocket
global.AbortController = AbortController;
global.fetch = fetch as any;
global.WebSocket = ws as any;

async function main() {
  const client = createTRPCClient<AppRouter>({
    links: [
      // call subscriptions through websockets and the rest over http
      splitLink({
        condition(op) {
          return !!op.context.noBatch;
        },
        left: httpBatchLink({
          url: `http://localhost:2022`,
        }),
        right: httpLink({
          url: `http://localhost:2022`,
        }),
      }),
    ],
  });

  // will be with batching
  const res1 = await client.query('hello');
  // skip batching
  const res2 = await client.query('hello', null, {
    context: {
      noBatch: true,
    },
  });

  console.log({ res1, res2 });
}

main();

```

PR #697, opened 2021-07-22
fix inference of unset input

### LINKED ISSUE #672: Client does not enforce 'no input' when an endpoint declared without 'input'
Example

`
const appRouter = trpc
  .router()
  .query('getData', {
    async resolve() {
      return data;
    },
  })
`

`
 await api.query('getData', 'some unneeded input');
`

Runtime error - 'no input expected'.

Expected Behaviour:
Client should enforce no input on using the compiler.

### LINKED ISSUE #698: `useQuery()`: `context` option not passed to links

KATT:
closes #672
closes #698

### DIFF
--- a/packages/server/src/router.ts
+++ b/packages/server/src/router.ts
@@ -62,10 +62,12 @@ function getDataTransformer(
 
 export type inferHandlerInput<TProcedure extends Procedure> =
   TProcedure extends ProcedureWithInput<any, infer TInput, any>
-    ? undefined extends TInput
-      ? [(TInput | null | undefined)?]
-      : [TInput]
-    : [(undefined | null)?];
+    ? undefined extends TInput // ? is input optional
+      ? unknown extends TInput // ? is input unset
+        ? [(null | undefined)?] // -> there is no input
+        : [(TInput | null | undefined)?] // -> there is optional input
+      : [TInput] // -> input is required
+    : [(undefined | null)?]; // -> there is no input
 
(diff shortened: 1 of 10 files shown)


packages/server/src/router.ts
```diff
@@ -62,10 +62,12 @@ function getDataTransformer(
...
-    ? undefined extends TInput
-      ? [(TInput | null | undefined)?]
-      : [TInput]
-    : [(undefined | null)?];
+    ? undefined extends TInput // ? is input optional
+      ? unknown extends TInput // ? is input unset
+        ? [(null | undefined)?] // -> there is no input
+        : [(TInput | null | undefined)?] // -> there is optional input
+      : [TInput] // -> input is required
+    : [(undefined | null)?]; // -> there is no input
```
KATT, line 70: @danielyogel this is where the magic happens

danielyogel: @KATT Looks good to me!

From what I understand, this solution is when no input required by server, the client still expects an optional `null | undefined` input, I would change that to `[]; // -> there is no input`, but I guess this is a breaking change, and also a very very minor issue...

danielyogel (approved):
This is not an intensive and deep review, because I still learning the source code, and also very busy till this Monday.

The only two comments I have for now are the one I've added inline, and and another general question - why so many type assertions? maybe I will help fixing inference in the future.

Thanks for fixing the issue I've opened !

KATT:
@all-contributors add @danielyogel for review & bug report

PR #718, opened 2021-07-24
[RFC] Add middleware `next()`-fn

### LINKED ISSUE #564: Middleware next?
Unless I'm missing something you can't have a middleware do pre/post style flow ala koa... 

```ts
.middleware(async ({ ctx, next }) => {
  const auth = checkForAuth(ctx)
  if (!auth) return handleUnauth(ctx)
  await next()
  const update = checkForUpdatedAuth(ctx)
  if (update) {
    ctx.res.headers['x-access-token'] = update.token
  }
})
```

KATT:
Middlewares will now have to call a `next()` function that you **have to** call.

### Example

> A middleware that measures the time of procedures

```ts
.middleware(async ({ ctx, path, next }) => {
  const start = Date.now()
  await next()
  const end = Date.now()
  console.log({path, took: end - start})
})
```
---

@mmkal @delaneyj - you've both championed this. I really don't like my implementation of the recursive async/await spaghetti, I wrote it in like 5 minutes, but the tests pass and it seems to work. You're both [added as contributors](https://github.com/trpc/trpc/invitations) so you can update the branch directly, would be great if one of you could clean up my implementation.

---

### Todos

- [x] First draft
- [x] Add tests
- [x] Make a prettier implementation
- [x] Update docs
- [x] Update examples

---
Closes #564  #586 #564


www/docs/server/middlewares.md
```diff
@@ -6,7 +6,7 @@ slug: /middlewares
 ---
 
 
-You can are able to add middlewares to a whole router with the `middleware()` method. The middleware(s) will be run before any of the procedures defined after are invoked & can be async or sync.
+You can are able to add middlewares to a whole router with the `middleware()` method. The middleware(s) will be run before any of the procedures are invoked & can be async or sync.
```
peterp, line 9:
```suggestion
You are able to add middleware(s) to a whole router with the `middleware()` method. The middleware(s) will be run before any of the procedures are invoked & can be async or sync.
```

simonedelmann:
I won't have time to review the code properly in the next days, sorry.

But I have a question about the purpose of the PR: 

Currently a middleware has 2 states: success (fn returns) and failure (fn throws). 

With this PR a middleware now gets a third state: whenever neither next() will be called nor an error will be thrown. I assume an unspecific error ("Missing next() call") will be thrown then? 

I am not sure if this improves the user experience of tRPC or just makes it more complicated. You still cannot reuse express middlewares for example, because in our middlewares you don't have access to req and res. 

I am sure there is some kind of benefit in this, which I am currently missing. ;)

mmkal:
@KATT thanks for adding me as a collaborator! I do have some untested ideas about how to implement this, as well as how to ensure that each middleware calls `next` exactly once at runtime and at the type-level. I was thinking of an approach using `.reduce`, but I will take a closer look at what you have here at some point this weekend. One note: I think it's important that the `next()` call _always_ has an `await` or `return`, since downstream middleware might be async.

@simonedelmann I requested this because it can enable a bunch of use cases:

- logging procedure calls in a standardized way to stdout
- automatically sending timing metrics to tools like datadog
- sending any thrown errors to sentry/other bug tracking tools
- adding context to [AsyncLocalStorage](https://nodejs.org/api/async_context.html)
- possibly even transforming responses, though there may be better ways to do that already

With the right design, each of these could be implemented as a separate middleware function, and any combination of them would work together.

But yes, these wouldn't be 1-1 with express middleware. IMO that's ok because express middleware are kinda badly designed, using callbacks not promises. Koa has some pretty good docs on why. This design seems closer to koa.

simonedelmann:
Thanks @mmkal ! Totally convinced ;)

KATT:
> One note: I think it's important that the `next()` call _always_ has an `await` or `return`, since downstream middleware might be async.

My implementation was that middlewares conform to `() => Promise<void> | void` - is this what you mean? 


> * logging procedure calls in a standardized way to stdout

💯 

> * automatically sending timing metrics to tools like datadog

💯 

> * sending any thrown errors to sentry/other bug tracking tools

This is not the role of the type of middleware implementation I thought and rather something I'd use [error reporting](https://trpc.io/docs/error-handling) for.

> * adding context to [AsyncLocalStorage](https://nodejs.org/api/async_context.html)

💯 

> * possibly even transforming responses, though there may be better ways to do that already

This should not be done in middlewares as far as I can tell and there's already support for this with [data transformers](https://trpc.io/docs/data-transformers).

A request can contain N requests to the same procedure which would all be passed through middlewares.

What you are looking for here is probably some over-arching middleware support on the request handler, so responses can be intersected before getting passed through `res.send()`. I've actually thought of a little refactoring here in order to be able to support [SvelteKit endpints](https://kit.svelte.dev/docs#routing-endpoints)


> But yes, these wouldn't be 1-1 with express middleware. IMO that's ok because express middleware are kinda badly designed, using callbacks not promises. Koa has some pretty good docs on why. This design seems closer to koa.

💯 This is also why I was okay with having the existing middleware implementation not having a `next()`-function, as the definition of a middleware is a bit wider than what we're used to in Node.js-land.

KATT:
@mmkal (when) do you think you will have a stab at this?

mmkal:
@KATT I've set aside some time on Wednesday. If you're ok to wait until then? If not totally your call of course!

mmkal:
@KATT I pushed a change. I squashed into a single commit so it's easy to revert if you don't like the implementation. I can also break this out into a separate branch if you'd like to do a side-by-side comparison.

The key change is: middlewares now have to pass on the result from their `next()` function. Doing this means we can rely on typescript for a lot more correctness-checking. `this.resolver` gets wrapped to become the last middleware, effectively - the one that provides the actual `output`.

We don't really need as much bookkeeping, like try/catch wrappers or to check for calling `next()` twice anymore, because doing that would already stick out as an error:

```ts
router
  .middleware(async ({ next }) => {
    const first = await next()
    const second = await next()

    // which one should be returned??
  })
```

I also added a test using `async_hooks` and it works as expected.

I also made middleware `next` functions _always_ return a Promise, since they can't know whether downstream middlewares are sync or async.

I updated the docs that I found that used `.middleware`, but please let me know if there are others that I missed.

___

Re error reporting, a pattern I've seen used before (admittedly for a probably overcomplicated architecture) was to throw a "recoverable" error - you could imagine a middleware that could do something like:

```ts
router
  .middleware(async ({ ctx, path, type, next }) => {
    try {
      return next()
    } catch (e) {
      if (e instanceof RecoverableError) {
        // maybe service a failed, but we can fall back on service b which is slower/less ideal in some way
        myErrorLogger.log(e, {ctx, path, type})
        return e.recover()
      }
      throw e
    }
  })
```

This should now be doable, although I think it's reasonable to let users figure out for themselves.

KATT:
@all-contributors add @mmkal for tests and code

KATT:
@mmkal this looks really impressive. I will give it a proper look tomorrow and get it in. Will probably start a v9-branch as there are other things I want to do that are also breaking changes (especially to do with supporting Svelte).

packages/server/src/router.ts
```diff
@@ -150,11 +150,30 @@ const defaultTransformer: CombinedDataTransformer = {
...
+export interface MiddlewareFunctionOpts<TContext> {
   ctx: TContext;
   type: ProcedureType;
   path: string;
-}) => Promise<void> | void;
+  input: unknown;
+}
+export type MiddlewareNextFn = () => Promise<MiddlewareResult>;
+
+export type MiddlewareFunction<TContext> = (
```
KATT, line 171: Is there any reason to actually export these? 🤔

mmkal: No, you're right, these can be reduced down. The only type we really need to import elsewhere is `MiddlewareFunction`, I'll un-export the rest.

www/docs/server/middlewares.md
```diff
@@ -47,3 +46,29 @@ trpc
...
+
+```ts
+trpc
+  .router<Context>()
+  .middleware(async ({ path, type, next }) => {
+    const start = Date.now()
+    const result = await next()
+    console.log('request timing:', {path, type, took: Date.now() - start})
+
+    return result
```
KATT, line 62:
```suggestion
    const start = Date.now();
    const result = await next();
    console.log('request timing:', {path, type, took: Date.now() - start});
    return result;
```

packages/server/src/procedure.ts
```diff
@@ -99,12 +100,25 @@ export abstract class Procedure<
     type,
     path,
   }: ProcedureCallOptions<TContext>): Promise<TOutput> {
-    for (const middleware of this.middlewares) {
-      await middleware({ ctx, type, path });
-    }
-    const input = this.parseInput(rawInput);
-    const output = await this.resolver({ ctx, input, type });
-    return output;
+    const opts = { ctx, type, path, input: this.parseInput(rawInput) };
```
KATT, line 103: I'm unsure about this --

Previously, middlewares had knowledge about `input` and we wouldn't even _get to_ the input parser unless all the middlewares had been run; was this an intentional change of yours or an unintentional side-effect? 

I was intentional with the previous behavior which made sure that nothing in the procedures were called if all middlewares didn't run first.

mmkal: It was intentional, or at least it was needed for `opts` to be passed to `this.resolver`. I like it though, partly for consistency - the `this.resolver` function expects an `input` prop, and is now effectively acting like a middleware. It's a little confusing for middlewares to receive everything _except_ input. But also it makes sense to me for parsing to happen first. Middleware could have side effects and it's helpful for trpc to filter out bad requests before they get to the middleware.

Example use case: a [fhir server](https://www.hl7.org/fhir) of some kind. On a previous team we used codegen to create io-ts codecs for every fhir resource. Something like that could be extended and plugged into trpc, and a middleware could log statistics per fhir resource type. For that to happen we'd need either `input` or `rawInput`, but it's more useful to pass `input`, especially since the parsing step might do some hydration of iso strings to dates, etc.

>I was intentional with the previous behavior which made sure that nothing in the procedures were called if all middlewares didn't run first.

In the new model, where middlewares are enclosing rather than preceding the procedure part, does that still apply?

I'm struggling to think of a use case where I'd want the middleware to still run on invalid input (in the world of http status codes for a regular server, I'd probably expect a malformed request which also had an invalid token to give me a 400, not a 401 or 403).

KATT: Thanks for spending so much time thinking on this! 🙏 

Why I think it should be the other way around:

- Intuitive: a middleware is declared before a resolver, hence, should always be called first
- The input parser could later be turned into a potentially expensive async method as well (think admin routes)
- The input parser could leak information about the internals of a protected route (again, think admin routes)

---

If you think we need to have `rawInput` (or `input`) in middlewares we can add that at a later stage when we need it; it's much easier to add code than to delete it. Potentially it could be returned from the `next` function.

---

Another thought - rather than dealing with `Promise` and `try/catch`, how about we return a discriminated union similar to [zod's `safeParse`](https://github.com/colinhacks/zod#safeparse)?

```ts
router
  .middleware(async ({ ctx, path, type, next }) => {
      const result = await next();
     if (!result.ok && result.error.originalError instanceof RecoverableError) {
        return result.e.recover();
     }
     return result;
  })
```
The interface could look like this

```
import { ProcedureType } from '../router';
import { TRPCError } from '../TRPCError';

export const middlewareMarker = Symbol('middlewareMarker');
interface MiddlewareResultBase {
  /**
   * All middlewares should pass through their `next()`'s output.
   * Requiring this marker makes sure that can't be forgotten at compile-time.
   */
  readonly marker: typeof middlewareMarker;
}

interface MiddlewareOKResult extends MiddlewareResultBase {
  ok: true;
  data: unknown;
  // this could be extended with `input`/`rawInput` later
}
interface MiddlewareErrorResult extends MiddlewareResultBase {
  ok: false;
  error: TRPCError; // we could guarantee it's always of this type
  // this could be extended with `input`/`rawInput` later
}

type MiddlewareResult = MiddlewareOKResult | MiddlewareErrorResult;

export type MiddlewareFunction<TContext> = (opts: {
  ctx: TContext;
  type: ProcedureType;
  path: string;
  next: () => Promise<MiddlewareResult>;
}) => Promise<MiddlewareResult>;
```

KATT: See 31efba0 suggestion

mmkal: >If you think we need to have rawInput (or input) in middlewares we can add that at a later stage when we need it

👍👍👍 Very good point, I'm absolutely ok with leaving out for now and possibly forever.

___

Re the discriminated union - will reply inline.

packages/server/src/internals/middlewares.ts
```diff
@@ -0,0 +1,31 @@
...
+   * All middlewares should pass through their `next()`'s output.
+   * Requiring this marker makes sure that can't be forgotten at compile-time.
+   */
+  readonly marker: typeof middlewareMarker;
+}
+
+interface MiddlewareOKResult extends MiddlewareResultBase {
+  ok: true;
+  data: unknown;
+  // this could be extended with `input`/`rawInput` later
```
mmkal, line 16: `input` would be on the `opts` of `MiddlewareFunction`, not here. Tbh I was convinced by your argument and don't think we even need this comment.

packages/server/src/internals/middlewares.ts
```diff
@@ -0,0 +1,31 @@
...
+export const middlewareMarker = Symbol('middlewareMarker');
+interface MiddlewareResultBase {
+  /**
+   * All middlewares should pass through their `next()`'s output.
+   * Requiring this marker makes sure that can't be forgotten at compile-time.
+   */
+  readonly marker: typeof middlewareMarker;
+}
+
+interface MiddlewareOKResult extends MiddlewareResultBase {
```
mmkal, line 13: Re the discriminated union idea:

I'm totally open to it, but as a user it might confuse me a little. If my middleware detects an error state, should I `throw ...` or should I return `{ ok: false, error: ... }`? And whatever the "correct" answer is, we can be sure that someone on my team will do it the other way anyway. I _personally_ prefer functional-style and pure code without `throw`s. But `throw` is a language feature that isn't going anywhere, so adding an error type to the middleware API increases the cognitive overhead. We can never get rid of the possibility that a middleware function might throw.

If it's for the sake of the "recoverable" error scenario, IMHO we could keep it simple and save for a future change if necessary (a nice thing about this implementation is it's backwards compatible). Especially since users are already able to "recover" via `catch`.

packages/server/src/procedure.ts
```diff
@@ -99,12 +101,49 @@ export abstract class Procedure<
...
+        try {
+          const input = this.parseInput(rawInput);
+          const data = await this.resolver({ ...opts, input });
+          return {
+            marker: middlewareMarker,
+            ok: true,
+            data,
+          };
+        } catch (_error) {
+          const error = getErrorFromUnknown(_error);
```
mmkal, line 119: Without the discriminated union thing, we wouldn't need this try/catch (only to be rethrown later), is that right? Implementation simplicity isn't the _most_ important consideration vs feature set, but maybe a factor.

mmkal (commented):
A couple of comments about the discriminated union, but I'm 👍 on this PR either way, just my $0.02!

packages/server/src/internals/middlewares.ts
KATT: I reverted the discriminated union, but I've not convinced myself yet, will let it simmer a bit; I'm not in a rush to get this in and I want to stop releasing so many majors.

mmkal: What if we set it up so that a breaking change wasn't required, maybe by refactoring `middlewareMarker` to `middlewareSuccessMarker`, leaving us open to adding a `middlewareFailureMarker` later, so the types now would be:

```ts
export const middlewareSuccess = Symbol('middlewareSuccess')

export interface MiddlewareResult {
  readonly marker: typeof middlewareSuccess
  data: unknown
}
```

And if you/the community decide to add in failure return values too, it could become:

```ts
export const middlewareSuccess = Symbol('middlewareSuccess')
export const middlewareFailure = Symbol('middlewareFailure')

interface MiddlewareSuccessResult {
  readonly marker: typeof middlewareSuccess
  data: unknown
}

interface MiddlewareFailureResult {
  readonly marker: typeof middlewareFailure
  error: TRPCError
}

export type MiddlewareResult = MiddlewareSuccessResult | MiddlewareFailureResult
```

Which is a non-breaking change, because all the middlewares out there in the wild would be returning `{ marker: middlewareSuccess, data: ... }` which is still valid. So wouldn't require a major bump.

Plus, I think it's marginally better having both a `marker` and an `ok` when only one is needed.

mmkal (approved):
Realised I didn't explicitly approve - @KATT let me know if there's any more you'd like me to look at but this LGTM!

I'd love to use this ASAP as I happen to have a project that won't be in production for a couple of months - if you want to hold off to collect more feedback would you be open to publishing a `9.0.0-next.1` prerelease or something?

mmkal:
@KATT I tried this and it seems to work very nicely!

anthonyshort:
My team is excited to use this, it'll clean up the logging wrappers we have around trpc :) Thanks for putting in the work.

KATT:
@anthonyshort Awesome! 

Do you have any opinion on the discussion regarding `next()` _throwing_ errors versus returning a `Promise<MiddlewareOKResult | MiddlewareErrorResult>`? 

See https://github.com/trpc/trpc/pull/718#discussion_r678935636 for discussion

anthonyshort:
> @anthonyshort Awesome!
> 
> Do you have any opinion on the discussion regarding `next()` _throwing_ errors versus returning a `Promise<MiddlewareOKResult | MiddlewareErrorResult>`?
> 
> See [#718 (comment)](https://github.com/trpc/trpc/pull/718#discussion_r678935636) for discussion

I prefer the union approach. It decouples the concept of an error and a result and prevents us from using try/catch for control flow. In the logging case, we'd be try/catching to determine what to log, which feels weird. And what happens if they don't rethrow the error?

While it might feel confusing to people at first I think it's the cleaner approach.

www/docs/server/middlewares.md
```diff
@@ -46,3 +46,28 @@ trpc
...
+
+In the example below timings for queries are logged automatically.
+
+```ts
+trpc
+  .router<Context>()
+  .middleware(async ({ path, type, next }) => {
+    const start = Date.now();
+    const result = await next();
+    console.log('request timing:', {path, type, took: Date.now() - start});
```
mmkal, line 60: I think the union thing makes this wrong. We now need to check if `ok` is true because otherwise we'll mess up our metrics with fast-throwing errors. Or if not wrong, then unexpected. I think most users would want to think about timing separately from errors.

packages/server/src/internals/middlewares.ts
```diff
@@ -0,0 +1,31 @@
+import { ProcedureType } from '../router';
+import { TRPCError } from '../TRPCError';
+
+export const middlewareMarker = Symbol('middlewareMarker');
+interface MiddlewareResultBase {
+  /**
+   * All middlewares should pass through their `next()`'s output.
+   * Requiring this marker makes sure that can't be forgotten at compile-time.
+   */
+  readonly marker: typeof middlewareMarker;
```
mmkal, line 10: If we're doing the union approach, maybe we don't need the `marker` prop. Just seems like overkill to have two "tag" type props.

mmkal (commented):
I can absolutely work with this - but I do stand by my earlier comments. As implemented, as far as I can tell:

For end users: this _doesn't_ give users the ability to return `{ ok: false, error: ... }` (we still need to `throw`), but it _does_ force them to add failure checks to virtually every middleware, not just ones that care about errors (see my comment on the timing logger middleware example in markdown).

For maintainers: more of a judgement call, but the code is now a bit more complex, so more surface area for bugs, harder to debug, etc.

Overall if it were my decision I would still opt against the union. It feels like it's going ~25% of the way to making this specific part of trpc functional-ish. Going down this route and covering all the edge cases, we might end up with a less battle-tested reimplementation of fp-ts inside this library (or at least `TaskEither`). So if it were me, I'd probably decide to stop worrying and love `catch`. But 1) I might have misunderstood something here and 2) maybe there aren't _that_ many edge cases.

www/docs/server/middlewares.md
KATT: Yeah, docs should probably be updated to reflect this. 


```suggestion
    const durationMs = Date.now() - start;
    res.ok 
      ? console.log('OK request timing:', {path, type, durationMs })
      : console.log('Non-OK request timing', {path, type, durationMs })
    
      
```

mmkal:
@KATT just to check - are you waiting on any more feedback from me on this? We're happily on the `9.0.0-alpha.0` version for now but don't want to miss out on other features!

Also re the conflicts, in case this is helpful, I use these commands sometimes to publish branch-specific canary releases from lerna without editing the actual package.json files/creating a commit. The `dist-tag` is the branch name. It generates a long, ugly version string including the branch, commit  date and hash, so you might want to tweak/YMMV:

```
    "commit-date": "git log -n 1 --date=format:'%Y-%m-%d-%H-%M-%S' --pretty=format:'%ad'",
    "canary-preid": "echo \"$(yarn --silent current-branch)-$(yarn --silent commit-date)\"",
    "current-branch": "echo \"$(git rev-parse --abbrev-ref HEAD)\" | sed -E 's/refs\\/heads\\///' | sed -E 's/\\W|_/-/g'",
    "publish-canary": "lerna publish --canary --preid $(yarn --silent canary-preid) --dist-tag $(yarn --silent current-branch)",
```

Sample `yarn publish-canary` output in my repo:

```
Found 1 package to publish:
 - @slonik/migrator => 0.8.5-lock-transaction-fix-2021-08-11-14-47-52.6+c9cb3ff

? Are you sure you want to publish these packages? (ynH) 
```

KATT:
I'm happy with this PR.

Why it's not merged is that I'm spending some time finding other things that requires some breaking changes here and there too and doing them as well.

See the v9 milestone. Also, if you've noticed any other structural stuff that you think we should change, this is a good opportunity to do so ☺

PR #730, opened 2021-07-28
Next.js handling proper http status codes

KATT:
This PR enables Next.js to send _actual_ `404` status codes for instance when doing SSR.

- Exposes `ssrContext` on components when server rendering
- Exposes `httpStatus` as part of error object as the default error shape

For a reference on how to use this you can do a component like the following: 

```tsx
import { TRPCClientError } from '@trpc/client';
import type { AppRouter } from 'server/routers/app';
import { trpc } from 'utils/trpc';
import NextError from 'next/error';

/**
 * Takes a TRPCClientError and renders a `next/error`-page
 * (this component might be added to `@trpc/next`)
 */
export function ErrorPage(props: {
  error: TRPCClientError<AppRouter>;
}): JSX.Element {
  const { error } = props;

  const statusCode = error.shape?.data.httpStatus ?? 500;
  if (!process.browser) {
    const res = utils.ssrContext?.res;
    if (res) {
      res.statusCode = statusCode;
    }
  }

  return (
    <NextError
      title={error.shape?.message ?? error.message}
      statusCode={statusCode}
    />
  );
}
```


simonedelmann (approved):
Looks good to me!

KATT:
Ignore that I merged this - reverted on `main` as I did it preemptively - see #811

PR #761, opened 2021-08-05
Add default server error handler as a fallback

### LINKED ISSUE #384: Log internal server errors during development by default
Currently, I’m using the following code snippet for Next.js at the `api/trpc/[trpc]` route:

```ts
import * as trpcNext from "@trpc/server/adapters/next";

const __DEV__ = process.env.NODE_ENV !== "production";

  router: appRouter,
  createContext,
  onError: __DEV__
    ? ({ error, type, path, input }) => {
        if (error.code === "INTERNAL_SERVER_ERROR") {
          // eslint-disable-next-line no-console
          console.error("Error:", {
            type,
            path,
            input,
            cause: error.originalError,
          });
        }
      }
    : undefined,
});
```

I think it would be useful to always log unhandled errors (where `error.code === "INTERNAL_SERVER_ERROR"`) to the server’s console like this during development, even when no `onError` is specified.

kripod:
Resolves #384.

As it turns out, errors should also be logged in production by default, so when a log drain is in use, no additional setup is required.

### DIFF
--- a/packages/server/src/http/requestHandler.ts
+++ b/packages/server/src/http/requestHandler.ts
@@ -84,8 +84,24 @@ export async function requestHandler<
     createContext: TCreateContextFn;
   } & BaseHandlerOptions<TRouter, TRequest>,
 ) {
-  const { req, res, createContext, teardown, onError, maxBodySize, router } =
-    opts;
+  const {
+    req,
+    res,
+    createContext,
+    teardown,
+    onError = ({ error, type, path, input }) => {
+      if (error.code === 'INTERNAL_SERVER_ERROR') {
+        console.error('API error:', {
+          type,
+          path,
+          input,
+          cause: error.originalError,
+        });
+      }
+    },
+    maxBodySize,
+    router,
+  } = opts;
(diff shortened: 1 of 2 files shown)


KATT (commented):
Unsure other devs want this as a default.

@anthonyshort / @peterp - what's your take on a default error logger?

anthonyshort:
I think it makes sense to have default logging.

However, I'm not sure what the plan is for default logging behaviour during dev or production would be, but if there was a bigger plan for it then maybe this small change to just `console.log` might not fit into that. 

For example, we've got a custom setup so that it looks like this during development:

image

Maybe if there was a really nice default logger that we use during development, we'd use the same one in production too, then this PR wouldn't make as much sense.

KATT:
Agree with @anthonyshort. 

I don't want tRPC to have an opinion on how/what logging you do or what logging lib you use (I like sticking to zero deps too).

I think it's enough [for now] that the default example is using an `onError`-callback; #718 will also make this easier to do on a more granular per-procedure basis.

KATT:
Thanks for the time spent on this PR, @kripod!

PR #811, opened 2021-08-20
Improve handling of HTTP statuses & headers

KATT:
This PR enables Next.js to send _actual_ `404` status codes for instance when doing SSR.

- Exposes `httpStatus` as part of error object as the default error shape
- `httpStatus` will be used to determine the actual HTTP Status code set
- Expose `TRPCClientError.data` as a shortcut to `error.shape.data` as that's the only one I find myself using
- Stricter `ErrorFormatter` - `data` has to be a `Record<string, unknown>`
- `TRPCError` we infer in `@trpc/react` is now `TRPCErrorLike` and is serialized for SSR in `@trpc/next`
- add new method `responseMeta()` that is triggered before a request ends - both as part of `@trpc/next` & `@trpc/server`
  - on `@trpc/server` receives `type`, `paths`, `errors`, and the unserialized `data[]` that is going out
  - on `@trpc/next` it client errors

**See docs at 👉**  https://www-git-feature-error-example.tmp.trpc.io/

---

### Next.js error example

For a reference on how to use this you can do a component like the following: 

#### In `withTRPC()`

```tsx

export default withTRPC<AppRouter>({
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
  config() { /* ... */ }
  /**
   * @link https://trpc.io/docs/ssr
   */
  ssr: true,
  /**
   * Set headers or status code when doing SSR
   */
  responseMeta({ clientErrors }) {
    if (clientErrors.length) {
      // grab the first error's httpStatus
      return {
        status: clientErrors[0].data?.httpStatus ?? 500,
      };
    }

    return {};
  },
})(MyApp);
```

#### In page

```tsx
import { useRouter } from 'next/dist/client/router';
import { trpc } from 'utils/trpc';
import NextError from 'next/error';

export default function PostViewPage() {
  const id = useRouter().query.id as string;
  const postQuery = trpc.useQuery(['post.byId', id]);
  if (postQuery.error) {
    const statusCode = postQuery.error.data?.httpStatus ?? 500;
    return (
      <NextError title={postQuery.error.message} statusCode={statusCode} />
    );
  }
  if (postQuery.status === 'loading') {
    return <>Loading...</>;
  }
  return (
    <>
      <h1>{postQuery.data?.title}</h1>

      <h2>Raw data:</h2>
      <pre>{JSON.stringify(postQuery.data ?? null, null, 4)}</pre>
    </>
  );
}
```


www/docs/server/error-formatting.md
```diff
@@ -41,10 +41,10 @@ function MyComponent() {
     mutation.mutate({ title: 'example' });
   }, []);
 
-  if (mutation.error?.shape?.data.zodError) {
+  if (mutation.error?.data.zodError) {
```
KATT, line 44:
```suggestion
  if (mutation.error?.data?.zodError) {
```

packages/server/test/errors.test.ts
```diff
@@ -220,6 +255,43 @@ describe('formatError()', () => {
...
+        },
+      },
+    );
+    const res = await fetch(`${httpUrl}/q`);
+
+    expect(res.ok).toBeFalsy();
+    expect(res.status).toBe(TEAPOT_ERROR_CODE);
+
+    close();
+  });
```
KATT, line 294: cc @mmkal re https://github.com/trpc/trpc/discussions/809

cyrus-za (approved):
Thanks for this. Can't wait to use it

PR #814, opened 2021-08-21
Add `beforeEnd:`-function on HTTP response handlers

KATT:
- add new method `beforeEnd()` that is triggered before a request ends
- receives `type`, `paths`, and the unserialized `data` that is going out
- See docs at 👉  https://www-git-feature-before-send.tmp.trpc.io/docs/caching#api-response-caching

### Questions for the reviewers

- unsure about the naming `beforeEnd`
- considering changing it to be `getResponseHeaders`- this will work better once we add svelte support as in Svelte the is no `res.headers` / `res.statusCode` to mutate. See alternative in #816
- ~unsure if it should actually live on the router - maybe it's better to put it as `beforeEnd:` on the API handler, similar to `createContext`? (i.e. the `[trpc].ts`-file)~ I think it's better to do it in api handler - refactored in #815

### DIFF
--- a/examples/next-prisma-starter/src/pages/api/trpc/[trpc].ts
+++ b/examples/next-prisma-starter/src/pages/api/trpc/[trpc].ts
@@ -26,4 +26,10 @@ export default trpcNext.createNextApiHandler({
   batching: {
     enabled: true,
   },
+  /**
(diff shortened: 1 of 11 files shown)


KATT:
Closing in favor of #816

PR #816, opened 2021-08-21
Add `responseMeta`-fn on HTTP response handlers

KATT:
- add new method `responseMeta()` that is triggered before a request ends - both as part of `@trpc/next` & `@trpc/server`
- on `@trpc/server` receives `type`, `paths`, and the unserialized `data[]` that is going out
- on `@trpc/next` receives all client errors
- See docs at 👉  https://www-git-feature-response-headers.tmp.trpc.io/docs/caching

### DIFF
--- a/www/docs/server/caching.md
+++ b/www/docs/server/caching.md
@@ -60,7 +60,7 @@ export default withTRPC({
 
 Since all queries are normal HTTP `GET`s we can use normal HTTP headers to cache responses, make the responses snappy, give your database a rest, and easier scale your API to gazillions of users.
 
-### Using `beforeEnd()` to cache responses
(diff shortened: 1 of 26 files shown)


www/docs/server/caching.md
```diff
@@ -60,7 +60,9 @@ export default withTRPC({
 
 Since all queries are normal HTTP `GET`s we can use normal HTTP headers to cache responses, make the responses snappy, give your database a rest, and easier scale your API to gazillions of users.
 
-### Example code
+### Using `beforeEnd` to cache responses
```
KATT, line 63:
```suggestion
### Using `getResponseHeaders ` to cache responses
```

KATT:
Merging to #811

Issue #810, opened 2021-08-20
tRPC v9 feedback before marking it as `9.0.0`


KATT:
I've just merged the #732 & #718 to `main` and released a new V9 draft. 

**Are there any other structural, potentially breaking changes that you would want to get into the next major?**

It should be production-ready so feel free to update your deps to `9.0.0-alpha.2`, play around with it, and give me feedback.

I have tried slowing down on the major version bumps, so I would ideally not do more breaking changes for a while.

Planning to release `9.0.0` early next week.

## Summary of changes so far

- Middlewares will now have to call a `next()` function that you **have to** call.
- Reduce `TRPCError`s - `METHOD_NOT_FOUND` removed & `PATH_NOT_FOUND` renamed to `NOT_FOUND` (#733)
- Align input and output serialization (#746)
- Remove deprecated `httpErrors`-helper 
- Remove support for `?input` to be an array - now assuming a `Record<number, unknown>`-style dict (changed with backward compat in #669)
- Remove support for `POST` calls being `{ input: x }` (originally from #671)

simonedelmann:
Great work, @KATT ! I am looking forward to update my app, when I find some time.

You asked about other things which could be changed for the next major and I think we could use the opportunity to improve the transformer syntax. (my favorite topic 😉 ) I am currently on vacation so I here is just a small idea. Feel free to implement, improve or forget... whatever you like.

For transformers currently we have the simple syntax, which is fine, and the combined syntax, which I still feel as a compromise to enable multiple transformers. (My fault, I made that PR back then.... 🙈)
```ts
const transformer = {
  serialize: (value) => value,
  deserialize: (value) => value,
}
const combinedTransformer = {
  input: transformer,
  output: transformer,
}
```

I was thinking about making transformers more generic and stackable. What if we have a stack of transformers, which is applied depending on some flags. 

```
Currently: User -> Input Transformer -> Server -> Output Transformer -> User

Suggestion: User -> Transformer A -> Transformer C -> Server -> Transformer C -> Transformer B -> User
```

A transformer stack could look like that:
```
const transformers = [
  {
    transform: (val) => someFancyOptimization(val),
    // if `resolve` is not defined, don't do anything for "deserializing"
    apply: OUTPUT,
  },
  {
    transform: (val) => escapeHtmlCharsForSecurityReasons(val),
    apply: INPUT,
  },
  {
    transform: (val) => superjson.serialize(val),
    resolve: (val) => superjson.deserialize(val),
[...]

PR #818, opened 2021-08-22
infer new context in middlewares

KATT:
- attempt to have a new middleware type that can infer the context - called `middlewareSwapContext` here
- managed to get the inference working (ish) with the simpler `swapContext` one
- ideally i'd just have `middleware` and if one would call `next({ctx})` it'd be inferred (kinda where I'm going with `middlewareSwapContext`

> I probably won't spend much more time on this as it's kinda a nice to have and has some other side effects, but if someone wants to pick it up and have a stab at it, please have a go

requested by @mmkal  in https://github.com/trpc/trpc/issues/810#issuecomment-902930210

### DIFF
--- a/packages/server/src/internals/middlewares.ts
+++ b/packages/server/src/internals/middlewares.ts
@@ -2,30 +2,61 @@ import { ProcedureType } from '../router';
 import { TRPCError } from '../TRPCError';
 
 export const middlewareMarker = Symbol('middlewareMarker');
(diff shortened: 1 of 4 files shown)


mmkal:
@KATT I have this mostly working on my local branch - ok if I push it tomorrow?

KATT:
Replaced by #915

PR #857, opened 2021-08-31
Fix prefetching infinite query

hypnodron:
I've encountered an issue with prefetching infinite queries where all page contents were replaced by the content of the first page while the number of pages remained the same. I narrowed it down to a `pageParam` being ignored when prefetching and fetching infinite queries.

The PR fixes the issue and adds tests.


KATT (approved):
Amazing work @hypnodron - thank you so much!

PR #871, opened 2021-09-03
Allow createContext to be optional if there's no context

### LINKED ISSUE #218: Make `createContext` optional
Default to `()=>({})`

Simplifies use cases where it isn't required. e.g if the user is passing a token in the input.

joaopaulobdac:
With this PR we can omit the `createContext` prop if there's no context. The prop becomes required again if a context type is specified.

---

Closes #218

### DIFF
--- a/packages/server/test/_testHelpers.ts
+++ b/packages/server/test/_testHelpers.ts
@@ -53,7 +53,7 @@ export function routerToServerAndClient<TRouter extends AnyRouter>(
     wss,
     router,
     createContext: ({ req, res }) => ({ req, res }),
-    ...(opts?.wssServer ?? {}),
+    ...(opts?.wssServer as any ?? {}),
   };
   const wssHandler = applyWSSHandler(applyWSSHandlerOpts);
   const wssUrl = `ws://localhost:${wssPort}`;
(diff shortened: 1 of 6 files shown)


packages/server/test/_testHelpers.ts
```diff
@@ -53,7 +53,7 @@ export function routerToServerAndClient<TRouter extends AnyRouter>(
     wss,
     router,
     createContext: ({ req, res }) => ({ req, res }),
-    ...(opts?.wssServer ?? {}),
+    ...(opts?.wssServer as any ?? {}),
```
joaopaulobdac, line 56: I had to type this as `any` because TS complains about the optional `createContext` prop coming from `Partial<WSSHandlerOptions<TRouter>>` (at least I think that's what it is) although the error doesn't happen with `Partial<CreateHttpHandlerOptions<TRouter>>` for some reason :thinking:

KATT:
@joaopaulobdac you have an invitation to join as a contributor so you don't have to push forks https://github.com/trpc/trpc/invitations

KATT (changes requested):
Looks good to me! Just accept the invitation and open a PR with this branch that isn't from a fork so the tests can be run properly.

https://github.com/trpc/trpc/invitations

packages/server/test/_testHelpers.ts
KATT: that's fine, i'm not sacred of the test code

PR #878, opened 2021-09-03
update `tsconfig.json` appropriate for node>=14

KATT:
- be explicit that we only support node `>=14`
- any downsides on this? i guess anyone using trpc already is on node>=14

https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping#node-14

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -10,6 +10,9 @@
     "url": "git+https://github.com/trpc/trpc.git",
     "directory": "packages/server"
   },
+  "engines": {
+    "node": ">=14"
+  },
   "eslintConfig": {
     "rules": {
       "explicit-module-boundary-types": "off"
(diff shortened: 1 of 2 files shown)


packages/server/package.json
```diff
@@ -10,6 +10,9 @@
     "url": "git+https://github.com/trpc/trpc.git",
     "directory": "packages/server"
   },
+  "engines": {
+    "node": ">=14"
```
mmkal, line 14: Most users won't be affected by this but it is technically a breaking change, is the plan to do a major release?

KATT: I actually didn't plan to - I never intended to support node `12.x` in the first place and that tsconfig was accidental. 

That said, you're right, it is a major change. I'm going to check if the lib currently even works in node 12 and then decide what to do.

KATT: Unfortunately, trpc seems to work with `12.x` which means some people might be using it so I might just merge #886 and wait to do this change until next major

KATT:
Closing in favor of #886 - we'll have to do this in next major

PR #880, opened 2021-09-04
refactor http internals

KATT:
- Expose lower-level function `resolveHTTPResponse()` to be used in place of `requestHandler()` when doing adapters
- Rename `requestRequestHandler` to `nodeHTTPRequestHandler` and move to `adapters/node-http` *
- Make naming consistent of `HTTP` rather than `Http` everywhere *
- Move standalone http server to `adapters/standalone` *
- Move websocket from `@trpc/ws` to `@trpc/adapters/ws` *

***)** with backwards compatibility and deprecation warnings

## why? 

- separation of "generic" HTTP vs `require('http')`-esque HTTP
- easier aws lambda support (#882)
- easier svelte support


packages/server/src/adapters/aws-lambda.ts
```diff
@@ -0,0 +1,10 @@
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from 'aws-lambda';
+export const lambdaHandler = async (
+  event: APIGatewayProxyEvent,
+): Promise<APIGatewayProxyResult> => {
+  const queries = JSON.stringify(event.queryStringParameters);
+  return {
+    statusCode: 200,
+    body: `Queries: ${queries}`,
+  };
+};
```
KATT, line 10: not done at all

packages/server/src/http/internals/HTTPHandlerOptions.ts
```diff
@@ -49,3 +57,13 @@ export type HTTPHandlerOptions<
          **/
         createContext: CreateContextFn<TRouter, TRequest, TResponse>;
       });
+
+export interface ResolveHTTPRequestOptions<
+  TRouter extends AnyRouter,
+  TRequest extends HTTPRequest,
+> extends HTTPHandlerOptionsBase<TRouter, TRequest> {
+  createContext: () => Promise<inferRouterContext<TRouter>>;
```
Lilja, line 65: 🤔 doesn't this mean that `createContext` can't receive any fun input? Perhaps it's possible to create a partial of a function and "load" stuff into it before sending it to `resolveHttpResponse` somehow?

KATT: No, it just means that the wrapper fn is responsible of arguments. Check `requestHandler` for reference

packages/server/src/http/internals/HTTPResponse.tsx
```diff
@@ -0,0 +1,17 @@
+import { Dict } from '@trpc/server';
+import { URLSearchParams } from 'url';
+
+export type HTTPHeaders = Dict<string | string[]>;
```
Lilja, line 4: Any reason why this isn't `Record<string, string>` instead of `{ [x: string]: string | string[] | undefined; }`?

The output typing to api-gateway hates the fact that some of these headers can be undefined.

https://i.imgur.com/iXKBtLe.png

KATT: Because it would be a lie and it's not really great practice to type anything as `Record<string, string>`

**Example:**

```ts
const obj: Record<string, string> = {
  foo: 'bar',
}


obj.bar.toLowerCase() // yay works
obj.nope.toLowerCase() // yay works (but it's actually a runtime error "Cannot read property 'toLowerCase' of undefined ")
```

KATT: When you do the gateway adapter you'll have to deal with this yourself

KATT: It's their lib that is wrong not to deal with that

Lilja: Interesting! Didn't realize how bad `Record<string, string>` could be.

KATT: Yeah, that's why I have that `Dict`-helper, it's more accurate to the truth :) 

Simply cast the headers as `any` in your adapter maybe

KATT (commented):
some notes to myself

packages/server/package.json
```diff
@@ -18,8 +18,11 @@
   "preconstruct": {
     "entrypoints": [
       "index.ts",
-      "adapters/next.ts",
       "adapters/express.ts",
+      "adapters/next.ts",
+      "adapters/node-http/index.ts",
```
KATT, line 23: maybe call it just `http` which would quite nicely reflect `require('http')`? 🤔 

```suggestion
      "adapters/http/index.ts",
````

PR #908, opened 2021-09-08
add `fetchOptions` on client runtime

KATT:


mmkal (commented):
This looks good, but I wonder if another way of getting similar capabilities without adding to the API would be to expose the `getFetch` function in the internals. Then we could do something like:

```ts
fetch: async (url, opts) => {
  const defaultFetch = getFetch()
  return defaultFetch(url, {
    ...opts,
    headers: {
      'x-foo': await getXFooHeaderFromSomewhere({
        heyCoolItCouldEvenBeBasedOnOtherHeaders: opts.headers?['x-something-else'],
        orUseTheUrlYouCanReallyDoWhatever: url,
      }),
    },
    // add whatever else is necessary here
  })
},
```

(I was thinking of finding a way of doing this either by reimplementing something like `getFetch` or reaching into trpc internals, before I saw #909, as a way of adding headers asynchronously)

mmkal:
@KATT not sure if this has been released already, but any thoughts on the comment above?

KATT:
@mmkal for that we'd need to do a `getFetch(): FetchLike | Promise<FetchLike>`... what's the use case apart from setting headers [which are about to be async]?

mmkal:
Users would be able to easily tweak fetch options based on inputs, e.g.

```ts
fetch: async (url, opts) => {
  const fetch = getFetch()
  return fetch(url, {
    ...opts,
    credentials: url.includes('/someCustomPath') ? 'include': opts.credentials,
  })
}
```

Or get headers based on inputs too, for more advanced use-cases

```ts
fetch: async (url, opts) => {
  const fetch = getFetch()
  return fetch(url, {
    ...opts,
    headers: {
      ...opts?.headers,
      'x-client-auth': url.includes('/admin') ? await getClientKey() : undefined,
    },
  })
}
```

But even without those clear cut use-cases, it would give you the same abilities as `fetchOptions`, but much more flexibility and without adding an extra property to the API. Nothing would need to change except exporting getFetch, I think it could exported as-is from [here](https://github.com/trpc/trpc/blob/b1654ea9ae559ebaf36a9dac2abbf2539a778738/packages/client/src/internals/fetchHelpers.ts#L13).

No _harm_ in having `fetchOptions` other than bigger API surface to maintain and for users to grok, and it's not really necessary IMO, and less flexible than wrapping fetch anyway.

KATT:
@mmkal i'm just gonna revert this PR, I don't have energy to fix that 😅 - what you're describing is possible but writing a your own "replacement" of fetch

PR #915, opened 2021-09-09
Middleware: swap context

mmkal:
From #818 

@KATT I created a new PR because I made a lot of (type-level) changes.

- Added TInputContext to Router and Procedure
- Tests to make sure context-swapping works and doesn't have unexpected side-effects
- Removed swapContext method for now - can be added later, but the use-case is supported via `return next({ ctx: ... })`
- Replaced `MiddlewareFunctionKeepContext` and `MiddlewareFunctionSwapContext` with a single `MiddlewareFunction` type with an overloaded `next` input
- Made `next({ ctx: ... })` a function which can pass anything to `ctx`. The type passed is inferred as the `TContext` for the middleware's return value - this gets the `expectTypeOf` test added in the previous PR working
- Removed a bunch of ` = any` in type args. It was hard to keep track of what needed to be changed when defaults were used. I don't think any were internal so I've left them like that, but worth taking a look
- Added a swap-context example to the docs

Notes for the future:

- The types in `procedure.ts` are a bit loose. It's internal, so that's ok as long as it's covered by tests, but I think the crux is that it uses a `middlewares` array. I haven't tried this at all, but I think we could get more type safety by requiring that a procedure has exactly one middleware, and adding some helpers which compose middlewares. That way we can properly keep track of input and output contexts, instead of procedures having a list of unknown -> unknown middlewares.
- Having `ProcedureWithInput` and `ProcedureWithoutInput` made this a bit harder than it feels like it should be. Maybe Procedure instances could always have inputs, where for convenience you can add queries/mutations/subscriptions to a router without, but the "input" defaults to `() => undefined` (which is a valid `ProcedureInputParserCustomValidatorEsque`, I believe)
- When classes/interfaces start having loads of generics like here, something I've tried before is to have a single generic with properties, so you can do something like `class Router<T extends TRouterOptions>{ ... }`. Then to get the context you do `T['context']`, to get queries it'd be `T['queries']`. This can make it easier to write helper functions and refactor things without having to duplicate the ever-growing `<TInputContext, TContext, TQueries, TMutations, TSubscriptions, TGodKnowsWhatElse>` typearg lists, or fall back on potentially dangerous `any` defaults.


KATT:
Wow, awesome! 🤩  Will review properly soon, right now just responding to your text!

> @KATT I created a new PR because I made a lot of (type-level) changes.

👌 

> * Added TInputContext to Router and Procedure

- Should we see this as a breaking change or not? I've had issues before when libs changing type defs without major deps. Haven't reflected in detail if it'd cause any issues.
- Haven't played with it yet, will ASAP - can I merge two a router with a router that has swapped the context? I also thought about the input/output context approach, I guess it should solve that.

> * Tests to make sure context-swapping works and doesn't have unexpected side-effects

💚 ✅ 

> * Removed swapContext method for now - can be added later, but the use-case is supported via `return next({ ctx: ... })`
> * Replaced `MiddlewareFunctionKeepContext` and `MiddlewareFunctionSwapContext` with a single `MiddlewareFunction` type with an overloaded `next` input

💯  - rip them out - I just had those there as a work in progress kinda thing.

> * Made `next({ ctx: ... })` a function which can pass anything to `ctx`. The type passed is inferred as the `TContext` for the middleware's return value - this gets the `expectTypeOf` test added in the previous PR working

Yay!!

> * Removed a bunch of ` = any` in type args. It was hard to keep track of what needed to be changed when defaults were used. I don't think any were internal so I've left them like that, but worth taking a look

👌 

> * Added a swap-context example to the docs

👌 👌 

> * The types in `procedure.ts` are a bit loose. It's internal, so that's ok as long as it's covered by tests, but I think the crux is that it uses a `middlewares` array. I haven't tried this at all, but I think we could get more type safety by requiring that a procedure has exactly one middleware, and adding some helpers which compose middlewares. That way we can properly keep track of input and output contexts, instead of procedures having a list of unknown -> unknown middlewares.

Agree with you - it's loose but it's kinda okay for now. The things we do in tRPC are possible thanks to TypeScript not being 100% pure and we are able to typecast stuff.

> * Having `ProcedureWithInput` and `ProcedureWithoutInput` made this a bit harder than it feels like it should be. Maybe Procedure instances could always have inputs, where for convenience you can add queries/mutations/subscriptions to a router without, but the "input" defaults to `() => undefined` (which is a valid `ProcedureInputParserCustomValidatorEsque`, I believe)

I'm impressed you worked this out on your own though, it was a bit of a bitch to make and I agree there should be a simpler approach.


> * When classes/interfaces start having loads of generics like here, something I've tried before is to have a single generic with properties, so you can do something like `class Router<T extends TRouterOptions>{ ... }`. Then to get the context you do `T['context']`, to get queries it'd be `T['queries']`. This can make it easier to write helper functions and refactor things without having to duplicate the ever-growing `<TInputContext, TContext, TQueries, TMutations, TSubscriptions, TGodKnowsWhatElse>` typearg lists, or fall back on potentially dangerous `any` defaults.

YES. I love options objects and I don't know why I haven't thought about doing this with generics!

KATT (changes requested):
Added a failing test in 9d2f4ef80d6b6f9ef2569812bdbc4f96b983de01

mmkal:
@KATT good catch, hadn't tried merging routers. Hope it doesn't get crazy.

I added c99e345 to expose `LegacyRouter` as a subclass with the old type signature. The "main" router is called `VNextRouter`, and `LegacyRotuer` is exported in `src/index.ts` as `Router` so anyone using it via `import { Router } from '@trpc/server'` will still get the same signature. I added a test to verify it behaves as expected at runtime and compile-time. I _think_ that's sufficient to make it backwards-compatible? I haven't deprecated it, because we could switch to the typearg-options-bag that way without confusing users.

packages/server/src/router.ts
```diff
@@ -609,6 +671,27 @@ export class Router<
   }
 }
 
-export function router<TContext = void>() {
-  return new Router<TContext, {}, {}, {}, DefaultErrorShape>();
+/** Subclass of `VNextRouter` with `TInputContext` and `TContext` set to the same type, for backcompat. */
+export class LegacyRouter<
```
KATT, line 675:
```suggestion
/** 
 * Subclass of `Router` with `TInputContext` and `TContext` set to the same type, for backcompat. 
 * @deprecated
 */
export class LegacyRouter<
```

packages/server/src/router.ts
```diff
@@ -192,11 +202,36 @@ const defaultTransformer: CombinedDataTransformer = {
...
+  TObj extends ProcedureRecord<any, any, any, any>,
+  TNewContext,
+> = format<
+  {
+    [P in keyof TObj]: SwapProcedureContext<TObj[P], TNewContext>;
+  }
+>;
+
+/** @experimental The type signature of this class may change without warning. */
 export class Router<
```
KATT, line 228: I've started using this https://api-extractor.com/pages/tsdoc/tag_internal/

Not sure what `@experimental` actually does 


```suggestion
/** 
 * @internal The type signature of this class may change without warning. 
 */
export class Router<
```

packages/server/src/index.ts
```diff
@@ -1,6 +1,6 @@
 export * from './assertNotBrowser';
 export * from './http';
-export * from './router';
+export * from './router.legacy';
```
KATT, line 3: I've used the `./deprecated`-folder for this but I'm with above too

```suggestion
export * from './router';
export * from './deprecated/LegacyRouter';
```

KATT (commented):
Some last few comments then I think it's good to go

packages/server/src/router.ts
mmkal: Only reason I didn't deprecate it yet is in case we want to switch to an options bag for typeargs before the next major. If that's the plan then it'd be not great to tell people that this one is deprecated and they should switch to something that will change again. But maybe both `Router` and `LegacyRouter` could be marked as `@internal` so that anyone using them in future does so at their own risk, what do you think?

KATT: I think it should be marked as internal in general -- I don't want people using `Router`, but it's ok if they use `AnyRouter` or `inferProcedureOutput<X>`, etc that we can maintain

mmkal: Actually, on further thought, you're right. I was thinking the `@deprecated` tag would point to the new `Router`, but how about we deprecate it with no replacement. I don't know of any use case for an external user using the `Router` class, with all its typeargs directly, and it's pretty iffy to have to support unknown use cases. I'd vote to make it not part of the API surface at all in the next major. If someone has a use-case they can raise an issue and maybe some type helper can be exposed.

But having it part of the API surface reduced the ability to refactor easily, and having so many typeargs means it makes more sense as internal generally.

So basically exactly what you suggested!

KATT (approved):
Great work - I did the last tweak in 741c90e - feel free to smash the merge button!

PR #923, opened 2021-09-10
Expose `getFetch`

mmkal:
Based on discussion in #908 /  #916

This basically just exports the `getFetch` function which was previously internal.

Note: this isn't super hard to do in user-land but if wrapping fetch it's nice to know that we can use trpc's default under the hood, and all of the `typeof window` checks are easy to miss edge cases in. My team are currently doing this by just using the global `fetch` which will break when we turn on SSR.

This also adds a `TRPCFetch` type which is a narrowing of the default `fetch` signature. trpc always uses the `fetch('string', { object }`) format. This makes that fact part of the API contract so that users can rely on it and do things like

```ts
fetch: (url, options) => {
  const fetch = getFetch()
  return fetch(
    url,
    url.includes('/admin')
      ? { credentials: 'include', ...options }
      : options
  )
}
```

without having to do `if (typeof url === 'string')` checks.


packages/client/src/links/core.ts
```diff
@@ -46,13 +46,25 @@ export interface HTTPLinkOptions {
...
+
 export type LinkRuntimeOptions = Readonly<{
   transformer: DataTransformer;
   headers: () => HTTPHeaders | Promise<HTTPHeaders>;
-  fetch: typeof fetch;
+  fetch: TRPCFetch;
   AbortController?: typeof AbortController;
 }>;
 
+export { getFetch } from '../internals/fetchHelpers';
```
KATT, line 66: Move `getFetch()` out of the `internals/` folder

packages/client/src/links/core.ts
```diff
@@ -46,13 +46,25 @@ export interface HTTPLinkOptions {
...
 export type HTTPHeaders = Record<string, string | string[] | undefined>;
+
+/**
+ * The default `fetch` implementation has an overloaded signature. By convention this library
+ * only uses the overload taking a string and options object.
+ */
+export type TRPCFetch = (
+  url: string,
+  options?: RequestInit,
+) => Promise<Response>;
```
KATT, line 57: 👌

KATT (changes requested):
LGTM apart from one comment

packages/client/src/links/core.ts
mmkal: Where do you think it should go? `src/links/` feels wrong, should I make a `getFetch.ts` in `src/`?

KATT: Pick one, I don't mind, it won't change the interface

KATT (approved):

PR #955, opened 2021-09-16
fix `inferProcedureInput`

### LINKED ISSUE #949: `inferProcedureInput` broken in published package
This is a super weird one. It may have been me who broke it, and I haven't been able to figure out how/why yet. My team's project hits this error trying to update to v9.6.0 (from v9.0.0-alpha.0, which was before #915 went in but after/somewhere in the middle of #718).

We're finding that *under certain conditions* `inferProcedureInput<>` is returning `unknown` (in fact, it's the type inference of router definition - the procedure itself is `Procedure<Context, Context, unknown, { ...result... }>`). Repro (note - this does _not_ repro in this repo, if you do `import * as trpc from '../src'` in a test, which is what I tried first. You can use this [playground link](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgUwB5mQYxgFQJ4YDyAZnAL5zFQQhwDkaG2AtDAcnQNwBQokscAFRwAhgGc4MKGEyVqtOgAEpMgPRjkUAG6a6vcNHgIAXhSo16xiABM93TBAB2Y+CLBgAShACuMTXABeSWlMADpqX00ACgBKUJBNAHNkKO44ehFrEGBHVToAGjTgmSL08J8-KFjSuHjfERhgJyi6PxcAWXrGpwLEGvScsF8ALjhjUIgAIwArLBgohH70yggIUfGXKBzE2MLl5bIYvf24KGQxCAAbHVHxPEdZBbhB33IYwIA+OAADABIEF4wULEVZkACE32O6UONXiwGs1kuyAA7iIzlE7g84E9HGh4IdPnBcah5jEjtwYjxuGwMHBOjAGk1nIFJOwIKQ3J4KppQgB9azIYh1BndZw8RhzfBEYgAHiKKjCOWImgACtRMMhrN4zgBJRxDGAy+mMpxiADadEy2VybRgxtFdAAuh9uB9YqEYBAAKIAR28IkuUuQJBlSBBazgm225DdlO4EuwQZD8pCoSVqvVmu1yD1BqNXSZ5stWRyqlt9qZTpdsdCjggQM9ACFkABVRwAazryMcsR4QA) to see it going wrong:

```ts
import {z} from 'zod'

const appRouter = trpc.router().merge(
  'admin/',
  trpc
    .router()
    .mutation('testMutation', {
      input: z.object({
        foo: z.string(),
      }),
      resolve: async ({ input }) => `${input.foo}!`,
    })
    .middleware(async ({ next }) => next()),
);

type Mutations = typeof appRouter._def.mutations;

expectTypeOf<
  trpc.inferProcedureInput<Mutations['admin/testMutation']>
>().toEqualTypeOf<{ foo: string }>();

expectTypeOf<
  trpc.inferProcedureInput<Mutations['admin/testMutation']>
>().not.toBeUnknown();
```

The two `expectTypeOf` tests fail. The second is just to show that the inferred type is `unknown` and that it isn't a typo/bad test setup.

Update: The tests succeed if you remove the `.middleware(async ({ next }) => next())` line (in this case, the middleware is pointless - we are seeing it for a similar middleware that throws under certain circumstances)

Update 2: It can also be repro'd by doing `import * as trpc from '../dist/trpc-server.cjs'` in this repo, which makes things much easier. Could there be some types that are being stripped at build time, maybe?

KATT:
closes #949

## How I did it

1. Random whack-a-mole trying to export everything to see if it changes anything
1. Noticed that if you put the `middleware()` *last* it does work
1. Added another procedure last and noticed that one was `Procedure<...>` and the other was `ProcedureWithInput<...>`
1. Dug into the `SwapContext` code and changed it to `ProcedureWithInput` 
1. Added regression test + made sure regression test actually breaks when reverting "good" commit

## On another note.. 

You seem to be a heavy user of tRPC - I'd love to chat with you on how you and your team are using it and get feedback on what you love & hate! 

Feel free to book [a slot](https://calendly.com/alexdotjs) in my calendar whenever suits you ("US" event one is a bit more flexible).


packages/server/src/router.ts
```diff
@@ -205,14 +205,14 @@ const defaultTransformer: CombinedDataTransformer = {
...
-> = TProcedure extends Procedure<
+> = TProcedure extends ProcedureWithInput<
   infer TInputContext,
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   infer _TOldContext,
   infer TInput,
   infer TOutput
 >
-  ? Procedure<TInputContext, TNewContext, TInput, TOutput>
+  ? ProcedureWithInput<TInputContext, TNewContext, TInput, TOutput>
```
KATT, line 215: 🔑

mmkal:
@KATT haven't had a read through this yet... I have a repro and fix for it too 🙃 should have checked GitHub. Btw my fix involved getting rid of `ProcedureWithInput` and `ProcedureWithoutInput` in favour of just one non-abstract `Procedure` class. There is a caveat which you may have avoided. I might put mine up as a draft PR now so you can decide which approach you'd prefer - and maybe it'd be possible to pick-and-choose bits of each. Will open it in the next 30 mins.

packages/server/test/regression/issue-949-inferProcedureInput.test.ts
```diff
@@ -0,0 +1,71 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+// IMPORTANT:
+// needs to be imported from compiled output otherwise we get a false-positive
+import * as trpc from '../../dist/trpc-server.cjs';
+import { expectTypeOf } from 'expect-type';
```
mmkal, line 5: Interstingly in debugging this I found that there's _also_ a bug in expect-type:

```
import { expectTypeOf } from 'expect-type'

class Foo {
  constructor(input: string) {}
}

class Bar {
  constructor(input: number) {}
}

expectTypeOf<Foo>().not.toEqualTypeOf<Bar>() // fails!
```

...so I'll need to push an update to that too.

packages/server/test/regression/issue-949-inferProcedureInput.test.ts
```diff
@@ -0,0 +1,71 @@
...
+import { z } from 'zod';
+
+// https://github.com/trpc/trpc/issues/949
+// https://github.com/trpc/trpc/pull/955
+test('inferProcedureFromInput regression', async () => {
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  type Context = {};
```
mmkal, line 12: Personal opinion: `{}` is fine and I usually have a good reason when I want to use it. I usually disable ban-types on repos where most people committing code are pretty familiar with typescript.

packages/server/test/regression/issue-949-inferProcedureInput.test.ts
```diff
@@ -0,0 +1,71 @@
...
+        }),
+      )
+      .query('hello-2', {
+        input: z.object({
+          in: z.string(),
+        }),
+        resolve() {
+          return 'out' as const;
+        },
+      }),
```
mmkal, line 46: I think it'd be worth having a test case for a no-input procedure too. Changing from `Procedure` -> `ProcedureWithInput` in the type helper might have made this _only_ work for procedures with input.

mmkal (commented):
If this works with no-input procedures too, then 👍 from me. My changes could be a follow-up to simplify down the whole `Procedure`/`ProcedureWithInput`/`ProcedureWithoutInput` thing.

KATT:
Will merge and push this and turn we can do follow-up!

mmkal:
(just read the PR description - I forget to about 80% of time time. put in 30 mins on your calendly next week!)

PR #956, opened 2021-09-16
Simplify procedure types/overloads

### LINKED ISSUE #949: `inferProcedureInput` broken in published package

mmkal:
~Fixes #949~

Follow-up to #955

This simplifies Procedure down to one type:

- ~`CreateProcedureWithInput`, `CreateProcedureWithoutInput`, `CreateProcedureOptions` -> `CreateProcedureOptions` (a single union type)~
- `ProcedureWithInput`, `ProcedureWithoutInput`, `Procedure` -> `Procedure` (a single non-abstract class which _always_ has a `TInput`. The input just happens to be undefined if it was created without an input parser specified)

~Note: there is one downside to this approach which is that for procedures with no inputs, the `input` prop passed to the resolver is typed as `unknown`, where ideally it'd be `undefined` or `never`. IMHO that's a pretty ok tradeoff given how much this simplifies both `procedure.ts` and `router.ts`.~

### DIFF
--- a/packages/server/src/router.ts
+++ b/packages/server/src/router.ts
@@ -7,12 +7,9 @@ import { MiddlewareFunction } from './internals/middlewares';
 import {
   createProcedure,
   CreateProcedureOptions,
-  CreateProcedureWithInput,
-  CreateProcedureWithoutInput,
   inferProcedureFromOptions,
   Procedure,
   ProcedureCallOptions,
-  ProcedureWithInput,
 } from './internals/procedure';
 import {
   TRPCErrorShape,
@@ -47,7 +44,7 @@ export type ProcedureRecord<
  */
 export type inferProcedureInput<
   TProcedure extends Procedure<any, any, any, any>,
-> = TProcedure extends ProcedureWithInput<any, any, infer Input, any>
+> = TProcedure extends Procedure<any, any, infer Input, any>
   ? Input
   : undefined;
 
@@ -90,7 +87,7 @@ function getDataTransformer(
  */
 export type inferHandlerInput<
   TProcedure extends Procedure<any, any, any, any>,
-> = TProcedure extends ProcedureWithInput<any, any, infer TInput, any>
+> = TProcedure extends Procedure<any, any, infer TInput, any>
   ? undefined extends TInput // ? is input optional
     ? unknown extends TInput // ? is input unset
       ? [(null | undefined)?] // -> there is no input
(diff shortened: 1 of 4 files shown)


KATT (changes requested):
I really like the change but I don't wanna lose the `undefined` type inference - added extra safeguards for this in #963

packages/server/src/router.ts
```diff
@@ -394,14 +396,11 @@ export class Router<
...
-  public subscription<
-    TPath extends string,
-    TInput,
-    TOutput extends Subscription<unknown>,
-  >(path: TPath, procedure: CreateProcedureOptions<TContext, TInput, TOutput>) {
+
+  public subscription(
+    path: string,
+    procedure: CreateProcedureOptions<TContext, any, any>,
+  ) {
```
mmkal, line 403: Note: for the implementation, the jsdoc and types are ignored so using `any` a bit more so using `any` and not duplicating the beta tag etc. here and in the equivalents for `query` and `mutation`.

mmkal:
@KATT updated, I brought back the overloads on `Router`, but removed the extensions of the `Procedure` class. So a smaller simplification, but it passes the new tests now. I also made procedure without input to have a `TInput` of `undefined | null` instead of `undefined`, since `undefined`-only would be a breaking change (one of the e2e tests caught this). Hopefully this change is ok - since the runtime check is `input != null`, it could be either at runtime so this seems appropriate to me. Up to you though if you think it needs another iteration though.

KATT (approved):
I moved the `undefined/null` logic into `inferProcedureInput` which feels more correct - let me know what you think @mmkal

mmkal:
LGTM!

PR #966, opened 2021-09-20
simplify merge a bit

KATT:
@mmkal has an idea on how to simplify this further by having an "options bag" on the `Router` so it'd be `Router<TSomething>` instead of `Router<x,y,z,m,forever>`


KATT:
meh - saw the test failed as the order of context was wrong - this doesn't actually improve anything, so will close it.

PR #1009, opened 2021-09-24
Allow passing no input on mutations with no input

### LINKED ISSUE #390: Allow passing no `input` on mutations with optional/no input
When no `input` is specified for a mutation on the back-end, the corresponding `useMutation` hook shouldn’t require any variables to be given. Unfortunately, this isn’t the case over here:

image

With code similar to:

```tsx
const { mutate: softCancelSubscription } = trpc.useMutation(["WithoutInput"]);

mutate(); // Type checking fails when no parameter is given
```

infix:
fixes #390

updates the `useMutation` mutate type to allow it to be without an input

### DIFF
--- a/packages/react/src/createReactQueryHooks.tsx
+++ b/packages/react/src/createReactQueryHooks.tsx
@@ -74,8 +74,8 @@ type UseTRPCMutationResult<
   ? Override<
       UseMutationResult<TOutput, TError, TInput>,
       {
-        mutate: (options?: TOptions) => TOutput;
-        mutateAsync: (options?: TOptions) => Promise<TOutput>;
+        mutate: (variables?: null, options?: TOptions) => TOutput;
+        mutateAsync: (variables?: null, options?: TOptions) => Promise<TOutput>;
       }
     >
   : UseMutationResult<TOutput, TError, TInput>;
--- a/packages/react/src/createReactQueryHooks.tsx
+++ b/packages/react/src/createReactQueryHooks.tsx
@@ -21,6 +21,7 @@ import {
   UseInfiniteQueryOptions,
   useMutation,
   UseMutationOptions,
+  UseMutationResult,
   useQuery,
   UseQueryOptions,
   UseQueryResult,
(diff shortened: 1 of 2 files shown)


KATT (changes requested):
@infix amazing! Thank you so much! I just [added you](https://github.com/trpc/trpc/invitations) as a collaborator, could you make a branch on the tRPC origin and make a PR through there instead?

Also not 100% sure this would work on `.nullish()` inputs.

If you want to, you can add some tests [here](https://github.com/trpc/trpc/blob/main/packages/server/test/react.test.tsx) but I know the tests are a bit horrible _(I haven't cared much about the code quality there)_, so I can also do that as soon as you've opened a branch on the origin.

infix:
Alright, will check `nullish`, and test cases, and create new PR tomorrow, also thank you for creating this awesome library

KATT:
@all-contributors add @infix for code

packages/react/src/createReactQueryHooks.tsx
```diff
@@ -60,6 +61,25 @@ interface UseTRPCMutationOptions<TInput, TError, TOutput>
...
+  TInput,
+  TOutput,
+  TError,
+  TOptions = UseTRPCMutationOptions<TOutput, TError, never>,
+> = TInput extends null | undefined
+  ? Override<
+      UseMutationResult<TOutput, TError, TInput>,
+      {
+        mutate: (variables?: null, options?: TOptions) => TOutput;
+        mutateAsync: (variables?: null, options?: TOptions) => Promise<TOutput>;
```
mmkal, line 78: Probably makes sense to do:

```suggestion
        mutate: (variables?: TInput, options?: TOptions) => TOutput;
        mutateAsync: (variables?: TInput, options?: TOptions) => Promise<TOutput>;
```

KATT: That'd fix nullish (if it was a problem which I'm not sure of)

PR #1012, opened 2021-09-25
Allow passing no input on mutations with no input

### LINKED ISSUE #390: Allow passing no `input` on mutations with optional/no input

infix:
fixes #390

updates the useMutation mutate type to allow it to be without an input

same as #1009 except created from tRPC origin, and updated types to accept nullish inputs, and add test cases for mutations called with optional inputs, and no inputs.

### DIFF
--- a/examples/next-prisma-todomvc/src/pages/[filter].tsx
+++ b/examples/next-prisma-todomvc/src/pages/[filter].tsx
@@ -305,7 +305,7 @@ export default function TodosPage({
             <button
               className="clear-completed"
               onClick={() => {
-                clearCompleted.mutate(null);
+                clearCompleted.mutate();
               }}
             >
               Clear completed
--- a/examples/next-prisma-todomvc/src/pages/[filter].tsx
+++ b/examples/next-prisma-todomvc/src/pages/[filter].tsx
@@ -305,7 +305,7 @@ export default function TodosPage({
             <button
               className="clear-completed"
               onClick={() => {
-                clearCompleted.mutate();
+                clearCompleted.mutate(null);
               }}
             >
               Clear completed
--- a/packages/react/src/createReactQueryHooks.tsx
+++ b/packages/react/src/createReactQueryHooks.tsx
@@ -16,7 +16,6 @@ import React, { ReactNode, useCallback, useEffect, useMemo } from 'react';
 import {
   hashQueryKey,
   QueryClient,
-  QueryKey,
   useInfiniteQuery,
   UseInfiniteQueryOptions,
   useMutation,
@@ -45,15 +44,15 @@ interface TRPCUseQueryBaseOptions extends TRPCRequestOptions {
   ssr?: boolean;
 }
 
-interface UseTRPCQueryOptions<TInput, TError, TOutput>
-  extends UseQueryOptions<TInput, TError, TOutput, QueryKey>,
+interface UseTRPCQueryOptions<TError, TOutput, TQueryKey extends unknown[]>
+  extends UseQueryOptions<TOutput, TError, TOutput, TQueryKey>,
     TRPCUseQueryBaseOptions {}
 
 interface UseTRPCInfiniteQueryOptions<
-  TInput = unknown,
-  TError = unknown,
-  TOutput = TInput,
-> extends UseInfiniteQueryOptions<TInput, TError, TOutput, TOutput, QueryKey>,
+  TError,
+  TOutput,
+  TQueryKey extends unknown[],
+> extends UseInfiniteQueryOptions<TOutput, TError, TOutput, TOutput, TQueryKey>,
     TRPCUseQueryBaseOptions {}
 
 interface UseTRPCMutationOptions<TInput, TError, TOutput>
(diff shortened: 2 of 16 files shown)


packages/react/src/createReactQueryHooks.tsx
```diff
@@ -60,6 +61,31 @@ interface UseTRPCMutationOptions<TInput, TError, TOutput>
...
     TRPCUseQueryBaseOptions {}
 
+type Override<A, B> = {
+  [K in keyof A]: K extends keyof B ? B[K] : A[K];
+};
+type Equals<X, Y> = (<T>() => T extends X ? 1 : 2) extends <T>() => T extends Y
+  ? 1
+  : 2
+  ? true
+  : false;
```
infix, line 71: `Override`, and `Equals` probably shouldn't live in `createReactQueryHooks`. not sure where to move them tho

packages/server/test/react.test.tsx
```diff
@@ -389,6 +399,90 @@ test('mutation on mount + subscribe for it', async () => {
   });
 });
 
+describe('useMutation()', () => {
```
infix, line 402: will probably need to rewrite the test cases later to allow for testing the hooks in isolation instead of rendering

packages/react/src/createReactQueryHooks.tsx
```diff
@@ -60,6 +61,31 @@ interface UseTRPCMutationOptions<TInput, TError, TOutput>
...
+
+type UseTRPCMutationResult<TInput, TOutput, TError, TOptions> = Equals<
```
mmkal, line 73: Is `Equals` definitely needed here? It seems quite big/complex. Maybe this would work:

```ts
[NonNullable<TInput>] extends [never]
  ? ... nullish...
  : ...not nullish...

Comment (infix) [reply packages/react/src/createReactQueryHooks.tsx:73]: I agree that Equals is a bit complex but for example if input is `z.string().nullish()` that means `TInput` would `string | null | undefined`, calling `NonNullable<TInput>` would just return `string`.

`Equals<TInput, NonNullable<TInput>>` seemed like the simplest way, and checking if `TInput extends null | undefined` narrows the type which results in `TInput` being useless for the nullish branch of the ternary.

also Equals looks worse than it actually is because of prettier formatting, it should actually look this https://github.com/Microsoft/TypeScript/issues/27024#issuecomment-421529650

**Comment by mmkal** [comment]
I haven't looked closely at the react-query types or even trpc/client so take this with a pinch of salt, but my initial feeling is that this could be better solved in react-query, or at least solved more simply. Does react-query only support unary mutation functions? If it supported an argument list input type, trpc night not need the Override helper. It could pass `[TInput?]` or `[]` as an argument list instead

**Comment by infix** [comment]
@mmkal I decided to take a step back and look at the types again. resulted in the type being simplified from the 30-something lines to 2 line changes :partying_face:

File: examples/next-prisma-todomvc/src/pages/[filter].tsx [hunk 4]
```diff
@@ -305,7 +305,7 @@ export default function TodosPage({
             <button
               className="clear-completed"
               onClick={() => {
-                clearCompleted.mutate(null);
+                clearCompleted.mutate();
```
Comment (KATT) [inline examples/next-prisma-todomvc/src/pages/[filter].tsx:308]: You need to do this change in a follow-up as the example deploy requires a new version to be published.

At the same time, having this here makes sure that you're not making any breaking changes

```suggestion
                clearCompleted.mutate(null);
```

packages/server/test/react.test.tsx
```diff
@@ -389,6 +399,92 @@ test('mutation on mount + subscribe for it', async () => {
...
+    await waitFor(() => {
+      expect(utils.container).toHaveTextContent('first post');
+      expect(utils.container).toHaveTextContent('second post');
+    });
+
+    await waitFor(() => {
+      expect(utils.container).not.toHaveTextContent('first post');
+      expect(utils.container).toHaveTextContent('second post');
+    });
+  });
```
KATT, line 485: We should add a test that we're not breaking existing `mutate(null)`-functionality + check that docs are updated

KATT (changes requested):
I might do a PR with these changes

infix:
will update the pr in couple of hours or after work

KATT (approved):

KATT:
> I haven't looked closely at the react-query types or even trpc/client so take this with a pinch of salt, but my initial feeling is that this could be better solved in react-query, or at least solved more simply. Does react-query only support unary mutation functions? If it supported an argument list input type, trpc night not need the Override helper. It could pass `[TInput?]` or `[]` as an argument list instead

@mmkal:

ReactQuery actually support both `useQuery('path')` and `useQuery(['path', input])` and tRPC used to have both and it worked in the tests. However, the autocompletion in VSCode breaks as soon as I added this

KATT:
In a future version of tRPC we might want to replace this `useQuery(['path', input])` with something similar to

```ts
useQuery('path', {
  input: input
})
```

And `input` can be `input?:` if we the input is `nullish`

PR #1058, opened 2021-10-03
Refactor queries new v2

KATT:


alii (approved):

mgreenw:
Even though it's a bit strange to use at first, I don't hate the array style of the current `trpc.useQuery` method -- not needing to type `input` on every query is pretty nice. That being said, this format is potentially a bit more friendly / prettier.

KATT:
> Even though it's a bit strange to use at first, I don't hate the array style of the current `trpc.useQuery` method -- not needing to type `input` on every query is pretty nice. That being said, this format is potentially a bit more friendly / prettier.

Yeah, it's that initial _"why is this an array (or more strictly, a ternary) thing here?"_ I want to get rid of - it's not very intuitive even if it's the way react-query deals with it. I decided to follow suit just to not create too much disparity between the two, but I have always thought it's sub-optimal.

To be clear, the proposed new API would only enforce `opts` to be passed if there's a non-nullable input.

packages/server/test/react.test.tsx
```diff
@@ -1667,3 +1667,71 @@ describe('withTRPC()', () => {
...
+  test('with input', async () => {
+    const { trpc, client } = factory;
+    function MyComponent() {
+      // @ts-expect-error
+      trpc.useQuery('paginatedPosts');
+      trpc.useQuery('paginatedPosts', {
+        input: {
+          limit: 1,
+          // we _should_ add a @ts-expect-error here
+          notExists: 'test',
```
KATT, line 1707: @mmkal 👋 

You think we can get this to be erroneous? Right now we allow extra args as `TInput` _`extends`_ something. Feel free to edit on this branch!

I'd like to change this to:

```suggestion
          // @ts-expect-error
          notExists: 'test',
```

packages/react/src/createReactQueryHooks.tsx
```diff
@@ -44,28 +45,87 @@ interface TRPCUseQueryBaseOptions extends TRPCRequestOptions {
...
+
+// interface UseTRPCInfiniteQueryOptionalInputOptions<
+//   TPath,
+//   TInput,
+//   TOutput,
+//   TError,
+// > extends UseInfiniteQueryOptions<TOutput, TError, TOutput, [TPath, TInput]>,
+//     TRPCUseQueryBaseOptions,
+//     TRPCUseQueryBaseOptionsOptionalInput<TInput> {}
+
```
KATT, line 96: Will be added at a later point - a bit trickier as infinite queries always are at least `{ cursor: any }`

TheoBr:
Array for queries and bare args for mutations always felt weird, definitely in favor of this option

Feels _slightly_ weird to be diverging from the `react-query` patterns but these differ for good reasons, as trpc functions are defined elsewhere

KATT:
> Feels _slightly_ weird to be diverging from the `react-query` patterns but these differ for good reasons, as trpc functions are defined elsewhere

@TheoBr - - 

An alternative would be to allow simple `useQuery(path)` for queries that have no required input and using a tuple for the ones that do. That would adhere to `react-query` more.

Do you think that would be better than this? It could make a bunch of other methods (invalidate etc) more intuitive than my proposition here & we wouldn't have to think about potentially making more intuitive apis for them

TheoBr:
> Do you think that would be better than this? 
@KATT
Honestly I have no strong opinions here. I'm happy as is but do see the benefit in diverging

Personally, I'd like "one way to do things" that's agreed upon, indifferent as to which

KATT:
@all-contributors add @TheoBr  for code review

PR #1059, opened 2021-10-03
feat: pass procedure input to middleware

mgreenw:
@KATT

In many routers, the authentication step takes place in the procedure resolver instead of middleware because data needed to perform the authentication is located in the procedure input, which is currently not passed to middleware. It would be helpful to have access to procedure input in middleware so a set of procedures can be authenticated all at once by a single middleware.

To solve this, this PR adds `input` as a parameter to the middleware function. Because the middleware cannot infer the type of procedure input because they act on many procedures, the `input` is passed as `unknown`.

Please let me know if this makes sense / is useful for others!

Example usage:

```ts
const router = createRouter()
  .middleware(({ ctx, input }) => {
    checkUserAuth(ctx, input.userId); // throws if the ctx does not have access to the given user
    return next();
  })
  .query("getUser", {
    input: z.object({ userId: z.string() }),
    async resolve({ input }) {
      // No additional auth step needed
      const user = await getUser(input.userId);
      return user;
    }
  })
  .query("getUserTasks", {
    input: z.object({ userId: z.string() }),
    async resolve({ input }) {
      // No additional auth step needed
      const tasks = await getUserTasks(input.userId);
      return tasks;
    }
  });
```

### DIFF
--- a/packages/server/src/internals/procedure.ts
+++ b/packages/server/src/internals/procedure.ts
@@ -102,11 +102,10 @@ export class Procedure<TInputContext, TContext, TInput, TOutput> {
   public async call(
     opts: ProcedureCallOptions<TInputContext>,
   ): Promise<TOutput> {
-    const input = this.parseInput(opts.rawInput);
-
     // wrap the actual resolver and treat as the last "middleware"
     const middlewaresWithResolver = this.middlewares.concat([
       async ({ ctx }: { ctx: TContext }) => {
+        const input = this.parseInput(opts.rawInput);
         const data = await this.resolver({ ...opts, ctx, input });
         return {
           marker: middlewareMarker,
@@ -122,9 +121,10 @@ export class Procedure<TInputContext, TContext, TInput, TOutput> {
       return async (nextOpts?: { ctx: TContext }) => {
         const res = await wrapCallSafe(() =>
           fn({
-            ...(opts as any),
-            ...(nextOpts as any),
-            input,
+            ctx: nextOpts ? nextOpts.ctx : opts.ctx,
+            type: opts.type,
+            path: opts.path,
+            rawInput: opts.rawInput,
             next: nextFns[index + 1],
           }),
         );
--- a/packages/server/src/internals/procedure.ts
+++ b/packages/server/src/internals/procedure.ts
@@ -102,10 +102,11 @@ export class Procedure<TInputContext, TContext, TInput, TOutput> {
   public async call(
     opts: ProcedureCallOptions<TInputContext>,
   ): Promise<TOutput> {
+    const input = this.parseInput(opts.rawInput);
+
     // wrap the actual resolver and treat as the last "middleware"
     const middlewaresWithResolver = this.middlewares.concat([
       async ({ ctx }: { ctx: TContext }) => {
-        const input = this.parseInput(opts.rawInput);
         const data = await this.resolver({ ...opts, ctx, input });
         return {
           marker: middlewareMarker,
@@ -123,6 +124,7 @@ export class Procedure<TInputContext, TContext, TInput, TOutput> {
           fn({
             ...(opts as any),
             ...(nextOpts as any),
+            input,
             next: nextFns[index + 1],
           }),
         );
--- a/packages/server/test/middleware.test.ts
+++ b/packages/server/test/middleware.test.ts
@@ -23,18 +23,31 @@ test('is called if def first', async () => {
         resolve() {
           return 'bar2';
         },
+      })
+      .query('foo3', {
+        input: String,
+        resolve() {
+          return 'bar3';
+        },
       }),
   );
 
-  expect(await client.query('foo1')).toBe('bar1');
   const calls = middleware.mock.calls;
+  expect(await client.query('foo1')).toBe('bar1');
   expect(calls[0][0]).toHaveProperty('type');
   expect(calls[0][0]).toHaveProperty('ctx');
   expect(calls[0][0].type).toBe('query');
+
   expect(await client.mutation('foo2')).toBe('bar2');
   expect(calls[1][0].type).toBe('mutation');
 
-  expect(middleware).toHaveBeenCalledTimes(2);
+  expect(await client.query('foo3', 'foo3input')).toBe('bar3');
+  expect(calls[2][0]).toHaveProperty('type');
+  expect(calls[2][0]).toHaveProperty('ctx');
+  expect(calls[2][0].type).toBe('query');
+  expect(calls[2][0].input).toBe('foo3input');
+
+  expect(middleware).toHaveBeenCalledTimes(3);
   close();
 });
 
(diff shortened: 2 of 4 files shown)


mmkal:
Something like this was previously proposed in https://github.com/trpc/trpc/pull/718#discussion_r678717533 - the same debate applies here, I think the summary of why it didn't go in last time is that it's surprising for parsing, which is defined in the resolver, to happen before a middleware, even though it's declared afterwards.

One thing I did notice though, is that because of the `...opts` here, `rawInput` is actually passed to each middleware. That was accidental (from my perspective at least) and is invisible in the types, we only noticed it by logging the whole object passed to a logging middleware. But we've actually found it very useful. So maybe this PR could be tweaked to just make `rawInput` an official prop on the middleware input? The `raw` prefix is also good I think - it makes it less confusing that it's `unknown` and is a hint that you should be a bit careful using it - the "proper" input is reserved for resolvers.

mgreenw:
> Something like this was previously proposed in #718 (comment) - the same debate applies here, I think the summary of why it didn't go in last time is that it's surprising for parsing, which is defined in the resolver, to happen before a middleware, even though it's declared afterwards.

Ah thanks for the context! I hadn't seen the previous proposal.

> One thing I did notice though, is that because of the `...opts` here, `rawInput` is actually passed to each middleware.

I noticed this as well and agree this might be a good middle ground. It is definitely less helpful to use `rawInput` as opposed to `input` here, but I see why `parseInput` should definitely not be called until just before the resolver is called s o that middleware before the resolver will be called regardless of the input validation status.

mgreenw:
@mmkal Updated with that tweak -- I also modified the params to the middleware fn to only be exactly what is expected by the types.

KATT:
> In many routers, the authentication step takes place in the procedure resolver instead of middleware because data needed to perform the authentication is located in the procedure input, which is currently not passed to middleware. It would be helpful to have access to procedure input in middleware so a set of procedures can be authenticated all at once by a single middleware.


In which cases would it be better to authenticate in an input object rather than a http header?

mgreenw:
> In which cases would it be better to authenticate in an input object rather than a http header?

Let's say that we have an application with `users`, `organizations`, and `roles` (like Github, for instance). A user can have multiple roles at different organizations. If a user wants to make a request on behalf on an organization, such as view the list of repositories for an organization, they would need to make a request like this:

```ts
const reposQuery = trpc.useQuery(["getRepos", { organizationId: "ABCD" }]);
```

When the request comes in, we need to verify that the user making the request has permission to access the repositories for that organization by checking if it has a role on that org and that role permits them to access that data. Maybe something like:

```ts
const router = createRouter()
  .query("getRepos", {
    input: z.object({ organizationId: z.string() }),
    async resolve({ input, ctx }) {
      checkThatUserCanAccessOrganization(input.organizationId, ctx.userId);
      // get and return the repos
    }
  })
```

In this case, it is important to provide the `organizationId` as `input` instead of via an http header for the purposes of caching / query dedeplication with `react-query` -- if we did not, the query key that `react-query` uses to cache queries would be duplicated across organizations (just `"getRepos"`), so we would have to bust our whole cache if we switch the organization we are currently browsing as. This approach also allows queries to be made on behalf of multiple organizations at once without the need to set a cookie / http header to indicate which one we are making the request on behalf of.

Please let me know if there is a better or different approach here that would make more sense with `react-query`!

KATT:
Sounds good to use `rawInput` in a middleware and pass it through the context with [context swapping](https://trpc.io/docs/middlewares#context-swapping).

Will have a look and try to get this merged tomorrow! Please update docs as well in the `/www`-folder) (with some warning text preferably)

mgreenw:
Updated with some docs - Docusaurus is awesome

www/docs/server/middlewares.md
```diff
@@ -105,6 +105,35 @@ trpc
   });
 ```
 
+## Raw input
```
KATT, line 108: Move to bottom

www/docs/server/middlewares.md
```diff
@@ -105,6 +105,35 @@ trpc
...
+```ts
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null;
+}
+
+trpc
+  .router<Context>()
+  .middleware(async ({ next, rawInput, ctx }) => {
+    const userId = isRecord(rawInput) ? rawInput.userId : undefined;
+    if (typeof userId !== "string") throw new TRPCError({ code: "BAD_REQUEST" });
```
KATT, line 125: Simplify & use zod (most of the docs uses zod)

```suggestion
trpc
  .router<Context>()
  .middleware(async ({ next, rawInput, ctx }) => {
    const shape = z.object({ userId: z.string() });
    const result  = shape.safeParse(rawInput)
    if (!result.success) {
      throw new TRPCError({ code: "BAD_REQUEST" });
    }
    const { userId } = result.data;
```

KATT (changes requested):
Thanks for this!

- Can you make _Raw input_ the last section? Right now it looks [a bit weird](https://www-pkoweupd6.tmp.trpc.io/docs/middlewares#context-swapping) - that helper should be under context swapping (or be a `##`)
- Make a test that corresponds to the docs (I usually make a test and copy+paste to docs)

packages/server/src/internals/procedure.ts
```diff
@@ -121,8 +121,10 @@ export class Procedure<TInputContext, TContext, TInput, TOutput> {
       return async (nextOpts?: { ctx: TContext }) => {
         const res = await wrapCallSafe(() =>
           fn({
-            ...(opts as any),
-            ...(nextOpts as any),
+            ctx: nextOpts ? nextOpts.ctx : opts.ctx,
+            type: opts.type,
+            path: opts.path,
+            rawInput: opts.rawInput,
```
KATT, line 127: Is this change actually needed?

packages/server/test/middleware.test.ts
```diff
@@ -23,18 +23,31 @@ test('is called if def first', async () => {
         resolve() {
           return 'bar2';
         },
+      })
+      .query('foo3', {
+        input: String,
+        resolve() {
+          return 'bar3';
+        },
```
KATT, line 31: This test doesn't seem to add anything?
```suggestion
```

www/docs/server/middlewares.md
```diff
@@ -105,6 +105,35 @@ trpc
...
+    // Verify user
+    return next({ ctx: { ...ctx, userId }})
+  })
+  .query('foo', {
```
KATT, line 129: I think this needs to be

```suggestion
  .query('foo', {
    input: z.object({ userId: z.string() });
```

as well

KATT: you could just reuse the same zod parser and declare it outside of the router

if some of the queries require more inputs you can [`.extend()`](https://github.com/colinhacks/zod#extend) it

packages/server/src/internals/procedure.ts
mgreenw: This change is _not_ strictly needed for this PR, but I think this is cleaner / clearer so we don't need to cast `opts` and `nextOpts` to `any`. I also think it's best to not pass params that are not in the types so clients don't begin to [rely on them just because they are there](https://www.hyrumslaw.com/).

www/docs/server/middlewares.md
mgreenw: 💯

KATT (approved):

KATT:
@all-contributors add @mgreenw for code, docs and tests

PR #1060, opened 2021-10-04
[RFC] Allow `useQuery(path)` for queries with no required input

KATT:
**ℹ️ Alternative to #1058** - diverges less from `react-query`.
 
---

- Allows you to do 
  - `useQuery(path)` for queries that have optional `input`
  - For queries with required input it's still `useQuery([path, input])`
- Fully backwards compatible

Other improvements

- `useInfiniteQuery()` can now only be used with queries that extends `{ cursor?: any }`
- [Update `useQuery()`-docs](https://www-git-refactor-queries-new-v2alt.tmp.trpc.io/docs/react-queries)
- [Tweak `useMutation()`-docs](https://www-git-refactor-queries-new-v2alt.tmp.trpc.io/docs/react-mutations)

### DIFF
--- a/www/docs/react/useQuery.md
+++ b/www/docs/react/useQuery.md
@@ -7,9 +7,10 @@ slug: /react-queries
 
 > The hooks provided by `@trpc/react` are a thin wrapper around React Query. For in-depth information about options and usage patterns, refer to their docs on [Queries](https://react-query.tanstack.com/guides/queries).
 
-You pass a `[path, input]`-tuple as the first argument. You'll notice that you get autocompletion on the `path` and automatic typesafety on the `input`.
+- If the query has an optional `input`, you pass only it's procedure name as the first argument
+- If the query requires an input, you have to pass a `[path, input]`-tuple as the first argument
 
-If an `input`-argument is optional, you can omit the `, input` part of the argument.
+You'll notice that you get autocompletion on the `path` and automatic typesafety on the `input`.
 
 ### Example
 
@@ -45,6 +46,8 @@ import { trpc } from '../utils/trpc';
 export function MyComponent() {
   // input is optional, so we don't have to pass second argument
   const helloNoArgs = trpc.useQuery(['hello']);
+  // const helloNoArgs = trpc.useQuery('hello'); // we can also pass the the path as a string
+
   const helloWithArgs = trpc.useQuery(['hello', { text: 'client' }]);
 
   return (
--- a/www/docs/react/useQuery.md
+++ b/www/docs/react/useQuery.md
@@ -1,7 +1,7 @@
 ---
 id: queries
-title: Queries
-sidebar_label: useQuery
+title: useQuery()
+sidebar_label: useQuery()
 slug: /react-queries
 ---
 
--- a/www/docs/react/useQuery.md
+++ b/www/docs/react/useQuery.md
@@ -7,6 +7,13 @@ slug: /react-queries
 
 > The hooks provided by `@trpc/react` are a thin wrapper around React Query. For in-depth information about options and usage patterns, refer to their docs on [Queries](https://react-query.tanstack.com/guides/queries).
 
+```tsx
+function useQuery(
+  pathAndInput: string | [string, TInput?],
+  opts?: UseTRPCQueryOptions;
+)
+```
+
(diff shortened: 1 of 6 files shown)


www/docs/react/useQuery.md
```diff
@@ -1,15 +1,23 @@
...
 
-You pass a `[path, input]`-tuple as the first argument. You'll notice that you get autocompletion on the `path` and automatic typesafety on the `input`.
+```tsx
+function useQuery(
+  pathAndInput: string | [string, TInput?],
+  opts?: UseTRPCQueryOptions;
+)
+```
+
+- If the query has an optional `input`, you pass only it's procedure name as the first argument
```
mmkal, line 17: Hyper-nit:
```suggestion
- If the query has an optional `input`, you pass only its procedure name as the first argument
```

mmkal (commented):
If I'm honest, I personally don't think this should go in. I get that it does make trpc seem more perfect, but a couple of factors make it not worth it for me:

- the maintenance cost. Adding features which provide more concrete value to the end-user will now be harder. `useInfiniteQuery` needing special case type fanciness is a good example of this happening already.
- it only affects queries with no input. The vast majority of real-life queries aren't like this, in my experience. And almost no projects will have exclusively this kind of query. Which ties into the next point:
- the user is now burdened with having to learn about both modes, and this PR, effectively. Someone joining a team with an existing trpc codebase will look at the `useQuery` usages, and instead of intuitively figuring out that `useQuery` requires braces, they'll get confused and think that it only _sometimes_ requires braces, and that there's something beyond aesthetics behind it. It's another thing to learn, which is an adoption risk, I think. 
- refactoring cost. Minor, but adding an input will usually add visual noise to the code
- performance. This is pure conjecture, but I suspect every mapped type will make the check and compile times a little slower. Trpc is already chock full of them - which is okay, because they are providing *huge* value. But this change will slow my compiler down without enabling me to do anything new, besides saving two keystrokes on about one in ten queries

KATT:
> If I'm honest, I personally don't think this should go in. I get that it does make trpc seem more perfect, but a couple of factors make it not worth it for me:

Thanks for this! I appreciate it 🙏  - do you feel the same about #1058?

> * the maintenance cost. Adding features which provide more concrete value to the end-user will now be harder. `useInfiniteQuery` needing special case type fanciness is a good example of this happening already.

I may note here that this behaviour is the default behaviour in react-query so I'm simply adhering to that.

> * it only affects queries with no input. The vast majority of real-life queries aren't like this, in my experience. And almost no projects will have exclusively this kind of query. Which ties into the next point:

This is true!

> * the user is now burdened with having to learn about both modes, and this PR, effectively. Someone joining a team with an existing trpc codebase will look at the `useQuery` usages, and instead of intuitively figuring out that `useQuery` requires braces, they'll get confused and think that it only _sometimes_ requires braces, and that there's something beyond aesthetics behind it. It's another thing to learn, which is an adoption risk, I think.

Haven't thought about this. Great!


> * performance. This is pure conjecture, but I suspect every mapped type will make the check and compile times a little slower. Trpc is already chock full of them - which is okay, because they are providing _huge_ value. But this change will slow my compiler down without enabling me to do anything new, besides saving two keystrokes on about one in ten queries

True!

mmkal:
> do you feel the same about #1058?

Yes, I was pretty happy with the existing behaviour! I'd definitely noticed it - I thought "I wonder why" but got it when I saw it with options, and never had any complaints.

> I may note here that this behaviour is the default behaviour in react-query so I'm simply adhering to that.

Which behaviour do you mean specifically? If it's passing a bare string as the first arg, that's true but the string could conceivably be something different in an unopinionated react-query world. react-query only knows about the client, so supports something like

```js
const q1 = useQuery('q1', impl)
const q2 = useQuery('q2', impl)
```

Whereas in trpc the query cache key is bound to the procedure that's being called (I think?). Incidentally, adding support for two calls to the same procedure with different cache keys is the sort of feature this or #1058 would make a little harder to add.

The `useInfiniteQuery`/`cursor` changes sound valuable either way though, maybe they could go into a separate PR (which would also make the diff here smaller)?

KATT:
@all-contributors add @mmkal for code review

PR #1062, opened 2021-10-04
rename `invalidateQuery()` to `invalidateQueries()`

KATT:
- Better alignment with `react-query`
- Deprecates `invalidateQuery()`

### DIFF
--- a/packages/react/src/internals/context.tsx
+++ b/packages/react/src/internals/context.tsx
@@ -72,11 +72,20 @@ export type TRPCContextState<TRouter extends AnyRouter> = {
     >,
   ) => Promise<void>;
 
+  /**
+   * @deprecated use `invalidateQueries`
+   */
   invalidateQuery: <
     TPath extends keyof TRouter['_def']['queries'] & string,
     TInput extends inferProcedureInput<TRouter['_def']['queries'][TPath]>,
   >(
     pathAndInput: [TPath, TInput?],
+  ) => Promise<void>;
+  invalidateQueries: <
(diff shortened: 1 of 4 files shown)


packages/react/src/internals/context.tsx
```diff
@@ -26,6 +26,9 @@ export type TRPCContextState<TRouter extends AnyRouter> = {
   client: TRPCClient<TRouter>;
   isPrepass: boolean;
 
+  /**
+   * @link https://react-query.tanstack.com/guides/prefetching
```
mmkal, line 30: Hmm what's the difference between this and `prefetchQuery` below? Based on the link here, is it the same? If so maybe this one should be deprecated too.

www/vercel.json
```diff
@@ -15,6 +15,11 @@
       "destination": "/docs/react-queries",
       "permanent": true
     },
+    {
+      "source": "/docs/invalidateQueries",
```
mmkal, line 19:
```suggestion
      "source": "/docs/invalidateQuery",
```
?

packages/react/src/internals/context.tsx
KATT: It's a weird one but quite nice DX-detail.

`prefetchQuery` **never** `throw`s so it's very useful for optimistic prefetch where you don't care about caring about error state so useful for things like what I outlined in #914

www/vercel.json
KATT: It's renamed to `invalidateQueries` to adhere to react-query

KATT: OH!

good catch!

PR #1079, opened 2021-10-04
fix type instantiation depth for big routers

KATT:
## Issue

- Showcase some underlying issues leading to _`Type instantiation is excessively deep and possibly infinite.`_
- Had a call with @mgreenw today who said he forked typescript to circumvent this 😮
- ~Possibly not an issue and solved by TypeScript 4.5~ it is an issue in 4.5 as well, but a bit higher threshold
- Also the router seem to start losing type inference on first procedures' outputs after a while (cb24f03718593f1f5377f982b1903049047cb8e4)
- With  TypeScript `^4.5.0-beta` the roof seems to be `65` for this to fail - b3a4356e36af3c808436e8ef3c6451eaaf7b4dea

## 🎉  Increased it to work with `> 500` procedures

`flatten<X, Y>` was the culprit!

### Changes

- add codegen for a big router + test assertions
- deleted `flatten` & other evil typescript utility methods that i barely understand

### How to test locally

- Change the local number `NUM_PROCEDURES_TO_GENERATE`, then run ` yarn codegen && yarn test --testPathPattern bigRouter`
- With  TypeScript `^4.5.0-beta` the roof seems to be `65` for this to fail - b3a4356e36af3c808436e8ef3c6451eaaf7b4dea


packages/react/src/createReactQueryHooks.tsx
```diff
@@ -81,13 +81,22 @@ type inferInfiniteQueryNames<TObj extends ProcedureRecord<any, any, any, any>> =
...
+
 export function createReactQueryHooks<TRouter extends AnyRouter>() {
   type TQueries = TRouter['_def']['queries'];
-  type TMutations = TRouter['_def']['mutations'];
   type TSubscriptions = TRouter['_def']['subscriptions'];
   type TError = TRPCClientErrorLike<TRouter>;
   type TInfiniteQueryNames = inferInfiniteQueryNames<TQueries>;
 
+  type TQueryValues = inferProcedures<TRouter['_def']['queries']>;
+  type TMutationValues = inferProcedures<TRouter['_def']['mutations']>;
```
KATT, line 98: Without this optimization `useQuery()` wouldn't work with 40 queries

scripts/generate-fixtures.ts
```diff
@@ -0,0 +1,30 @@
+import fs from 'fs';
+import path from 'path';
+function getBigRouter() {
+  const str = [`trpc.router()`];
+  for (let num = 1; num <= 40; num++) {
```
KATT, line 5: Increasing this number to over `40` will fail the build

KATT (commented):
cc @mgreenw @mmkal

alii (approved):

heybereket (approved):

package.json
```diff
@@ -34,7 +34,9 @@
     "test": "cd packages/server && yarn jest",
     "lint": "eslint --ext \".js,.ts,.tsx\" --ignore-path .gitignore .",
     "lint-fix": "yarn lint --fix && manypkg fix",
-    "postinstall": "preconstruct dev",
+    "codegen": "yarn ts-node scripts/generate-big-f-router.ts",
+    "ts-node": "ts-node --compiler-options \"{\\\"module\\\":\\\"commonjs\\\"}\"",
```
mmkal, line 38: ts-node options might be able to go in tsconfig these days. IMO transpileOnly should be turned on too, since there are already CI checks for type correctness. Failing that, I think there could be one fewer escape with single quotes
```suggestion
    "ts-node": "ts-node --compiler-options '{\"module\":\"commonjs\"}'",
```

packages/server/src/router.ts
```diff
@@ -283,10 +280,8 @@ export class Router<
   ): Router<
     TInputContext,
     TContext,
-    flatten<
-      TQueries,
-      Record<TPath, inferProcedureFromOptions<TInputContext, typeof procedure>>
-    >,
+    TQueries &
+      Record<TPath, inferProcedureFromOptions<TInputContext, typeof procedure>>,
```
mmkal, line 284: I wonder if this breaks routers which redefine queries with a duplicate path, since flatten made the lhs "win"? Probably a good thing if so since it would almost always be a mistake, maybe?

KATT: Duplicate queries are already handled and throws an error

KATT: https://github.com/trpc/trpc/blob/2c1341f182df825509bcb7bd546fd724fd1f58a6/packages/server/src/router.ts#L486

KATT: Let me see though

KATT: Duplicate queries should already throw an error!

KATT:
@all-contributors add @mgreenw for bug reporting

PR #1124, opened 2021-10-15
add Vue (Nuxt.js 3)  example

KATT:
>  Currently blocked by #1129 

Adds Nuxt.js 3 example.

## Todo

- [x] Fix `yarn dev`
- [x] Fix `yarn build`
- [x] Get rid of [ugly long `import`](https://github.com/trpc/trpc/pull/1124/files/d0e0ea1d138f52fbe125459474a7a6aaaa322d69#diff-ca4c9db10cf79546f3a70276cfe6dfb5ad90b231961b652190e6dc3553a22c0c)
- [x] ❓ `.vscode/` recommendations?
- [ ] Fix ESModules <-> CommonJS interop - #1129
- [ ] ❓ `/post/[id]` detail page?
- [ ] ❓ Add Prisma & superjson?

## How-to try it out

For anyone who'd like to help:

```bash
git clone git@github.com:trpc/trpc.git
cd trpc
git checkout katt/nuxt
yarn
cd examples/nuxt
yarn dev
```

PR #1127, opened 2021-10-16
deprecate multiple entrypoints in `@trpc/server`

KATT:
- Experiment to delete entrypoints of `@trpc/server` and prefer using i.e. `import { nextjsAdapter } from "@trpc/server"` over `import * as nextjsAdapter from "@trpc/server/adapters/next"`.
- Background - Better interoperability with ESModules: https://github.com/trpc/trpc/pull/1124/files/d0e0ea1d138f52fbe125459474a7a6aaaa322d69#r730175908
- Multiple entrypoints on node.js projects might be unnecessary (bundle sizes aren't sensitive) and since we don't have any explicit dependencies in any of them they should work anyway. We have only implicit by depending on the type declarations only.
- Multiple entrypoints is really fucking hard to get right with ESModules + Webpack + TypeScript (that isn't actually supporting it until 4.5)

---

- It seems to work fine, I also tested with a package outside of the project: https://github.com/KATT/typescript.careers/pull/66
- Hopefully, this change is not needed by somehow defining `exports` in `package.json`

---

## Todos

- [ ] Deprecation warning on old `@trpc/adapters/*`-imports
- [ ] Update docs
- [ ] Update reference projects (follow-up PR)
- [ ] Release notes with migration path

### DIFF
--- a/examples/express-server/tsconfig.json
+++ b/examples/express-server/tsconfig.json
@@ -13,6 +13,7 @@
     "experimentalDecorators": true,
     "sourceMap": true,
     "resolveJsonModule": true,
+    "skipLibCheck": true,
     /* Strict Type-Checking Options */
     "strict": true /* Enable all strict type-checking options. */,
     "noImplicitAny": true /* Raise error on expressions and declarations with an implied 'any' type. */,


KATT:
Fixed by c8f2379

PR #1129, opened 2021-10-16
Add ESModules interoperability in Node.js

KATT:
- Adds support for ESModules interoperability
- Required for #1124 

## State of affairs

See response from preconstruct-authors at https://github.com/preconstruct/preconstruct/issues/251#issuecomment-945085478

### `"exports": "{{string}}.cjs.js"`

https://github.com/trpc/trpc/blob/ba96ce8f51fbf7e018c92f19fa81a27fa272b75f/packages/client/package.json#L23-L31

- Seemingly works OK at first glance
- Breaks TypeScript imports when using as a dep - see build connected to [this(https://github.com/KATT/typescript.careers/pull/70/commits/474aaface716a1040e329bfe2cc941c59960a6d8)

## `"exports": {...}`

https://github.com/trpc/trpc/blob/dc43a35e69261eee89aebfb25a335466ef5bf1ba/packages/client/package.json#L23-L61

- Doesn't work in Next.js without the TypeScript 4.5 beta
- Nuxt still wants to import it as CJS because (I think) it's not an `.mjs`-ending


KATT:
> This looks good to me. Only note that there may some issues with older version of TypeScript (the latest beta does understand `exports`), and anyone consuming this library with webpack4 may also struggle. I'll add an example of an escape hatch if you want one.

Thanks for the feedback! Curious of the escape hatch.

- tRPC requires `>= 4.1.0` already.
- Will do a sanity check with Webpack 4 - but that's what it already relies on for all Next.js-stuff already - webpack is a bit smarter on this than Node.js it seems

sachinraja:
Just a note: this PR is not adding esm support, it's just forcing users to use cjs.

KATT:
Fuck me - total catch-22 scenario.

- `"exports": "<<string>>"` fucks webpack - proven in https://github.com/KATT/typescript.careers/pull/70
- `"exports": {...}"` fucks something in Nuxt.js - it sems to think it's importing a CJS module but it's actually importing a ESM-module

KATT:
You were both right - I'm at complete loss how to fix it. Right with this solution the Vuxt can actually build but `yarn dev` is broken in #1124.

KATT:
Maybe a solution is to make tRPC ESM-only -- anyone using it will use TypeScript which should(?) resolve it for you

sachinraja:
> "exports": "<<string>>" fucks webpack

I would think webpack would resolve that just fine. What was the issue?

KATT:
An alternative would also be to just skip using multiple entrypoints for now until all of these ESM-issues are ironed out (#1127) but might still be a problem if one wanted to use `@trpc/client/[something]` in an ESM-env

sachinraja:
Not sure why the types are erroring, but 
> but might still be a problem if one wanted to use `@trpc/client/[something]` in an Node.js ESM env

Using cjs is not a problem in a Node.js esm env. That import should work fine.

sachinraja:
Got a few ideas for how to advance this in no particular order:
* Stay with preconstruct and wait for [preconstruct/preconstruct#435](github.com/preconstruct/preconstruct/pull/435) to be merged, which should allow for ESM. Personally I don't really like how preconstruct imposes a lot of control over your directory structure and the output structure is a little odd to me, but others may feel differently.
* Switch to [unbuild](https://github.com/unjs/unbuild) (ironically a Nuxt contrib project), which has the same passive watcher system with stubs that preconstruct uses but supports ESM with multiple entry points and does not control your directory structure. It will emit warnings for a few things if they are out of place though. It also supports bundling with Rollup (including TS declarations). [This is what the new Blitz.js toolkit uses](https://github.com/blitz-js/blitz/pull/3129).
* Directly import the packages into the examples and alias the packages to their actual source. Use something like Rollup for a production build. [This is what I am currently doing in tRPC Playground](https://github.com/sachinraja/trpc-playground/blob/main/apps/vite/vite.config.ts), but it may not be the best for tRPC's (development) use case.

KATT:
> Got a few ideas for how to advance this in no particular order:
> 
> * Stay with preconstruct and wait for [preconstruct/preconstruct#435](//github.com/preconstruct/preconstruct/pull/435) to be merged, which should allow for ESM. Personally I don't really like how preconstruct imposes a lot of control over your directory structure and the output structure is a little odd to me, but others may feel differently.
I quite like it as I don't have to think about it too much myself

> * Switch to [unbuild](https://github.com/unjs/unbuild) (ironically a Nuxt contrib project), which has the same passive watcher system with stubs that preconstruct uses but supports ESM with multiple entry points and does not control your directory structure. It will emit warnings for a few things if they are out of place though. It also supports bundling with Rollup (including TS declarations). [This is what the new Blitz.js toolkit uses](https://github.com/blitz-js/blitz/pull/3129).

Create idea. Would you want to have a stab at it?

> * Directly import the packages into the examples and alias the packages to their actual source. Use something like Rollup for a production build. [This is what I am currently doing in tRPC Playground](https://github.com/sachinraja/trpc-playground/blob/main/apps/vite/vite.config.ts), but it may not be the best for tRPC's (development) use case.

Unsure about this, feels a bit dirty?

sachinraja:
> Would you want to have a stab at it?

Definitely! Will send in a PR soon.

> Unsure about this, feels a bit dirty?

Yeah, I think it's better for frontend projects though (the HMR is nice).

KATT:
> > Would you want to have a stab at it?
> 
> Definitely! Will send in a PR soon.

Can you do it against the `next` branch? That's going to be the new active development branch

KATT:
I probably won't add support for ESM until v10

sachinraja:
> I probably won't add support for ESM until v10

Makes sense, it'll probably be a breaking change anyway because adding `exports` restricts what you can import from the package.

sachinraja:
Should this be closed as it's been done for v10 (implemented on the `next` branch)?

PR #1172, opened 2021-10-26
[RFC] Easier reusability in example for input/output inference

KATT:
I want to ship some nicer ts utility in the example apps for using input/output from the API - I find writing `inferQueryOutput<'myQuery'>` quite fiddly and repetitive - it'd be easier to have pre-set record where I could just use what I want straight-away.

> Both implementations are part of this PR if you want to play with them yourself.

## Current / Option 0 - `inferQueryOutput<'x'>` / `inferQueryInput<'x'>`

```tsx
import { inferQueryOutput, inferQueryInput } from 'utils/trpc';

type Post = inferQueryOutput<'post.byId'> // or `inferQueryInput['post.byId']` for getting input
function PostItem(props: { post: Post }) {
 // [..]
}
```

## Option 1 - `TQueries`

```tsx
import { TQueries } from 'utils/trpc';

type Post = TQueries['post.byId']['output'] // or `TQueries['post.byId']['input']` for getting input
function PostItem(props: { post: Post }) {
 // [..]
}
```

## Option 2 - `TQueryInputs` / `TQueryOutputs`

```tsx
import { TQueryOutputs, TQueryInputs } from 'utils/trpc';

type Post = TQueryOutputs['post.byId'] // or `TQueryInputs['x']` for getting input
function PostItem(props: { post: Post }) {
 // [..]
}
```
## (✨ New) Option 3 using a shadow object

- Suggested by @mmkal [here](https://github.com/trpc/trpc/pull/1172#issuecomment-952881784)

```tsx
import { trpc } from '../utils/trpc'

type Post = typeof trpc.types['post.byId'].output; // or:
type Post = typeof trpc.types.post.byId.output;
function PostItem(props: { post: Post }) {
 // [..]
}
```

## (✨ New) Option 4 - template literal strings

> `type TProcedures could` would something like `${'query' | 'mutation' | 'subscription'}.${TPath}.${'input' | 'output' | 'inputParsed'}`

- Autocompletion on all different values straight away

```tsx
import { TProcedures } from 'utils/trpc';

type Post = TProcedures['query.post.byId.output']
function PostItem(props: { post: Post }) {
 // [..]
}
```
Screenshot 2021-11-03 at 16 50 41

## Option XX - suggestions?

Any other good ideas?


mgreenw:
+1 for "Option 1 - TQueries" -- I like only having to import one helper type

KATT:
I had another idea yesterday 💡  - how about we could to full out on string literals?

We could do something like this:

```tsx
import { TProcedures } from 'utils/trpc';

type Post = TProcedures['query.post.byId.output']
function PostItem(props: { post: Post }) {
 // [..]
}
```

This would also get you instant autocompletion of all the available things to select.

`type TProcedures could` would something like `${'query' | 'mutation' | 'subscription'}.${TPath}.${'input' | 'output'}` (and potentially `| 'inputParsed'` after #1189)

KATT:
> One other option would be to add some compile time only props to the helper exported from `@trpc/react` (and/or `@trpc/client` and/or `@trpc/server` even), then take advantage of `typeof`. Then, the helper is the only thing that needs to be imported by components, plus no awkward `<'...'>` or `['...']` prop syntax:
> 
> ```ts
> import { trpc } from '../utils/trpc'
> 
> const Profile = ({ user }: { user: typeof trpc.types.getUser.output }) => {
>   return <div>{user.name}</div>
> }
> 
> export default () => {
>   const userQuery = trpc.useQuery(['getUser', {}])
>   
>   return userQuery.data
>     ? <Profile user={userQuery.data} />
>     : <div>no user</div>
> }
> ```
> 
> Downside: it implies that you could actually use `trpc.types.getUser.output` to do something at runtime but I don't think that's likely to trip people up.

I quite like this - however, it's quite hard to pull-off without having sub-routers to be end up something like `type Post = typeof trpc.types['post.byId'].output`

kimroen:
I was looking for exactly this last week and didn't find `inferQueryOutput`, so at the very least there might be some IA issue here that meant I didn't find it.

Of these options, I like the last one @KATT shared here. Naming-wise, I also like that one because including "query" in the name kind of implies you'd also need a "mutation" version, and I don't think that's necessarily true.

In case it's useful, when I was looking for this I was imagining I'd find something like this (both imported from `trpc` itself):

```ts
type Post = TReturnType<AppRouter, 'getPost'>
type CreatePostInput = TInputType<AppRouter, 'createPost'>
```

or as something imported from utils:

```ts
type Post = TReturnType<'getPost'>
type CreatePostInput = TInputType<'createPost'>
```

Which is fairly similar to what we already have before this PR.

mmkal:
> I quite like this - however, it's quite hard to pull-off without having sub-routers to be end up something like `type Post = typeof trpc.types['post.byId'].output`

Good point. Although VSCode still gives helpful autocomplete for non-identifier-style keys even if using `.`:

image

When you select an option it adds the `['...']` automatically:

image

My nit against anything involving `AppRouter` is that I usually want this as props to some react component, and importing an `AppRouter` there feels wrong - I don't care about routing, I just want to use the output type of an API. That's why I like the idea of piggy-backing on `import { trpc } from '../utils/trpc'` which I'm already doing, usually.

KATT:
@mmkal what's your feeling on string literals (option 4 above) / [this](https://github.com/trpc/trpc/pull/1172#issuecomment-956099480) comment?

mmkal:
Tbh my feeling is it's a little _too_ clever, I'm not a huge fan of template literal type tricks unless they enable doing something otherwise impossible (like trpc server router merging, for example). If nothing else, by virtue of seeming weird it could put people off trpc. But it'd certainly do the trick!

PR #1188, opened 2021-10-29
safe wrapper around `TRPCClient` to return `Either<>`-style responses

KATT:
Not intended to be merged, just a demo from https://github.com/trpc/trpc/pull/1188


danielyogel (commented):
Interesting approach!
May use this my self, but still different than explicitly returning expected errors as a kind in a union type (and transforming any thrown Error be one of the union’s kind) - in two ways: 
1. With inferred Either, a route may return only a subset of tRPC errors, so less branches may be handled in the client than required in the wrapSafe way. 
2. A route may throw different errors than what wrapSafe casts to, which may cause a bug…

So, because the “inferred union” has some disadvantages as well, still haven’t decided which way to use in the future…

KATT:
Closed this, do what you want with it @danielyogel

PR #1189, opened 2021-10-29
Input schemas defined with Zod does not use the inferred input type

### LINKED ISSUE #1182: Input schemas defined with Zod does not use the inferred input type
### Summary
```ts
trpc.router()
  .query('greeting', {
    input: z.object({
      hi: z.string().transform(s => s.length),
    }),
    resolve({ input }) {
      input.hi // => `number`, as expected
    },
  });

// Somewhere else

client.query('greeting', { hi: 'my greeting' }); // => ERROR: expected `number`
```

[CodeSandbox](https://codesandbox.io/s/heuristic-leakey-bjd51?file=/src/client.ts)

When defining an input schema using zod, I would expect the call site to accept the `z.input` version of the schema.

### Details

When you define a schema with zod, like this simple one:

```ts
const GreetingSchema = z.object({
  hi: z.string(),
});
```

You can infer its type with `z.infer` like this:

```ts
type Greeting = z.infer<typeof GreetingSchema>;

// Equals

type Greeting = {
  hi: string;
}
```

However, in zod you can also add transformers that change the type of a field, which means the input and output types of the schema are different.

Here's an example of that:

```ts
const GreetingSchema = z.object({
  hi: z.string().transform(s => s.length),
});
```

`z.infer` infers the *output* type of this:

```ts
type InferredGreeting = z.infer<typeof GreetingSchema>;

// Equals

type InferredGreeting {
  hi: number;
}
```

To get at the input type, zod also has a `z.input` helper (and `z.output`, which `z.infer` is an alias of), meaning you can get at the version we want:

```ts
type InputGreeting = z.input<typeof GreetingSchema>;
type OutputGreeting = z.output<typeof GreetingSchema>;

// Equals

type InputGreeting {
  hi: string;
}

type OutputGreeting {
  hi: number;
}
```

This is also mentioned [in the README of Zod](https://github.com/colinhacks/zod#what-about-transforms).

This concept applied to tRPC:

```ts
trpc.router()
  .query('greeting', {
    input: z.object({
      hi: z.string().transform(s => s.length),
    }),
    resolve({ input }) {
      input.hi // => `number`, as expected
    },
  });

// Somewhere else

client.query('greeting', { hi: 'my greeting' }); // => ERROR: expected `number`
```

In other words, for tRPC I would expect the type required to call a function to be the `z.input` type, and the type available in the `resolve` function to be the `z.output` type.

KATT:
Closes #1182

Will make something like making a REST-adapter a lot easier as one could do `param: z.union([z.string(), z.number()]).transform(v => Number(v))` to handle query params.

**tl;dr** Allows us to do this below:

```tsx
trpc.router()
  .query('greeting', {
    input: z.object({
      hi: z.string().transform(s => s.length),
    }),
    resolve({ input }) {
      input.hi // => `number`, as expected
    },
  });

client.query('greeting', { hi: 'my greeting' }); 
```


KATT:
cc @mmkal - I merged this as it works, I'd love to see you having a stab at simplifying it if you feel inclined some day 😅

kimroen:
Oh perfect! I was just trying to get this running locally so I could verify that it works and give some feedback here, but I couldn't figure out how to run it 🙃  I'll try this now!

PR #1193, opened 2021-10-30
an example of a more fun SSG helper

KATT:


examples/next-prisma-starter/src/pages/post/[id].tsx
```diff
@@ -1,21 +1,60 @@
...
+  });
+
+  return {
+    props: {
+      trpcState: ssg.dehydrate(),
+      id,
+    },
+    revalidate: 1,
+  };
+});
```
Comment (KATT) [inline examples/next-prisma-starter/src/pages/post/[id].tsx:48]: cc @mmkal

www/docs/nextjs/ssg.md
```diff
@@ -14,41 +14,83 @@ Static site generation requires executing tRPC queries inside `getStaticProps` o
...
+import superjson from 'superjson';
+import { trpc } from 'utils/trpc';
 
-// Statically fetch the data in getStaticProps
-export const getStaticProps = async (
-  context: GetStaticPropsContext<{ filter: string }>,
-) => {
-  const ssg = createSSGHelpers({
+
+export async function getStaticProps(
```
mmkal, line 32: Nit: should this go after `getStaticPaths`? Since the paths param is an input to it.

www/docs/nextjs/ssg.md
```diff
@@ -14,41 +14,83 @@ Static site generation requires executing tRPC queries inside `getStaticProps` o
...
+  const ssg = await createSSGHelpers({
     router: appRouter,
-    transformer, // optional
     ctx: {},
+    transformer: superjson, // optional - adds superjson serialization
+  });
+  const id = context.params?.id as string;
+  await ssg.fetchQuery('post.byId', {
+    id,
   });
```
mmkal, line 43: I wonder if it's worth encouraging a kind of ad-hoc pattern here, or thinking about whether trpc could help in some way making sure what's statically fetched matches what's passed to `useQuery` later. e.g.

```ts
const postByIdQueryParams: trpc.SomeHelperType<'post.byId'> = params => ['post.byId', params]

//...

export const getStaticProps = async context => {
  //...
  await ssg.fetchQuery(...postByIdQueryParams(context.params))
  //...
}

export default function PostViewPage(props) {
  const postQuery = trpc.useQuery(postByIdQueryParams(props))
  //...
}
```

Having written that... it's a bit much for this PR.

PR #1199, opened 2021-11-02
make `@trpc/*` deps peer deps

KATT:


packages/next/package.json
```diff
@@ -30,16 +30,16 @@
     }
   },
   "peerDependencies": {
+    "@trpc/client": "^9.12.1-alpha.0",
+    "@trpc/react": "^9.12.0",
+    "@trpc/server": "^9.12.0",
```
KATT, line 35: didn't get bumped automatically :(

PR #1226, opened 2021-11-08
service oriented / microservices approach 

KATT:
# Service Oriented Architecture (SOA)

Example of a Service Oriented Architecture.

Uses two different servers, but they are combined through a custom link.


anthonyshort:
This is pretty interesting. The way we've done this is by just having multiple trpc clients. We typically only want there to be one BFF layer that we're going through, then we have some shared APIs as a second endpoint. We've found that once we made the process of setting up providers and trpc clients super easy it felt cleaner to just do that rather than putting it into one client.

Here's an example:

```ts
export const TRPCProvider = createProvider(() => {
  const getToken = useAccessTokenCallback(); // From our auth provider

  return {
    appClient: createTRPCClient<AppRouter>({
      url: '/api/trpc',
      getToken,
    }),
    anotherClient: createTRPCClient<AppRouter>({
      url: '/api/other',
      getToken,
    }),
  }
});
```

Then in the components:

```ts
const { appClient } = useProvider(TRPCProvider);

const response = useQuery(appClient, ['greeting', {
  name: 'Tom'
}])
```

KATT:
cc @anthonyshort 

In theory, the two should be the same in terms of the requests, the approach I'm suggesting above will just feel a bit more seamless. As long as you don't need to have auth headers that differ on the two, the above would work great.

> ```ts
> const { appClient } = useProvider(TRPCProvider);
> 
> const response = useQuery(appClient, ['greeting', {
>   name: 'Tom'
> }])
> ```

How have you made `useQuery` type-safe in your implementation? Seems a bit tricky if you haven't dug around in the internals.

KATT:
Will close this with the icebox label as I prefer non-SOA-oriented approaches generally :) was mainly curious if it was feasible in an easy way!

KATT:
See #2986 for a v10 variant.

Issue #1254, opened 2021-11-12
[TRP-62] Kitchen Sink Wishlist


KATT:
# Kitchen Sink

- **Live Site:** https://kitchen-sink.trpc.io/
- **Source:** https://github.com/trpc/examples-kitchen-sink

The goal of the Kitchen Sink is to provide a good collection of useful design patterns when working with tRPC. It should be easy to navigate [kitchen-sink.trpc.io](https://kitchen-sink.trpc.io) and have it as a reference site — the northern star would be the [Tailwind Docs](https://tailwindcss.com/docs).

I will probably replace the main website with this site & I want to experiment with monetization. Maybe this could be the "tailwindui" of trpc. 

### Overall site improvements

- [ ] Nicer home page
- [x] Nicer code example pages
- [ ] Interactive code examples with [Monaco](https://github.com/suren-atoyan/monaco-react) or similar
- [x] Implement code syntax highlighter for source code
- [x] Copy to clipboard 
- [x] Nicer view, sort of (preview/code tab bar?) 
- [ ] Link / embed / generate codesandbox on each example? 
- [ ] Maybe some nicer way of handling features? Some code generation to make it easier to work on one feature in isolation?
- [ ] Login with GitHub

### Features

- [x] `getStaticProps` [🎉 Implemented](https://kitchen-sink.trpc.io/ssg)
  - [x] Basic examples
  - [ ] Improve with splitting up into a "list" page and a `[id].tsx` page
- [ ] getServerSideProps 
- [ ] Infinite queries
- [ ] Paginated pages 
- [ ] Optimistic prefetching
- [ ] Good patterns with Suspense
- [ ] Error handling
- [x] Integration with react-hook-form -[ 🎉 Implemented](https://kitchen-sink.trpc.io/react-hook-form)
  - [x] Reusing validation schema server/client
  - [x] Invalidate queries
- [ ] API response caching
- [ ] `QueryCell` pattern - (#1734)
- [ ] Integration testing
- [x] User auth
- **... + 🙏  ideas welcome!**

### Stretch goals

- Make this into the actual trpc.io main site?
- Potentially I'll limit view of examples to contributors & sponsors - maybe it's possible to make money off OSS?

### Clone and start the Kitchen sink:

```bash
git clone git@github.com:trpc/trpc.git
cd trpc/examples/kitchen-sink
yarn && yarn dx
```

### Adding an example

1. Create a new folder in `src/feature/X` by duplicating some of the existing ones
[...]

PR #1256, opened 2021-11-13
Minor changes to nextjs-prisma starter

Nutlope:
- No need to include `next.config.js` if you have no custom configuration
- Noticed you had a lib folder with two files that weren't being used anywhere so I deleted it

### DIFF
--- a/examples/next-prisma-starter/next.config.js
+++ b/examples/next-prisma-starter/next.config.js
@@ -1,4 +1,0 @@
-/**
- * @link https://nextjs.org/docs/api-reference/next.config.js/introduction
- */
-module.exports = {};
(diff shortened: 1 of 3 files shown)


KATT:
Thanks for the PR, but I want to keep them there for reference & I need at least one test to see that jest runs correctly. 

Maybe that sum file should have a comment in it that it can be deleted.

Next config often needs editing eventually, then it's nice to just have it there.

KATT:
If you're keen to contribute, check out #1254 which I started yesterday, could really needs some eyes/hands/ideas there. ☺

Nutlope:
> **Edit:** I will take this into consideration and maybe end up merging it once I have given it some more thought. Maybe included test could be changed to be more useful -- i.e. doing an integration test of the backend or something.

Makes total sense! Feel free to close or do as you see fit :)

May take a look at #1254 once I get a little more comfortable with trpc. Seems pretty cool at first glance!

KATT:
Closing this - I quite like having them there even if I should make them more useful

PR #1275, opened 2021-11-17
alias `useMutation([path])` with `useMutation(path)`

KATT:
DX Improvement suggestion -

- `useQuery` uses a `useQuery([path, input?])`-tuple
- I personally trip on mutations and keep on writing `useMutation([path])` by accident all the time
- Would lead to less hiccups (I think) to just make them the aliased on `useMutation` as well

~Should we update all docs to refer to `useMutation([path])` too?~ done


simonedelmann:
Hm... I do get the point, but I am not sure if this improves the DX or makes it worse. (People new to tRPC will only know about that syntax then...) 

I personally am fine with the current solution and I think I would continue to use the old syntax (less finger sport). But honestly I don't mind that much. 

I guess it is not possible to do `useQuery(path, props, ...)` for some reasons?

peterp:
We've just started using the client side library, and I do recall getting blocked on the difference between the query vs mutation params.

KATT:
> Hm... I do get the point, but I am not sure if this improves the DX or makes it worse. (People new to tRPC will only know about that syntax then...) 

Consistency is good! If I as the creator trip on it, I'm sure others do too 

> I personally am fine with the current solution and I think I would continue to use the old syntax (less finger sport). But honestly I don't mind that much. 

Obviously it'll be backwards compat. 

> I guess it is not possible to do `useQuery(path, props, ...)` for some reasons?


See #1060

PR #1313, opened 2021-11-26
remove `CACHE_KEY_*` suffixes in `react-query` cache

### LINKED ISSUE #1312: Get rid of `CACHE_KEY_QUERY`,  `CACHE_KEY_INFINITE_QUERY` suffixes
I've always added `CACHE_KEY_QUERY`, `CACHE_KEY_INFINITE_QUERY` to `react-query`'s query cache to prevent collisions of things that are fetched as an infinite query can't collide.

Turns out, this creates quite a bit of ugliness and tech debt all around tRPC so I am pretty sure I want to remove it. I don't think this should be a responsibility of tRPC and rather live in react-query.

KATT:
I added `CACHE_KEY_QUERY`, `CACHE_KEY_INFINITE_QUERY` to `react-query`'s query cache to prevent collisions if you fetch the same procedure as an infinite query and as a normal query. This was a bad idea, it creates quite a bit of ugliness and tech debt all around tRPC want to remove it. I don't think this should be a responsibility of tRPC and rather live in react-query (or in developer's implementation).

---

- This PR removes a lot of complexity and enables our `react-query` wrappers to truly be type sugar on top of `react-query`.
- ⚠️⚠️  Will publish as a minor even though it might be a breaking change in some rare cases where people tamper with the react-query cache manually. ⚠️⚠️

closes #1312


packages/react/src/internals/context.tsx
```diff
@@ -7,16 +7,16 @@ import type {
 } from '@trpc/server';
 import { createContext } from 'react';
 import {
+  CancelOptions,
```
mmkal, line 10: Was this file changed deliberately?

mmkal (commented):
LGTM - does this mean useQuery and useInfiniteQuery can take advantage of each other's cached results? 👍 If so

packages/react/src/internals/context.tsx
KATT: I feature creeped some other stuff into this that I moved out to #1319, the only thing left is that the imports are alphabetically sorted -- I don't care enough to have that perfect

KATT:
> LGTM - does this mean useQuery and useInfiniteQuery can take advantage of each other's cached results? 👍 If so

It should, but react-query might cache useInfiniteQuery and useQuery a bit different... I think they might do, so it might be a problem. I just did a prerelease, please try it!

```

Successfully published:
 - @trpc/client@9.15.1-remove-cache-keys-2021-11-26-14-35-08.12+81442243
 - @trpc/next@9.15.1-remove-cache-keys-2021-11-26-14-35-08.12+81442243
 - @trpc/react@9.15.1-remove-cache-keys-2021-11-26-14-35-08.12+81442243
 - @trpc/server@9.15.1-remove-cache-keys-2021-11-26-14-35-08.12+81442243
```

TheoBr (approved):
Looks solid to me, we don't use `queryClient` directly too often but this wouldn't break our 5ish uses :)

KATT:
Just waiting for confirmation from @mmkal that it didn't break his app and then I'll merge this and #1319 

This PR will simplify making tRPC feature-complete with react-query.

anthonyshort (approved):
This won't affect any use cases for us across all of our apps, so LGTM :)

mmkal (approved):
I haven't tried this on my team's codebase but we don't use infinite queries yet and don't have any plans to so I don't want to hold this up - go for it! If it's a worrisome change, it might be worth investigating and documenting exactly what cache keys react-query uses by default, but personally I think the test coverage is good enough.

Issue #887, opened 2021-09-05
[RFC] Next major overview


KATT:
So. We're making a new major. 

## Goals

- **More ergonomic API for creating procedures** and building out your backend
- **CMD+Click** from your frontend and jump straight into the backend procedure. This will work with `react-query` as well!
- **Better scaling** than the current structure - the TypeScript server starts choking a bit when you get close to 100 procedures in your backend
- **Fully backward compatible** for the 99% use-case with a migration path which you can do incrementally.  

## Todos

> **Important for contributors:** All of the work here is related to the [`next`](https://github.com/trpc/trpc/tree/next) branch.

Tracked in https://github.com/orgs/trpc/projects/1

- [x] New Router API
  - [x] Tracked in [trpc/v10-playground](https://github.com/trpc/v10-playground) - the current winning proposal is https://github.com/trpc/v10-playground/pull/33
  - [x] Error formatting in new API - [`formatError()`](https://trpc.io/docs/error-formatting)-equivalent. 
  - [x] `meta`-equivalent
  - [x] Maybe we need a `createRootRouter()` that additionally accepts `{ transformer: X, formatError: Y}`
  - [x] Implement new API in tRPC
  - [x] Make old routers compatible (`.legacy()` at the end to transform them to the new version)
- [x] New client API with backward compatibility 
  - Maybe old can be `client.query('postById', input, options)` whilst new stuff can be `client.queries.postById({ input, ...options})`
- [x] New react-query API with backward compat
- [ ] Make actual issues for all of these
- [x] Add all old projects in a folder (copy/paste `examples/*` to `examples/.test` so we can track breaking changes easily and make sure we remain backward compatible for a long time
- [ ] Rewrite `TRPCClient` to be functional instead of a class but reuse the same interface
- [ ] Write a migration guide
  - [ ] New router API
  - [ ] New links
- [x] Move transformers back to `runtime` to make it backward compatible
- [ ] Explore: Add [`stripInternal`](https://www.typescriptlang.org/tsconfig#stripInternal)-flag
- [x] Simplify HTTP-responses
- [x] New links architecture (#1366)
- [ ] Refactor observables implementation to have `unknown` errors or make it guaranteed to be `TRPCClientError` of some sort. 
- [ ] Make transformers dynamically opt-in similar to how a batch request is indicated by `?batch=1` maybe a transformed request could be `?transform=1` or `transform=TRANSFORMER_NAME`?
- [ ] Write a blog post about new major and API.
[...]

PR #1366, opened 2022-01-08
New link architecture

KATT:
> **Breaking change** for next major! Full WIP of ideas for next major in #887 

## New link architecture

- Heavily inspired by RxJS
- Each link needs to return an observable which is responsible for subscribing to `next()` observable - https://www-git-katt-observable-new-trpc.vercel.app/docs/links#creating-a-custom-link
- The data in the links also receive "Error responses" from the backend - they aren't considered "errors" by the links themselves, that's up to the interpretation of the `TRPCClient`
- The data in the links are now wrapped in response object `{ context, data }` to links that can add metadata (like the original `Response`) to
- Data transformers are moved to being a link - https://www-git-katt-observable-new-trpc.vercel.app/docs/data-transformers#3-add-to-createtrpcclient-or-setupnext-config
- Add new `setupTRPC()` method to `packages/next` to simplify setup - https://www-git-katt-observable-new-trpc.vercel.app/docs/nextjs#3-create-trpc-hooks

### What it (hopefully) enables

- Should enable us to write a `@trpc/react-lite`-package that doesn't depend on `react-query` where each feature is composable by a link. **This is not the goal in itself**, but it's good that the architecture now (should) allow it.
- Enable better integration with other frameworks that might not have stuff like `react-query`
- More power for people using `@trpc/client`
- Ability to access response metadata (#1353)

### Open questions

- Should the `next()`-call in a link maybe be called `forward()` instead to not to confuse with observables' next concept?
- Should we just use RxJS instead of having our own implementation?
- Should response errors be errors in the chain? Could be a `TRPCResponseError` - might work easier with things like retry

### Notes from @KATT

- I'm quite new to observables/RxJS, so keen to get feedback on people who've used it more

### Q&A

> This is a **fat** PR - where can I look to get a feel of it?

- https://github.com/trpc/trpc/blob/katt/observable-new/packages/client/src/links/dedupeLink.ts
- https://github.com/trpc/trpc/blob/katt/observable-new/packages/client/src/links/retryLink.ts

> OK, cool, where's the core juice at that I can review?

- https://github.com/trpc/trpc/blob/katt/observable-new/packages/client/src/rx
- https://github.com/trpc/trpc/blob/katt/observable-new/packages/client/src/links/internals/createChain.ts

> Why is it not [ES Observable](https://github.com/tc39/proposal-observable)-compliant?

- I hate `this` in JS and try to avoid it all cost.
- _If_ we think this a good idea further on, backwards compatibility should be achievable be added by mapping `observable` to `new Observable()`-factory fn

> Why not use RxJS/x?

- I don't want tRPC to have dependencies
- RxJS is quite heavy
- RxJS doesn't allow me to specify a generic for `Error`s

> Fuck me, this is a lot, I can't possibly review this

Happy to organize a community call / 1on1 Zoom call for anyone interested in helping out.


danielyogel:
Interesting!  🚀
Just read the comment and a few parts of the code, I need much more time until I understand this whole PR... 🙂

> Enable better integration with other frameworks that might not have stuff like react-query
Does this improve somehow the integration with MobX?

Related to the internal Rx library, I'm really not an expert with observables, but did you checkout [https://github.com/callbag/callbag](https://github.com/callbag/callbag).

packages/client/src/rx/observable.ts
```diff
@@ -0,0 +1,70 @@
...
+          observer.error?.(err);
+          unsubscribe();
+        },
+        complete() {
+          if (isDone) {
+            return;
+          }
+          isDone = true;
+          observer.complete?.();
+          unsubscribe();
```
SimonRichardson, line 53: Auto unsubscribing on (error|complete) seems to make sense if you follow the following rxjs grammar. See: http://reactivex.io/rxjs/manual/overview.html#executing-observables

packages/client/src/rx/operators/tap.ts
```diff
@@ -0,0 +1,26 @@
+import { MonoTypeOperatorFunction, Observer } from '../types';
+
+export function tap<TValue, TError>(
```
SimonRichardson, line 3: Not sure about the name of this, but I don't have a better one atm (other than tee).

KATT: The naming here is from https://rxjs.dev/api/operators/tap

packages/client/src/rx/operators/share.ts
```diff
@@ -0,0 +1,68 @@
...
+          }
+        },
+      });
+    }
+    function resetIfNeeded() {
+      if (!running) {
+        return;
+      }
+      // "resetOnRefCountZero"
+      if (refCount === 0 && subscription) {
```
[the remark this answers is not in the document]
KATT: THANK YOU! I had missed to reset to `running = false`. Fixed in cd5c198b4b4542962bbe7d8c75d3f6e92e6ddcb2

benlesh:
Hi @KATT! This is very interesting, and I'm very flattered you've been inspired by Rx to take this approach!

That said, I would _strongly_ advise against using a reinvented observable here. Especially if you're exposing it as a public API.

RxJS is:
- tree-shakable (you only pay for what you use)
- widely used
- battle-tested
- covered by more than 3,000 integration tests
- covered by hundreds of typescript type/API tests.
- It's interoperable with many common libraries and supported natively by Angular, Svelte, Vue, and XState, to name a few.

Apollo GraphQL is also looking into moving over to use RxJS for these reasons. Not having dependencies for the sake of not having dependencies may be a mistake here. It's more likely to add weight than remove any, given that the RxJS implementation won't be much larger (if at all), and anyone that is already using RxJS (downloaded 31M times this week) will end up paying the cost of your implementation.

At the very least, if you do decide to go with your own observable implementation, please don't use `rx` in the name, as it implies it's RxJS, which it is not.

I'm happy to pair with you on this if you'd like.

KATT:
You're making many good points, @benlesh, I'm going to let this marinate a bit. 

> At the very least, if you do decide to go with your own observable implementation, please don't use `rx` in the name, as it implies it's RxJS, which it is not.

100% will do. 

> I'm happy to pair with you on this if you'd like.

Extremely generous offer. I think I can swap them around in a matter of minutes, unless the swap showcases some underlying issues 😅

KATT:
Will merge this for now and come back to it later. This will be part of next major release which isn't expected for another month or so anyway.

PR #1386, opened 2022-01-11
Make `invalidateQueries()` to align with `react-query`

### LINKED ISSUE #1383: `invalidateQueries` does not work as expected
Welcome back @KATT, I hope you enjoyed your holidays 😀

The [`invalidateQueries`](https://trpc.io/docs/invalidateQueries) docs say:

> [See react-query docs](https://react-query.tanstack.com/guides/query-invalidation) if you want more fine-grained control.

In turn, the react query docs say:

- `invalidateQueries()` will invalidate every query (I found this on some other page)
- `invalidateQueries('todos')` will invalidate every query with a key that starts with `'todos'`
- `invalidateQueries({ predicate: query => boolean })` allows fine-grain control

However, if you now try the following with trpc:

```ts
import { trpc } from '../utils/trpc'

const context = trpc.useContext()

// Invalid
context.invalidateQueries()

// Works differently than expected:
//   Will invalidate the exact query 'todos'
//   Will not clear queries that start with 'todos'!
context.invalidateQueries('todos') 

// Invalid
context.invalidateQueries({ predicate: () => true })

const { queryClient } = context

// Doesn't seem to work for me. Does nothing.
queryClient.invalidateQueries('todos')

// Will invalidate the exact query 'todos', if it has no params
queryClient.invalidateQueries(['todos'])
```

If we want to invalidate all queries it must be done as so:

```ts
queryClient.invalidateQueries()
```

If we want to invalidate queries that start with `'todos'` it must be done like so:

```ts
queryClient.invalidateQueries({
  predicate({ queryKey }) {
    const name = queryKey[0]
    return typeof name === 'string' && name.startsWith('todos')
  },
})
```

For example, when a user logs off in my application, I clear all private queries like so:

```ts
queryClient.removeQueries({
  predicate({ queryKey }) {
    const name = queryKey[0]
    return typeof name !== 'string' || !name.includes('public')
  },
})
```

Solutions I propose:

1. The distinction should be made in the docs that trpc uses arrays as keys, not strings, and thus `invalidateQueries` can't be used in certain ways.

2. The typing of `invalidateQueries` needs to be fixed (at least) to allow empty args and `{ predicate: (query) => boolean }`.

3. Maybe there should be a separate `invalidateQueries` / `removeQueries` methods returned directly from `createReactQueryHooks` that allow others to implement query invalidation in a way that makes better sense with trpc. If the typings of `invalidateQueries` are completely fixed, this change would not be necessary. Looking for feedback on whether anyone else thinks this is a good idea.

    ```ts
    // Maybe something like this
    import { trpc } from '../utils/trpc'
    
    // Will correctly invalidate every query with a key that starts with `'todos'`
    trpc.invalidateQueries(['todos'])

    // Clear all non-public queries
    trpc.removeQueries({
      predicate([key, params]) {
        return !key.includes('public')
      },
    })
    ```

KATT:
Closes #1383

### DIFF
--- a/packages/react/src/internals/context.tsx
+++ b/packages/react/src/internals/context.tsx
@@ -28,10 +28,10 @@ interface TRPCFetchInfiniteQueryOptions<TInput, TError, TOutput>
   extends FetchInfiniteQueryOptions<TInput, TError, TOutput>,
     TRPCRequestOptions {}
 
-export type TRPCContextState<
+export interface TRPCContextState<
   TRouter extends AnyRouter,
   TSSRContext = undefined,
-> = {
+> {
   queryClient: QueryClient;
   client: TRPCClient<TRouter>;
   isPrepass: boolean;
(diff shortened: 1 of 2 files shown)


packages/react/src/internals/context.tsx
```diff
@@ -105,22 +105,26 @@ export type TRPCContextState<
...
+  >(
+    pathAndInput?: [TPath, TInput?] | TPath,
+    filters?: InvalidateQueryFilters,
+    options?: InvalidateOptions,
+  ): Promise<void>;
+  /**
+   * @link https://react-query.tanstack.com/guides/query-invalidation
+   */
+  invalidateQueries(
+    filters?: InvalidateQueryFilters,
```
KATT, line 124: This `InvalidateQueryFilters` takes a genric but I didn't bother typing it for now, feel free to do a PR with that.. it'll be quite tricky

MarcGuiselin:
Awesome, thanks for getting to this so quickly!

KATT:
> Awesome, thanks for getting to this so quickly!

No problem! Please review it so I can merge & release it :)

MarcGuiselin:
I fixed the `invalidate-5-predicate` test which was never clicked.

MarcGuiselin (approved):
See my fix.

Seems good otherwise. The type-checking will ensure `invalidateQueries` is only used with valid trpc queries, so others won't run into the same issues I was. Thanks!

KATT:
Not at computer right now, I will merge and release asap

PR #1394, opened 2022-01-14
feat: dehydration of SSG helpers to serialize errors

### LINKED ISSUE #1393: Router throwing on server fetch results in odd state
Hi! I noticed some funky query state which didn't make sense when using SSR. I've tracked it down to occurring when a router throws. It results in a query looking like this:

```
{
  "status": "loading",
  "isLoading": true,
  "isSuccess": false,
  "isError": false,
  "isIdle": false,
  "dataUpdatedAt": 0,
  "error": {
    "name": "TRPCError"
  },
  "errorUpdatedAt": 1641955658792,
  "failureCount": 1,
  "isFetched": true,
  "isFetchedAfterMount": false,
  "isFetching": true,
  "isRefetching": false,
  "isLoadingError": false,
  "isPlaceholderData": false,
  "isPreviousData": false,
  "isRefetchError": false,
  "isStale": true
}
```

as you can see, the query is simultaneously `isLoading` and `error`.

Here's a minimal repro: https://2frfr.sse.codesandbox.io/post/5c03994c-fc16-47e0-bd02-d218a370a078 

*edit: CodeSandbox source: https://codesandbox.io/s/hardcore-poitras-2frfr

thanks!

KATT:
Closes #1393

- Add logic to loop through `react-query`'s cache and serialize errors when using SSG-helpers the same way they are serialized in SSR
- Did a prerelease and added it to a fork of @vinspee's repro: https://codesandbox.io/s/frosty-voice-kvbbc 
- Unsure if I want to add this as it adds quite a bit of complexity

### DIFF
--- a/examples/.wip/next-prisma-starter--big/package.json
+++ b/examples/.wip/next-prisma-starter--big/package.json
@@ -65,7 +65,7 @@
     "jest-playwright-preset": "^1.4.5",
     "npm-run-all": "^4.1.5",
     "playwright": "^1.14.1",
-    "prettier": "^2.3.2",
+    "prettier": "^2.5.1",
     "prisma": "^3.7.0",
     "ts-jest": "^27.0.5",
     "ts-node": "^10.3.0",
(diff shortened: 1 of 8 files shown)


VinSpee:
> Unsure if I want to add this as it adds quite a bit of complexity

I get what you mean, but IMO it was very unexpected behavior for me - I expected to to feel like a react-query response and it was like a halted version.

KATT:
Introduces a lot of complexity and it hasn't gotten more people asking for it so will close this for now.

PR #1466, opened 2022-01-28
Bundle with Rollup

sachinraja:
Use [Rollup](https://github.com/rollup/rollup) to bundle. This setup emits ESM and CJS. It *should* have everything that the current setup has. It still uses Babel and produces declaration maps.


packages/server/adapters/ws/index.d.ts
```diff
@@ -0,0 +1 @@
+export * from '../../dist/adapters/ws';
```
KATT, line 1: Are these needed? How come?

KATT:
Something fails with the release I did...
Screenshot 2022-01-28 at 16 34 54

sachinraja:
Looks like `react-query` does not have proper bundling. I've updated the imports to import from `react-query` instead of `react-query/hydration` (the hydration utilities were moved to core anyway), which should fix the issue. See [react-query docs](https://react-query.tanstack.com/reference/hydration#dehydrate) and [this issue](https://github.com/tannerlinsley/react-query/issues/2924).

sachinraja:
> https://vercel.com/trpc/todomvc/CRj5xFQdp675eKDzpDbVdX1tnbZg

I can't actually see that, presumably because I don't have the proper permissions.

KATT:
> > https://vercel.com/trpc/todomvc/CRj5xFQdp675eKDzpDbVdX1tnbZg
> 
> I can't actually see that, presumably because I don't have the proper permissions.

Is there a request access button? If so, please click it

sachinraja:
There is not (logged in with GitHub)

KATT:
@sachinraja could you make sense of the error at all?

sachinraja:
@KATT Yes, it should be fixed now as per https://github.com/trpc/trpc/pull/1466#issuecomment-1024353372

KATT:
Hmmmm

```


> Build error occurred
--
21:27:11.469 | TypeError: _trpc_server__WEBPACK_IMPORTED_MODULE_1__.router is not a function
21:27:11.470 | at createRouter (/vercel/path1/.next/server/pages/[filter].js:537:54)
21:27:11.470 | at /vercel/path1/.next/server/pages/[filter].js:571:97 {
21:27:11.470 | type: 'TypeError'
21:27:11.470 | }
21:27:11.520 | error Command failed with exit code 1.


```

sachinraja:
This is a very strange error. I'm not able to reproduce it locally (with the sqlite database). Wish I could see the webpack output for the above command.

sachinraja:
Looked into this a lot and not able to find the fix. Might have to create an issue for Next.js or Webpack on this. It works locally with yarn because Next.js bundles packages that are imported outside of node_modules and it follows the workspace `@trpc` package symlinks in `node_modules` to their source directories. It doesn't work with an actual install because webpack is not awaiting the `import("@trpc/server")` correctly so the function does not exist because the import promise is still resolving when `router()` is called.

sachinraja:
This is probably **not** the problem but I noticed that the individual dist directories (preconstruct artifacts) were still published for each entrypoint: <https://unpkg.com/browse/@trpc/react@10.0.0-alpha.4/ssg/dist/>. These should be removed.

sachinraja:
After a ton of experimenting I managed to reproduce this with [Chalk](https://github.com/chalk/chalk) (an esm-only dep). It has to do with the circular dependency between [`src/pages/api/trpc/[trpc].ts`](https://github.com/trpc/trpc/blob/7d2afa50da40fa1ea8560c0bc1f8d7be47136356/examples/next-prisma-todomvc/src/pages/api/trpc/%5Btrpc%5D.ts) and [`src/routers/todo.ts`](https://github.com/trpc/trpc/blob/main/examples/next-prisma-todomvc/src/routers/todo.ts). `todo.ts` imports `createRouter` from `[trpc].ts` file and `[trpc].ts` imports `todoRouter` (which is created using `createRouter`). This error gets fixed when I abstract `createRouter` to its own file to avoid the circular dependency (although there's another, different error that I'm getting after it).

KATT:
Oh wow, how did you come to that conclusion? That what a horrible one. I'm very grateful for your work here @sachinraja! ♥

sachinraja:
> how did you come to that conclusion?

Many hours of throwing everything I could think of at it 😅

Should I update the example to fix the circular deps in this PR?

Also next up on this PR: a prerendering error!

KATT:
> > how did you come to that conclusion?
> 
> Many hours of throwing everything I could think of at it 😅
> 
> Should I update the example to fix the circular deps in this PR?
Yes please!! Just push it here 

> Also next up on this PR: a prerendering error!

🙀

sachinraja:
Tested locally and everything works great. I think the prerendering error is a weird edge case on my local setup because of mixing package managers and me messing with a few things.

sachinraja:
@KATT don't know if you're busy but this is ready for review

KATT:
> @KATT don't know if you're busy but this is ready for review

Dang. Doing this in the morning ♥

packages/server/tsconfig.json
```diff
@@ -3,6 +3,6 @@
   "include": ["src", "test"],
   "compilerOptions": {
     "target": "ES2019",
-    "lib": ["ES2020.Promise"]
+    "lib": ["es5", "es6", "dom", "dom.iterable", "ES2020.Promise"]
```
KATT, line 6: Might revisit this later. I want to do next major as node > 14, so don't think all of it is needed

scripts/rollup.js
```diff
@@ -0,0 +1,85 @@
...
+        }),
+        babel({
+          babelHelpers: 'runtime',
+          configFile: path.join(__dirname, '..', '..', 'babel.config.js'),
+          extensions,
+        }),
+      ],
+    },
+  ]);
+};
```
KATT, line 85: 🥳

KATT (approved):
This is amazing! ❤️

PR #1515, opened 2022-02-14
docs: the "createRouter" function must be explicitly defined

ahhshm:
It took me a few minutes to realize that the "createRouter" was a helper function, not a trpc function.


KATT (approved):
♥ x 2

PR #1518, opened 2022-02-15
Add trpc-sveltekit

icflorescu:
Not a full-feature set yet (i.e. subscriptions), but it's a start ;-)

### DIFF
--- a/www/docs/further/awesome-trpc.mdx
+++ b/www/docs/further/awesome-trpc.mdx
@@ -18,6 +18,7 @@ slug: /awesome-trpc
 | Turborepo + Expo + tRPC Starter                                                                  | https://github.com/gunnnnii/turbo-expo-trpc-starter                     |
 | tRPC in a Cloudflare Worker                                                                      | https://github.com/esamattis/trpc-cloudflare-worker                     |
 | tRPC on Commerce Cloud Managed Runtime                                                           | https://github.com/danechitoaie/commerce-cloud-managed-runtime-trpc-poc |
+| tRPC with SvelteKit                                                                              | https://github.com/icflorescu/trpc-sveltekit                            |
 
 ## 🏁 Open-source projects using tRPC
 


KATT (commented):
😍😍😍😍😍

Do you have an example app too?

KATT:
Does it work with Svelte's ssr too? I had troubles with not understanding Svelte's fetch helper when trying to get it to work

KATT:
@all-contributors add @icflorescu for examples

icflorescu:
Sure, here's the example app: [icflorescu/trpc-sveltekit-example](https://github.com/icflorescu/trpc-sveltekit-example).
Works in CodeSandbox too.

I had no issues with SSR. In fact, I've updated the example above to render 3 pages with SSR and everything seems to be in order. It's only a matter of configuring the tRPC client... something like this:

```ts
import { browser } from '$app/env';
import type { Router } from '$lib/server/trpc';
import * as trpc from '@trpc/client';

const client = trpc.createTRPCClient<Router>({
  url: browser ? '/trpc' : 'http://localhost:3000/trpc', // 👈
});
```

As you can see in the example app, I've only tested it with the "vanilla" client so far, but now I'm thinking to play around a little with [Svelte Query](https://sveltequery.vercel.app/), and maybe even come up with a wrapper.

I'd also like to find the time to look into subscriptions at some point...

www/docs/further/awesome-trpc.mdx
```diff
@@ -18,6 +18,7 @@ slug: /awesome-trpc
 | Turborepo + Expo + tRPC Starter                                                                  | https://github.com/gunnnnii/turbo-expo-trpc-starter                     |
 | tRPC in a Cloudflare Worker                                                                      | https://github.com/esamattis/trpc-cloudflare-worker                     |
 | tRPC on Commerce Cloud Managed Runtime                                                           | https://github.com/danechitoaie/commerce-cloud-managed-runtime-trpc-poc |
+| tRPC with SvelteKit                                                                              | https://github.com/icflorescu/trpc-sveltekit                            |
```
KATT, line 21: Please move this to "🧩 Extensions & community add-ons" and add your example project here instead :)

KATT: Feel free to move the extensions section closer to the top too :)

icflorescu:
I'm gonna close this in favor of https://github.com/trpc/trpc/pull/1522

PR #1520, opened 2022-02-15
docs: use unused variable

ahhshm:
`createContext` is defined but never used so either it should be used in the `createExpressMiddleware` (that's what I did) or should be removed.


KATT:
Thank you!

PR #1540, opened 2022-02-20
Fix typo

p6l-richard:
create-next-app was missing a `-`


KATT (changes requested):
Hmmmm --- tested this, left side works for me and right side doesn't? Which version of yarn are you on?

KATT (approved):
Nvm, i was being silly. Thanks!

PR #1544, opened 2022-02-20
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/examples/next-prisma-starter-websockets/test/playwright.test.ts
+++ b/examples/next-prisma-starter-websockets/test/playwright.test.ts
@@ -11,7 +11,7 @@ test('send message', async () => {
   const nonce = Math.random()
     .toString(36)
     .replace(/[^a-z]+/g, '')
-    .substr(0, 6);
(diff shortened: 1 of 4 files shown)


KATT (approved):

KATT:
@all-contributors add @CommanderRoot for code

TheoBr:
Fancy seeing you here @CommanderRoot 👀

CommanderRoot:
> Fancy seeing you here @CommanderRoot 👀
Your Tweets have lead me to this repo 👍

PR #1549, opened 2022-02-21
merge `main` -> `next`

KATT:
@skarab42, I need your help here! There's **1** test that fails, any only on CI and it's something with the subscriptions. Would you mind having a look? 

For context, `next` is the current WIP branch for next major version and here I'm merging `main` into it.


skarab42:
I don't understand what's going on, I'll look further into it tomorrow.

skarab42:
The issue on CI is related to an error I made in the tests where the WS server creation process is not awaited which means we're not sure it's started before querying it. From various tests I made, on CI and slow running machines, the test will always fail whereas on faster machines, it may start quickly enough for the test to randomly pass sometimes.

I can easily fix this issue by awaiting the server creation process, altho, what should be in your opinion the correct behavior if a subscription is triggered when the server is not yet online? Should I also add a test for this case?

https://github.com/trpc/trpc/blob/0afd3122f713b38a629173080bfad516be7dc84c/packages/server/test/adapters/fastify.test.ts#L158-L159

KATT:
Thanks! I sent you a monthly coffee  ☕  :)

skarab42:
> Thanks! I sent you a monthly coffee ☕ :)

Awesome, thank you very much  <3

Issue #1562, opened 2022-02-25
Pass request object to the teardown function in createExpressMiddleware


olehmisar:
I need to access some request information in the teardown function:
```typescript
app.use(
  trpcExpress.createExpressMiddleware({
    router: appRouter,
    teardown() {
      // need access to request object here
    }
  })
);
```
  
I would like to use it like this:
```typescript
teardown(req) {
  console.log(req.url, "done")
}
```

KATT:
Hey @olehmisar 

What do you need this function for? Are you sure you don't want to have a logger middleware?

https://trpc.io/docs/middlewares#logging

olehmisar:
Hi @KATT. Looks like it is exactly what I need! Thanks.

PR #1568, opened 2022-02-26
feat(express-adapter): pass req instance to teardown function

ahhshm:
related to #1562


KATT:
Skipping this PR as the @olehimisar seem to have been looking in the wrong place from what they wanted to do 

https://github.com/trpc/trpc/issues/1562#issuecomment-1059991190

PR #1583, opened 2022-03-01
fix(examples): avoid destructuring 'process.env'. it's not safe

ahhshm:
[Next.js docs:
](https://nextjs.org/docs/basic-features/environment-variables)

> Note: In order to keep server-only secrets safe, Next.js replaces `process.env.*` with the correct values at build time. This means that process.env is not a standard JavaScript object, so you’re not able to use [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment). Environment variables must be referenced as e.g. `process.env.PUBLISHABLE_KEY`, not `const { PUBLISHABLE_KEY } = process.env`.

### DIFF
--- a/examples/next-prisma-starter-websockets/src/pages/api/auth/[...nextauth].ts
+++ b/examples/next-prisma-starter-websockets/src/pages/api/auth/[...nextauth].ts
@@ -4,10 +4,9 @@ import GithubProvider from 'next-auth/providers/github';
 import CredentialsProvider from 'next-auth/providers/credentials';
 
 let useMockProvider = process.env.NODE_ENV === 'test';
-const { GITHUB_CLIENT_ID, GITHUB_SECRET, NODE_ENV, APP_ENV } = process.env;
 if (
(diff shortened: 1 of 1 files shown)


KATT (changes requested):
This is a pure backend function, so that comment does not apply here

KATT:
Or maybe it does and I closed this too hastly - can you expand on why it would be unsafe?

ahhshm:
I did a little research and it revealed that you are right. I was blindly following the docs. only environment variables that start with `NEXT_PUBLIC_` couldn't be destructured. Next.js uses something called [`DefinePlugin`](https://webpack.js.org/plugins/define-plugin/) (as you can see [here](https://github.com/vercel/next.js/blob/721dc7bccd0bcdd6f8d6600b0dc30192822a0f3d/packages/next/build/webpack-config.ts#L1288)) to replace all occurrences of `process.env.NEXT_PUBLIC_*` with the actual value so only the public environment variables will be bundled to the client.

sorry for taking your valuable time.

PR #1592, opened 2022-03-05
rewrite links with rxjs

### LINKED ISSUE #1420: [v10] Rewrite links with RxJS
Do further iterations on links #1366, currently implemented on the [`next`](https://github.com/trpc/trpc/tree/next)-branch.

- Currently using my own implementation of observables
- Use RxJS instead (see https://github.com/trpc/trpc/pull/1366#issuecomment-1010485835).

sachinraja:
closes #1420

Switch from the trpc rxjs implementation to actual rxjs.


KATT:
Thanks, @sachinraja! ❤️ 

I sent you another $100 - this was not by any means an easy rewrite and I'm super impressed by your work. The only thing I fixed was in d33319f -- `this` in JS unfortunately sucks and it's this stuff is the main reason why I try to avoid JS classes at all cost and hence a reason why I didn't want to use RxJS in the first place (but that's really a criticism of [the spec](https://github.com/tc39/proposal-observable) and not RxJS itself).

@benlesh will likely be very happy with this, feel free to review if you feel inclined to :)

KATT:
@benlesh - the only things bogging me now:

- Errors are untyped `any`
- When calling `.unsubscribe()` I'd expected/want my `complete()` to be called and this is why one of the tests is failing - what is the concept I'm missing here? I'm sure this is a common misconception and I'm not finding the right thing in the docs

sachinraja:
Thank you so much @KATT!

> this in JS unfortunately sucks and it's this stuff is the main reason why I try to avoid JS classes at all cost

Well that's annoying, nice catch on that.

One thing that might be improved in the current implementation is a replacement for [`observableToPromise`](https://github.com/trpc/trpc/blob/d33319f014a741b78490f6a05687d6388a1f79ee/packages/client/src/rx.ts#L17-L52).

KATT (dismissed):
LGTM. Curious what Ben has to say

packages/client/package.json
```diff
@@ -48,30 +53,20 @@
...
-    "rx"
+    "src"
   ],
   "scripts": {
     "build": "rollup -c"
   },
   "dependencies": {
-    "@babel/runtime": "^7.9.0"
+    "@babel/runtime": "^7.9.0",
+    "rxjs": "^7.5.4"
```
KATT, line 69: Reduce risk of multiple installations of rxjs

```suggestion
    "rxjs": "^7.0.0"
```

bogdansoare:
@sachinraja great job on this PR.

@KATT the thing that worries me with adding rxjs is the increased bundle sized. 

I saw that you just published `v10.0.0-alpha.10`, so I checked it on [bundlephobia](https://bundlephobia.com/package/@trpc/client@10.0.0-alpha.10) and the bundle size for `@trpc/client` is about 168.5 kb minified according to bundlephobia. But maybe it's something wrong with bundlephobia, because it's quite a steep increase in bundle size.

sachinraja:
I think there's an issue with rxjs's bundling. Also @bogdansoare I would recommend using https://bundle.js.org as it tends to be more accurate than bundlephobia. I'm seeing an increase of about 30 KBs from [@trpc/client@10.0.0-alpha.9](https://bundle.js.org/?q=@trpc/client@10.0.0-alpha.9&treeshake=[{+createTRPCClient+}]) to [@trpc/client@10.0.0-alpha.10](https://bundle.js.org/?q=@trpc/client@10.0.0-alpha.10&treeshake=[{+createTRPCClient+}]).

When I import `Observable` from [rxjs](https://bundle.js.org/?q=https://unpkg.com/rxjs@7.5.4/dist/esm5/index.js&treeshake=[{+Observable+}]) it's 5 KBs more than when I import from [rxjs/internal/Observable](https://bundle.js.org/?q=https://unpkg.com/rxjs@7.5.4/dist/esm5/internal/Observable.js&treeshake=[{+Observable+}]).

KATT:
Eeeek. 😬 

I really underestimated the impact this would have. I thought RxJS was quite light & modular.

- [`@trpc/client@10.0.0-alpha.9`](https://bundle.js.org/?q=@trpc/client@10.0.0-alpha.9&treeshake=[{+createTRPCClient+}]): 17.6 kB -> 5.23 kB
- [`@trpc/client@10.0.0-alpha.10`](https://bundle.js.org/?q=@trpc/client@10.0.0-alpha.9&treeshake=[{+createTRPCClient+}]): 175 kB -> 39.3 kB

It's "only" 25KB more but that's 25KB is a **7.5x** increase in bundle size. 

Are we misinterpreting anything here or using it wrongly? @benlesh

utrolig:
I did some small tests, and the difference in a basic app from `alpha9` to `alpha10` was 17kb
```
dist/assets/vendor.952a33f2.js     191.34 KiB / gzip: 57.80 KiB - alpha.9
dist/assets/vendor.6be1e5d9.js     208.34 KiB / gzip: 63.16 KiB - alpha.10
```
so I think it also depends on the actual tree-shaking that happens when you import the client in your application.

sachinraja:
Is that Vite? The difference there might be caused by different bundlers/configurations/targets.

utrolig:
> Is that Vite? The difference there might be caused by different bundlers/configurations/targets.

Yes that was vite. About the same result with `create-react-app`

```
61.37 kB  build\static\js\main.fa9472d2.js <- alpha.9
66.32 kB (+4.95 kB)  build\static\js\main.7a682bb9.js <- alpha.10
```

KATT:
Thanks for chipping in everyone! 

React and surrounding stuff skews the stats, so I'd prefer if we could do the comparison on vanilla projects without other dependencies.

Would y'all think that [the numbers](https://github.com/trpc/trpc/pull/1592#issuecomment-1060035010) from bundle.js.org accurate?

sachinraja:
Got a visualization using `esbuild` locally [here](https://github.com/sachinraja/esbuild-rxjs-visualize). About half of the bundle seems to be `tslib`.

packages/client/src/links/dedupeLink.ts
```diff
@@ -1,16 +1,15 @@
 import { AnyRouter } from '@trpc/server';
-import { observable } from '../rx/observable';
-import { share } from '../rx/operators';
-import { Observable } from '../rx/types';
-import { TRPCLink } from './types';
+import { Observable } from 'rxjs';
+import { share } from 'rxjs';
```
benlesh, line 3: You have two separate `from 'rxjs'` clauses here.

packages/client/src/links/dedupeLink.ts
```diff
@@ -21,31 +20,33 @@ export function dedupeLink<
...
+          const next$ = next(op).subscribe({
+            next(v) {
+              observer.next(v);
+            },
+            error(e) {
+              reset();
+              observer.error(e);
+            },
+            complete() {
+              reset();
```
benlesh, line 40: note: (you don't need to change this here) but even with your previous observable, when `observer.error` or `observer.complete` is called, The teardown below is called automatically, so `reset()` is technically called twice. It doesn't seem like that's ever mattered, just calling it out.

packages/client/src/rx.ts
```diff
@@ -9,9 +12,7 @@ export class ObservableAbortError extends Error {
 }
 
 /** @internal */
-export function observableToPromise<TValue>(
-  observable: Observable<TValue, unknown>,
-) {
+export function observableToPromise<TValue>(observable: Observable<TValue>) {
```
benlesh, line 15: Note: This whole function can be replaced with RxJS's `firstValueFrom`.

benlesh (changes requested):
Just a couple of small things that I noticed. Mostly that there are inconsistent imports (rxjs/operators vs rxjs). We should try to import everything from `rxjs`, and that requires ^7.2.0. Honestly I'd recommend ^7.5.4 if possible, because it's non-breaking from 7.2, but has some really nice performance improvements.

examples/fastify-server/src/client/index.ts
```diff
@@ -50,13 +50,13 @@ async function start() {
   let randomNumberCount = 0;
 
   await new Promise<void>((resolve) => {
-    const unsub = anon.client.subscription('sub:randomNumber', null, {
+    const subscription = anon.client.subscription('sub:randomNumber', null, {
       next(data) {
         console.log('>>> anon:sub:randomNumber:received:', data);
         randomNumberCount++;
 
         if (randomNumberCount > 3) {
```
benlesh, line 58: I've noticed this pattern in more than one spot. Consider using `take(3)` instead. It's the same basic idea.

packages/client/src/internals/TRPCClient.ts
```diff
@@ -18,11 +18,9 @@ import {
   OperationLink,
   TRPCLink,
 } from '../links/types';
-import { UnsubscribeFn } from '../rx';
-import { inferObservableValue } from '../rx/observable';
-import { share } from '../rx/operators';
-import { Observer } from '../rx/types';
-import { observableToPromise } from '../rx/util/observableToPromise';
+import { inferObservableValue, observableToPromise } from '../rx';
+import { share } from 'rxjs/operators';
```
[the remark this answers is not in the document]
sachinraja: Could this be what's causing the increased bundle size?

sachinraja:
Looks like there's a related issue at https://github.com/ReactiveX/rxjs/issues/6321 and a fix is on the [roadmap for v8](https://github.com/ReactiveX/rxjs/issues/6367):
> Ensure we're publishing only ES2021 or later and CJS. Drop tslib.

I believe that will fix the bundle size problem but looks like we will unfortunately have to wait until v8 for it.

benlesh:
All of this is fine, and a great metric/discussion. Is it possible that I can get some source map explorer output to analyze? I'm interested in figuring out if there's some imports that are unnecessary or some low-hanging fruit to help reduce size. It would be super helpful for our library and our other users.

benlesh:
I'm wondering if there's a few small wins that could be had. For example, rxjs's `share` operator is kind of a "one share to rule them all" sort of solution that also includes `Subject` and some other things you're not necessarily using, and it might be that your existing share operator would work fine, and you don't need RxJS's copy.

benlesh:
Just as a talking point: I created an `angular-cli` app and installed `rxjs@7.5.5`, then only included  the bits you were using (`Observable`, `map`, `tap`, and `share`), and _even including the stuff that `@angular/router` et al brings in_ - The total size of RxJS in the bundle was ~13.7kB. So seems unlikely there could be a size increase greater than that. Especially given that multiple files were deleted from this library.

image

Something feels off about the other measurements, and I'd love to be able to take the same measurements to see what the differences are.

packages/client/src/rx.ts
sachinraja: Even the abort functionality?

sachinraja:
@benlesh got a visual with rollup at https://github.com/sachinraja/rxjs-rollup-bundle. 
image

Note that this is unminified. Minified via terser, it's ~20 KB.

Also the angular webpack config is resolving the `es2015` export condition but most bundlers setups will not do that. When I setup rollup to resolve the `es2015` condition and minify it's ~15 KB, which is much closer to what angular produces. It seems like we could further reduce this by removing `tslib` and trpc using its own `share` implementation like you said.

benlesh:
So that called out a bug in our code that was including the wrong function and requiring a bunch of unnecessary stuff. Fixing that reduces the size another 5kB in your test.

image


HOWEVER.. by excluding `share`, we can remove an additional ~18kb.

image

But that means we still need a lightweight `share`, which I was able to roll in less than 1K:

image

benlesh:
For reference, the "share light" is what's below. I didn't test it or anything, it's just a simple implementation, and I cheat a bit and used an internal function (that you're not _supposed_ to use, but I know isn't going to change in version 7)

```ts
import { operate } from "rxjs/internal/util/lift";

export function share() {
  const subscribers = new Set();
  let connection = null;
  return operate((source, subscriber) => {
    subscribers.add(subscriber);
    subscriber.add(() => {
      subscribers.delete(subscriber);
      if (subscribers.size === 0) {
        connection.unsubscribe();
        connection = null;
      }
    });
    if (!connection) {
      connection = source.subscribe({
        next: (value) => {
          for (const subscriber of subscribers) {
            subscriber.next(value);
          }
        },
        error: (error) => {
          for (const subscriber of subscribers) {
            subscriber.error(error);
          }
        },
        complete: () => {
          for (const subscriber of subscribers) {
            subscriber.complete();
          }
        },
      });
    }
  });
}

```

sachinraja:
Wow that is very helpful. Great job reducing the bundle size. I just tried switching out for the lighter `share` and it looks good, but we can't reach into any of the internal functions because there are no cjs paths for it (like how there is a folder with `package.json` in it for `rxjs/operators`). So it will only work in environments that understand package.json `exports`.

benlesh:
It can be built as a ordinary operator too, feel free to not use that internal function, It's just a small wrapper around `(source) => new Observable()`.

I'm very grateful for this thread. It pointed out a flaw that was easily fixed. RxJS will only get smaller as time goes on.

benlesh:
> [@trpc/client@10.0.0-alpha.9](https://bundle.js.org/?q=@trpc/client@10.0.0-alpha.9&treeshake=%5B%7B+createTRPCClient+%7D%5D): 17.6 kB -> 5.23 kB
[@trpc/client@10.0.0-alpha.10](https://bundle.js.org/?q=@trpc/client@10.0.0-alpha.9&treeshake=%5B%7B+createTRPCClient+%7D%5D): 175 kB -> 39.3 kB

FYI: I think that the numbers from bundle.js.org can be dismissed. It seems it's not properly tree-shaking the rxjs dependency from what you're using, as it looks like it's [including literally everything from rxjs](https://bundle.js.org/?q=rxjs) (147 kB -> 32.8 kB)

So instead of including 44kB from rxjs (still not amazing) it was including all 147kB for whatever reason.

I feel that if you take the advice above, you should be able to land RxJS and not see significant bundle size changes. The win mostly being that you don't have to maintain your own Observable type, and anyone that happens to be using RxJS won't have to pay the cost of your Observable on top of theirs.

In any case, this has been a hugely interesting thing to watch/help with.

sachinraja:
> FYI: I think that the numbers from bundle.js.org can be dismissed. It seems it's not properly tree-shaking the rxjs dependency from what you're using, as it looks like it's including literally everything from rxjs (147 kB -> 32.8 kB)

There does seem to be an issue with it. I believe 32 KB might actually be the gzipped size rather than the minified size so it's not comparable with our other experiments either.

> All of that said. If you're truly being size/performance conscious, I'd recommend removing any abstractions like this at all from your internals entirely. It can be more bug prone and difficult to write certain things that way, but if the goal is speed and size, that's probably the way to go

I think this is something we should definitely consider. I'm also not sure if rxjs is the best solution or way to model links either. @KATT is this something you're interested in discussing?

benlesh:
FWIW: I'm willing to help to some degree with any of this. I will say though, that if you're exporting any type that is supposed to push streaming updates, I _highly_ recommend having it adhere to an observable contract (which is simple enough to implement in a naive way). Basically:

```ts
class ObservableLike<T> {
   subscribe(observer: Partial<Observer<T>>): Unsubscribable {
     // setup here 
     // Just be careful:
     // 1. don't call non-existant handlers
     // 2. don't call handlers out of order.
     // 3. Make sure you teardown everything after error or complete!
     
     // return a "subscription" for consumers to cancel with.
     return {
        unsubscribe: () => {
            // teardown here
        }
     }
     
     // Now it's compatible with RxJS and dozens of other libraries
     [Symbol.observable ?? '@@observable']() {
       return this;
     }
   }
}
```

RxJS observable (or any "good" observable implementation) will provide all of the safety for you. So you need to be careful. But you'll have to be careful if you're doing anything like this with callbacks, too.

sachinraja:
@KATT can you do another publish so everyone can check the new bundle size? I updated it with the lighter `share`.

KATT:
Sorry for being off here.

> @KATT can you do another publish so everyone can check the new bundle size? I updated it with the lighter `share`.
Done!

> I think we should move ahead with rxjs and consider using a generic Observable later.

It didn't do much difference.... I really find it hard to stomach a multiple X increase in bundle size when we already have an implementation that is on par for what we're doing.

I think what would make sense is that we'd have interoperability with RxJS in our own implementation, so we can easily swap in the future once RxJS has gotten smaller.

sachinraja:
superseded by #1644

PR #1593, opened 2022-03-05
Correcting usage of yarn create next-app.

kevinlangleyjr:
The syntax to use `create-next-app` with `yarn`, is `yarn create next-app` and not `yarn create-next-app` like it states in the docs.

Small typo but could definitely trip someone up if they weren't paying attention! 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ There are a few [examples](https://trpc.io/docs/example-apps) that you can use f
 
 ```sh
 # yarn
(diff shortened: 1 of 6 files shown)


sachinraja (approved):
This is definitely the right way as [it's suggested in the Next.js docs](https://nextjs.org/docs/api-reference/create-next-app), but it looks like the opposite was done in #1540.

KATT:
thank you!

KATT:
@all-contributors add @kevinlangleyjr for docs

PR #1602, opened 2022-03-07
Better schema for starters

FarazPatankar:
This PR improves the DB Schema for a few starters by:
- Removing the `unique` constraint from the `createdAt` and `updatedAt` fields
- Adding the `@updatedAt` flag on the `updatedAt` field so it actually updates on every update


FarazPatankar:
@KATT Cool to make the same changes for `next-prisma-starter-websockets` and `next-prisma-todomvc`?

KATT:
cc @FarazPatankar it has to be a new migration and to change the previous one - you may see that the deployment fail

FarazPatankar:
Hmm, any reason why? I feel like if I was using a starter from a repository, I'd want a single `init` migration. Also feels cleaner that way.

KATT (approved):

PR #1604, opened 2022-03-07
`next-prisma-starter`: add integration test example

KATT:
- Switch to using `paths` instead of `baseUrl`
- Add integration test example for post router


examples/next-prisma-starter/jest.config.ts
```diff
@@ -1,4 +1,24 @@
 import type { Config } from '@jest/types';
```
sachinraja, line 1: Maybe we should switch this to use [next/jest](https://nextjs.org/docs/testing#setting-up-jest-with-the-rust-compiler) so it uses the swc compiler.

examples/next-prisma-starter/src/utils/trpc.ts
```diff
@@ -18,3 +18,15 @@ export const trpc = createReactQueryHooks<AppRouter>();
...
+  TRouteKey extends keyof AppRouter['_def']['queries'],
+> = inferProcedureInput<AppRouter['_def']['queries'][TRouteKey]>;
+
+export type inferMutationOutput<
+  TRouteKey extends keyof AppRouter['_def']['mutations'],
+> = inferProcedureOutput<AppRouter['_def']['mutations'][TRouteKey]>;
+
+export type inferMutationInput<
+  TRouteKey extends keyof AppRouter['_def']['mutations'],
+> = inferProcedureInput<AppRouter['_def']['mutations'][TRouteKey]>;
```
sachinraja, line 32: Maybe we should consider moving this logic to trpc. It could also help in case the signature of `AppRouter` has to change in the future.

examples/next-prisma-starter/tsconfig.json
```diff
@@ -18,18 +14,13 @@
...
-    "types": [
-      "@types/jest",
-    ],
-    "baseUrl": "./src",
-    "incremental": true
+    "types": ["@types/jest"],
+    "incremental": true,
+    "baseUrl": ".",
+    "paths": {
+      "~/*": ["./src/*"]
```
sachinraja, line 21:
```suggestion
      "~/*": ["src/*"]
```
not needed when `baseUrl` is already `.`.

sachinraja (commented):
Looks good, just saw a couple of things that might be improved.

examples/next-prisma-starter/jest.config.ts
KATT: > "next/jest" is currently experimental. https://nextjs.org/docs/messages/experimental-jest-transformer

skipped it because above warning when doing it

examples/next-prisma-starter/src/utils/trpc.ts
KATT: Agreed

examples/next-prisma-starter/tsconfig.json
KATT: Fixed

sachinraja (approved):

PR #1645, opened 2022-03-16
bug: SSR hydration with error doesn't set status `error`

KATT:
When server-side-rendering the status stays at `loading` even when the query fails.

Here's an awful hack because I'm unsure what else to do. Any other thing I can think of will lead to hydration warnings.. I could also dive into the `optimisticResults`-property of `react-query` to see if there's a way to trick it to become `error`.

### This PR:

- Adds a new `SSRState`-property
- Bypasses default options and defaults to `refetchOnMount: false` when hydrating SSR until it's been mounted on the client

### For more context

How we do the SSR prepass to enable SSR:

1. Run through the app tree until there are no pending queries: https://github.com/trpc/trpc/blob/f27680fe9ccc444da6282f3e118ba82c3b0b2095/packages/next/src/withTRPC.tsx#L173-L192
2. In each query hook, we call `prefetchQuery` when doing the prepass in order to fetch it on the server: https://github.com/trpc/trpc/blob/f27680fe9ccc444da6282f3e118ba82c3b0b2095/packages/react/src/createReactQueryHooks.tsx#L243-L251

### DIFF
--- a/packages/react/src/createReactQueryHooks.tsx
+++ b/packages/react/src/createReactQueryHooks.tsx
@@ -252,14 +252,8 @@ export function createReactQueryHooks<
       TError
     >,
   ): UseQueryResult<TQueryValues[TPath]['output'], TError> {
-    const {
-      client,
-      isPrepass,
-      queryClient,
-      prefetchQuery,
-      ssrReady,
-      ssrEnabled,
-    } = useContext();
+    const { client, isPrepass, queryClient, prefetchQuery, ssrReady } =
+      useContext();
 
     if (
       typeof window === 'undefined' &&
@@ -275,13 +269,12 @@ export function createReactQueryHooks<
      * Hack to make sure errors return `status`='error` when doing SSR
      * @link https://github.com/trpc/trpc/pull/1645
      */
-    const actualOpts =
-      ssrEnabled && !ssrReady
-        ? {
-            retryOnMount: false,
-            ...opts,
-          }
-        : opts;
+    const actualOpts = !ssrReady
+      ? {
+          retryOnMount: false,
+          ...opts,
+        }
+      : opts;
     return __useQuery(
       pathAndInput as any,
       () => (client as any).query(...getClientArgs(pathAndInput, actualOpts)),
@@ -372,14 +365,8 @@ export function createReactQueryHooks<
     >,
   ): UseInfiniteQueryResult<TQueryValues[TPath]['output'], TError> {
     const [path, input] = pathAndInput;
-    const {
-      client,
-      isPrepass,
-      prefetchInfiniteQuery,
-      queryClient,
-      ssrEnabled,
-      ssrReady,
-    } = useContext();
+    const { client, isPrepass, prefetchInfiniteQuery, queryClient, ssrReady } =
+      useContext();
 
     if (
       typeof window === 'undefined' &&
@@ -395,13 +382,12 @@ export function createReactQueryHooks<
      * Hack to make sure errors return `status`='error` when doing SSR
      * @link https://github.com/trpc/trpc/pull/1645
      */
-    const actualOpts =
-      ssrEnabled && !ssrReady
-        ? {
-            retryOnMount: false,
-            ...opts,
-          }
-        : opts;
+    const actualOpts = !ssrReady
+      ? {
+          retryOnMount: false,
+          ...opts,
+        }
+      : opts;
 
     return __useInfiniteQuery(
       pathAndInput as any,
--- a/packages/react/src/createReactQueryHooks.tsx
+++ b/packages/react/src/createReactQueryHooks.tsx
@@ -256,6 +256,7 @@ export function createReactQueryHooks<
   ): TQuery {
     const { queryClient, ssrEnabled } = useContext();
     const isMounted = useIsMounted();
+
     if (
       isMounted ||
       !query.error ||
@@ -264,8 +265,8 @@ export function createReactQueryHooks<
     ) {
       return query;
     }
-    query.status = 'error';
-    return query;
+
+    return { ...query, status: 'error' };
   }
   function useQuery<TPath extends keyof TQueryValues & string>(
     pathAndInput: [path: TPath, ...args: inferHandlerInput<TQueries[TPath]>],
--- a/packages/react/src/createReactQueryHooks.tsx
+++ b/packages/react/src/createReactQueryHooks.tsx
@@ -97,6 +97,7 @@ function useIsMounted() {
   }, []);
   return ref.current;
 }
+
 export function createReactQueryHooks<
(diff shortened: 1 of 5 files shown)


packages/react/src/createReactQueryHooks.tsx
```diff
@@ -229,6 +246,28 @@ export function createReactQueryHooks<
...
+    query: TQuery,
+  ): TQuery {
+    const { queryClient, ssrEnabled } = useContext();
+    const isMounted = useIsMounted();
+
+    if (
+      isMounted ||
+      !query.error ||
+      !ssrEnabled ||
+      !queryClient.getQueryCache().find(queryKey)?.meta?._trpcSSR
```
KATT, line 264: Any reason why `meta` isn't available as `useQuery().meta`, @TkDodo?

TkDodo: meta is part of the options, not part of the result. Same as the queryKey for example.

KATT: Could it be both so I don't have to reach into the query cache to find it?

KATT: > just for my understanding: why are queries not in `error` state when they have an error ?


How we do the SSR prepass to enable SSR:

- Run through the app tree until there are no pending queries: https://github.com/trpc/trpc/blob/f27680fe9ccc444da6282f3e118ba82c3b0b2095/packages/next/src/withTRPC.tsx#L173-L192
- In each query hook, we call `prefetchQuery` when doing the prepass in order to fetch it on the server: https://github.com/trpc/trpc/blob/f27680fe9ccc444da6282f3e118ba82c3b0b2095/packages/react/src/createReactQueryHooks.tsx#L243-L251


Since the `useQuery()` is mounted with some (in this case, erroneous) data in the cache, it will go through the `optimisticResults`-flow which never really sets `status='error'` [as far as I can tell](https://github.com/tannerlinsley/react-query/blob/16783feed17a9361e4263edc8fecf564bc127de7/src/core/queryObserver.ts#L439-L605) (I think I link the right place here).


This is AFAIK the only way to trigger fetch of the query's data on the server, since hooks won't be called there. An alternative would be to do `suspense: true` but I don't want to go down that route yet until React 18 is stable (and then I might enable suspense as the _only_ way to query.

KATT: In essence - unsure how I should correctly hydrate/dehydrate the query cache when doing SSR in order for errors to actually be status=error.

TkDodo: this is pretty cool - I have seen a [discussion](https://github.com/tannerlinsley/react-query/discussions/3270#discussioncomment-2120752) about react-query and ssr-prepass, and I'm still hoping someone can come up with an example for the docs.

> Since the useQuery() is mounted with some (in this case, erroneous) data in the cache, it will go through the optimisticResults-flow which never really sets status='error' [as far as I can tell](https://github.com/tannerlinsley/react-query/blob/16783feed17a9361e4263edc8fecf564bc127de7/src/core/queryObserver.ts#L439-L605) (I think I link the right place here).

That I don't quite understand. If a query fails to prefetch, it should be in the cache in the correct error state. Yes, it runs through the optimistic result process, but that should only set the state optimistically to `loading` if it doesn't have data yet and will fetch once the effect that starts fetching kicks in. This just avoids that every `useQuery` call starts off in `idle` state, followed by a `loading` state 🤔 . Maybe this is an issue in RQ ?

TkDodo: we'd need to verify with a test case. is it like:

- we prefetch with `prefetchQuery`, which fails
- then we mount the query with `useQuery` and would expect it to be in `error` status?

I think it would indeed go to `loading` state because the query would refetch due to `retryOnMount`, which defaults to `true`:

> retryOnMount: boolean
If set to false, the query will not be retried on mount if it contains an error. Defaults to true.

or did I get the flow wrong?

KATT: I haven't thought of `retryOnMount` - gonna do some more experimentation.

packages/react/src/createReactQueryHooks.tsx
```diff
@@ -352,15 +397,28 @@ export function createReactQueryHooks<
...
+     * @link https://github.com/trpc/trpc/pull/1645
+     */
+    const actualOpts =
+      ssrEnabled && !isMounted
+        ? {
+            ...opts,
+            retry: false,
+            retryOnMount: false,
+          }
+        : opts;
```
KATT, line 411: CC @TkDodo here's the new hack to fix this. 

Until the app is mounted on the client, if SSR is enabled, I `retry`/`retryOnMount`

TkDodo: as I commented here:
- https://github.com/tannerlinsley/react-query/pull/3411#discussion_r829007744

turning off `retryOnMount` should be enough

KATT: ✅ 

If you have time and you understand the context, please skim through my changes here :)

packages/react/src/createReactQueryHooks.tsx
```diff
@@ -111,20 +117,28 @@ export function createReactQueryHooks<
...
+    ssrEnabled = false,
   }: {
     queryClient: QueryClient;
     client: TRPCClient<TRouter>;
     children: ReactNode;
     isPrepass?: boolean;
     ssrContext?: TSSRContext | null;
+    ssrEnabled?: boolean;
   }) {
+    const [ssrReady, setSSRReady] = useState(() => !ssrEnabled);
```
KATT, line 129:
```suggestion
    const [ssrReady, setSSRReady] = useState(!ssrEnabled);
```

TkDodo (dismissed):

sachinraja (approved):

packages/react/src/createReactQueryHooks.tsx
```diff
@@ -111,20 +117,28 @@ export function createReactQueryHooks<
...
+    ssrEnabled = false,
   }: {
     queryClient: QueryClient;
     client: TRPCClient<TRouter>;
     children: ReactNode;
     isPrepass?: boolean;
     ssrContext?: TSSRContext | null;
+    ssrEnabled?: boolean;
   }) {
+    const [ssrReady, setSSRReady] = useState(!ssrEnabled);
```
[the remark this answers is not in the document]
KATT: I swapped this naming around so many times.

I also considered to introduce an ssrState variable that could be something like `nossr prepass | mounting | mounted`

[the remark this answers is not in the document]
sachinraja: That'd definitely be easier to understand

PR #1651, opened 2022-03-17
remove tslib and "gitHead" field

sachinraja:
`tslib` is an unused dependency (even in the `dist`) and `"gitHead"` is a property that is supposed to be temporary, created by lerna during the publish process (see lerna/lerna#1880).


KATT (approved):
👌

Issue #1661, opened 2022-03-21
Refactor `react.test.tsx`


KATT:
Related to the [`next`](https://github.com/trpc/trpc/tree/next) branch.

- It's a big mess. 
- It should be split into logical files.

PR #1667, opened 2022-03-21
fix: `useMutation()` with `context:`

KATT:


packages/server/test/react.test.tsx
```diff
@@ -580,6 +591,40 @@ describe('useMutation()', () => {
...
+      }, []);
+
+      return <pre>{deletePostsMutation.isSuccess && '___FINISHED___'}</pre>;
+    }
+
+    const utils = render(
+      <App>
+        <MyComponent />
+      </App>,
+    );
```
KATT, line 617: Nicer approach than the rest cc #1661

PR #1679, opened 2022-03-24
Dependabot/npm and yarn/next 12.1.0

KATT:
So, the problem with this one seems to be that the `adapters/next.ts` imports from `next` which in turn nowadays some `global.d.ts` which affects the whole build.

Giving us this output in `packages/server/dist/declarations/src/adapters/ws.d.ts:2:23 `:

```ts
/// <reference types="node" />
/// <reference types="./types/global" /> // <--------------- this is bad and is coming from node_modules/next
import http from 'http';
import ws from 'ws';
import { BaseHandlerOptions } from '../internals/BaseHandlerOptions';
import { AnyRouter } from '../router';
import { NodeHTTPCreateContextOption } from './node-http';
/**
 * Web socket server handler
 */
export declare type WSSHandlerOptions<TRouter extends AnyRouter> =
  BaseHandlerOptions<TRouter, http.IncomingMessage> & {
    wss: ws.Server;
    process?: NodeJS.Process;
  } & NodeHTTPCreateContextOption<TRouter, http.IncomingMessage, ws>;
export declare function applyWSSHandler<TRouter extends AnyRouter>(
  opts: WSSHandlerOptions<TRouter>,
): {
  broadcastReconnectNotification: () => void;
};
//# sourceMappingURL=ws.d.ts.map
```

Now, why it shows up in that declaration and not in like `packages/server/dist/declarations/src/adapters/standalone.d.ts` is very confusing to me.

Only solution I can think of right now is to move the whole adapter to `packages/next`...

@sachinraja - any ideas here?

### DIFF
--- a/examples/standalone-server/tsconfig.json
+++ b/examples/standalone-server/tsconfig.json
@@ -6,7 +6,6 @@
     "target": "ES2019",
     "declaration": true,
     "esModuleInterop": true,
-    "skipLibCheck": true,
     "strict": true
   },
   "include": ["src"],
--- a/examples/standalone-server/tsconfig.json
+++ b/examples/standalone-server/tsconfig.json
@@ -6,6 +6,7 @@
     "target": "ES2019",
     "declaration": true,
     "esModuleInterop": true,
+    "skipLibCheck": true,
     "strict": true
   },
   "include": ["src"],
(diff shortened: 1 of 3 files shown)


sachinraja:
pushed a fix @KATT

sachinraja (approved):

KATT:
Wow, thanks! ♥

I had to read it four times to see the change

Issue #755, opened 2021-08-05
[RFC] Using tRPC for public-facing APIs (OpenAPI/Swagger/etc)


KATT:
It would be nice if there was a way to easily expose tRPC-endpoints as REST (or GraphQL) for people who want to use tRPC for a public API and is a bit more intuitive. Could potentially drive adoption a bit more.

See #752 for an example of what gets confusing.

... been thinking a bit about it, but it's not obvious how it could work. 

Maybe doing some sort of API-builder... or something where you define your API schema which then does automatic mapping to tRPC. 

I'm also thinking that it might not be needed - `JSON.stringify(encodeURIComponent(x))` is hard when you toy around with manual API-requests, but it's not that bad when you're actually working with the API, as it's only needed once.

## Current implementation, HTTP Methods <-> Type mapping

| HTTP Method | Mapping           | Notes                                                                                                  |
| ----------- | ----------------- | ------------------------------------------------------------------------------------------------------ |
| `GET`       | `.query()`        | Input JSON-stringified in query param.<br/>_e.g._ `?input=${JSON.stringify(encodeURIComponent(input))` |
| `POST`      | `.mutation()`     | Input as post body.           |                      

## Considerations with OpenAPI

- Resources are usually `/{resource}/{id}?param1=x&param2=y`-style - tRPC is [currently] with `{resource}?input=JSON.stringify(encodeURIComponent(input))`-style
- JSON-RPC based response shape might not be the ideal response shape for OpenAPI
- An output schema usually have a `$ref`-schema which would be possible to do automatically / first feature people would request is to make different paths request the same input type
- Actually using zod or any other validation on a resolver's `output` would slow down API outputs.

## Related

https://github.com/trpc/trpc/discussions/271

mmkal:
For local development I've played around with an express middleware that just mutates `req.query` to make it more "traditional". It only really works for very simple input schemas that are extensions of `Record<string, string>`, and there may be a whole host of edge-cases, but it's been handy for using postman etc.

[...]

PR #1699, opened 2022-03-25
Add output response parser

jlalmes:
## Added output validation to router queries & mutations

I have a goal to try and auto-generate OpenAPI docs for a TRPC router (cc. https://github.com/trpc/trpc/issues/755). The feature in this PR will enable us to generate JSON schemas for request and response payloads. This output parser is **opt-in only** and should not introduce any breaking changes.

### API
```
const appRouter = new trpc.router<Context>()
  .query({
    input: z.string(),
    output: z.object({
      input: z.string(),
    }),
    resolve: ({ input }) => {
      return { input };
    }
  });
```

- Added a new optional `output` property.
- When `output` is provided, the `resolve` function will expect a return type inferred from the `output` parser.
- The response data will be validated against the `output` parser, unless `NODE_ENV` is `production`. 
- If the response data fails validation a `TRPCError INTERNAL_SERVER_ERROR` error will be thrown.

### Todo
- Get some feedback (cc. @KATT)
- Write some comprehensive output parser tests.
- Update documentation.

### DIFF
--- a/packages/server/src/internals/procedure.ts
+++ b/packages/server/src/internals/procedure.ts
@@ -43,11 +43,11 @@ export type ProcedureResolver<TContext, TInput, TOutput> = (opts: {
   type: ProcedureType;
 }) => Promise<TOutput> | TOutput;
 
-interface ProcedureOptions<TContext, TInput, TOutput> {
+interface ProcedureOptions<TContext, TInput, TOutput, TParsedOutput> {
   middlewares: Array<MiddlewareFunction<any, any>>;
   resolver: ProcedureResolver<TContext, TInput, TOutput>;
   inputParser: ProcedureParser<TInput>;
-  outputParser: ProcedureParser<TOutput>;
+  outputParser: ProcedureParser<TParsedOutput>;
 }
 
 /**
@@ -103,15 +103,18 @@ export class Procedure<
   TParsedInput,
   TOutput,
   TParsedOutput,
+  TFinalOutput = unknown extends TParsedOutput ? TOutput : TParsedOutput,
 > {
   private middlewares: Readonly<Array<MiddlewareFunction<any, any>>>;
-  private resolver: ProcedureResolver<TContext, TParsedInput, TParsedOutput>;
+  private resolver: ProcedureResolver<TContext, TParsedInput, TOutput>;
   private readonly inputParser: ProcedureParser<TParsedInput>;
   private parseInputFn: ParseFn<TParsedInput>;
-  private readonly outputParser: ProcedureParser<TParsedOutput>;
-  private parseOutputFn: ParseFn<TParsedOutput>;
+  private readonly outputParser: ProcedureParser<TFinalOutput>;
+  private parseOutputFn: ParseFn<TFinalOutput>;
 
-  constructor(opts: ProcedureOptions<TContext, TParsedInput, TParsedOutput>) {
+  constructor(
+    opts: ProcedureOptions<TContext, TParsedInput, TOutput, TFinalOutput>,
+  ) {
     this.middlewares = opts.middlewares;
     this.resolver = opts.resolver;
     this.inputParser = opts.inputParser;
@@ -131,10 +134,10 @@ export class Procedure<
     }
   }
 
-  private async parseOutput(rawOutput: unknown): Promise<TParsedOutput> {
+  private async parseOutput(rawOutput: TOutput): Promise<TFinalOutput> {
     try {
       if (process.env.TRPC_SKIP_OUTPUT_VALIDATION) {
-        return rawOutput as TParsedOutput;
+        return rawOutput as any;
       }
       return await this.parseOutputFn(rawOutput);
     } catch (cause) {
@@ -152,7 +155,7 @@ export class Procedure<
    */
   public async call(
     opts: ProcedureCallOptions<TInputContext>,
-  ): Promise<TOutput> {
+  ): Promise<TFinalOutput> {
     // wrap the actual resolver and treat as the last "middleware"
     const middlewaresWithResolver = this.middlewares.concat([
       async ({ ctx }: { ctx: TContext }) => {
@@ -204,7 +207,7 @@ export class Procedure<
       throw result.error;
     }
 
-    return result.data as TOutput;
+    return result.data as TFinalOutput;
   }
 
   /**
@@ -216,7 +219,7 @@ export class Procedure<
   ): this {
     const Constructor: {
       new (
-        opts: ProcedureOptions<TContext, TParsedInput, TParsedOutput>,
+        opts: ProcedureOptions<TContext, TParsedInput, TOutput, TFinalOutput>,
       ): Procedure<
         TInputContext,
         TContext,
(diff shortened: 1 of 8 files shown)


KATT (commented):
Not able to look it through until tomorrow at the earliest but it is great! I would love for optional output validation

KATT:
I think if there is an output validation, it should always run, we can add the optionality as a feature later!

KATT:
Also setup github sponsors so I can send you 💷💷!

sachinraja:
> I have a goal to try and auto-generate OpenAPI docs for a TRPC router

Let me know if there's any way I can help! I have some experience in this area with [`zod-to-ts`](https://github.com/sachinraja/zod-to-ts) and [tRPC Playground](https://github.com/sachinraja/zod-to-ts).

packages/server/test/inference.test.ts
```diff
@@ -16,6 +17,25 @@ test('infer input & output', async () => {
...
+      async resolve({ input }) {
+        return { input };
+      },
+    })
+    .query('preferOutput', {
+      input: z.string(),
+      output: z.object({
+        input: z.string(),
+      }),
+      // @ts-ignore - ensuring types from "output" are expected
```
KATT, line 35:
```suggestion
      // @ts-expect-error - ensuring types from "output" are expected
```

packages/server/test/inference.test.ts
```diff
@@ -16,6 +17,25 @@ test('infer input & output', async () => {
       async resolve({ input }) {
         return { input };
       },
+    })
+    .query('withOutput', {
+      input: z.string(),
+      output: z.object({
+        input: z.string(),
+      }),
```
KATT, line 25: Maybe tests to do with this should be in it's own file? Like `output-parser.test.ts`

packages/server/src/internals/procedure.ts
```diff
@@ -97,22 +86,26 @@ function getParseFn<TParsedInput>(
...
 
-  constructor(opts: ProcedureOptions<TContext, TParsedInput, TOutput>) {
+  constructor(opts: ProcedureOptions<TContext, TInput, TOutput>) {
     this.middlewares = opts.middlewares;
     this.resolver = opts.resolver;
     this.inputParser = opts.inputParser;
-    this.parse = getParseFn(this.inputParser);
+    this.parseInputFn = getParseFn(this.inputParser);
+    this.outputParser = opts.outputParser;
+    this.parseOutputFn = this.outputParser && getParseFn(this.outputParser);
```
KATT, line 103: This could maybe be simplified for usage it just has a passthrough fn

```suggestion
    this.parseOutputFn = this.outputParser ?? (val) => val;
```

packages/server/src/internals/procedure.ts
```diff
@@ -121,6 +114,20 @@ export class Procedure<TInputContext, TContext, TInput, TParsedInput, TOutput> {
     }
   }
 
+  private async parseOutput(rawOutput: unknown): Promise<TOutput> {
+    try {
+      if (process.env.NODE_ENV === 'production' || !this.parseOutputFn) {
+        return rawOutput as TOutput;
+      }
+      return await this.parseOutputFn(rawOutput);
```
KATT, line 122: Would be nice if it was:

```suggestion
  return await this.parseOutputFn(rawOutput);
```

packages/server/src/internals/procedure.ts
```diff
@@ -121,6 +114,20 @@ export class Procedure<TInputContext, TContext, TInput, TParsedInput, TOutput> {
...
+    } catch (cause) {
+      throw new TRPCError({
+        code: 'INTERNAL_SERVER_ERROR',
+        cause,
+      });
```
KATT, line 127: Definitely, it's an internal server error if this is wrong. Unsure if we should do more to guide the developer.

Maybe something like this?:

```suggestion
      throw new TRPCError({
        code: 'INTERNAL_SERVER_ERROR',
        cause,
        message: 'Output validation failed',
      });
```

KATT (commented):
This all looks really sound to me. GOOD JOB!! 🙏 🙏 

Some suggestions:

- Make the output parser stuff into its own test file
- Don't allow bad outputs regardless of the environment
- Some smaller comments inline

KATT:
You might've accidentally removed the #1182 feature BTW ☺

jlalmes:
> #1182

Rolled it back - will make sure that any transforms in the output zod schemas are also handled!

KATT:
> > #1182
> 
> Rolled it back - will make sure that any transforms in the output zod schemas are also handled!

Seems like an edge case to have transformers in the output but why not!

sachinraja (changes requested):

packages/server/test/inference.test.ts
jlalmes: Done 👍

packages/server/src/internals/procedure.ts
jlalmes: outdated

packages/server/src/internals/procedure.ts
jlalmes: outdated

packages/server/src/internals/procedure.ts
jlalmes: done 👍

packages/server/test/inference.test.ts
jlalmes: I have kept these high-level type inference related tests in this file. However, I have also added `outputParser.test.ts` with a more comprehensive set of tests relating to this PR, hope this is ok. 👍

packages/server/src/internals/procedure.ts
```diff
@@ -100,29 +96,49 @@ function getParseFn<TParsedInput>(
 export class Procedure<TInputContext, TContext, TInput, TParsedInput, TOutput> {
   private middlewares: Readonly<Array<MiddlewareFunction<any, any>>>;
   private resolver: ProcedureResolver<TContext, TParsedInput, TOutput>;
-  private readonly inputParser: ProcedureInputParser<TParsedInput>;
-  private parse: ParseFn<TParsedInput>;
+  public readonly inputParser: ProcedureParser<TParsedInput>;
+  private parseInputFn: ParseFn<TParsedInput>;
+  public readonly outputParser: ProcedureParser<TOutput> | undefined;
+  private parseOutputFn: ParseFn<TOutput>;
```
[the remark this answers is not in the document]
jlalmes: I have changed `inputParser` and `outputParser` back to `private` for now as this is out of scope for this PR. I imagine we will change them to `public` to generate openapi documentation in a future PR though 👍

packages/server/src/internals/procedure.ts
```diff
@@ -233,19 +256,14 @@ export type CreateProcedureOptions<
...
-    >
+  | CreateProcedureWithTransformInput<TContext, TInput, TParsedInput, TOutput>
   | CreateProcedureWithoutInput<TContext, TOutput>;
 
 export function createProcedure<TContext, TInput, TParsedInput, TOutput>(
   opts: CreateProcedureOptions<TContext, TInput, TParsedInput, TOutput>,
 ): Procedure<unknown, TContext, TInput, TParsedInput, TOutput> {
   const inputParser =
-    'input' in opts
+    'input' in opts && opts.input
```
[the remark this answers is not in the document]
jlalmes: done 👍

jlalmes:
## Update

Hi @KATT & @sachinraja - I have just updated this PR & resolved all outstanding comments, please review again and leave any feedback!

### Changes to this PR
* Ensure transformers are handled correctly in zod output schemas (https://github.com/trpc/trpc/pull/1699#issuecomment-1079649264)
* Removed `NODE_ENV === "production"` output validation bypass
* Added env var `TRPC_SKIP_OUTPUT_VALIDATION === "true"` output validation bypass
* Added `"Output validation failed"` error message on validation failure
* Reverted `outputParser` & `inputParser` back to `private` class members of Procedure (outside scope of this PR)
* Added a comprehensive tests for this PR to `outputParser.test.ts` file
* Refactoring

### Outstanding tasks
* API documentation
* Add examples

.

jlalmes:
Just a thought - Currently `output` is omitted from `subscription()` method. 

Do you think we should allow `output` in `subscription()` and ensure that we are validating the `emit.data(output)` output payload?

KATT (commented):
Played around with this and it works great as far as I can see.

Last thing - could you add a section in the docs? `/www` folder

KATT (commented):
Will merge this tomorrow and do a release.

I removed the `TRPC_SKIP_OUTPUT_VALIDATION` because I don't want magic runtime variables - this should either be explicitly implemented per procedure or on the app router similarily to transformers/error formatters.

KATT (approved):

PR #1703, opened 2022-03-27
Docs formatting

colinhacks:
Probably the easiest win in terms of docs readability. The fact that docusaurus doesn't apply a max-width on the main content column is bizarre. This makes the tRPC docs feel intimidating, like a wall of text. I think this helps dramatically.

Before
Screen Shot 2022-03-26 at 5 13 28 PM

After
Screen Shot 2022-03-26 at 5 13 12 PM

### DIFF
--- a/www/docs/main/introduction.md
+++ b/www/docs/main/introduction.md
@@ -1,6 +1,7 @@
 ---
 id: introduction
 title: tRPC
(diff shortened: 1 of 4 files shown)


KATT (approved):
Love some nice looking dark mode

Issue #1712, opened 2022-03-29
Update HTTP headers handling of `undefined`


ArnaudBarre:
The current typing of headers in `CreateTRPCClientOptions` allow undefined, which at runtime sends `"undefined"`, because the browser expects `Record<string, string>`. 

The simple solution is to remove undefined from typing, the other is to filter out undefined headers (which could be a nice DX when merging auth headers with always sent headers, at the cost to be not very consistent with other libs)

jlalmes:
Opened a PR, please review @ArnaudBarre https://github.com/trpc/trpc/pull/1713

PR #1713, opened 2022-03-29
Remove undefined HTTP headers from TRPCClient

jlalmes:
Closes: https://github.com/trpc/trpc/issues/1712

## Remove undefined HTTP headers from TRPCClient

Added a transform to delete headers with `undefined` values before HTTP requests are sent by the `TRPCClient`.

```javascript
{ "x-defined": "xyz", "x-undefined": undefined } 
// is transformed to
{ "x-defined": "xyz" }
```

### Changes
* Added headers transformer to `httpRequest` method
* Removed `undefined` typing from `HTTPHeaders` record
* Added test case for removing undefined headers 
* Updated @trpc/client imports to relative path in tests

### DIFF
--- a/packages/client/src/internals/httpRequest.ts
+++ b/packages/client/src/internals/httpRequest.ts
@@ -59,12 +59,13 @@ export function httpRequest<TResponseShape = TRPCResponse>(
     const url = getUrl();
 
     Promise.resolve(rt.headers())
-      .then((headers) => {
-        Object.keys(headers).forEach((key) => {
-          if (headers[key] === undefined) {
-            delete headers[key];
+      .then((rawHeaders) => {
+        const headers = Object.keys(rawHeaders).reduce((acc, key) => {
+          if (rawHeaders[key] !== undefined) {
+            return { ...acc, [key]: rawHeaders[key] };
           }
-        });
+          return acc;
+        }, {});
 
         return rt.fetch(url, {
           method: method[type],
--- a/packages/client/src/internals/httpRequest.ts
+++ b/packages/client/src/internals/httpRequest.ts
@@ -62,11 +62,12 @@ export function httpRequest<TResponseShape = TRPCResponse>(
       .then((rawHeaders) => {
         const headers: HeadersInit = { 'content-type': 'application/json' };
         for (const key in rawHeaders) {
-          if (rawHeaders[key] !== undefined) {
-            if (Array.isArray(rawHeaders[key])) {
-              headers[key] = (rawHeaders[key] as string[]).join(',');
+          const header = rawHeaders[key];
+          if (header !== undefined) {
+            if (Array.isArray(header)) {
+              headers[key] = header.join(',');
             } else {
-              headers[key] = rawHeaders[key] as string;
+              headers[key] = header;
             }
           }
         }
--- a/packages/client/src/internals/httpRequest.ts
+++ b/packages/client/src/internals/httpRequest.ts
@@ -59,8 +59,14 @@ export function httpRequest<TResponseShape = TRPCResponse>(
     const url = getUrl();
 
     Promise.resolve(rt.headers())
-      .then((headers) =>
-        rt.fetch(url, {
+      .then((headers) => {
+        Object.keys(headers).forEach((key) => {
+          if (headers[key] === undefined) {
+            delete headers[key];
+          }
+        });
+
+        return rt.fetch(url, {
           method: method[type],
           signal: ac?.signal,
           body: getBody(),
@@ -68,8 +74,8 @@ export function httpRequest<TResponseShape = TRPCResponse>(
             'content-type': 'application/json',
             ...headers,
           },
-        }),
-      )
+        });
+      })
       .then((res) => {
         return res.json();
       })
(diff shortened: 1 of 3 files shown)


packages/client/src/links/core.ts
```diff
@@ -45,7 +45,7 @@ export interface HTTPLinkOptions {
  */
 export type HttpLinkOptions = HTTPLinkOptions;
 
-export type HTTPHeaders = Record<string, string | string[] | undefined>;
+export type HTTPHeaders = Record<string, string | string[]>;
```
KATT, line 48: This is a breaking change, so better if we filter them out

jlalmes: Ok agreed! I have rolled back changes to the `HTTPHeaders` typing, `undefined`s will still be filtered out. 👍

KATT: Also, setup github sponsors, I wna send you money 😉

jlalmes: @KATT - I've just setup github sponsors 😊

packages/client/src/internals/httpRequest.ts
```diff
@@ -59,17 +59,23 @@ export function httpRequest<TResponseShape = TRPCResponse>(
     const url = getUrl();
 
     Promise.resolve(rt.headers())
-      .then((headers) =>
-        rt.fetch(url, {
+      .then((headers) => {
+        Object.keys(headers).forEach((key) => {
+          if (headers[key] === undefined) {
+            delete headers[key];
```
KATT, line 65: I like this solution but one shouldn't mutate an argument from a function

KATT: Instead, create a new object

jlalmes: Good point! Have amended 🎉

packages/server/test/headers.test.tsx
```diff
@@ -13,9 +13,13 @@ test('pass headers', async () => {
...
-        return {
-          'x-special': ctx.headers['x-special'],
-        };
+        // return only custom headers
+        return Object.keys(ctx.headers).reduce((acc, key) => {
+          if (key.startsWith('x-')) {
+            return { ...acc, [key]: ctx.headers[key] };
+          }
+          return acc;
+        }, {});
```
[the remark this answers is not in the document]
jlalmes: This is because previously `query('hello', ...)` was only responding with `{ 'x-special': string | undefined }`, however the new test I have added below would like to check the `'x-defined'` & `'x-undefined'` headers. I have therefore added this block to return all custom headers back to client (only if they start with 'x-'). Hope this is ok?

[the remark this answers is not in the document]
jlalmes: Could change to this if you prefer?

```
return {
  'x-special': ctx.headers['x-special'],
  'x-defined': ctx.headers['x-defined'],
  'x-undefined': ctx.headers['x-undefined'],
};
```

packages/client/src/internals/httpRequest.ts
```diff
@@ -59,17 +59,24 @@ export function httpRequest<TResponseShape = TRPCResponse>(
...
     Promise.resolve(rt.headers())
-      .then((headers) =>
-        rt.fetch(url, {
+      .then((rawHeaders) => {
+        const headers = Object.keys(rawHeaders).reduce((acc, key) => {
+          if (rawHeaders[key] !== undefined) {
+            return { ...acc, [key]: rawHeaders[key] };
+          }
+          return acc;
+        }, {});
```
ArnaudBarre, line 68: You remove most of spreads to avoid creating new object by starting with the default headers and mutating it.

Also this would probably surface the question on how to handle `string[]` values, because the type from the fetch api is `type HeadersInit = string[][] | Record<string, string> | Headers;`

```ts
        const headers: Record<string, string> = { 'content-type': 'application/json' }
        for (const key in rawHeaders) {
          if (rawHeaders[key] !== undefined) {
            headers[key] = rawHeaders[key] ;
          }
        }  
```

jlalmes: Thanks for this @ArnaudBarre, I have updated this PR to use the proposed code above! Regarding the `string[]` values, I have added logic to join all array values with a `","` (this is how `fetch` natively handles `[]` header values).

packages/client/src/internals/httpRequest.ts
```diff
@@ -59,17 +59,25 @@ export function httpRequest<TResponseShape = TRPCResponse>(
...
-      .then((headers) =>
-        rt.fetch(url, {
+      .then((rawHeaders) => {
+        const headers: HeadersInit = { 'content-type': 'application/json' };
+        for (const key in rawHeaders) {
+          if (rawHeaders[key] !== undefined) {
+            if (Array.isArray(rawHeaders[key])) {
+              headers[key] = (rawHeaders[key] as string[]).join(',');
+            } else {
+              headers[key] = rawHeaders[key] as string;
```
ArnaudBarre, line 69: You can remove TS assertion by using a local reference:
```suggestion
          const value = rawHeaders[key];
          if (value !== undefined) {
            if (Array.isArray(value)) {
              headers[key] = value.join(',');
            } else {
              headers[key] = value;
```

ArnaudBarre: You can remove the last one by setting header before the undefined check

KATT (approved):

Issue #1720, opened 2022-03-31
Update example links


KATT:
In docs and on website in various places

- Rather than linking to e.g. https://github.com/trpc/trpc/tree/main/examples/next-prisma-starter, let's link to https://github.com/trpc/examples-next-prisma-starter which is the same thing but it's easy to just click the "Use template button"
- I think we can start using stackblitz now since https://github.com/stackblitz/webcontainer-core/issues/45 is closed

soulsam480:
Should I migrate all of our examples to stackblitz and put the links in the docs in a PR ?

KATT:
That would be great! Make sure they work first though, I'm not sure it will

soulsam480:
I checked, any example involving prisma won't work as prisma uses binaries and stackBlitz skips them. more info https://github.com/stackblitz/core/issues/1470 . But I was able to run the non-prisma examples which are fastify, express and bare http server. I see one error with the base HTTP server. need you help with that. Opening a PR with the links

blntrsz:
Hi, replaced all occurrence of 'next-prisma-starter-websockets' and 'next-prisma-todomvc' in https://github.com/trpc/trpc/pull/1967. Other improvement suggestions are in the description.

BLamy:
Can [devcontainer.json](https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/introduction-to-dev-containers#devcontainerjson) be used to launch all the examples in codespaces?


https://github.blog/2022-04-20-codespaces-multi-repository-monorepo-scenarios/

KATT:
> Can [devcontainer.json](https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/introduction-to-dev-containers#devcontainerjson) be used to launch all the examples in codespaces?
> 
> https://github.blog/2022-04-20-codespaces-multi-repository-monorepo-scenarios/

Never tried, would be cool! Would you wna make a proof of concept?

nawed2611:
Hey, is this still open to work on?

KATT:
This is done!

PR #1726, opened 2022-04-02
docs: add stackblitz links  for some example apps

soulsam480:
Following #1720 , I've added some StackBlitz links for  `bare node`, `expressjs` and `fastify` server examples. There is one issue with the bare node server example on StackBlitz which I'm unable to fix . If anyone can look into that it'd be a great help. Here is the link  https://stackblitz.com/edit/trpc-standalone-server?file=README.md . Thx

### DIFF
--- a/www/docs/main/example-apps.md
+++ b/www/docs/main/example-apps.md
@@ -92,7 +92,7 @@ Here's some example apps:
       <td>
         <ul>
           <li><a href="https://githubbox.com/trpc/trpc/tree/main/examples/standalone-server">CodeSandbox</a></li>
-          <li><a href="https://stackblitz.com/edit/trpc-standalone-server?file=README.md">StackBlitz</a></li>
+          <li><a href="https://stackblitz.com/edit/trpc-standalone-server">StackBlitz</a></li>
           <li><a href="https://github.com/trpc/trpc/tree/main/examples/standalone-server">Source</a></li>
         </ul>
       </td>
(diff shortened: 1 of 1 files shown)


sachinraja:
Nodemon doesn't seem to be working correctly either. The server isn't stopped after a reload, so when Nodemon restarts, it will error because there is already something listening on that port.

www/docs/main/example-apps.md
```diff
@@ -102,6 +103,7 @@ Here's some example apps:
       <td>
         <ul>
           <li><a href="https://githubbox.com/trpc/trpc/tree/main/examples/express-server">CodeSandbox</a></li>
+          <li><a href="https://stackblitz.com/edit/trpc-express-server?file=README.md">StackBlitz</a></li>
```
KATT, line 106: This should be something like

https://stackblitz.com/github/trpc/trpc/tree/main/examples/express-server

www/docs/main/example-apps.md
```diff
@@ -92,6 +92,7 @@ Here's some example apps:
       <td>
         <ul>
           <li><a href="https://githubbox.com/trpc/trpc/tree/main/examples/standalone-server">CodeSandbox</a></li>
+          <li><a href="https://stackblitz.com/edit/trpc-standalone-server?file=README.md">StackBlitz</a></li>
```
KATT, line 95: See next comment

www/docs/main/example-apps.md
```diff
@@ -112,6 +114,7 @@ Here's some example apps:
       <td>
         <ul>
           <li><a href="https://codesandbox.io/s/github/trpc/trpc/tree/main/examples/fastify-server">CodeSandbox</a></li>
+          <li><a href="https://stackblitz.com/edit/trpc-fastify-server?file=README.md">StackBlitz</a></li>
```
KATT, line 117: See previous comment

KATT (changes requested):

soulsam480:
> Nodemon doesn't seem to be working correctly either. The server isn't stopped after a reload, so when Nodemon restarts, it will error because there is already something listening on that port.

Yes it's only that example which is having issues. Can you help ?

www/docs/main/example-apps.md
soulsam480: fixing them

KATT:
see #1761 for an example

Issue #1596, opened 2022-03-06
Rewrite subscriptions with new observables


KATT:
Related to #1592 / the [`next`](https://github.com/trpc/trpc/tree/next) branch.

Now that we have rxjs added, we might want to use it for our subscription support.

### Before

https://github.com/trpc/trpc/blob/ed62629326ceaa5d56ab91fe3b09ccda25a4dd6c/examples/next-prisma-starter-websockets/src/server/routers/post.ts#L130-L140

### After

```ts
  .subscription('onAdd', {
    resolve() {
      return observable<Post>((emit) => {
        const onAdd = (data: Post) => emit.data(data);
        ee.on('add', onAdd);

        return () => {
          ee.off('add', onAdd);
        };
      });
    },
  })
```

---

The underlying event emitter in the example could also be observables:

https://github.com/trpc/trpc/blob/ed62629326ceaa5d56ab91fe3b09ccda25a4dd6c/examples/next-prisma-starter-websockets/src/server/routers/post.ts#L13-L28

john-schmitz:
Just for context; why was the issue blocked?
I'm very new to the repository and trcps internals

sachinraja:
We were evaluating whether or not to use rxjs and decided not to (see discussion in #1592 for more details).

KATT:
Yeah, what @sachinraja said.

Additionally, this would include moving the observables from `@trpc/client/observables` to server

PR #1727, opened 2022-04-03
refactor using globals and fix misc type errors

### LINKED ISSUE #1588: Type error in @trpc/server with TS 4.6
Hi @KATT,  
Thank you for this great library.  

I am reporting that the server types need updates after the latest typescript release (4.6):  

```bash
node_modules/@trpc/server/dist/declarations/src/TRPCError.d.ts:7:14 - error TS2416: Property 'cause' in type 'TRPCError' is not assignable to the same property in base type 'Error'.
  Type 'unknown' is not assignable to type 'Error | undefined'.

7     readonly cause?: unknown;
               ~~~~~


Found 1 error in node_modules/@trpc/server/dist/declarations/src/TRPCError.d.ts:7
```

image

Best regards

### LINKED ISSUE #1597: Decouple `@trpc/server` from node
Related to the [`next`](https://github.com/trpc/trpc/tree/next) branch.

---

Currently, `@trpc/server` is using a lot of Node-based dependencies, i.e. `url`, `http`, etc.

It would be great to decouple `@trpc/server` completely from a node in order to easier deploy tRPC to i.e. CF workers (#1375).

## Proposal

> Please DM me on Discord before addressing this, there might be better implementations ideas, below are my first thoughts.

Rough steps:

- Rename `@trpc/server` to `@trpc/node` (and update all examples, etc)
- Create a new package, `@trpc/core`
- Gradually extract all non-node-specific stuff into a package named `@trpc/core` or similar
  - This requires a lot of refactoring as there are a lot of functions that like [`resolveHTTPResponse`](https://github.com/trpc/trpc/blob/next/packages/server/src/http/resolveHTTPResponse.ts) should still exist as part of `@trpc/core` but with a slightly different signature.
- For ease of installation, `@trpc/node` could have a hard dependency (rather than a peerDep) of `@trpc/core` and `export * from '@trpc/core'`
- Make a CF example in the `./examples` folder
- Make deno.land-module
- This should be done in the [`next`](https://github.com/trpc/trpc/tree/next)-branch

## Payment

I'll pay $200 if someone addresses it. If you think that isn't enough, feel free to contact me. Payment will be through GitHub sponsors - see #1599.

sachinraja:
closes #1597
closes #1588

Refactor globals:
  - use `URL` and `URLSearchParams` global instead of importing from `url`

Type errors:
  - update the `cause` property to conform to https://github.com/tc39/proposal-error-cause
  - fix various type errors in tests

This could have been two separate PRs but both are needed for Deno support.

### DIFF
--- a/packages/server/src/adapters/standalone.ts
+++ b/packages/server/src/adapters/standalone.ts
@@ -2,7 +2,6 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import http from 'http';
-import url from 'url';
 import { AnyRouter } from '../router';
 import {
   NodeHTTPCreateContextFnOptions,
@@ -22,7 +21,8 @@ export function createHTTPHandler<TRouter extends AnyRouter>(
   opts: CreateHTTPHandlerOptions<TRouter>,
 ) {
   return async (req: http.IncomingMessage, res: http.ServerResponse) => {
-    const endpoint = url.parse(req.url!).pathname!.slice(1);
+    const endpoint = new URL(req.url!).pathname.slice(1);
+
     await nodeHTTPRequestHandler({
       ...opts,
       req,
--- a/packages/server/src/adapters/ws.ts
+++ b/packages/server/src/adapters/ws.ts
@@ -1,9 +1,9 @@
-import http from 'http';
+import { IncomingMessage } from 'http';
 import ws from 'ws';
 import { TRPCError } from '../TRPCError';
 import { BaseHandlerOptions } from '../internals/BaseHandlerOptions';
 import { callProcedure } from '../internals/callProcedure';
-import { getErrorFromUnknown } from '../internals/errors';
+import { getCauseFromUnknown, getErrorFromUnknown } from '../internals/errors';
 import { transformTRPCResponse } from '../internals/transformTRPCResponse';
 import { AnyRouter, ProcedureType, inferRouterContext } from '../router';
 import {
@@ -84,11 +84,11 @@ function parseMessage(
  */
 export type WSSHandlerOptions<TRouter extends AnyRouter> = BaseHandlerOptions<
   TRouter,
-  http.IncomingMessage
+  IncomingMessage
 > & {
   wss: ws.Server;
   process?: NodeJS.Process;
-} & NodeHTTPCreateContextOption<TRouter, http.IncomingMessage, ws>;
+} & NodeHTTPCreateContextOption<TRouter, IncomingMessage, ws>;
 
 export function applyWSSHandler<TRouter extends AnyRouter>(
   opts: WSSHandlerOptions<TRouter>,
@@ -232,7 +232,7 @@ export function applyWSSHandler<TRouter extends AnyRouter>(
       } catch (cause) {
         const error = new TRPCError({
           code: 'PARSE_ERROR',
-          cause,
+          cause: getCauseFromUnknown(cause),
         });
 
         respond({
--- a/packages/server/src/assertNotBrowser.ts
+++ b/packages/server/src/assertNotBrowser.ts
@@ -3,6 +3,7 @@
 export function assertNotBrowser() {
   if (
     typeof window !== 'undefined' &&
+    !('Deno' in window) &&
     process.env.NODE_ENV !== 'test' &&
     process.env.JEST_WORKER_ID === undefined
   ) {
--- a/packages/server/src/http/internals/types.ts
+++ b/packages/server/src/http/internals/types.ts
@@ -1,15 +1,14 @@
+import { TRPCError } from '../../TRPCError';
+import { BaseHandlerOptions } from '../../internals/BaseHandlerOptions';
 import {
   AnyRouter,
-  Dict,
   ProcedureType,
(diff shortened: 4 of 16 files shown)


packages/server/src/assertNotBrowser.ts
```diff
@@ -3,6 +3,7 @@
 export function assertNotBrowser() {
   if (
     typeof window !== 'undefined' &&
+    !('Deno' in window) &&
```
sachinraja, line 6: Needed for Deno support

packages/server/src/http/internals/types.ts
```diff
@@ -1,15 +1,14 @@
...
+import { BaseHandlerOptions } from '../../internals/BaseHandlerOptions';
 import {
   AnyRouter,
-  Dict,
   ProcedureType,
-  ResponseMeta,
-  TRPCError,
   inferRouterContext,
   inferRouterError,
-} from '@trpc/server';
```
sachinraja, line 9: Deno/URL imports/CDNs don't support this

packages/server/src/adapters/ws.ts
```diff
@@ -232,7 +232,7 @@ export function applyWSSHandler<TRouter extends AnyRouter>(
       } catch (cause) {
         const error = new TRPCError({
           code: 'PARSE_ERROR',
-          cause,
+          cause: getCauseFromUnknown(cause),
```
KATT, line 235: 👌

packages/server/src/http/internals/types.ts
KATT: 👌

KATT: Is there a way of asserting this in a build step or something? Easy to make a mistake like this

packages/server/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "../../tsconfig.json",
   "include": ["src", "test"],
   "compilerOptions": {
-    "lib": ["ES2020"],
+    "lib": ["ES2020", "DOM"],
```
KATT, line 5: Hmmmm. How come is this needed?

KATT (commented):
This looks great!!

packages/server/tsconfig.json
sachinraja: For the tests and the `window` object in `assertNotBrowser`:
https://github.com/trpc/trpc/blob/b838ef2035435719230d895525a6f82adf93c684/packages/server/src/assertNotBrowser.ts#L5-L6

I can suppress the errors in `assertNotBrowser` and create a separate tsconfig for the tests if you like that better

KATT: Ah, didn't know window was a global in deno. All good then.

Is there a way to do a "deno build" as part of ci to ensure it all works?

packages/server/src/http/internals/types.ts
sachinraja: I can add an eslint rule for it ([no-restricted-imports](https://eslint.org/docs/rules/no-restricted-imports)).

packages/server/tsconfig.json
sachinraja: Considering it relies on a CDN to work with Deno, probably not - at least not without doing some sort of publish. We *could* do a separate build for Deno but that would definitely complicate things (although I'm open to looking into it if you want to go in that direction).

sachinraja:
Adding a cloudflare workers example needs #1596 because subscriptions currently uses the node builtin `events`.

packages/server/src/adapters/standalone.ts
```diff
@@ -22,6 +22,7 @@ export function createHTTPHandler<TRouter extends AnyRouter>(
   opts: CreateHTTPHandlerOptions<TRouter>,
 ) {
   return async (req: http.IncomingMessage, res: http.ServerResponse) => {
+    // cannot use WHATWG URL because it doesn't support relative paths
```
KATT, line 25: We could do something like

```ts
const url = req.url!.startsWith('/') ? `http://127.0.0.1${req.url}` : req.url;
```

packages/server/src/http/internals/types.ts
KATT: Do it!

packages/server/tsconfig.json
KATT: Could we use github as a cdn and have a dynamic deno script that is created and run in CI?

Not asking you to do this, just spitballing if it would work

E.g. create a file creates:

```ts
import trpc from `https://raw.githubusercontent.com/trpc/trpc/{{CURRENT_COMMIT_HASH}}/packages/server/src/index.ts`


// [..] start deno server with a little router
```

Then it could ping itself or something.

KATT: Example: https://raw.githubusercontent.com/trpc/trpc/919758d/packages/server/src/index.ts

sachinraja: There would still be other issues, Deno can work with relative file imports anyway. One thing is that in Deno, all imports must have file extensions (`.ts`).

KATT (approved):
Thanks for this!

PR #1730, opened 2022-04-05
`typeof window !== 'undefined'` to test for client

tianhuil:
Since `process.browser` is deprecated, the recommended workaround for determining if we are on the client or server on nextjs is `typeof window !== 'undefined'`: 
https://github.com/zeit/next.js/issues/5354#issuecomment-520305040

Thanks for quickly accepting this @KATT!  🙏


KATT (approved):

KATT:
Thanks!

Issue #1724, opened 2022-04-01
tRPC to OpenAPI generator


KATT:
**Note:** I do not want to have this as part of the library itself, but an example would be good

## Overview

With the addition of output parser (#1699), we should now have all the pieces to automatically create an OpenAPI-compliant endpoint

### Some rough ideas:

- Create a new endpoint `/api/openapi.yml.ts` that can use the method earlier to generate a schema
- Create a catch-all `/api/openapi/[...params].ts` that can dynamically read and call tRPC queries
- https://github.com/StefanTerdell/zod-to-json-schema#readme can be useful
- Example how to create JSON schema from tRPC routes: https://gist.github.com/KATT/993fbcbaf123f08098a154631e5af68c
- This package can be of good inspo for the openapi output https://github.com/RobinTail/express-zod-api

A starting point for the api handler:

```ts
import { ProcedureType, TRPCError } from '@trpc/server';
import { NextApiRequest, NextApiResponse } from 'next';
import { createContext } from '~/server/context';
import { appRouter } from '~/server/routers/_app';

export default async function swag(req: NextApiRequest, res: NextApiResponse) {
  const { params, ...queryParams } = req.query;
  if (req.method !== 'POST' && req.method !== 'GET') {
    res.status(405).json({
      message: 'Invalid method',
    });
    return;
  }
  const path = Array.isArray(params) ? params.join('/') : params;
  const ctx = await createContext({ req, res });
  const type: ProcedureType = req.method === 'POST' ? 'mutation' : 'query';
  try {
    const caller = appRouter.createCaller({ ctx });

    const fn = caller[type] as any;
    const result = await fn(path, queryParams);

    res.json(result);
    return;
  } catch (error) {
    if (!(error instanceof TRPCError)) {
      res.status(500).json({
        message: 'Something went wrong',
      });
      return;
    }
    const formatted = appRouter.getErrorShape({
      error,
      path,
      ctx,
      input: params,
      type,
    });
    // TODO: use https://github.com/trpc/trpc/blob/main/packages/server/src/http/internals/getHTTPStatusCode.ts#L55-L59
    // Feel free to make them publicly exposed
    const status = 500;
    res.status(500).json(formatted);
  }
}
```

LouisHaftmann:
Would this be used to generate mutation/query routes that comply to the OpenAPI specification or just to generate a swagger schema?

KATT:
[...]

PR #1743, opened 2022-04-07
Router methods input type narrowing

jlalmes:
## Router methods input type narrowing

Ensure router method `TInput` type (`input:`) satisfies `TInput` type defined in the `ProcedureRecord<...>` for each `TQueries | TMutations | TSubscriptions`. 

This non-breaking change is required for composing type-safe OpenAPI compliant tRPC routers in a future PR.

```typescript
// Create a Router that will only allow `input` type of `string` on `router.[method](...)`
const narrowedRouter = new Router<
    TContext,
    TContext,
    TMeta,
    ProcedureRecord<TContext, TContext, TMeta, string, any, any, any>,
    ProcedureRecord<TContext, TContext, TMeta, string, any, any, any>,
    ProcedureRecord<TContext, TContext, TMeta, string, unknown, Subscription<any>, unknown>,
    TErrorShape
  >();

// The following will throw a TypeScript compilation error. (Expects `input` to be typeof `string`)
narrowedRouter.query('invalidInputType', {
  input: z.number(),
  resolve: ({ input }) => {
    return { input };
  }
});
```

### Additional changes
* Added export of the `Router` type (returned by `router()`) as `TRPCRouter`.


LouisHaftmann (approved):

KATT:
Is this really _required_? It could just automatically omit the ones that don't fulfill the requirement for swagger?

jlalmes:
I suppose it's not _really_ required in the context of OpenAPI if we go with the approach shared here https://github.com/trpc/trpc/issues/1724#issuecomment-1091768926. But this change doesn't affect current functionality nor introduces much of a TS overhead, and type narrowing is always good. Won't be offended if you want to close this one for now.

PR #1761, opened 2022-04-09
Update README.md

KATT:
cc @soulsam480 

This is how it needs to be

- Link the `/github/*`-url so it gets updated automatically
- Add `?file=x,y` to make sure both files are added
- If it's not a frontend-example, do `&view=editor` to hide the explorer
- `&terminal=dev` to trigger the dev script in package.json


KATT:
Merging to test if it works

Issue #1778, opened 2022-04-12
Simplify tRPC Request/Response/Message Interfaces


jlalmes:
Originally posted by @KATT 

> See @jlalmes' PR https://github.com/trpc/trpc/pull/1759
> 
> Potential things. Not definitive, just ideas.
> 
> 
> ## Remove  unused fields
> 
> Before:
> 
> ```js
> [
>   {
>     "id": null,
>     "result": {
>       "type": "data",
>       "data": {
>         "id": "1",
>         "title": "Hello tRPC",
>         "body": "..."
>         // ...
>       }
>     }
>   },
> ]
> ```
> 
> After:
> 
> ```js
> [
>   {
>     "result": {
>       "data": {
>         "id": "1",
>         "title": "Hello tRPC",
>         "body": "..."
>         // ...
>       }
>     }
>   },
> ]
> ```
> 
> ## Changed response to a mapped object
> 
> Potential ease of allowing for streaming... maybe we could start sending response as procedures are finalised and not wait for _all of them_ to be finalized before returning... but it also breaks HTTP-codes so not sure.
> 
> Probably not a good idea, just a thought
> 
> ```js
> {
>   // 0 means index 0 of the batch
>   "0": { 
>     "id": null,
>     "result": {
>       "type": "data",
>       "data": {
>         "id": "1",
>         "title": "Hello tRPC",
>         "body": "..."
>         // ...
>       }
>     }
>   },
> ]
> ```
> 
> ## Remove JSONRPC 
> 
> Remove JSNORPC stuff on the request/response for internal requests...
> 
>

jlalmes:
> cc @KATT

## Documenting the current tRPC API

Thought this would be good for visibility.

#### HTTP Queries
```jsonc
// HTTP Query
/* --> GET /sayHello?input=%7B%22name%22%3A%22tRPC%22%7D */ 
// no body sent
/* <-- */ 
{ "id": null, "result": { "type": "data", "data": { "greeting": "Hello tRPC" } } }

// HTTP Query with request batching
/* --> GET /sayHello,sayHello?batch=1&input=%7B%220%22%3A%7B%22name%22%3A%22tRPC%22%7D%2C%221%22%3A%7B%22name%22%3A%22tRPC2%22%7D%7D */ 
// no body sent
/* <-- */ 
[
  { "id": null, "result": { "type": "data", "data": { "greeting": "Hello tRPC" } } },
[...]

PR #1780, opened 2022-04-13
Simplify tRPC request/response/message interfaces

jlalmes:
Closes https://github.com/trpc/trpc/issues/1778

## Simplify tRPC request/response/message interfaces
 
  * Remove `"id": null` from HTTP response envelopes
  * Remove `"type": "data"` from HTTP response envelopes
  * Remove `"jsonrpc": "2.0"` from WS message envelopes

### DIFF
--- a/packages/client/src/links/wsLink.ts
+++ b/packages/client/src/links/wsLink.ts
@@ -2,8 +2,9 @@ import { AnyRouter, ProcedureType } from '@trpc/server';
 import {
   TRPCClientIncomingMessage,
   TRPCClientIncomingRequest,
-  TRPCRequest,
-  TRPCResponse,
+  TRPCClientOutgoingMessage,
+  TRPCRequestMessage,
+  TRPCResponseMessage,
 } from '@trpc/server/rpc';
 import { TRPCClientError } from '../TRPCClientError';
 import { retryDelay } from '../internals/retryDelay';
@@ -30,7 +31,7 @@ export function createWSClient(opts: WebSocketClientOptions) {
   /**
    * outgoing messages buffer whilst not open
    */
-  let outgoing: TRPCRequest[] = [];
+  let outgoing: TRPCClientOutgoingMessage[] = [];
   /**
    * pending outgoing requests that are awaiting callback
    */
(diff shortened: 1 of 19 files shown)


packages/server/src/rpc/envelopes.ts
```diff
@@ -36,74 +15,114 @@ export interface TRPCErrorShape<
...
-      type: 'data';
-      data: TData;
-    }
-  | {
-      type: 'started';
-    }
-  | {
-      type: 'stopped';
-    };
+export namespace JSONRPC2 {
```
KATT, line 20: Nice

packages/client/src/links/wsLink.ts
```diff
@@ -30,7 +31,7 @@ export function createWSClient(opts: WebSocketClientOptions) {
   /**
    * outgoing messages buffer whilst not open
    */
-  let outgoing: TRPCRequest[] = [];
+  let outgoing: TRPCClientOutgoingMessage[] = [];
```
KATT, line 34: This file is a complete mess - I'm happy you figured it out  😅

KATT (commented):
Thoughts

- I'm thinking we might need to return the `jsonrpc`-thing in WS if the client sends it in order to be compliant-when-asked to JSON-RPC
- Docs needs updating

jlalmes:
### Changes
* If incoming websocket message includes `"jsonrpc"` key, it will be sent back with the outgoing response message.
* Updated API payloads in `/docs/rpc` & `/docs/subscriptions`

KATT:
This is good - making a follow-up task

packages/server/src/adapters/ws.ts
```diff
@@ -74,11 +74,7 @@ function parseMessage(
   const { input: rawInput, path } = params;
   assertIsString(path);
   const input = transformer.input.deserialize(rawInput);
-  return {
-    id,
-    method,
-    params: { input, path },
-  };
+  return { jsonrpc, id, method, params: { input, path } };
```
KATT, line 77: We need a test that actually asserts this otherwise I will break it eventually, guaranteed

Here's a websocket raw client test we can use as a base 

https://github.com/trpc/trpc/blob/97039a761b8704b980fe1a62eaadb5af84942754/packages/server/test/websockets.test.ts#L473-L510

jlalmes: Will do!

KATT: #1785

PR #1786, opened 2022-04-14
Added websocket `"jsonrpc"` message envelope test

jlalmes:
## Include `"jsonrpc"` in response if sent with message

* Bug fix `"jsonrpc"` in websocket message envelopes
* Added test

@KATT - Glad you asked for a test (found a bug).

### DIFF
--- a/packages/server/test/websockets.test.ts
+++ b/packages/server/test/websockets.test.ts
@@ -13,7 +13,11 @@ import { wsLink } from '../../client/src';
 import * as trpc from '../src';
 import { TRPCError } from '../src';
 import { applyWSSHandler } from '../src/adapters/ws';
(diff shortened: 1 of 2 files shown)


packages/server/test/websockets.test.ts
```diff
@@ -670,3 +674,169 @@ test('regression - badly shaped request', async () => {
   rawClient.close();
   t.close();
 });
+
+describe('include "jsonrpc" in response if sent with message', () => {
+  test('queries & mutations', async () => {
```
KATT, line 679: you could flatten this 

```suggestion
test('include "jsonrpc" in response if sent with message', async () => {
```

but it's not like the tests are very well architected so i don't mind :D

KATT (approved):

PR #1790, opened 2022-04-15
fix: error with cause

skarab42:
Related to #1773


skarab42:
@KATT I think it's OK to upgrade to 4.6.3

skarab42:
I wonder if this is a breaking change for people who publish plugins outside the tRPC eco system?

sachinraja:
This is fixed on next due to it being a slight breaking change (see #1727)

KATT:
Thanks for this @skarab42, closing as it's fixed on the `next`-branch

PR #1812, opened 2022-04-21
docs: add note about fastify param length limit

kindlyfire:
This adds a note to prevent people from encountering an error when tRPC batches too many queries, which generates long HTTP request URI queries, which Fastify will reject by default.


KATT:
@all-contributors add @kindlyfire for docs

PR #1834, opened 2022-04-27
fix middleware context swapping bug

### LINKED ISSUE #1833: Context swapped in by middleware is lost after chaining through another middleware
### Scenario:
- Add a middleware on a router that has context set to A
  - Call `next` with a new context B
- Stack another middleware on top
  - Call `next` without passing on a specific context
- Add a query and use the context

### Expected result:
The query's context is context B

### Actual result:
The query's context is context A (but in Typescript, the context type now suggests it is context B)

### Code sample:
```
const router = createRouter()
    .middleware(({ next }) => {
        return next({ ctx: "modified" });
    })
    .middleware(({ ctx, next }) => {
        // ctx is "modified"
       return next();
    })
    .query("foo", {
        resolve({ ctx }) {
            // ctx is now "original" again
        }
    });

app.use(
  '/trpc',
  trpcExpress.createExpressMiddleware({
    router,
    createContext() {
        return "original";
    },
  })
);
```

KATT:
Closes #1833 

### Changes

- Rewrite middleware call stack with recursion that can keep track of the `ctx` from the previous call
- Feel free to take this PR over and refactor it. I wrote this after a few beers and I'm not proud of the code.


mmkal (commented):
Erk, sorry for missing this case. There may be a slightly cleaner or smaller change that could fix this (thoughts: maybe we just need to make sure we always pass opts, maybe .reduce could avoid tracking indexes and maybe we don’t need the anys) but if this fixes the bug then I say 👍

KATT:
Feel free to make a PR to rewrite it with reduce or map :)

mgreenw:
+1 to what @mmkal said!

PR #1836, opened 2022-04-27
Move imports from react-query/hydration to react-query

lukahartwig:
This entrypoint will be removed in `react-query@4` and currently breaks `react-query@beta` because of the missing entry in the export map. See https://react-query-beta.tanstack.com/guides/migrating-to-react-query-4#separate-hydration-exports-have-been-removed

PR #1838, opened 2022-04-27
move imports from react-query/hydration to react-query

lukahartwig:
As requested [here ](https://github.com/trpc/trpc/pull/1836) I opened a PR from within the repo


packages/next/src/withTRPC.tsx
```diff
@@ -18,8 +18,13 @@ import {
   NextPageContext,
 } from 'next/dist/shared/lib/utils';
 import React, { createElement, useState } from 'react';
-import { QueryClient, QueryClientProvider } from 'react-query';
-import { DehydratedState, Hydrate, dehydrate } from 'react-query/hydration';
+import {
+  DehydratedState,
+  Hydrate,
```
KATT, line 23: Do you know at which minor this was changed in react-query 3? We would need to update `peerDependencies` to that version as well in

KATT (commented):
Thanks for this! Looks great, just left one question

KATT:
@all-contributors add @lukahartwig for code

packages/next/src/withTRPC.tsx
lukahartwig: The change was made in https://github.com/tannerlinsley/react-query/releases/tag/v3.22.0. I double checked the `peerDependencies`, all of them are above that.

KATT: Perfect

KATT (approved):

PR #1852, opened 2022-05-01
add fetch adapter

### LINKED ISSUE #1374: Feature request: Add Fetch API adapter aka "Web Platform support"
Eg. support for the [Request][] and [Response][] objects specified in the [Fetch Standard][]. It would allow deploying to Remix routes for example.

Unfortunately I don't have the time to work an a proper PR but I created a userland adapter that can be used with Remix:

https://github.com/valu-digital/npm-packages/tree/master/packages/trpc-fetch-api-adapter

The [implementation](https://github.com/valu-digital/npm-packages/blob/9482a8c71b99660ec9625ccdbe33a1e0ebb80dcb/packages/trpc-fetch-api-adapter/src/adapter.ts) is probably very naive but at least it is some short of starting point.

[request]: https://developer.mozilla.org/en-US/docs/Web/API/Request
[response]: https://developer.mozilla.org/en-US/docs/Web/API/Response
[fetch standard]: https://fetch.spec.whatwg.org/
[remix]: https://remix.run/
[cloudflare workers]: https://workers.cloudflare.com/
[deno]: https://deno.land/

sachinraja:
closes #1374
part of #1597

Adds a [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) adapter to `@trpc/server`.

Also adds a [Cloudflare Workers](https://workers.cloudflare.com/) example. Right now tests and the actual router are just copied from the Express example. Preferably both examples would use the same files so an update to the code only has to happen in one place, but not sure how to make that happen unless I move the files out of the individual example directories.

Let me know if I should add myself as a codeowner.

cc @icflorescu who may be interested as the author of [trpc-sveltekit](https://github.com/icflorescu/trpc-sveltekit) (would love a review if you have time)
also cc @esamattis as the author of [trpc-fetch-api-adapter](https://github.com/valu-digital/npm-packages/tree/master/packages/trpc-fetch-api-adapter) (would also love a review)

Thanks to both of you for your packages! Took a lot of inspiration from them.


KATT:
Cc @esamattis

packages/server/src/adapters/fetch/fetchRequestHandler.ts
```diff
@@ -0,0 +1,62 @@
...
+export async function fetchRequestHandler<TRouter extends AnyRouter>(
+  opts: FetchHandlerRequestOptions<TRouter>,
+): Promise<Response> {
+  const createContext = async () => {
+    return opts.createContext?.({ req: opts.req });
+  };
+
+  const url = new URL(opts.req.url);
+  const path = url.pathname.slice(opts.endpoint.length + 1);
+  console.log(path);
```
KATT, line 23:
```suggestion
```

packages/server/src/adapters/fetch/fetchRequestHandler.ts
```diff
@@ -0,0 +1,62 @@
...
+  const req: HTTPRequest = {
+    query: url.searchParams,
+    method: opts.req.method,
+    headers: Object.fromEntries(opts.req.headers),
+    body: await opts.req.text(),
+  };
```
KATT, line 29: Later, we should probably invert this lower down so `fetch` is the default and that the "old stuff" are mapped instead

packages/server/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "../../tsconfig.json",
   "include": ["src", "test"],
   "compilerOptions": {
-    "lib": ["ES2020", "DOM"],
+    "lib": ["ES2020", "DOM", "DOM.Iterable"],
```
KATT, line 5: Do we need this when we are targeting node 14? 🤔 

Related: https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping#node-14

examples/cloudflare-worker/src/server.ts
```diff
@@ -0,0 +1,111 @@
...
+        if (ctx.user?.name !== 'alex') {
+          throw new TRPCError({ code: 'FORBIDDEN' });
+        }
+        return {
+          secret: 'sauce',
+        };
+      },
+    }),
+  )
+  .merge('messages.', messages);
```
KATT, line 109: Maybe let's make this example more minimal, we maybe only need 1 mutation and 1 query. Less maintainence

KATT (changes requested):
Looks really good!

A few pieces I'd want to improve:

- [ ] Add follow-up ticket to invert `resolveHTTPResponse` so that modern handling is the default and old handler
- [ ] Docs for the adapter
- [ ] Add example project to the list of example projects
- [x] Simplify example project to be just a mutation and a query?

packages/server/tsconfig.json
sachinraja: Yes, it is needed to add `Symbol.iterator` to `Headers` so we can use `Object.fromEntries`:
https://github.com/trpc/trpc/blob/dcc48da36ca3056b2747fb95303ed9edf1c97ce5/packages/server/src/adapters/fetch/fetchRequestHandler.ts#L27

Note that Node does not implement the Fetch spec.

packages/server/src/adapters/fetch/fetchRequestHandler.ts
sachinraja: Good idea, definitely

packages/server/src/adapters/fetch/fetchRequestHandler.ts
sachinraja: Should I add a lint rule for this?

KATT: Sure! Can he useful in examples though

sachinraja: Added it only for `packages/**/*`

KATT (approved):
LGTM - @jlalmes, what you think?

jlalmes (approved):
Not very familiar with Cloudflare Workers, but LGTM! ✅

sachinraja:
Thanks for reviewing! As a side note, this is much more than just CF Workers. Deno (Deploy), SvelteKit, Astro, Next.js middleware, Remix, and others are also using the Fetch API.

Issue #1801, opened 2022-04-19
Idea: Add `maxBatchURILength` config option to `httpBatchLink`


jlalmes:
https://trpc.io/docs/links#setting-a-maximum-batch-size

## Idea: 
The maximum number of requests that can be batched at one time is currently limited by the `maxBatchSize` config option. There are cases where queries contain large input payloads and therefore, when batched, exceed the maximum URI length of the deployment environment before being split up by the `maxBatchSize` limit (cc. https://github.com/trpc/trpc/discussions/1792)

### Proposal
Add a new config option `maxBatchURILength`(?) to the `httpBatchLink` that will split the batch into smaller chunks before sending requests if the initial batch URI exceeds the limit.

KATT:
Yep. Would be better, this was initially what I wanted to do but it required a lot of refactoring so I took the shortcut. 

Happy to do a breaking change on this for the next major - both features do not need to exist.

See the original issue at #506

KATT:
I changed it to "Maybe" as it's a nice-to-have but feel free to attack it, @jlalmes!

jlalmes:
Note to self: #1812

PR #1855, opened 2022-05-03
Move transformers back to `runtime`

jlalmes:
Closes https://github.com/trpc/trpc/issues/1807.
Supersedes https://github.com/trpc/trpc/pull/1669.

## Move transformers back to `runtime`

This PR removes the `transformerLink` (introduced in https://github.com/trpc/trpc/pull/1366), and adds `transformer` back to the `TRPCClientRuntime`.

PR #1858, opened 2022-05-04
Add `maxURLLength` config option to `httpBatchLink`

jlalmes:
Closes https://github.com/trpc/trpc/issues/1801.
Please merge https://github.com/trpc/trpc/pull/1855 first.

## Add `maxURLLength` config option to `httpBatchLink`

This PR sets a sensible default maximum URL length of batched requests to `2048` (cc. https://www.sistrix.com/ask-sistrix/technical-seo/site-structure/url-length-how-long-can-a-url-be). This can be overridden by setting `maxURLLength` on the `httpBatchLink`.

If a batch only contains one query but it is still larger than the `maxURLLength`, the request is sent by itself regardless. Should we do something like this?
```ts
if (process.env.NODE_ENV !== 'production') {
  console.warn('tRPC batched request was over the specified maxURLLength: ${maxURLLength}. Find out more here: ${docsURL}');
}
```

### Changes
* Added optional `maxURLLength` config to `httpBatchLink`.
* Removed `maxBatchSize` config from `httpBatchLink`.
* Added tests for new `maxURLLength` config option.

Let me know any thoughts & I will update docs before merging.


KATT (commented):
One thing I'd change with the behaviour - rather than splitting off the batch when adding, can we do the splitting when we dispatch?

This makes it so a procedure that is fired and cancelled within the same millisecond won't have a `fetch` call done.

KATT (approved):
[Could be improved slightly](https://github.com/trpc/trpc/pull/1858#pullrequestreview-964910115) but think it works well as a first stab - @sachinraja, what do you think?

jlalmes:
Thanks for the feedback @KATT - I didn't notice that edge case. Don't merge this yet I will add a new test and make the suggested change!

jlalmes:
### Changes

* Implemented request batching from inside `dataLoader.dispatch` function.
* Added test for batched request cancellation

Regarding `DEFAULT_MAX_URL_LENGTH`:

```
Maximum URL length for popular browsers
The following maximum URL lengths apply to commonly used browsers. Since all of these browsers have a substantial market share, website operators have to stick to the lowest common denominator and limit their URLs to 2,083 characters in length.

Microsoft Internet Explorer: 2,083 characters
Microsoft Edge: 2,083 characters
Google Chrome: 32,779 characters
Mozilla Firefox: more than 64,000 characters
Apple Safari: more than 64,000 characters
Google Android: 8,192 characters
```

Happy to change the default to **2,083**, but I we should chose stay on the safe side and document the fact that you might want to increase this limit on a case-by-case basis.

www/docs/client/links.md
```diff
@@ -25,9 +25,9 @@ const somePosts = await Promise.all([
 
 > The below examples assuming you use Next.js, but the same as below can be added if you use the vanilla tRPC client
 
-### Setting a maximum batch size
+### Setting a maximum URL length
 
-This limits the number of requests that can be sent together in batch ( useful to prevent the url from getting too large and run into [HTTP error 413](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413) ).
+When sending batch requests, sometimes the URL can become too large causing [HTTP error 413](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413). The `maxURLLength` option will limit the number of requests that can be sent together in a batch. This value defaults to 2,083 but you may wish to increase or decrease this dependent on your deployment environment & use case. You can read more about URL lengths [here](https://www.sistrix.com/ask-sistrix/technical-seo/site-structure/url-length-how-long-can-a-url-be).
```
sachinraja, line 30: I would've thought it would be [HTTP error 414](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/414) "URI Too Long." Maybe we should also make a note for that?

sachinraja:
IE is EOL so I wouldn't worry about that. For Microsoft Edge I'm not sure why it's so different from Chrome. Maybe they're referring to the older version of Edge and not the Chromium based one?

www/docs/client/links.md
jlalmes: Hmm, it also looks like it will sometimes result in a 404 error.. https://github.com/trpc/trpc/discussions/1792

sachinraja: Maybe something like this:
```
When sending batch requests, sometimes the URL can become too large causing HTTP errors like ["413 Payload Too Large"](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413), ["414 URI Too Long"](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/414), and ["404 Not Found"](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404).
```

Just to make it easier to find the section if someone gets the error.

jlalmes:
Ok you've convinced me 😂. Should we set the default to `Infinity`? @KATT @sachinraja

sachinraja (approved):

sachinraja:
> If a batch only contains one query but it is still larger than the maxURLLength, the request is sent by itself regardless. Should we do something like this?

Yes, I think we should warn the user there.

PR #1882, opened 2022-05-14
Create plugin for Cloudflare Pages

toyamarinyon:
On May 14, Cloudflare released Pages Plugin that allows to reuse chunk of runtime code that can be incorporated anywhere within some Pages application.

[Introducing Pages Plugins](https://blog.cloudflare.com/cloudflare-pages-plugins/)

So, I created Pages Plugin for tRPC.

This plugin is a thin wrapper for fetch-adapter, so there is not much difference between with and without plugin. Especially for those familiar with tRPC, it may be easier to make adjustments if it is not used.

But on first time to use tRPC, I think that it more friendly with plugin.

And I think that number of Cloudflare user will increase significantly, so tRPC become more popular once tRPC will support Cloudflare Pages and Cloudflare Pages Functions.

Thanks for @sachinraja(creator of fetch-adapter #1852 ).  Since there is fetch-adapter in tRPC, I could create this plugin fast.


sachinraja:
Wow this is great! Glad to see the fetch adapter enabling these kind of integrations. Personally I think the pages plugin should be a userland package though.

toyamarinyon:
@sachinraja  Thank you for looking!

> Personally I think the pages plugin should be a userland package though.

Actually, I had initially created it as a userland package.

But I thought perhaps it would be more convenient to include it in tRPC, so I created it here.  Of course, userland would be fine.

@KATT what do you think?

KATT:
I think this is great!I definitely don't mind the examples folder but I do not want to maintain another package - so I'd prefer that to be userland plugin and having it linked on the awesome trpc page.

toyamarinyon:
@KATT OK 👍 I'll provide it as userland plugin.

By the way, there isn’t `fetch-adapter` at `^10.0.0-alpha.21` so can you tag `^10.0.0-alpha.22` on current `next` branch?

KATT:
> By the way, there isn’t `fetch-adapter` at `^10.0.0-alpha.21` so can you tag `^10.0.0-alpha.22` on current `next` branch?

Did it now!

KATT:
FYI, the next major will contain quite a lot of breaking changes that haven't landed yet, that's why I haven't shouted much about it and tried to get people to use the alpha :)

KATT:
@all-contributors add @toyamarinyon for code

toyamarinyon:
I published [cloudflare-pages-plugin-tprc](https://github.com/toyamarinyon/cloudflare-pages-plugin-trpc) 🎉 

You can see demo site on https://cloudflare-pages-plugin-trpc.pages.dev.

PR #882, opened 2021-09-04
Add lambda gateway handler

Lilja:
to do:
- [ ] Add tests
- [ ] Fix example
  - [ ] add `yarn dev` / `build`
  - [ ] add `test-start` / `test dev` 
  - [ ] add to [workflows.yml](https://github.com/trpc/trpc/blob/main/.github/workflows/examples.yml)
  - [ ] mv app code to `src` folder
- [ ] Remove TODO:s in code

closes #713

PR #1907, opened 2022-05-19
API Gateway handler

Lilja:
Supersedes https://github.com/trpc/trpc/pull/882, https://github.com/trpc/trpc/pull/1901

### TODO

- [x] Add tests.
- [x] Remove todos in code.

### Rationale
I hade a lot of trouble trying to reason with the `TRequest` generic, which is why I stopped with the other PR. I think it makes more sense for the other node frameworks like Express and Fastify to have it, but not really for api-gateway/lambda. Hence, I removed all of the references to it for the types of API gateway.  I tried to refactor it out to the extent where two hacks were introduced:

* One to reach the `ResponseFnMeta`. // TODO: hack 2
* Skip `HTTPBaseHandlerOptions`, which includes the `TRequest` type. // TODO: hack 1

### Example library
Just a normal server/client written to be used in node. It also uses my own library to mock an API gateway locally with hot reloading capabilities, something `aws-sam` can't do.

### To discuss:
* ~Is `LambdaContext` narrow enough? In my experiences I've never touched the context blob so I'm not sure how many people use it. Perhaps good enough for now?~ Fixed it by implementing the actual Type for it. 
* Is the use of `AWSLambdaOptions` ok? It's not implementing basing it's type of `HTTPBaseHandlerOptions` like `NodeHTTPHandlerOptions` and friends are doing. I've decided to skip it because it doesn't really make sense to introduce the `TRequest` generic.
* In my old pr I made a remark about workflow. Since it's a nodejs client and not really an app, does it really make sense to us e that?


Lilja:
That lint job, I cannot make sense of it.

KATT:
> That lint job, I cannot make sense of it.

Try `yarn lint-fix`


It's something about duplicated dependencies in the monorepo

packages/server/src/adapters/aws-lambda.ts
```diff
@@ -0,0 +1,80 @@
...
+        opts?.onError?.({
+          ...o,
+          req: event,
+        });
+      },
+    });
+
+    const resp: APIGatewayProxyResult = {
+      statusCode: response.status,
+      body: response.body ?? '',
```
sachinraja, line 75: I don't think it makes sense for this to be an empty string. If we're using payload format v2, [body can be `undefined`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/05ca376629ce89983fb50d6af08c85d81a111734/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L236).

packages/server/src/adapters/lambda-utils.ts
```diff
@@ -0,0 +1,30 @@
+import type { APIGatewayProxyEvent } from 'aws-lambda';
```
sachinraja, line 1: To be consistent with the other adapters, let's create a separate directory for this and put `aws-lambda` and `lambda-utils` inside it. As a side note, tRPC uses a camelCase file naming convention. So the renaming should be like this:
```ts
adapters/aws-lambda.ts -> adapters/lambda/lambdaRequestHandler.ts
adapters/lambda-utils.ts -> adapters/lambda/types.ts
```

packages/server/src/adapters/aws-lambda.ts
```diff
@@ -0,0 +1,80 @@
...
+  return obj;
+}
+
+export function createApiGatewayHandler<TRouter extends AnyRouter>(
+  opts: AWSLambdaOptions<TRouter>,
+): (
+  event: APIGatewayProxyEvent,
+  context: LambdaContext,
+) => Promise<APIGatewayProxyResult> {
+  return async (event: APIGatewayProxyEvent, context: LambdaContext) => {
```
sachinraja, line 47:
```suggestion
  return async (event, context) => {
```
inferred from return annotation

packages/server/src/adapters/aws-lambda.ts
```diff
@@ -0,0 +1,80 @@
...
+  for (const [key, value] of Object.entries(headers)) {
+    if (typeof value === 'undefined') {
+      continue;
+    }
+    obj[key] = Array.isArray(value) ? value.join(', ') : value;
+  }
+  return obj;
+}
+
+export function createApiGatewayHandler<TRouter extends AnyRouter>(
```
sachinraja, line 41:
```suggestion
export function lambdaRequestHandler<TRouter extends AnyRouter>(
```

packages/server/src/adapters/lambda-utils.ts
```diff
@@ -0,0 +1,30 @@
...
+  };
+  // TODO: Hack 1.
+  onError?: (options: Record<string, unknown>) => void;
+  responseMeta?: ResponseMetaFn<TRouter>;
+} & {
+  /**
+   * @link https://trpc.io/docs/context
+   **/
+  createContext: LambdaCreateContextFn<TRouter>;
+};
```
sachinraja, line 30:
```suggestion
export type LambdaCreateContextOption<TRouter extends AnyRouter> =
  unknown extends inferRouterContext<TRouter>
    ? {
        /**
         * @link https://trpc.io/docs/context
         **/
        createContext?: LambdaCreateContextFn<TRouter>;
      }
    : {
        /**
         * @link https://trpc.io/docs/context
         **/
        createContext: LambdaCreateContextFn<TRouter>;
      };

export interface AWSLambdaOptions<TRouter extends AnyRouter>
  extends HTTPBaseHandlerOptions<TRouter, APIGatewayProxyEvent> & LambdaCreateContextOption<TRouter>
}
```
refer to https://github.com/trpc/trpc/blob/next/packages/server/src/adapters/fetch/types.ts

sachinraja (dismissed):

sachinraja:
> Is LambdaContext narrow enough? In my experiences I've never touched the context blob so I'm not sure how many people use it. Perhaps good enough for now?

Does `@types/aws-lambda` provide a type for it? If so, then let's use that.

> Is the use of AWSLambdaOptions ok? It's not implementing basing it's type of HTTPBaseHandlerOptions like NodeHTTPHandlerOptions and friends are doing. I've decided to skip it because it doesn't really make sense to introduce the TRequest generic.

See my review comment - you can just pass the concrete type instead of implementing a generic.

> In my old pr I made a remark about workflow. Since it's a nodejs client and not really an app, does it really make sense to use that?
Could you elaborate? I'm not sure what this is referring to.

packages/server/src/adapters/aws-lambda.ts
```diff
@@ -0,0 +1,80 @@
...
+import { HTTPHeaders, HTTPRequest } from '../http/internals/types';
+import { AnyRouter, inferRouterContext } from '../router';
+import type { AWSLambdaOptions, LambdaContext } from './lambda-utils';
+
+export type { CreateLambdaContextOptions } from './lambda-utils';
+
+function lambdaEventToHTTPRequest(event: APIGatewayProxyEvent): HTTPRequest {
+  const query = new URLSearchParams();
+  for (const [key, value] of Object.entries(
+    event.multiValueQueryStringParameters ?? {},
```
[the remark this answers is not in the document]
Lilja: I actually did not know that there were two different version, which sadly affects this. I'll change to `queryStringParameters`.

I have only experience building Rest APIs(as opposed to HTTP APIs) in API Gateway and I _would assume_ the 2.0 format is for HTTP apis. I cannot change to 2.0 in my existing API gateway deployment. I'm not sure either if/how you could check what format. That documentation page lists a "version" string in the json blob but I can't find it in actuality(checked the Lambda Event property)

examples/lambda-api-gateway/src/client.ts
```diff
@@ -0,0 +1,21 @@
...
+(async () => {
+  try {
+    const q = client.query('/greet', {
+      name: 'Erik',
+    });
+    console.log(await q);
+  } catch (error) {
+    console.log('error', error);
+  }
+})();
```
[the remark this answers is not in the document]
KATT: None of the other examples uses top-level await, I haven't gotten into that yet 😅

packages/server/src/adapters/aws-lambda.ts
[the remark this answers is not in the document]
sachinraja: I don't think we should support both formats. We should make it clear that the adapter will only work with v2.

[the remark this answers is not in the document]
mmkal: As far as I know trpc clients won't be making any requests with multi-value query string parameters anyway. I suspect by using `event.queryStringParameters`  you can simplify lines 11-18 to:

```ts
const query = new URLSearchParams(event.queryStringParameters)
```

Or this whole function body can just become:

```ts
  return {
    method: event.httpMethod,
    query: new URLSearchParams(event.queryStringParameters),
    headers: event.headers,
    body: event.body,
  };

File: packages/server/src/adapters/aws-lambda.ts [hunk 8]
```diff
@@ -0,0 +1,80 @@
...
+): APIGatewayProxyResult['headers'] {
+  const obj: APIGatewayProxyResult['headers'] = {};
+
+  for (const [key, value] of Object.entries(headers)) {
+    if (typeof value === 'undefined') {
+      continue;
+    }
+    obj[key] = Array.isArray(value) ? value.join(', ') : value;
+  }
+  return obj;
```
mmkal, line 38: Nit: not sure if mutating objects is the convention in this repo but usually `Object.fromEntries` is preferred these days:
```suggestion
  return Object.fromEntries(
    Object.entries(headers)
      .filter((_name, value]) => typeof value !== 'undefined') 
      .map(([name, value]) => Array.isArray(value) ? value.join(',') : value)
  );
```

(not sure if excluding `undefined` is strictly necessary, if not the `.filter` line can be deleted)

packages/server/test/adapters/aws-lambda.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+  delete (result as any).body;
+  expect(result).toMatchInlineSnapshot(`
+Object {
+  "headers": Object {
+    "Content-Type": "application/json",
+  },
+  "statusCode": 200,
+}
+`);
+  expect(body).toMatchInlineSnapshot(`
```
mmkal, line 67: Can avoid the `any`/usage of `delete` here:

```suggestion
  const { body, ...result } = await handler(
    mockAPIGatewayProxyEvent({
      body: JSON.stringify({}),
      headers: { 'Content-Type': 'application/json', 'X-USER': 'Lilja' },
      method: 'GET',
      path: 'hello',
    }),
    {},
  );
  expect(result).toMatchInlineSnapshot(`
Object {
  "headers": Object {
    "Content-Type": "application/json",
  },
  "statusCode": 200,
}
`);
  expect(JSON.parse(body)).toMatchInlineSnapshot(`
```

Same for the others in this file.

packages/server/src/adapters/aws-lambda.ts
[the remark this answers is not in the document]
Lilja: I have spent some time actually getting both formats to be supported. I need some more time to test it.

I don't agree just v2 should be supported as v1 is not deprecated.

www/docs/server/api-gateway.md
```diff
@@ -0,0 +1,89 @@
...
+### 3. Use the Amazon API Gateway adapter
+
+tRPC includes an adapter for API Gateway out of the box. This adapter lets you run your routes through the API Gateway handler.
+
+```ts title='server.ts'
+import * as trpcAPIGW from '@trpc/server/adapters/api-gateway';
+
+const appRouter = /* ... */;
+
+const app = express();
```
mmkal, line 69:
```suggestion
const appRouter = /* ... */;
```

examples/lambda-api-gateway/src/client.ts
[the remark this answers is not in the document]
Lilja: Prettier dislikes that :(
```
/Users/lilja/code/trpc/examples/lambda-api-gateway/src/client.ts
  11:15  error  Parsing error: 'await' is only allowed within async functions  prettier/prettier
```

mmkal (commented):
Excited to see this getting picked up again! @KATT if you cut a canary release of this I could test it out in a real application.

KATT (changes requested):
Thanks so much for this Lilja, I haven't scrutinized the code yet (Sachin did that very well!)

I spy with my eye... zero tests and no docs -- 

- [ ] Address @sachinraja's comments
- [ ] Add tests ([see here](https://github.com/trpc/trpc/tree/main/packages/server/test/adapters) for inspo)
- [ ] Add docs to `/www` - (see #1511 for inspo)

examples/lambda-api-gateway/src/client.ts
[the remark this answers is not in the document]
sachinraja: Caused by https://github.com/trivago/prettier-plugin-sort-imports/issues/119

packages/server/src/adapters/aws-lambda.ts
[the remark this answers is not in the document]
mmkal: v1 should work fine with queryStringParameters as long as trpc doesn’t rely on multi-value query string params. @KATT is that a safe assumption to make? As far as I’m aware the query string params will just be `input` and `batch` and they’re both single value, is that right?

packages/server/src/adapters/aws-lambda.ts
sachinraja: None of the other adapters actually create an object for headers. The current way of mutating objects reads the closest to other adapters though.

KATT:
> Excited to see this getting picked up again! @KATT if you cut a canary release of this I could test it out in a real application.

```
Successfully published:
 - @trpc/client@9.23.5-lambda-support-2022-05-22-18-18-13.17+e0e1b299
 - @trpc/next@9.23.5-lambda-support-2022-05-22-18-18-13.17+e0e1b299
 - @trpc/react@9.23.5-lambda-support-2022-05-22-18-18-13.17+e0e1b299
 - @trpc/server@9.23.5-lambda-support-2022-05-22-18-18-13.17+e0e1b299
```

packages/server/src/adapters/aws-lambda.ts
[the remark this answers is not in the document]
Lilja: TS hates `new URLSearchParams(event.queryStringParameters)`

image. I'll try to work around it I guess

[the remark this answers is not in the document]
mmkal: If the only problem is `Type 'null' is not assignable to type '...'` would `new URLSearchParams(event.queryStringParameters || {})` work?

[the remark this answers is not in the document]
Lilja: image

Lilja:
@KATT can you add me to vercel? I cannot [view](https://vercel.com/trpc/next-prisma-starter/Duc8NT1ZvRh2tafZPuvZhFr4Cbk9).

KATT:
> @KATT can you add me to vercel? I cannot [view](https://vercel.com/trpc/next-prisma-starter/Duc8NT1ZvRh2tafZPuvZhFr4Cbk9).

Yeah, if there's a request button you can click it or you can give me your email

Lilja:
> > @KATT can you add me to vercel? I cannot [view](https://vercel.com/trpc/next-prisma-starter/Duc8NT1ZvRh2tafZPuvZhFr4Cbk9).
> 
> Yeah, if there's a request button you can click it or you can give me your email

I can't find such a button, you can add my email `erik@rkllj.io`

Lilja:
Looks like what is blocking vercel to build is something in [zod](https://github.com/colinhacks/zod/issues/1161)

jlalmes (dismissed):
Looks great @Lilja - thanks! I think there's a bug with path resolution, so I've suggested some changes. Let me know what you think. (cc @mmkal)

examples/lambda-api-gateway/README.md
```diff
@@ -0,0 +1,9 @@
+# tRPC + AWS Api Gateway Lambda
+
+For running locally, this project uses `@funnel-io/lalr` to run a mocked API gateway locally, which is invoked through `start-server`. You can read more about the tool [here](https://github.com/funnel-io/lalr).
```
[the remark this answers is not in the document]
Lilja: No offense taken! My last PR I used serverless, so I'll change it.

packages/server/src/adapters/lambda/index.ts
```diff
@@ -0,0 +1,134 @@
...
+    return resp as TResult;
+  } else if (isPayloadV2(event)) {
+    const resp: APIGatewayProxyStructuredResultV2 = {
+      statusCode: response.status,
+      body: response.body ?? undefined,
+      headers: transformHeaders(response.headers ?? {}),
+    };
+    return resp as TResult;
+  } else {
+    throw UNKNOWN_PAYLOAD_FORMAT_VERSION;
```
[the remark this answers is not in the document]
Lilja: I changed the core Error, but didn't threw a new `TRPCError` every time.

colinhacks:
Sorry, Zod broke this temporarily. Re-running the CI with zod@latest should fix it

packages/server/test/adapters/lambda.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+      body: JSON.stringify({}),
+      headers: { 'Content-Type': 'application/json', 'X-USER': 'Lilja' },
+      method: 'GET',
+      path: 'hello',
+      queryStringParameters: {},
+    }),
+    {},
+  );
+  const parsedBody = JSON.parse(body || '');
+  delete (result as any).body;
```
mmkal, line 73: Not needed now you're destructuring above
```suggestion
```

packages/server/test/adapters/lambda.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+      body: JSON.stringify({ who: [[]] }),
+      headers: { 'Content-Type': 'application/json' },
+      method: 'GET',
+      path: 'echo',
+      queryStringParameters: {},
+    }),
+    {},
+  );
+  const parsedBody = JSON.parse(body || '');
+  delete (result as any).body;
```
mmkal, line 106: Not needed now you're destructuring above
```suggestion
```

packages/server/src/adapters/lambda/index.ts
```diff
@@ -0,0 +1,134 @@
...
+  : T extends APIGatewayProxyEventV2
+  ? APIGatewayProxyStructuredResultV2
+  : never;
+export function lambdaRequestHandler<
+  TRouter extends AnyRouter,
+  TEvent extends APIGatewayEvent,
+  TResult extends APIGWReturn<TEvent>,
+>(
+  opts: AWSLambdaOptions<TRouter, TEvent>,
+): (event: TEvent, context: LambdaContext) => Promise<TResult> {
```
mmkal, line 106: Should there be some documented way for an end-user to say they'll only be using APIGatewayEventV2? What would that look like now? Would it just be a matter of providing the generic typeargs explicitly like 

```ts
export const handler = lambdaRequestHandler<
  typeof appRouter,
  APIGatewayEventV2,
  APIGatewayProxyStructuredResultV2
>({
  ...
})
```
?

mmkal (commented):
Couple of very minor comments and a question. LGTM!

packages/server/src/adapters/lambda/index.ts
Lilja: You raise a good point which I forgot about. Some thoughts:

* Internally, it doesn't matter if V1 or V2 is picked. It can deal with both.
* Ideally, what `TEvent` and `TResult` should be should be inferred from the context. I couldn't find how with my current typescript generic skills. I got half way by inferring the return type based on the second parameter. Could `inferRouterContext<TRouter>["event"]` somehow be used?

mmkal:
Yes will try to this evening!

packages/server/src/adapters/lambda/index.ts
```diff
@@ -0,0 +1,149 @@
...
+> = AWSLambdaOptions<TRouter, TEvent>['createContext'] extends NonNullable<
+  AWSLambdaOptions<TRouter, TEvent>['createContext']
+>
+  ? Parameters<AWSLambdaOptions<TRouter, TEvent>['createContext']>[0]['event']
+  : APIGatewayEvent;
+
+/** 1:1 mapping of v1 or v2 input events, deduces which is which.
+ * @internal
+ **/
+type DeduceAPIGWReturnType<T> = T extends APIGatewayProxyEvent
```
mmkal, line 110: Minor: I think a better name would be 
```suggestion
type APIGWResult<T> = T extends APIGatewayProxyEvent
```

(Deduce is fine just seemed unnecessary - main thing is that it’s not technically a function return type, but a result type, which is an Aws rather than typescript concept)

packages/server/src/adapters/lambda/utils.ts
```diff
@@ -0,0 +1,102 @@
...
+        unknownEvent.version,
+      )
+    ) {
+      return unknownEvent.version as APIGatewayPayloadFormatVersion;
+    } else {
+      return 'custom';
+    }
+  }
+}
+export enum DefinedAPIGatewayPayloadFormats {
```
mmkal, line 89: My personal opinion: enums are bad and should be avoided. You can achieve the equivalent type safety just with a `'1.0' | '2.0'` type. Not sure if trpc already uses enums. If it does, pls ignore me! If it doesn’t I’d personally vote against adding the first (although I wouldn’t be offended if ignored anyway)!!

packages/server/test/adapters/lambda.utils.ts
```diff
@@ -0,0 +1,128 @@
+import type {
+  APIGatewayProxyEvent,
+  APIGatewayProxyEventV2,
+  Context,
+} from 'aws-lambda';
+
+export const mockAPIGatewayProxyEvent = ({
```
mmkal, line 7: Nit: worth suffixing function name with V1 would make sense I think

mmkal (approved):
Left some comments but am approving because I don’t want to hold this up - comments are pretty minor and it basically looks good!

edit: I should note I haven’t tried the canary version since I don’t think it’s based on the latest changes, but I think pretty low risk to publish

packages/server/src/adapters/lambda/utils.ts
sachinraja: Agreed, we do not use them anywhere. String literal types are better

packages/server/src/adapters/lambda/index.ts
sachinraja: Maybe `inferAPIGWResult`? We use `infer` in other places like this.

packages/server/src/adapters/lambda/utils.ts
```diff
@@ -0,0 +1,102 @@
...
+    );
+
+export function isPayloadV1(
+  event: APIGatewayEvent,
+): event is APIGatewayProxyEvent {
+  return determinePayloadFormat(event) == '1.0';
+}
+export function isPayloadV2(
+  event: APIGatewayEvent,
+): event is APIGatewayProxyEvent {
```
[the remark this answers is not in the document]
Lilja: Great catch

Lilja:
Re-ran last Code action job, build running again!

packages/server/src/adapters/lambda/index.ts
```diff
@@ -0,0 +1,156 @@
...
+  if (isPayloadV1(event)) {
+    return event.httpMethod;
+  } else if (isPayloadV2(event)) {
+    return event.requestContext.http.method;
+  } else {
+    throw new TRPCError({
+      code: 'INTERNAL_SERVER_ERROR',
+      message: UNKNOWN_PAYLOAD_FORMAT_VERSION_ERROR_MESSAGE,
+    });
+  }
```
KATT, line 51: Some unnecessary if/else

```suggestion
  if (isPayloadV1(event)) {
    return event.httpMethod;
  }
  if (isPayloadV2(event)) {
    return event.requestContext.http.method;
  }
  throw new TRPCError({
    code: 'INTERNAL_SERVER_ERROR',
    message: UNKNOWN_PAYLOAD_FORMAT_VERSION_ERROR_MESSAGE,
  });
```

packages/server/src/adapters/lambda/index.ts
```diff
@@ -0,0 +1,156 @@
...
+  if (isPayloadV1(event)) {
+    return event.path.slice(1);
+  } else if (isPayloadV2(event)) {
+    return event.rawPath.slice(1);
+  } else {
+    throw new TRPCError({
+      code: 'INTERNAL_SERVER_ERROR',
+      message: UNKNOWN_PAYLOAD_FORMAT_VERSION_ERROR_MESSAGE,
+    });
+  }
```
KATT, line 63: Also has unnecessary `else`

packages/server/src/adapters/lambda/index.ts
```diff
@@ -0,0 +1,156 @@
...
+    };
+    return resp as TResult;
+  } else if (isPayloadV2(event)) {
+    const resp: APIGatewayProxyStructuredResultV2 = {
+      statusCode: response.status,
+      body: response.body ?? undefined,
+      headers: transformHeaders(response.headers ?? {}),
+    };
+    return resp as TResult;
+  } else {
```
KATT, line 94: Unnecessary `else`s

packages/server/src/adapters/lambda/index.ts
```diff
@@ -0,0 +1,156 @@
...
+export function lambdaRequestHandler<
+  TRouter extends AnyRouter,
+  TEvent extends inferAPIGWEvent<TRouter, TEvent>,
+  TResult extends inferAPIGWReturn<TEvent>,
+>(
+  opts: AWSLambdaOptions<TRouter, TEvent>,
+): (event: TEvent, context: APIGWContext) => Promise<TResult> {
+  return async (event, context) => {
+    const req = lambdaEventToHTTPRequest(event);
+    const path = getPath(event);
```
KATT, line 131: It looks like if these fail, we don't get a response sent to the client, right?

mmkal (approved):

packages/server/src/adapters/lambda/index.ts
Lilja: No, that's correct. But what types of failures would you expect?

mmkal: I’m not 100% sure but I believe Aws maps rejected promises to some kind of 5xx response

KATT:
@mmkal are you happy with this? :)

Lilja:
Just a friendly ping on this :)

KATT:
Oh sorry, it wasn't in my review requests so I forgot about it haha

KATT (approved):

KATT (approved):

KATT:
@all-contributors add @lilja for docs,code,tests

KATT:
@all-contributors add @lilja for docs, code, tests

Issue #1911, opened 2022-05-20
Next example fails Hydration, throws uncaught error


nilskj:
Hey, fantastic project! 
Did not find this bug reported. I get these errors when navigating or refreshing the browser at a /post/id route, running the NextJS example:
When navigating from the home page:
- Setup a new trpc project with the nextjs template.
- Click `view more` to navigate to a page route.
At this point no error appears... but then if either navigating to the post/id route (or hitting browser refresh) I get the following error: 
<img width="1792" alt="image" src="https://user-images.githubusercontent.com/495782/169589363-69e7c8c8-29af-4cd1-b4fe-9bcf1155aa44.png">

This error appears in [prod](https://nextjs.trpc.io/) as well with [minified nextjs error](https://reactjs.org/docs/error-decoder.html/?invariant=418)r! . so for example just browsing to an id like [this one](https://nextjs.trpc.io/post/ef4a7383-8ae1-4a44-85f6-eb42c222780a) will expose the error in console. 
- Might be something with React 18?

nilskj:
I think I found the error. Since no locale is set on the Date.toLocaleString options it may differ between formatting on client and server. I submitted PR #1912 with a fix!

KATT:
Fixed by #1912! Let us know if you get more like this1

PR #1912, opened 2022-05-20
Fix hydration issue with Date.prototype.toLocaleDateString

nilskj:
See #1911 we get some hydration errors with the NextJS starter example.
This fixes the hydration by setting a locale in options. Was probably different between server and client.


KATT (commented):
Thank you! Culd you please submit this as a fork of `main` instead of `next`? 🙏

nilskj:
Aha! I did actually fork main and then changed the target to next because I read that somewhere. Is changing target back to main enough or do I need to re-fork and redo? 😅

KATT (approved):

PR #1923, opened 2022-05-22
update `CONTRIBUTING.md`

KATT:
Did a quick pass on improving docs for contributors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,10 +27,18 @@ In another terminal, you can for instance navigate to `examples/next-prisma-star
 ```bash
 # in project root directory
 yarn test --watch
+
+# example if you want to test a specific test file:
+yarn test --watch --testPathPattern react
 ```
 
 Testing is currently coalesced in [./packages/server/test](./packages/server/test); we import the different libs from here, this makes it easier for us to do integration testing + getting test coverage on the whole codebase.
 
+### Linting
+
+```bash
+yarn lint-fix
+```
 
 ### Documentation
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,26 +1,53 @@
...
+This will start a watcher in parallel which builds all `packages/*` on any file change.
+
+**Terminal 2:**
+
 In another terminal, you can for instance navigate to `examples/next-prisma-starter` and run `yarn dev` & it will update whenever code is changed in the packages.
 
-## Testing
+### Testing
+
+> Note: you will want to have `yarn dev` running in parallel in another terminal
```
sachinraja, line 31: Is this needed? Don't the tests run the actual source TS files?

KATT: If they're setup with mock properly yes, unsure if all test files import the _packages thing

KATT: If they're setup with mock properly yes, unsure if all test files import the _packages thing

sachinraja: I think we just import the files directly instead of using the `@trpc/client`, `@trpc/react imports. [Example](https://github.com/trpc/trpc/blob/c4138897578949d910dd0bce9e1471a6ae73137e/packages/server/test/errors.test.ts#L9)

KATT: Yeah, but you can import something from the client that then uses the `@trpc/*` which then requires the mock to be setup first ☺

KATT: I'll merge it as is for now and we can revisit later!

sachinraja (approved):
Love how it shows the dual terminal workflow!

PR #1924, opened 2022-05-22
chore(lint): enforce camel-case filenames

mmkal:
Minor thing: I noticed a comment about filename casing in another PR. As this repo gets more contributors, thought it might help to reduce noise on pull requests by enforcing conventions at lint-time so people see early.

eslint-plugin-unicorn has loads of great rules along these lines but thought it better to introduce just this one rather than unleashing `unicorn/recommended` on a perfectly-happy repo.

I updated a few filenames that just seemed like oversights, but put in exceptions where it felt sensible (e.g. `whatever-tool-config.js` where `whatever-tool`'s convention overrides this repo's, and `issue-1234-blahBlah.test.ts` where (maybe?) `issue-1234` is a convention of its own. Happy to change more or fewer though based on feedback.


package.json
```diff
@@ -92,6 +92,7 @@
     "eslint-plugin-prettier": "^4.0.0",
     "eslint-plugin-react": "^7.25.1",
     "eslint-plugin-react-hooks": "^4.2.0",
+    "eslint-plugin-unicorn": "40.1.0",
```
mmkal, line 95: This isn't the latest version of eslint-plugin-unicorn, because versions 41 onwards depend on eslint 8+.

KATT (approved):

Issue #1939, opened 2022-05-25
Support for procedure method override


jlalmes:
_Originally posted by @KATT in https://github.com/trpc/trpc/discussions/1936#discussioncomment-2818345_

## Bonus, allow for method overriding

I made [this tweet](https://twitter.com/alexdotjs/status/1528854148011130880) yesterday and the main feedback that came back was to be able to sometimes have massive bodies when doing queries.

```
// interpret as query, even if it's a post
POST http://localhost?type=query
```

<sub>[T-70](https://linear.app/trpc/issue/T-70/support-for-procedure-method-override)</sub>

## Funding
* You can sponsor this specific effort via a [Polar.sh](https://polar.sh) pledge below
* We receive the pledge once the issue is completed & verified

<a href="https://polar.sh/trpc/trpc/issues/1939">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/1939/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/trpc/trpc/issues/1939/pledge.svg">
</picture>
</a>

PR #1940, opened 2022-05-25
Support for procedure method override

jlalmes:
Closes https://github.com/trpc/trpc/issues/1939.

## Support for procedure method override

By passing an optional `method?: 'GET' | 'POST'` property into `TRPCRequestOptions`, you can override the HTTP transport method used by queries/mutations. This is great for circumventing URL length limits when sending large input payloads to `query` procedures (there are use-cases for sending a `mutation` via GET too).

### Implementation
If `method` override is present the original `type` is added as a query parameter (i.e. `?type=query`). The `@trpc/server` will then resolve the `input` as the `POST` body but proceed to handle the request as a `query`. 

### Example
Query as POST: `client.query('getData', { bigPayload: [...] }, { method: 'POST' })`

### Additional
`httpLink` & `httpBatchLink` will now throw an error when sent a `subscription` operation.


packages/client/src/internals/httpRequest.ts
```diff
@@ -15,17 +14,17 @@ function arrayToDict(array: unknown[]) {
 export function httpRequest<TResponseShape = TRPCResponse>(
   props: {
     runtime: LinkRuntimeOptions;
-    type: ProcedureType;
+    type: 'query' | 'mutation';
+    method: 'query' | 'mutation';
```
mmkal, line 18: I think `'get' | 'post'` Are more intuitive here don’t you think?

jlalmes: Option A
```suggestion
    method: 'GET' | 'POST';
```
Option B
```suggestion
    typeOverride: 'query' | 'mutation';
```

Thoughts?

mmkal: I prefer A, it more directly tells the end user what the property will actually do. Aside from changing get to post it doesn’t somehow turn a query into a mutation. It’s purely an http thing right?

jlalmes:
Just updated description and added tests - ready for review!

KATT (commented):
This could be used for DoS-attacks, could we make it opt-out on the server the same wait we can opt-out of batching?

sachinraja (approved):
Looks good - I agree with Alex's comment and just left a few code quality comments. I think we should couple this with good docs to prevent unnecessary usage and hopefully the new transformer API can also help here. I still think it's useful for production apps despite opening up DoS attacks and some providers allow you to cache POST requests.

jlalmes:
* Normalized method in `TRPCClient.ts`.
* Deduped HTTP subscription error message.
* Added opt-out for server via `methodOverride.enabled: false`.
* Fixed tests & added test for opt-out.

packages/client/src/internals/executeChain.ts
```diff
@@ -62,6 +63,8 @@ export function executeChain<
       },
     });
   }
+
   walk({ index: 0, op: opts.op, stack: [] });
+
```
KATT, line 68:
```suggestion
```

KATT:
Closing because of the many conflicts it has, feel free to pick it up though

Issue #391, opened 2021-05-18
Context type of queries and mutations isn’t inferred when using React Query


kripod:
When [doing updates optimistically](https://react-query.tanstack.com/guides/optimistic-updates), having a `context` value to roll back failed mutations with is crucial. Unfortunately, the `TContext` type used within React Query isn’t inferred in tRPC’s React Query client, and so the value of `context` will always be `unknown`, e.g. in `onError` handlers.

Providing a fourth generic type parameter, `TContext`, to the underlying `useQuery` and `useMutation` hooks might help – please see the related code snippet for details: https://github.com/trpc/trpc/blob/39129c54eb9ede1542b4cd2093d85bf371cdfa8e/packages/react/src/createReactQueryHooks.tsx#L188-L223

KATT:
That's not a bug, that's a feature, PR very welcome :)

KATT:
Closing as it's stale and no one else has bumped it

PR #1944, opened 2022-05-28
feat(react): add useMutation context support

luixo:
Fixes https://github.com/trpc/trpc/issues/391

### DIFF
--- a/packages/react/src/createReactQueryHooks.tsx
+++ b/packages/react/src/createReactQueryHooks.tsx
@@ -62,8 +62,12 @@ export interface UseTRPCInfiniteQueryOptions<TPath, TInput, TOutput, TError>
     >,
     TRPCUseQueryBaseOptions {}
 
-export interface UseTRPCMutationOptions<TInput, TError, TOutput>
-  extends UseMutationOptions<TOutput, TError, TInput>,
+export interface UseTRPCMutationOptions<
+  TInput,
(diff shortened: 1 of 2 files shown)


packages/react/src/createReactQueryHooks.tsx
```diff
@@ -62,8 +62,12 @@ export interface UseTRPCInfiniteQueryOptions<TPath, TInput, TOutput, TError>
     >,
     TRPCUseQueryBaseOptions {}
 
-export interface UseTRPCMutationOptions<TInput, TError, TOutput>
-  extends UseMutationOptions<TOutput, TError, TInput>,
+export interface UseTRPCMutationOptions<
+  TInput,
+  TError,
+  TOutput,
+  TContext = unknown,
```
KATT, line 69: Let's restrict it to a `Record<string, unknown>`

```suggestion
  TContext extends Record<string, unknown> = {},
```

KATT (changes requested):

1. Close this PR


packages/react/src/createReactQueryHooks.tsx
luixo: I object to that.
Original type of `TContext` in `react-query` doesn't extend anything and I personally use primitive there sometimes, so I think there should be no restriction like that.

Issue #623, opened 2021-07-04
@trpc/server - Handling CORS and CORS preflight requests


lleyton:
When using the standalone server, it might be useful to handle CORS preflight requests. Currently the request just falls with a 405, Method Not Allowed. I'm proposing the addition of a `cors` field on the `CreateHttpHandlerOptions` object, which would contain an array of allowed origins, or a boolean of whether to allow all origins. Thoughts?

KATT:
You can use the http handler straight-off. 

### Example code

```ts
import * as trpc from '@trpc/server';
import { z } from 'zod';
import http from 'http';

type Context = {};

export const appRouter = trpc.router<Context>().query('hello', {
  input: z
    .object({
      name: z.string(),
    })
    .optional(),
  resolve: ({ input }) => {
    return {
      text: `hello ${input?.name ?? 'world'}`,
    };
  },
});

export type AppRouter = typeof appRouter;

// create handler
const handler = trpc.createHttpHandler({
  router: appRouter,
  createContext() {
    return {};
  },
});

const server = http.createServer((req, res) => {
  // do whatever you want here
  // Set CORS headers
  res.setHeader('Access-Control-Allow-Origin', '*');
  res.setHeader('Access-Control-Request-Method', '*');
  res.setHeader('Access-Control-Allow-Methods', 'OPTIONS, GET');
  res.setHeader('Access-Control-Allow-Headers', '*');
  if (req.method === 'OPTIONS') {
    res.writeHead(200);
    res.end();
    return;
  }
  handler(req, res);
});

server.listen(2022);
```

PR #1946, opened 2022-05-30
replace standalone example

Crazytieguy:
I had trouble understanding how tRPC worked, and this example was too complex for me. I created a simpler example that will be useful to a wider audience. I hope you like it :)

It's possible it should be its own example, I wasn't sure what you would prefer


jlalmes:
Thanks @Crazytieguy, while I agree that this `standalone-example` could be tidied up - I think this example could be a little _too_ simple. The [docs](https://trpc.io) should cover the all of basics for new devs to get started. 

Additionally I think we should be encouraging users to use `zod` input/output validators.

Crazytieguy:
Hi @jlalmes :)

The problem I had was that the docs don't have a working minimal example at all, and they recommend looking at this examples folder in order to get started (in fact it's at the top of the first page of the docs). As someone who didn't know yesterday what zod is and was struggling to understand which dependencies are mandatory I think it's really nice to have an example where the only dependency is tRPC (and node-fetch to simulate the browser environment). It's really nice to see how changing the validation function changes the inferred type of the api - it really helped me understand how tRPC works. But I'm open to adding in zod if you think it's worth it, I can see the advantage. Maybe adding another api with a different type signature might also be nice (maybe an api that takes a number, objects are a bit hard without zod)

Also - following a short exchange with @KATT I changed the PR such that the old example is preserved.

Crazytieguy:
Perhaps the code in this example should match the code in the quickstart. Or maybe the other way around. Or maybe there should be a new docs page for people who want to understand the basic premise of tRPC. Or maybe it should match the nice video on the main readme, so that people can reproduce the video easily

Crazytieguy:
Oh and I would have liked the client to run in the browser instead of node, but then I would need to enable CORS for the trpc standalone httpserver and I didn't find a way to do that. Is there a way?

jlalmes:
Great argument and some excellent ideas here - I'm all for adding a `minimal-example`! There are plans to greatly improve the docs with interactive examples etc. (coming soon™)

Standalone HTTP server CORS setup: https://github.com/trpc/trpc/issues/623#issuecomment-878639248

Crazytieguy:
I gave up on making the client run in the browser because it was too complex and doesn't align with any of the other examples. Also I switched to yarn to solve an issue with npm workspaces

Crazytieguy:
As far as I'm concerned the PR is ready :)

KATT (changes requested):
Thanks! Please do the below, otherwise the builds won't run:

1. Close this PR

Crazytieguy:
@KATT Done https://github.com/trpc/trpc/pull/1948

PR #1948, opened 2022-05-30
add minimal example

Crazytieguy:
See https://github.com/trpc/trpc/pull/1946


Crazytieguy:
@KATT PR is ready :)

KATT (commented):
@Crazytieguy I did a bunch of tweaks, whatcha think?

Crazytieguy:
@KATT I really like the changes, and I'm glad I got to know these cool npm scripts :).
I played around and added a couple more small changes

Crazytieguy:
I added the example to the table in the relevant docs page.

Crazytieguy:
@KATT PR is ready

examples/minimal/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "test-dev": "start-server-and-test 'ts-node server' 2022 'ts-node client'",
+    "test-start": "start-server-and-test 'node server' 2022 'node client'"
+  },
+  "workspaces": [
+    "client",
+    "server"
+  ],
+  "devDependencies": {
+    "@types/node": "^17.0.10",
+    "@types/node-fetch": "^2.5.11",
```
sachinraja, line 20: Shouldn't this only be a dependency of the `client` workspace?

examples/standalone-server/package.json
```diff
@@ -12,18 +12,14 @@
     "test-start": "start-server-and-test 'node dist/server' 2022 'node dist/client'"
   },
   "dependencies": {
-    "@trpc/client": "^9.25.2",
-    "@trpc/react": "^9.25.2",
-    "@trpc/server": "^9.25.2",
+    "@trpc/client": "^9.23.6",
+    "@trpc/server": "^9.23.6",
```
sachinraja, line 16: These should be on the latest version

examples/minimal/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "compilerOptions": {
+    "moduleResolution": "node",
+    "esModuleInterop": true,
+    "strict": true
+  },
+  "include": ["server", "client"],
+  "exclude": ["**/node_modules"]
```
sachinraja, line 8: `node_modules` are excluded by default but it's fine if you want to be explicit

examples/minimal/client/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "@examples/minimal-client",
```
sachinraja, line 2: Let's add `"private": true` to these too

examples/minimal/.gitignore
```diff
@@ -0,0 +1,2 @@
+index.js
```
sachinraja, line 1: This seems weird to me. We should just build to a `dist` folder so we can ignore that.

sachinraja (changes requested):

examples/minimal/.gitignore
Crazytieguy: I think it would get a bit messy because of the separate dependencies

examples/minimal/package.json
Crazytieguy: fixed

examples/standalone-server/package.json
Crazytieguy: fixed

examples/minimal/tsconfig.json
Crazytieguy: fixed

examples/minimal/client/package.json
Crazytieguy: fixed

examples/minimal/.gitignore
ixahmedxi: I agree with @sachinraja, built JS files shouldn't be alongside source code, better to have them in some sort of folder like `dist` or `build`.

Crazytieguy: I was wrong, it wasn't a mess 😅

Crazytieguy:
Thanks for letting me know :) sounds very reasonable

KATT:
I don't see the big value over https://github.com/trpc/trpc/tree/main/examples/standalone-server, so going to close this for now, feel free to add improvements on the standalone server on the `next` branch

Crazytieguy:
@KATT When I told a friend about trpc, he instinctively disliked it because it felt like magic to him and he didn't understand how it works. I couldn't explain it without making my own minimal example, because standalone-server has some unnecessary dependencies and the code is much longer than is necessary. Also - this could turn into a one page quickstart guide that can actually be run, unlike the current quickstart which doesn't have any client code.

Crazytieguy:
You can see https://github.com/trpc/trpc/pull/1946 for the original discussion, and I think we chatted a bit on discord

KATT:
Apologies that I didn't merge this. I am don't want to do stuff on the `main`-branch right now as they also mean I have to migrate them to v10, I'm more inclined stuff that is pointing to [`next`](https://github.com/trpc/trpc/tree/next)

Maybe we should do this? _(all on the `next` branch)_

- Rename standalone example to `minimal`? AFAICT, basically includes the same thing with different structure
- Restructure it so it's easier to grok?
- _Potentially_ extract the websocket bits into its own example

Crazytieguy:
If you truly think it's valuable I'm willing to migrate this example to v10 and open a PR on next, but I don't want to start refactoring standalone-server right now. If you ask me that one can be deleted.

KATT:
I think it is valuable! ☺

Issue #1779, opened 2022-04-13
Add support react-query data selectors


KATT:
I just realized we don't support react-query's [Data Selectors](https://react-query.tanstack.com/guides/migrating-to-react-query-3#query-data-selectors) yet

KaiSpencer:
Hi @KATT,
Appears to be working for me, is there some functionality i've overlooked that is missing?

nilskj:
I have investigated this a bit. It feels kind of similar to this issue: [Properly type useInfiniteQuery](https://github.com/tannerlinsley/react-query/pull/3625). As an example, getting the correct type when using `select` seems to work fine in react query:
```tsx
  const selectQuery = useQuery(
    ["posts"],
    async (): Promise<Array<Post>> => {
      const { data } = await axios.get(
        "https://jsonplaceholder.typicode.com/posts"
      );
      return data;
    },
    {
      select: (posts) => ({
        hello: "world"
      })
    }
  );
  console.log(selectQuery?.data?.hello);
  ```

But in our wrapped hook in trpc, I don't get the correct return type of the `select` function (if provided) in the same fashion.
<img width="1023" alt="image" src="https://user-images.githubusercontent.com/495782/170783080-019b7c36-3804-4292-a903-ca1c3c08140d.png">
```tsx
  const postQueryTitle = trpc.useQuery(['post.byId', { id }], {
    select: (post) => ({
      hello: 'one',
    }),
  });
  console.log(postQueryTitle?.data?.hello); //here data is inferred as post data, select gives type error.
```

Maybe we need to dive in how the type is inferred conditionally if using a `select` property? Is the `select` conditionally lost in the wrapping somehow?

bradennapier:
select works as expected for `useQuery` - with `useInfiniteQuery` the problem comes in because your `select` receives unmodified data every time a new page comes in and can return whatever it wants - but that makes it impossible to infer the data properly without getting type hints.

Personally I almost always prefer using the obj syntax for `react-query` as it makes infer much more straight forward based on how they work.

```ts
const result = useQuery({
    queryKey: ['posts'] as const,
    queryFn: useCallback(async (): Promise<Array<Post>> => {
      const { data } = await axios.get(
        'https://jsonplaceholder.typicode.com/posts'
      );
      return data;
    }, []),
    select: useCallback(
      (posts) => ({
        hello: 'world',
      }),
      []
    ),
  });

```

This will give us

```ts
useQuery<Post[], unknown, {
    hello: string;
[...]

PR #1943, opened 2022-05-27
feat(react): fix typing for useQuery to honor select fn result

bradennapier:
This will fix the issue specifically defined in the issue linked, but a thorough investigation to make sure typing is being done in a way that honors the patterns behind `react-query` would be a good idea.  At the moment the `infinite queries` are actually broken (I have a PR there to resolve it), so those are probably less important to investigate until that PR is merged.

closes #1779

PR #1950, opened 2022-05-31
Honor type when useQuery select fn is used

bradennapier:
This will fix the issue specifically defined in the issue linked, but a thorough investigation to make sure typing is being done in a way that honors the patterns behind react-query would be a good idea. At the moment the infinite queries are actually broken (I have a PR there to resolve it), so those are probably less important to investigate until that PR is merged.

closes https://github.com/trpc/trpc/issues/1779
original PR: #1943


KATT (approved):
Perfect

KATT:
@all-contributors add @bradennapier for code and test

PR #1966, opened 2022-06-10
Added "exports" to package.json

JuanM04:
Added ["exports"](https://nodejs.org/api/packages.html#conditional-exports) field in package.json


sachinraja:
Thanks for the PR! However, we have full ESM compatibility in the `next` branch (released as an alpha version) because it's technically breaking and this version isn't fully ESM compliant for a multitude of reasons.

PR #1972, opened 2022-06-12
Correct import in API Gateway documentation

skovhus:
The new documentation for the API Gateway adapter includes an invalid import. Note that 9.25.0 actually includes a folder `adapters/api-gateway` but it seems like some left over that should have been cleaned before 9.25.0 was released.

Before people start using the adapter it could be worth considering if it should be called `aws-lambda` or `aws-api-gateway` instead. Prefixing with `aws` makes it clear where this adapter can be used. Note that the documentation consistency calls this an Amazon API Gateway adapter, but the code mainly calls it `lambda`. 

FYI @Lilja

### DIFF
--- a/www/docs/server/api-gateway.md
+++ b/www/docs/server/api-gateway.md
@@ -66,7 +66,7 @@ export type AppRouter = typeof appRouter;
 tRPC includes an adapter for API Gateway out of the box. This adapter lets you run your routes through the API Gateway handler.
 
 ```ts title='server.ts'
-import * as trpcAPIGW from '@trpc/server/adapters/api-gateway';
+import { CreateLambdaContextOptions, lambdaRequestHandler } from '@trpc/server/adapters/lambda';
 
 const appRouter = /* ... */;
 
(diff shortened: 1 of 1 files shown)


sachinraja:
Ah, yes I think we should rename to `aws-lambda`/`awsLambdaRequestHandler`.

KATT:
> Ah, yes I think we should rename to `aws-lambda`/`awsLambdaRequestHandler`.

Oh... feel free to do a PR with both supported and update all docs, we can remove the `/lambda` one for next major

skovhus:
> Oh... feel free to do a PR with both supported and update all docs, we can remove the /lambda one for next major

As lambda was just related 2 hours ago, couldn't we see this as a bug fix instead of keeping both `lambda` and `aws-lambda` around?

skovhus:
The documentation is also a bit inconsistent and calls the adapter "API Gateway". In reality this is both an adapter for the API Gateway AND Lambda, but calling this an "AWS Lambda" adapter seems more standard than an "API Gateway".

PR #1974, opened 2022-06-12
Rename lambda adapter to aws-lambda

skovhus:
Follow up on https://github.com/trpc/trpc/pull/1972. 

Here I suggest:
1) rename the new `lambda` adapter to `aws-lambda` to make it clear where this adapter can be used and for consistency with other adapter (e.g. fastify and next)
2) use the same name in the documentation instead of "API Gateway"

Note that 1. is a breaking change, but I guess this could also be seen as a bug fix on the feature that was just released 2 hours ago... If we don't want to rename this to `aws-lambda` then I still suggest we update the documentation to at least be consistent around the name of the adapter. :) 

---

**Update from @KATT:** I updated this PR to be backwards-compatible which makes the diff a bit harder to read - all things was moved to `/aws-lambda` with backwards-compatible exports in `/lambda`


KATT:
I'm very OCD about backwards compatability, so I pushed some changes that I think ensure it is backward compat.

KATT:
@Lilja please review today if possible

Lilja (approved):
Naming is hard!

KATT (approved):
I agree this is better, waiting for @Lilja and/or @sachinraja and/or  @mmkal to sanity check

KATT:
@Lilja @skovhus feel free to keep retrying CI during the day and ping me once it passes 🙃

KATT:
@all-contributors add @skovhus for docs, code

PR #1784, opened 2022-04-14
rewrite subscriptions with observables

sachinraja:
closes #1596
part of #1597

new API:
```ts
.subscription('onAdd', {
  resolve() {
    return observable<Post>((emit) => {
      const onAdd = (data: Post) => emit.next(data);
      ee.on('add', onAdd);

      return () => {
        ee.off('add', onAdd);
      };
    });
  },
})
```

This *should* allow tRPC to be used in most (all?) JS runtimes.
- observable moved to server (`@trpc/client/observable` -> `@trpc/server/observable`)

I'm not sure how `observable` can replace `EventEmitter` in the examples so would appreciate some help with that.

PR #1999, opened 2022-06-15
Expose various socket and subscriptions lifecycle methods

ArnaudBarre:
My current need is to be able to a detect deployment of the server by using onOpen/onClose on the main WebSocket client.

I also think exposing subscription lifecycles would be useful too in the future. Than change onNext -> onData is a breaking change but I found that the naming is more consistent for the end user, `onNext` feeling like a leak of the implementation details of observables.

What do you think?


jlalmes:
Hi @ArnaudBarre.

Please be aware that Subscriptions have undergone a complete rewrite to observables in the `next` branch (cc. https://github.com/trpc/trpc/pull/1784)

You should already be able to detect `open`/`close` events. Take a look at the following example:

```ts
const wsClient = createWSClient({ url: 'ws://localhost:8081' });
const trpcWsClient = createTRPCClient<WsRouter>({
  links: [wsLink({ client: wsClient })],
});

// "open" event
wsClient.getConnection().addEventListener('open', () =>  {
  ...
})
// "close" event
wsClient.getConnection().addEventListener('close', () =>  {
  ...
})
```

ArnaudBarre:
Hi,

Looking at the PR it doesn't seems to conflict with this PR. I think that at the React level (useSubscription), dealing with `onStarted`, `onData` & `onStopped` is nicer than an `onNext`.

For the client level, I had this idea you suggested, but it only fires for the first connection, and not after reconnection.

jlalmes:
_Heads up: I just changed the base branch to `next`._

Out of interest, what is your use-case for needing to detect if the server has redeployed? There may be a better solution.

I don't mind the idea of changing `{ next: ({ type, data }) => {} }` to `{ started: () => {}, stopped: () => {}, data: (data) => {} }`, but bear in mind we should probably change `emit.next(...)` to `emit.data(...)` on the server-side too.

jlalmes:
So I propose that we keep the `observable` implementation as-is but we change the `client.subscription()` and `trpc.useSubscription()` interfaces as follows:

```ts
client.subscription(path, input, {
  onStarted: () => void;
  onStopped: () => void;
  onData: (data: TRPCResultMessage<TOutput>) => void;
  onError: (error: TRPCClientError<TRouter>) => void;
  onComplete: () => void;
})
``` 

```ts
trpc.useSubscription([path, input], {
  enabled: boolean;
  onStarted: () => void;
  onStopped: () => void;
  onData: (data: TRPCResultMessage<TOutput>) => void;
  onError: (error: TRPCClientError<TRouter>) => void;
  onComplete: () => void;
})
``` 

Bear in mind it appears that `onStopped` and `onComplete` are _basically_ the same so we may wish to omit one of them? Any thoughts on this @KATT @sachinraja?

KATT (changes requested):

1. Close this PR


packages/react/src/createReactQueryHooks.tsx
```diff
@@ -343,8 +344,11 @@ export function createReactQueryHooks<
     ],
     opts: {
       enabled?: boolean;
-      error?: (err: TError) => void;
-      next: (data: TOutput) => void;
```
KATT, line 347: Don't touch these, please

ArnaudBarre: So you're against the API change?

Should I just add start end stop callback so?

KATT: Add stuff, don't modify existing.

jlalmes: I think we are ok to remove these properties if following the design I propose here: https://github.com/trpc/trpc/pull/1999#issuecomment-1159030370. (👈 cc @sachinraja any thoughts?)

Regardless, it's probably best to let the new API evolve and settle before making these changes though.

PR #2000, opened 2022-06-16
feat(www): migrates docusaurs app to use typescript

ixahmedxi:
This PR brings the following changes:

- Adds typescript into the docusaurus website and configures it.
- Migrates the landing page from `index.js` to `index.tsx`
- Upgrades `react-github-btn` to contain children as a prop in the type defs (see: https://github.com/ntkme/react-github-btn/commit/aae7f3db6f30139e9c70d11ef0412e7f1ee5ee18)
- Types the docusaurus configuration file so that autocomplete works.
- Removes imageUrl from the `Feature` component as that none of the features use the prop and it leads to typescript errors.

### DIFF
--- a/www/tsconfig.json
+++ b/www/tsconfig.json
@@ -1,0 +1,3 @@
+{
+  "extends": "@tsconfig/docusaurus/tsconfig.json"
+}
--- a/www/docusaurus.config.js
+++ b/www/docusaurus.config.js
@@ -1,3 +1,6 @@
+// @ts-check
+
+/** @type {import('@docusaurus/types').Config} */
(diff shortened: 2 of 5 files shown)


sachinraja (approved):

KATT (approved):
🔥

www/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "@tsconfig/docusaurus/tsconfig.json"
```
jlalmes, line 2:
```jsonc
"compilerOptions": {
  "strict": true
}
```

PR #2024, opened 2022-06-20
add custom headers to `httpLinks` / `httpBatchLink`

### LINKED ISSUE #2018: Procedure specific custom request headers
### Discussed in https://github.com/trpc/trpc/discussions/2017

<sup>Originally posted by **skovhus** June 17, 2022</sup>
From the documentation and code example it seems like Client side HTTP headers can only be configured in the tRPC client and not call site when doing a query or mutation. 

But there are use cases where you want to pass in specific headers for a single procedure. What is the recommended way here? Is that something we lack support for? 

Example use case: Elevated privileges that would update the authorization header for a single procedure. </div>

KATT:
> **Note:** Not decided yet if I want to add this on v9. Might just be a v10-thing.

Closes #2018 

- [ ] Backwards compatability
- [ ] Deprecated old
- [ ] Code review
- [ ] Update docs (help plz!)

### DIFF
--- a/packages/client/src/internals/httpRequest.ts
+++ b/packages/client/src/internals/httpRequest.ts
@@ -1,6 +1,10 @@
 import { ProcedureType } from '@trpc/server';
 import { TRPCResponse } from '@trpc/server/rpc';
-import { LinkRuntimeOptions, PromiseAndCancel } from '../links/core';
+import {
+  HTTPHeaders,
+  LinkRuntimeOptions,
+  PromiseAndCancel,
+} from '../links/core';
 
 // https://github.com/trpc/trpc/pull/669
 function arrayToDict(array: unknown[]) {
@@ -14,10 +18,11 @@ function arrayToDict(array: unknown[]) {
 
 export function httpRequest<TResponseShape = TRPCResponse>(
   props: {
-    runtime: LinkRuntimeOptions;
+    runtime: Pick<LinkRuntimeOptions, 'transformer'>;
     type: ProcedureType;
     path: string;
     url: string;
+    headers: () => Promise<HTTPHeaders>;
   } & ({ inputs: unknown[] } | { input: unknown }),
 ): PromiseAndCancel<TResponseShape> {
   const { type, runtime: rt, path } = props;
--- a/packages/client/src/internals/httpRequest.ts
+++ b/packages/client/src/internals/httpRequest.ts
@@ -1,10 +1,6 @@
-import { ProcedureType } from '@trpc/server';
+import { DataTransformer, ProcedureType } from '@trpc/server';
 import { TRPCResponse } from '@trpc/server/rpc';
-import {
-  HTTPHeaders,
-  LinkRuntimeOptions,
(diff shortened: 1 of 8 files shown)


packages/client/src/internals/httpRequest.ts
```diff
@@ -14,24 +14,27 @@ function arrayToDict(array: unknown[]) {
...
 export function httpRequest<TResponseShape = TRPCResponse>(
   props: {
-    runtime: LinkRuntimeOptions;
+    transformer: DataTransformer;
     type: ProcedureType;
     path: string;
     url: string;
+    AbortController?: typeof AbortController;
+    fetch: TRPCFetch;
+    headers: () => HTTPHeaders | Promise<HTTPHeaders>;
```
sachinraja, line 23: nice

packages/client/src/links/httpBatchLink.ts
```diff
@@ -1,14 +1,20 @@
 import { AnyRouter, ProcedureType } from '@trpc/server';
 import { TRPCResponse } from '@trpc/server/rpc';
+import { getFetch } from '..';
```
sachinraja, line 3:
```suggestion
import { getFetch } from '../getFetch';
```
better for other runtimes (Deno) and code run without tree shaking

packages/client/src/links/httpLink.ts
```diff
@@ -1,9 +1,19 @@
 import { AnyRouter } from '@trpc/server';
+import { getFetch } from '..';
```
sachinraja, line 2:
```suggestion
import { getFetch } from '../getFetch';
```

sachinraja (approved):
Looks good! I would also change `getWindow` to just use [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) but that issue wasn't introduced in this PR.

sachinraja (changes requested):
meant to request changes, still needs docs as you noted

packages/client/src/links/core.ts
```diff
@@ -58,8 +65,17 @@ export type TRPCFetch = (
 
 export type LinkRuntimeOptions = Readonly<{
   transformer: DataTransformer;
+  /**
+   * @deprecated use `headers` on your `httpLink` / `httpBatchLink` instead
```
skovhus, line 69: For most users I believe it would be inconvenient to have to define both links in order to add headers. Wouldn’t us make sense to preserve this as base headers. 🤔

skovhus (commented):
Awesome to see this come to life!

packages/client/src/links/core.ts
KATT: It doesn't make sense to have both places, I don't wanna support 2 APIs that do the same thing. One of the non-articulated design principles of tRPC is that it's better to be closer to the metal than to offer inflexible shortcuts.

KATT:
Will close this and use it as a reference when adding it to v10

PR #2031, opened 2022-06-22
fix(types): export internal types for type definition files generation

jgoux:
As users of `@trpc/server`, when we create our own type definition files using `tsc`, the compiler needs to access specific internal types that are inferred but not exported.

I added these exports so the compilation works again.

Feel free to close this PR if you think it's not immediately relevant. 👍


KATT (commented):
Happy to merge this if you get your PR up-to-date @jgoux ☺

jgoux:
It should be good. I only tested on simple cases, other internal types might be exported later. I'll do other PRs if I encounter failures. 👌

sachinraja (approved):

KATT:
@jgoux  this is published as `10.0.0-alpha.38`

PR #2038, opened 2022-06-23
:memo: Adding custom header

makyfj:
Documentation update to customize headers in config based on this [#1408](https://github.com/trpc/trpc/discussions/1408) discussion.

### DIFF
--- a/www/docs/client/header.md
+++ b/www/docs/client/header.md
@@ -1,0 +1,48 @@
+---
+id: header
+title: Custom header
+sidebar_label: Create Custom Header
+slug: /header
+---
+
+The headers option can be customize in config when using `withTRPC` in nextjs or `createClient` in react.js.
+
+`headers` can be both an object or a function. If it's a function it will gets called dynamically every http request.
+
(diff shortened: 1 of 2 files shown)


www/docs/client/header.md
```diff
@@ -0,0 +1,48 @@
+---
+id: header
+title: Custom header
+sidebar_label: Create Custom Header
+slug: /header
+---
+
+The headers option can be customize in config when using `withTRPC` in nextjs or `createClient` in react.js.
+
```
ixahmedxi, line 9: In my opinion, this feels like a duplication of content, we already supply an example of setting headers here: https://trpc.io/docs/react#3-add-trpc-providers. I think what we need is documentation on the `onSuccess` and how it can be used to dynamically set headers on a mutation's success.

makyfj: Yes, I think that makes sense. Examples for NextJS and ReactJS would be better.

KATT:
Thanks so much for doing this, @makyfj!


@all-contributors add @makyfj for docs

PR #2050, opened 2022-06-28
Aws Lambda adapter support prefix path styles

Ankcorn:
I ran into a bug setting up the aws-lambda adapter with the serverless framework when I set the path to contain a prefix before the wildcard.
```json
{"id":null,"error":{"message":"No \"query\"-procedure on path \"api/trpc/getUser\"","code":-32004,"data":
```
I think this will fix that scenario


packages/server/src/adapters/aws-lambda/index.ts
```diff
@@ -52,10 +52,10 @@ function getHTTPMethod(event: APIGatewayEvent) {
 }
 function getPath(event: APIGatewayEvent) {
   if (isPayloadV1(event)) {
-    return event.path.slice(1);
+    return event.path.split('/')[event.path.split('/').length - 1];
```
mmkal, line 55: I’m not sure it’s safe to assume all users want to do this, I think it’s dependent on how api gateway is configured - which _might_ be in serverless framework, but it also might be serverless-components, or aws-cdk, or aws-sam, or cloudformation, or pulumi, or terraform, or point and click, or serverless-stack or…

mmkal: Also won’t this break for any users who use `/` in their procedure paths?

mmkal (changes requested):
@Lilja would be great to get your input here. I’m afraid I still haven’t tried this adapter out as we’re using aws-cdk, and have a couple of express middlewares we depend on so will take some learning and refactoring to integrate into my work project.

But from eyeballing the code I’m not sure this is ok to merge.

I’d be more comfortable with a configuration option of a prefix which should be stripped. The default could be `/`. In your case you could set it to `/api/trpc/`. But, ideally even that wouldn’t be necessary, I haven’t dug in enough to find the most elegant solution.

packages/server/src/adapters/aws-lambda/index.ts
Ankcorn: > Can’t api gateway be configured with something like api/foo/bar/trpc/{proxy+}

That's exactly what this handles :) With the current code api/foo/bar/trpc/xxxx is passed to trpc as the query

>also capturing everything after the last slash wouldn’t work for my team’s router for example, which has procedures like abc/def/doSomething (usually via .merge('abc/', defRouter) or similar). This change would change the path to doSomething and would throw out the abc/def/ wouldn’t it?

This would do exactly that, thanks! You are right I can't see another way of doing this with path or rawPath without breaking your router. I suspect there is another API gateway event property I should be using here

Lilja:
Let's see here. I haven't tried this adapter in production, only through the serverless framework.

I suppose you have you don't have `${proxy+}` wildcards in your api gateway deployment? But rather like a:

```Python
x = api.root.resource_for_path("/api/trpc/{proxy+}")
```

We have one lambda through API GW in production today, they are using a custom domain name. 
It looks like this:
`https://x.execute-api.us-east-1.amazonaws.com/global/` and is aliased as `https://example.com` with the base-path-maping of `empty` => `global`.

Curling `https://example.com/test` results in a `/test` as `event.path`(v1 API), curling the raw endpoint(execute-api URL) also gave me a `/test` as `event.path`.

During development of this adapter, this was my sole case to test against. I didn't realize that people might not be using `${proxy+}`, *only*.

-------------

I like what @mmkal is suggesting, a common prefix to "replace" the path-string with. By default we are removing the leading slash.
The problem and what we should look into is if we can figure out what prefix we need to strip "automagically". If we can get how the lambda who has called has been configured and remove the "routing-boilerplate" then we would be golden.

When I'm calling `/test`, I'm getting a `"${proxy+}"` as `event.resource`. I wonder if you @Ankcorn get `"/api/trpc/${proxy+}"` as `event.resource`?
I wonder if we can use this information somehow?

Ankcorn:
I have a solution for http-apis now

When the config is 
```yml
functions:
  trpc:
    handler: src/server.handler
    events:
      - httpApi: '*'
```
The input event looks like
```javascript
{
  version: '2.0',
  routeKey: '$default',
  rawPath: '/trpc/greet',
  rawQueryString: 'input={%22name%22:%22thomas%22}',
  headers: {
    accept: 'text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8',
    'accept-encoding': 'gzip, deflate, br',
    'accept-language': 'en-GB,en;q=0.5',
    'cache-control': 'no-cache',
    'content-length': '0',
    dnt: '1',
    host: 'lg0kwnkcw7.execute-api.us-east-1.amazonaws.com',
    pragma: 'no-cache',
    'sec-fetch-dest': 'document',
    'sec-fetch-mode': 'navigate',
    'sec-fetch-site': 'cross-site',
    'upgrade-insecure-requests': '1',
    'user-agent': 'Mozilla/5.0 (X11; Linux x86_64; rv:101.0) Gecko/20100101 Firefox/101.0',
    'x-amzn-trace-id': 'Root=1-62bb6ec6-1acb8bc150155d8308c5250d',
    'x-forwarded-for': '188.214.12.16',
    'x-forwarded-port': '443',
    'x-forwarded-proto': 'https'
  },
  queryStringParameters: { input: '{"name":"thomas"}' },
  requestContext: {
    accountId: '260235234243',
    apiId: 'lg0kwnkcw7',
    domainName: 'lg0kwnkcw7.execute-api.us-east-1.amazonaws.com',
    domainPrefix: 'lg0kwnkcw7',
    http: {
      method: 'GET',
      path: '/trpc/greet',
      protocol: 'HTTP/1.1',
      sourceIp: '188.214.12.16',
      userAgent: 'Mozilla/5.0 (X11; Linux x86_64; rv:101.0) Gecko/20100101 Firefox/101.0'
    },
    requestId: 'Uc4_BguhIAMEVJQ=',
    routeKey: '$default',
    stage: '$default',
    time: '28/Jun/2022:21:12:38 +0000',
    timeEpoch: 1656450758328
  },
  isBase64Encoded: false
}
```

When the function config is this

```yml
functions:
  trpc:
    handler: src/server.handler
    events:
      - httpApi:
          method: any
          path: /trpc/{param+}
```
The input event looks like
```javascript
{
  version: '2.0',
  routeKey: 'ANY /trpc/{param+}',
  rawPath: '/trpc/a/b',
  rawQueryString: '',
  headers: {
    accept: 'text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8',
    'accept-encoding': 'gzip, deflate, br',
    'accept-language': 'en-GB,en;q=0.5',
    'content-length': '0',
    host: 'it7w1lr0r1.execute-api.us-east-1.amazonaws.com',
    'sec-fetch-dest': 'document',
    'sec-fetch-mode': 'navigate',
    'sec-fetch-site': 'none',
    'sec-fetch-user': '?1',
    'upgrade-insecure-requests': '1',
    'user-agent': 'Mozilla/5.0 (X11; Linux x86_64; rv:101.0) Gecko/20100101 Firefox/101.0',
    'x-amzn-trace-id': 'Root=1-62bb7d45-0bb0ea047a82899433e5451a',
    'x-forwarded-for': '188.214.12.16',
    'x-forwarded-port': '443',
    'x-forwarded-proto': 'https'
  },
  requestContext: {
    accountId: '260235234243',
    apiId: 'it7w1lr0r1',
    domainName: 'it7w1lr0r1.execute-api.us-east-1.amazonaws.com',
    domainPrefix: 'it7w1lr0r1',
    http: {
      method: 'GET',
      path: '/trpc/a/b',
      protocol: 'HTTP/1.1',
      sourceIp: '188.214.12.16',
      userAgent: 'Mozilla/5.0 (X11; Linux x86_64; rv:101.0) Gecko/20100101 Firefox/101.0'
    },
    requestId: 'UdCC6jkboAMEPSQ=',
    routeKey: 'ANY /trpc/{param+}',
    stage: '$default',
    time: '28/Jun/2022:22:14:29 +0000',
    timeEpoch: 1656454469649
  },
  pathParameters: { param: 'a/b' },
  isBase64Encoded: false
}
```

That means the code can look something like (I still need to figure out a sensible way to grab the key name as I guess people can call it anything. Is their a valid use case for multiple paths?
```javascript
return event.routeKey === '$default' ? event.rawPath.splice(1) : event.pathParameters[keyName]
```

What are your thoughts on this approach?

Lilja:
It could work, I guess we need to adjust it for v1 since `routeKey` doesn't exist in v1.

Ankcorn:
So I actually ran into some issues with the rest api

https://apx6avil82.execute-api.us-east-1.amazonaws.com/no-prefix/greet?input={"name":"thomas"}

For some reason it always bugs out (no-prefix is the stage) with no error message

But the http api version works fine

https://lg0kwnkcw7.execute-api.us-east-1.amazonaws.com/greet?input={"name":"thomas"}

Lilja:
How is the rest-api configured?

Lilja:
Very relevant documentation from AWS: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-routes.html

adamsullovey:
I ran into this path issue while using the Serverless CDK (not Serverless framework) and worked around it like this:

```ts
export const handler: APIGatewayHandler = async (event, context) => {
  const trpcHandler = awsLambdaRequestHandler({
    /** the usual config stuff */
   })

  /**
   * Modify the Event's path before passing it to TRPC to convince it
   * it is at the root. This prevents errors about queries like
   * `api/trpc/create-signed-url` when I want to use the query `create-signed-url`.
   * TRPC does not strip out the extra parts of the path for me.
   * Will this cause issues later? Who knows!
   */
  const shortenedPath = '/' + event.rawPath.split('/')[3] // [3] is very specific to my use-case
  event.routeKey = `${event.requestContext.http.method} ${shortenedPath}`
  event.rawPath = shortenedPath
  event.requestContext.http.path = shortenedPath
  event.requestContext.routeKey = event.routeKey
  // console.log('modfied event for TRPC', event)

  return trpcHandler(event, context)
}

I considered adding it to the docs, but I'm happy to see some discussion here first.  Passing in a path to strip out sounds reasonable.

**Comment by Ankcorn** [comment]
I got my branch in a bit of a pickle, so nuked my fork and am starting again on this, I am convinced passing a path can be avoided for http apis. To make the change easier to test I would prefer to fix for http apis then look at rest apis later

**Comment by Lilja** [comment]
Here you go:

<details>

<summary>Click to expand</summary>

```
{
  body: null,
  headers: {
    'content-type': 'application/json',
    Accept: '*/*',
    'User-Agent': 'node-fetch/1.0 (+https://github.com/bitinn/node-fetch)',
    'Accept-Encoding': 'gzip,deflate',
    Connection: 'close',
    Host: 'localhost:4050'
  },
  httpMethod: 'GET',
  isBase64Encoded: false,
  multiValueHeaders: {
    'content-type': [ 'application/json' ],
    Accept: [ '*/*' ],
    'User-Agent': [ 'node-fetch/1.0 (+https://github.com/bitinn/node-fetch)' ],
    'Accept-Encoding': [ 'gzip,deflate' ],
    Connection: [ 'close' ],
    Host: [ 'localhost:4050' ]
  },
  multiValueQueryStringParameters: { batch: [ '1' ], input: [ '{"0":{"name":"Erik"}}' ] },
  path: '/leading/prefix//greet',
  pathParameters: { proxy: '/greet' },
  queryStringParameters: { batch: '1', input: '{"0":{"name":"Erik"}}' },
  requestContext: {
    accountId: 'offlineContext_accountId',
    apiId: 'offlineContext_apiId',
    authorizer: {
      claims: undefined,
      scopes: undefined,
      principalId: 'offlineContext_authorizer_principalId'
    },
    domainName: 'offlineContext_domainName',
    domainPrefix: 'offlineContext_domainPrefix',
    extendedRequestId: 'cl5htiwtg0000h2tw6mh1cc7d',
    httpMethod: 'GET',
    identity: {
      accessKey: null,
      accountId: 'offlineContext_accountId',
      apiKey: 'offlineContext_apiKey',
      apiKeyId: 'offlineContext_apiKeyId',
      caller: 'offlineContext_caller',
      cognitoAuthenticationProvider: 'offlineContext_cognitoAuthenticationProvider',
      cognitoAuthenticationType: 'offlineContext_cognitoAuthenticationType',
      cognitoIdentityId: 'offlineContext_cognitoIdentityId',
      cognitoIdentityPoolId: 'offlineContext_cognitoIdentityPoolId',
      principalOrgId: null,
      sourceIp: '127.0.0.1',
      user: 'offlineContext_user',
      userAgent: 'node-fetch/1.0 (+https://github.com/bitinn/node-fetch)',
      userArn: 'offlineContext_userArn'
    },
    operationName: undefined,
    path: '/leading/prefix//greet',
    protocol: 'HTTP/1.1',
    requestId: 'cl5htiwtg0001h2tw3l8w3l98',
    requestTime: '12/Jul/2022:08:54:33 +0200',
    requestTimeEpoch: 1657608873887,
    resourceId: 'offlineContext_resourceId',
    resourcePath: '/dev/leading/prefix/{proxy*}',
    stage: 'dev'
  },
  resource: '/leading/prefix/{proxy*}',
  stageVariables: null
}

```
</details>

```TypeScript
//server.ts
.query("greet", { input: z.object({name: z.string()}), resolve(req) { return req.inp.name })

// client.ts
client("http://localhost:4050/dev/leading/prefix/").query("greet", { name: "Erik" })
```

```
# serverless.yml
rest-api-leading-prefix:
    # ...
    events:
        - http
            path: /leading/prefix/{proxy+}
            method: any 

```

PR #2066, opened 2022-07-01
enforce meta

### LINKED ISSUE #2065: Enforce properties in procedures
Hello,

We've recently started to migrate a large project over to tRPC and we have decided on enforcing the `meta` property as a way to define per-procedure access control (roles and misc. settings).
This control is being handled in by a middleware, however, the "issue" is that `meta` is still an optional property and it would be nice if this could be caught at compile time, rather than having the middleware throw an error at runtime.

```ts
interface Meta {
  auth: {
    /* .... */
  };
}
export const appRouter = trpc
  .router<Context, Meta /* 👈 when this is defined, `meta` should be required in procedures   */>()
  // handles error formatting to client
  .formatError(errorHandler)
  // makes sure procedure meta information like
  // procedure authorization guards etc. is correctly parsed
  .middleware(metaMiddleware)
  // all routes
  .merge('auth/', authRouter)
  .merge('invitation/', invitationRouter)
  .merge('equipment/', equipmentRouter);

```

Perhaps it would also be possible to make `input` and `output` properties non-optional, maybe an opt-in flag/generic?
The main reasoning behind this is to provide consistent code and enforce endpoints to be fully written.

sachinraja:
@jlalmes @KATT appreciate your thoughts on this

This makes it so if `TMeta` is not optional, then it must be set in the procedure.
closes #2065 (for v10)

another idea is to just omit the resolve functions (query/mutation/subscription) when the meta isn't set or show the "errors" there instead


KATT (commented):
I really like this direction, sorry for not reviewing ahead of merging the dependent branch, feel free to turn this into a real PR if you want :)

sachinraja:
will pick this up later in another PR

PR #2082, opened 2022-07-03
add path matching for aws-lambda api gateway v2

Ankcorn:
Big thanks to @mmkal and @Lilja for guiding me on this in https://github.com/trpc/trpc/pull/2050

Currently trpc only works when configured as the $default route, this makes it harder to integrate trpc into an existing aws project without the workaround mentioned by @adamsullovey in this [comment](https://github.com/trpc/trpc/pull/2050#issuecomment-1172907254) 

I did not alter the api gateway v1 behaviour, I don't use it so I am not really motivated to work on it.

What the codechange does: 

First I find all the matches in the routeKey against this regex `\{(.*?)\}`. This looks for anything between { }.

If no match is found an empty iterable is returned so it skips to the old logic.

the routeKey can look like

* $default
* any /trpc/{path+}
* any /trpc/{abc}/{path+}
* any /trpc/{path}

for each match we are looking for the greedy route (the one with the + in)
If we find the greedy route that will be the path for the trpc query so we return it. If the path is undefined we return an empty string. This will make trpc throw a 404 and the error message saying no query procedure in path ''.

If there is no greedy route we use the existing logic because we are assuming the user wants to match against the whole path in their project.


packages/server/test/adapters/awsLambda.test.tsx
```diff
@@ -181,6 +189,96 @@ Object {
 `);
 });
 
+test('test v2 format with multiple / in query key', async () => {
```
Ankcorn, line 192: I have added a test case to make sure I don't break the multiple / in query key case with this change

packages/server/test/adapters/awsLambda.test.tsx
```diff
@@ -181,6 +189,96 @@ Object {
...
+  });
+  const { body, ...result } = await handler2(
+    mockAPIGatewayProxyEventV2({
+      body: JSON.stringify({}),
+      headers: { 'Content-Type': 'application/json', 'X-USER': 'Lilja' },
+      method: 'GET',
+      routeKey: 'ANY /trpc/{a}/{path+}',
+      path: 'trpc/abc/hello',
+      queryStringParameters: {},
+      pathParameters: { a: 'abc', path: 'hello' },
```
Ankcorn, line 257: I'm not totally happy with having to handcraft this aws implementation detail in the test but I can't think of a better alternative without resorting to deploying and running against real infrastructure.

KATT (changes requested):

1. Close this PR


Ankcorn:
Thanks @KATT

PR #2085, opened 2022-07-05
add utility to remove invalid headers for node 18 `fetch()`

### LINKED ISSUE #2084: Node 18 utility for filter out blocklisted request headers
**Update:** culprit is this: https://github.com/trpc/trpc/pull/2085#issuecomment-1175630821

---

After integrating the [WebSocket example](https://github.com/trpc/trpc/tree/main/examples/next-prisma-starter-websockets) into my own project I had noticed that server-side/SSR batched `fetch` calls fail. This behavior can also be observed with the official example itself (I had to remove some conflicting/incompatible deps and bump the rest to their latest versions). To illustrate, here are the (server-side) logs when rendering a page which queries the `healthz` key:

```
<< query #1 healthz {
  input: undefined,
  result: TRPCClientError: fetch failed
      at Function.from (.../node_modules/@trpc/client/dist/TRPCClientError-bfee2bf5.cjs.dev.js:69:14)
      at .../node_modules/@trpc/client/links/httpBatchLink/dist/trpc-client-links-httpBatchLink.cjs.dev.js:208:50 {
    originalError: TypeError: fetch failed
        at Object.processResponse (node:internal/deps/undici/undici:5555:34)
        at node:internal/deps/undici/undici:5877:42
        at node:internal/process/task_queues:140:7
        at AsyncResource.runInAsyncScope (node:async_hooks:203:9)
        at AsyncResource.runMicrotask (node:internal/process/task_queues:137:8) {
      cause: [InvalidArgumentError]
    },
    shape: undefined,
    data: undefined,
    isDone: false,
    [cause]: TypeError: fetch failed
        at Object.processResponse (node:internal/deps/undici/undici:5555:34)
        at node:internal/deps/undici/undici:5877:42
        at node:internal/process/task_queues:140:7
        at AsyncResource.runInAsyncScope (node:async_hooks:203:9)
        at AsyncResource.runMicrotask (node:internal/process/task_queues:137:8) {
      cause: [InvalidArgumentError]
    }
  },
  elapsedMs: 12,
  context: {}
}
TRPCClientError: fetch failed
    at Function.from (.../node_modules/@trpc/client/dist/TRPCClientError-bfee2bf5.cjs.dev.js:69:14)
    at .../node_modules/@trpc/client/links/httpBatchLink/dist/trpc-client-links-httpBatchLink.cjs.dev.js:208:50 {
  originalError: TypeError: fetch failed
      at Object.processResponse (node:internal/deps/undici/undici:5555:34)
      at node:internal/deps/undici/undici:5877:42
      at node:internal/process/task_queues:140:7
      at AsyncResource.runInAsyncScope (node:async_hooks:203:9)
      at AsyncResource.runMicrotask (node:internal/process/task_queues:137:8) {
    cause: InvalidArgumentError: invalid connection header
        at processHeader (node:internal/deps/undici/undici:2082:15)
        at new Request (node:internal/deps/undici/undici:1965:13)
        at [dispatch] (node:internal/deps/undici/undici:2891:25)
        at Client.dispatch (node:internal/deps/undici/undici:408:33)
        at [dispatch] (node:internal/deps/undici/undici:627:32)
        at Pool.dispatch (node:internal/deps/undici/undici:408:33)
        at [dispatch] (node:internal/deps/undici/undici:4174:27)
        at Agent.dispatch (node:internal/deps/undici/undici:408:33)
        at node:internal/deps/undici/undici:6255:83
        at new Promise (<anonymous>) {
      code: 'UND_ERR_INVALID_ARG'
    }
  },
  shape: undefined,
  data: undefined,
  isDone: false,
  [cause]: TypeError: fetch failed
      at Object.processResponse (node:internal/deps/undici/undici:5555:34)
      at node:internal/deps/undici/undici:5877:42
      at node:internal/process/task_queues:140:7
      at AsyncResource.runInAsyncScope (node:async_hooks:203:9)
      at AsyncResource.runMicrotask (node:internal/process/task_queues:137:8) {
    cause: InvalidArgumentError: invalid connection header
        at processHeader (node:internal/deps/undici/undici:2082:15)
        at new Request (node:internal/deps/undici/undici:1965:13)
        at [dispatch] (node:internal/deps/undici/undici:2891:25)
        at Client.dispatch (node:internal/deps/undici/undici:408:33)
        at [dispatch] (node:internal/deps/undici/undici:627:32)
        at Pool.dispatch (node:internal/deps/undici/undici:408:33)
        at [dispatch] (node:internal/deps/undici/undici:4174:27)
        at Agent.dispatch (node:internal/deps/undici/undici:408:33)
        at node:internal/deps/undici/undici:6255:83
        at new Promise (<anonymous>) {
      code: 'UND_ERR_INVALID_ARG'
    }
  }
}
```

Judging from the error message, trpc is attempting to set the `connection` header via Node's built-in `fetch` (since v18, internally undici), which [is reserved](https://github.com/nodejs/undici/issues/1470) (`invalid connection header`). There are suggestions to [allow user-provided connection headers](https://github.com/nodejs/undici/issues/1307), but it looks like that change is going to take some time. I assume it is being set to `keep-alive` for `node-fetch` compatibility - in any case, it would be very forthcoming if a workaround could be implemented, as it looks like I'm not able to use trpc at the moment without downgrading node (yay).

KATT:
Closes #2084


.github/workflows/examples-node18.yml
```diff
@@ -0,0 +1,53 @@
...
+      PLAYWRIGHT_BROWSERS_PATH: ${{ github.workspace }}/playwright
+      DATABASE_URL: postgresql://postgres:@postgres:5432/trpcdb
+      NEXTAUTH_SECRET: supersecret
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
+    continue-on-error: true
+    strategy:
+      matrix:
+        dir: [standalone-server-node18]
+        node-start: ['18.x']
```
KATT, line 18: This makes sure it starts with node 18

examples/standalone-server-node18/src/client.ts
```diff
@@ -0,0 +1,69 @@
...
+import { httpLink } from '@trpc/client/links/httpLink';
+import { splitLink } from '@trpc/client/links/splitLink';
+import { createWSClient, wsLink } from '@trpc/client/links/wsLink';
+import ws from 'ws';
+import type { AppRouter } from './server';
+
+// polyfill WebSocket
+const globalAny = global as any;
+globalAny.WebSocket = ws;
+// 🎉 `fetch` doesn't need polyfill b/c this is node 18
```
KATT, line 12: Notice the absence of `fetch` polyfill

KATT (commented):
cc @3x071c

examples/standalone-server-node18/sandbox.config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "view": "console",
+  "container": {
+    "node": "14"
```
KATT, line 4:
```suggestion
    "node": "18"
```

examples/lambda-api-gateway/src/payloadFormatVersionClient.ts
```diff
@@ -2,7 +2,9 @@ import { createTRPCClient } from '@trpc/client';
 import fetch from 'node-fetch';
 import type { AppRouter } from './server';
 
-global.fetch = fetch as any;
+const globalAny = globalThis as any; // 🚨 unsafe, don't do this in production
```
KATT, line 5: How come? isn't that how polyfills are added usually?

examples/lambda-api-gateway/src/payloadFormatVersionClient.ts
```diff
@@ -2,7 +2,9 @@ import { createTRPCClient } from '@trpc/client';
...
+if (!globalAny.AbortController) globalAny.AbortController = AbortController;
+if (!globalAny.fetch) globalAny.fetch = fetch; // 🙌 Node v18 has built-in fetch
```
KATT, line 7: I dig this, thanks for adding

KATT (commented):
@3x071c 

- Please revert the CI-jobs setup to similar setup to before
  - We only want to test a subset of examples with ALL node versions b/c we don't wanna wait for for hours.
  - We have a [`next`](https://github.com/trpc/trpc/tree/next) with a lot of pending changes that will cause collissions with this stuff
  - The names are sensitive in the branch protection settings and I don't wanna update that
  - I do like the retry thing you added

3x071c:
@KATT We can of course revert, but right now testing is split into three separate workflow files all essentially doing the same thing - IMO it's a lot cleaner to do that in a single file with a matrix. Adding more tests won't increase the overall testing time as they are run in parallel. Regarding the `next` branch: I would recommend revising the testing pipeline in that branch too (side note: this would be easier if there [weren't multiple release branches](https://trunkbaseddevelopment.com)). The branch protection settings can be changed and shouldn't be a blocker if executed correctly.
Sorry if I'm involving myself too much with your design decisions, that's what I obsess about in my free time 😅

KATT:
> @KATT We can of course revert, but right now testing is split into three separate workflow files all essentially doing the same thing - IMO it's a lot cleaner to do that in a single file with a matrix. Adding more tests won't increase the overall testing time as they are run in parallel. 

They do increase the time. GitHub has a max concurrency limit (which I forgot)

> Regarding the `next` branch: I would recommend revising the testing pipeline in that branch too (side note: this would be easier if there [weren't multiple release branches](https://trunkbaseddevelopment.com)). The branch protection settings can be changed and shouldn't be a blocker if executed correctly. Sorry if I'm involving myself too much with your design decisions, that's what I obsess about in my free time 😅

Agreed your approach is cleaner in general! I would **love** for you to be super opinionated and do a PR for the `next`-branch instead with all of your opinions. I'm just fed up dealing with the merge conflicts and the next major is so close now that I can almost smell it

KATT:
Also, feel free to join our [Discord](https://trpc.io/discord) and `@` me there :) would love to have more of your ideas onboard

3x071c:
@KATT That's it, the node v18 test is now correctly reproducing the problem with the `next-prisma-starter-websockets` example. I will fix up the project structure in a separate PR when I find the time, available to chat via discord too. Cheers 😁

3x071c:
I have injected some dirty JS into the appropriate `node_modules` files to log what's happening. Here's an example fetch (wrapped in a `Promise`) that gets resolved from https://github.com/trpc/trpc/blob/7739c934130bbb2cea2586523b1af552f560a749/packages/client/src/internals/httpRequest.ts :

```
============================================================ ✅ PREPARING FETCH ============================================================
return rt.fetch(http://localhost:3000/api/trpc/health?batch=1&input=%7B%220%22%3A%7B%22json%22%3Anull%2C%22meta%22%3A%7B%22values%22%3A%5B%22undefined%22%5D%7D%7D%7D, {
    method: GET,
    signal: [object AbortSignal],
    body: undefined,
    headers: {"content-type":"application/json","host":"localhost:3000","connection":"keep-alive","sec-ch-ua":"\"Chromium\";v=\"103\", \".Not/A)Brand\";v=\"99\"","sec-ch-ua-mobile":"?0","sec-ch-ua-platform":"\"macOS\"","dnt":"1","upgrade-insecure-requests":"1","user-agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.5060.53 Safari/537.36","accept":"text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9","sec-fetch-site":"none","sec-fetch-mode":"navigate","sec-fetch-user":"?1","sec-fetch-dest":"document","accept-encoding":"gzip, deflate, br","accept-language":"de,en-US;q=0.9,en;q=0.8","x-ssr":"1"}
});
============================================================ ☑️ PREPARED FETCH ============================================================
```

After some digging, I found the source of the `rawHeaders` from the `runtime` parameter as well: https://github.com/trpc/trpc/blob/7739c934130bbb2cea2586523b1af552f560a749/examples/next-prisma-starter-websockets/src/pages/_app.tsx#L75-L84

So this isn't strictly an internal problem after all 😅 Should the reserved headers (`"connection": "keep-alive"` etc) be filtered out here or in the `httpRequest` internals? @KATT

KATT:
Interesting, will respond on the other issue, easier to keep a conversation away from the code. Nice you found the culprit!

3x071c:
@KATT I have included a preliminary `stripHeaders` implementation. Judging from the built type declarations, the function can be accessed with a simple `@trpc/client` import (would be easier to evaluate if the examples would properly reference monorepo dependencies directly via paths ;) )

packages/client/src/stripHeaders.ts
```diff
@@ -0,0 +1,41 @@
...
+export const stripHeaders = async (
+  headers: HTTPHeaders | (() => HTTPHeaders | Promise<HTTPHeaders>),
+): Promise<HTTPHeaders> =>
+  Object.fromEntries(
+    // wrap tuple array back into an object
+    Object.entries(
+      // unwrap object into tuple array so we can iterate/map over it
+      typeof headers === 'function' ? await headers() : headers, // resolve headers (call function if necessary)
+    ).filter(([k]) => !forbiddenHeaderNames.includes(k)), // filter out all keys `k` which are included in `forbiddenHeaderNames`
+  );
```
KATT, line 41: Please add a `stripHeaders.test.ts` :)

3x071c:
@KATT Had to disable an ESLint rule in the `interference.test.ts` file (no involvement):
https://github.com/trpc/trpc/blob/f19ede4a67721d882602d75471c12ff506a3857c/packages/server/test/inference.test.ts#L1
The react testing file has a bunch of problems, this was the least I could do to get it to run:
https://github.com/trpc/trpc/blob/f19ede4a67721d882602d75471c12ff506a3857c/packages/server/test/react.test.tsx#L676

examples/lambda-api-gateway/src/payloadFormatVersionClient.ts
3x071c: Thanks 😸

examples/next-prisma-starter-websockets/src/pages/_app.tsx
```diff
@@ -72,16 +73,17 @@ export default withTRPC<AppRouter>({
...
+        stripHeaders(() => {
+          if (ctx?.req) {
+            // on ssr, forward client's headers to the server
+            return {
+              ...ctx.req.headers,
+              'x-ssr': '1',
+            };
+          }
+          return {};
+        }),
```
KATT, line 86: Can we skip the callback here? Makes it quite hard to read

```suggestion
      headers: () => {
        if (ctx?.req) {
          // on ssr, forward client's headers to the server
          return {
            ...stripHeaders(ctx.req.headers),
            'x-ssr': '1',
          };
        }
        return {};
      },
```

packages/client/src/stripHeaders.ts
```diff
@@ -0,0 +1,41 @@
...
+  'via',
+];
+
+/**
+ * Strips an HTTP header object of internally reserved keys (for undici/Node.js built-in fetch since v18)
+ * @param headers Your HTTP header function/object
+ * @returns Stripped headers without blacklisted keys
+ */
+export const stripHeaders = async (
+  headers: HTTPHeaders | (() => HTTPHeaders | Promise<HTTPHeaders>),
```
KATT, line 33: Simplify

````suggestion
  headers: HTTPHeaders,
```

examples/next-prisma-starter-websockets/src/pages/_app.tsx
3x071c: @KATT More aesthetically pleasing, but less safe. This introduces the potential for runtime bugs when the developer forgets to wrap all returns with `stripHeaders` (in the same line of thinking: what if `x-ssr` is a reserved header? This would cause a server crash here), versus just having to wrap the entire `headers` function once.

packages/client/src/stripHeaders.ts
3x071c: @KATT Related to the convo above. Should we keep this function as flexible as possible regardless of individual developer preference?

KATT: No. Keep it minimal and stupid. If people want something else they can roll their own wrapper

examples/next-prisma-starter-websockets/src/pages/_app.tsx
KATT: They can wrap the whole `return stripHeaders({})` as well, but if you _actually_ explicitly use one of the blocked headers it's better to have the server crash than to automagically strip it and lead people down a several hours down rabbit hole

3x071c:
@KATT Resolved

examples/next-prisma-starter-websockets/src/server/wssDevServer.ts
```diff
@@ -1,11 +1,10 @@
 import { createContext } from './context';
 import { appRouter } from './routers/_app';
 import { applyWSSHandler } from '@trpc/server/adapters/ws';
-import fetch from 'node-fetch';
 import ws from 'ws';
 
 if (!global.fetch) {
-  (global as any).fetch = fetch;
+  (global as any).fetch = require('node-fetch');
```
sachinraja, line 7: This doesn't make sense. You can't mix `import` and `require` (following the spec, but I guess webpack/Next allows it). I suggest keeping the `node-fetch` import and adding a comment saying to remove it instead.

packages/client/src/stripHeaders.ts
```diff
@@ -0,0 +1,36 @@
+import { HTTPHeaders } from './links/core';
+
+// from https://github.com/nodejs/undici/blob/9e5316c8b04a7b35522d0d5b8de71f67fa2a3c34/lib/fetch/constants.js
+const forbiddenHeaderNames = [
```
sachinraja, line 4: nice

packages/client/src/stripHeaders.ts
```diff
@@ -0,0 +1,36 @@
...
+  'upgrade',
+  'via',
+];
+
+/**
+ * Strips an HTTP header object of internally reserved keys (for undici/Node.js built-in fetch since v18)
+ * @param headers Your HTTP header object
+ * @returns Stripped headers without blacklisted keys
+ */
+export const stripHeaders = (headers: HTTPHeaders): HTTPHeaders =>
```
sachinraja, line 32: I don't like this name. Maybe `stripInvalidHeaders`?

sachinraja (changes requested):

examples/next-prisma-starter-websockets/src/server/wssDevServer.ts
3x071c: > Webpack allows it

That's the point. Right now, the example runs on all Node versions, while importing `node-fetch` fails on v18. The alternative is to use a dynamic `import()`, but without `topLevelAwait` (via webpack, once again :P) the whole server would need to be wrapped in an async IIFE of sorts.

packages/client/src/stripHeaders.ts
sachinraja: I wouldn't go with `undici` because they're just following the Fetch API spec. [MDN calls them "forbidden"](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name) so I'm leaning towards `stripForbiddenHeaders`

KATT:
@all-contributors add @3x071c for code, test

packages/client/src/stripHeaders.ts
KATT: > Forbidden header names start with Proxy- or Sec-, or are one of the following names:

Cool, let's use that list instead of unidici then as well and make sure we strip headers starting with `proxy-` or `sec-` too?

examples/next-prisma-starter-websockets/src/pages/about.tsx
```diff
@@ -9,9 +9,12 @@ export default function AboutPage() {
       setNumber(n);
     },
   });
+  const healthz = trpc.useQuery(['healthz']);
+
   return (
     <div>
       Here&apos;s a random number from a sub: {num} <br />
+      Here&apos;s a query result ({healthz.status}): {healthz.data} <br />
```
KATT, line 17:
```suggestion
```

examples/next-prisma-starter-websockets/src/pages/about.tsx
```diff
@@ -9,9 +9,12 @@ export default function AboutPage() {
       setNumber(n);
     },
   });
+  const healthz = trpc.useQuery(['healthz']);
+
```
KATT, line 13:
```suggestion
```

packages/server/test/react.test.tsx
```diff
@@ -673,7 +673,7 @@ describe('useMutation()', () => {
       trpc.useMutation(['deletePosts'], {
         onMutate: () => 'foo' as const,
         onSuccess: (_data, _variables, context) => {
-          expectTypeOf(context).toMatchTypeOf<'foo'>();
+          expectTypeOf(context).toMatchTypeOf<'foo' | undefined>();
```
KATT, line 676: Revert? Is this related?

packages/client/src/stripHeaders.ts
3x071c: @KATT Good point, I'd stick with the array from undici (unless it's missing something) and add a check for the forbidden string prefixes

packages/server/test/react.test.tsx
3x071c: > Tests appear to be failing on unchanged files? (https://github.com/trpc/trpc/pull/2085#issuecomment-1176059766)

[trpc/trpc@`f19ede4` (#2085)](https://github.com/trpc/trpc/pull/2085/commits/f19ede4a67721d882602d75471c12ff506a3857c)

KATT:
> There's something odd happening here, Undici doesn't have any constraints for forbidden headers in versions >=5.4.0 (see [docs](https://github.com/nodejs/undici#forbidden-and-safelisted-header-names)). And it never errored on them, it only filtered them out. I think this is a specific error happening with the `connection` header that needs more investigation.

Interesting!

3x071c:
@sachinraja



> Undici doesn't have any constraints for forbidden headers in versions >=5.4.0 (see [docs](https://github.com/nodejs/undici#forbidden-and-safelisted-header-names)) [...] I think this is a specific error happening with the `connection` header that needs more investigation.

The source code on main says otherwise: https://github.com/nodejs/undici/blob/c1a0490f20760ce700b28ddcddd493758d3f54a1/lib/core/request.js#L279-L304 I wouldn't rely too heavily on conflicting documentation, considering the numerous open issues on the repo that have been pointing out errors when setting forbidden headers and suggestions to update the docs.
It looks like not all forbidden headers actually lead to an error (for example, I was able to `await fetch("https://example.com", { headers: { "dnt": "1" } })` from the node 18 console without a problem), but it's good to be on the safe side (as backed by MDN) and filter out all blacklisted headers.

sachinraja:
I recommend we just strip out the five headers that Undici throws an error on and rename this to `omitUndiciHeaders`.

sachinraja:
> As long as undici declares the array of headers "forbidden" (literally!), the utility should filter them out.

That code doesn't exist anymore. The only checks currently are what you linked above: https://github.com/nodejs/undici/blob/c1a0490f20760ce700b28ddcddd493758d3f54a1/lib/core/request.js#L279-L304.

I'm not going to block this though.

KATT:
cc @sachinraja @3x071c 

It seems like the *only* thing we need to exclude is the `connection` header. 

See here for a fix in a example project: https://github.com/KATT/unidici-fetch-reserved-header-crap/pull/1

trevoro:
We've ran into this issue at work and bumping node to v18.15 seems to have addressed the issue. 
I suspect it's related to this changelog item or some combination of other commits made since v18.12 at least.
In either case, we no longer get this _error_ in our application. That being said, it's not clear what the behaviour is. If you're counting on being able to set a header that is being stripped, it would be nice to get a warning but (IMO) that's not really up to tRPC.

Issue #2108, opened 2022-07-07
Simplify rollup config and submodules


KATT:
Currently, when we define a new export in the packages we add it in several places:

**Example** for `@trpc/server/observables`:

- https://github.com/trpc/trpc/blob/8c5499cb42697db3c0b580e92aa176cfef5b2475/packages/server/rollup.config.js#L14
- https://github.com/trpc/trpc/blob/8c5499cb42697db3c0b580e92aa176cfef5b2475/packages/server/package.json#L82-L86
- https://github.com/trpc/trpc/blob/8c5499cb42697db3c0b580e92aa176cfef5b2475/packages/server/package.json#L98
- https://github.com/trpc/trpc/blob/8c5499cb42697db3c0b580e92aa176cfef5b2475/packages/server/observable/index.js#L1
- https://github.com/trpc/trpc/blob/8c5499cb42697db3c0b580e92aa176cfef5b2475/packages/server/test/___packages.ts#L4-L5

This is very fiddly and easy to get wrong. It would be nice if we could automate most of the parts of this.

I did a first attempt of this with trying to automate the rollup stuff in https://github.com/trpc/trpc/blob/next/scripts/rollup.js, but it's a terrible hacky script & currently only works for `@trpc/client` and doesn't cover edge-cases enough to work for `/react`, `/next`, `/server`.

If someone likes build-tools and has some good ideas, feel free to submit a PR to the `next` branch with automating this as much as possible. I do still miss preconstruct which we have on `main` which automatically did this for us.

PR #2109, opened 2022-07-07
add shared proxy helper

### LINKED ISSUE #2111: Fix abstractions for proxies and align across React + raw
Currently, there's an inconsistency between `@trpc/react` and `@trpc/client`

I think they should be a wrapper around the client and be i.e:

```ts
// raw:
const client = createTRPCClient<AppRouter>({ /* ... */ })
const rawProxy = createClientProxy(client)

// react
const trpcReact = createReactHooks<AppRouter>({ /* ... */} )
const reactProxy = createReactProxy(trpcReact)
```

KATT:
- Create `createProxy` helper in `@trpc/server/shared`
- Use said helper in `@trpc/react` + `@trpc/client` rather than rolling bespoke one
- Refactor `createReactQueryProxy` to take the result of `createReactQueryHooks` as an argument (funky typescript shit to infer that stuff)
- Refactor `createTRPCClient` to _not_ create a proxy by default and instead separate that in a specific `createTRPCClientProxy()`-fn

Closes #2111


packages/server/package.json
```diff
@@ -88,6 +88,11 @@
       "import": "./dist/subscription.mjs",
       "require": "./dist/subscription.js",
       "default": "./dist/subscription.js"
+    },
+    "./shared": {
+      "import": "./dist/shared.mjs",
+      "require": "./dist/shared.js",
+      "default": "./dist/shared.js"
```
KATT, line 95: Cc @sachinraja do you think we can throw `observables` and `rpc` into this as well rather than importing them separately?

Then it's slightly easier to create shared utilities, esp considering #2108.

Any modern bundler would tree-shake unusued stuff but I don't know the case for [Fresh](https://fresh.deno.dev/) etc

sachinraja: The observables are already imported in the client links so that doesn't matter. Our rpc stuff is not, but that's mostly just types, so it should be fine. Deno/Fresh already imports all of the links unnecessarily because of [this file](https://github.com/trpc/trpc/blob/next/packages/client/src/links/index.ts). This might be harsh but I could also care less about optimizing for non-tree-shaking environments.

sachinraja: Wait it should be completely fine, looks like Fresh does do tree-shaking, just at request time: <https://github.com/denoland/fresh/blob/main/src/server/bundle.ts>.

KATT: How the fuck does it do tree-shaking without a build step? Is it some sort of JIT-buildstep?

KATT: That sounds so slow

KATT: Created #2110

PR #2119, opened 2022-07-07
feat: create new trpc context for each createReactQueryHooks call

### LINKED ISSUE #2121: Using multiple TRPC clients in the same app
### Discussed in https://github.com/trpc/trpc/discussions/1990

<sup>Originally posted by **dodas** June 14, 2022</sup>
I am trying to have multiple `TRPCRouter`s on backend and separate `TRPCClient`s to access them.
Looking at [the way react hooks / contexts are set up](https://github.com/trpc/trpc/blob/main/packages/react/src/createReactQueryHooks.tsx), seems like there is no easy way to have 2 separate instances, since they reuse the same `TRPCContext` singleton.
I am expecting to be able to use these two routers/clients completely separately, without them conflicting with each other.
Is this a viable use-case? It seems that the current design does not allow this.

Thanks!</div>

dodas:
See https://github.com/trpc/trpc/discussions/1990

Closes #2121


packages/react/src/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from '@trpc/client';
 
 export { createReactQueryHooks } from './createReactQueryHooks';
+export type { CreateReactQueryHooks } from './createReactQueryHooks';
```
KATT, line 4: `export type` was the magic :)

KATT:
I chatted to @dodas about this the other day

I think this is a good approach forward:

- `withTRPC` + `createReactQueryHooks` gets `@deprecated` but continues using a global context for backward compatibility. Otherwise, `createReactQueryHooks` and `withTRPC` would have different root contexts which messes up `react-query`'s caching
- We add a new `export to `@trpc/react` `createReactQueryHooksWithContext` which becomes the new default, we probably want another name. `setupReact`?
- `setupNext` calls  `createReactQueryHooksWithContext` as part of the call.


@jlalmes @sachinraja @mmkal - WDYT?

mmkal:
That sounds good to me and I'd happily switch to the new functions - although if the existing functions can be updated in a backwards-compatible way (which it looks like this PR is going for), isn't that preferable? Or would we want to take the opportunity to rename anyway?

KATT:
I don't wanna change the behavior of existing ones

> Otherwise, createReactQueryHooks and withTRPC would have different root contexts which messes up react-query's caching

sachinraja:
That all sounds good to me @KATT!

KATT:
Hey @dodas, I'll close this now as it has a bunch of conflicts

PR #2125, opened 2022-07-07
Fixed typo for function parameter

hmzakhalid:
Small typo fix for example consistency.


KATT:
We're rewriting the docs for next major so I won't merge this b/c of conflicts, but thanks so much for this fix!

PR #2129, opened 2022-07-08
reset `./interop/examples` according to `main`

KATT:
Get rid of `setupNext`, etc


examples/.interop/next-prisma-starter-websockets/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@examples/interop-next-starter-websockets",
+  "name": "@examples/legacy-trpc-next-starter-websockets",
```
KATT, line 2:
```suggestion
  "name": "@examples/interop-next-starter-websockets",
```

Issue #2152, opened 2022-07-11
Rewrite all `./examples/.interop` to v10


KATT:
Feel free to submit PRs with migrating all of the examples here to the new version: https://github.com/trpc/trpc/tree/next/examples/.interop

 
1. copy folder to `/examples`
2. update `package.json` and remove `-interop` in the name
3. go through code and change it to `initTRPC()`-style stuff and remove old routers

All examples should run with `yarn dev` after `yarn` have ben run once in the root

Note: I want to keep all the existing interop examples -- they are there to ensure that the backward compatibility work

---
- [x] `cloudflare-workers`
- [x] `express-server`
- [ ] `express-trpc-shield`
- [x] `fastify-server` (needs update after #2114)
- [x] [`kitchen-sink`](https://github.com/trpc/examples-kitchen-sink) ~(blocked by #2113)~
- [x] `lambda-api-gateway` 🚧  #2165
- [x] `next-minimal-starter` 🚧  #2169
- [x] `next-prisma-starter` (blocked by #2113 + #2114)
- [x] `next-prisma-starter-websockets` ~(blocked by #2113)~
- [x] `next-prisma-todomvc` (blocked by #2113)
- [x] `standalone-server` (needs update after #2114)
- [x] `trpc-openapi`  (@jlalmes)
- [ ] [`zart`](https://github.com/trpc/zart) - outlined in #2118

Nsttt:
Adding a basic zart example. I see there's none at the moment.

KATT:
> Adding a basic zart example. I see there's none at the moment.

@Nsttt see #2118, @zomars is working on it :)

Nsttt:
Ah, got it 😊. I'll move into other stuff then.

Nsttt:
> > Adding a basic zart example. I see there's none at the moment.
> 
> @Nsttt see #2118, @zomars is working on it :)

Ah, got it 😊. I'll move into other stuff then.

KATT:
> > > Adding a basic zart example. I see there's none at the moment.
> > 
> > 
> > @Nsttt see #2118, @zomars is working on it :)
> 
> Ah, got it 😊. I'll move into other stuff then.

Or [reach out to zomars](http://trpc.io/discord) to see if you can help, he's on our discord!

AlvaroAquijeDiaz:
is `next-prisma-starter` updated to v.10 yet? And should the new version be used in production or wait for an official release?

KATT:
[...]

Issue #2114, opened 2022-07-07
Create `Proxy`-API for subscriptions


KATT:
Currently, the subscription API is unchanged and is using template literals. It should be  something like:

```tsx
// React:
function MyComponent() {
  proxy.post.onMessage.useSubscription({ /* ... */ })
}


// Raw:
proxy.onMessage.subscription({ /* ... */ });
```

jlalmes:
Related: https://github.com/trpc/trpc/pull/1999

PR #2156, opened 2022-07-11
chore: update fastify-server example to v10

juliusmarminge:
Resolves (partially) #2152 

Router and Client updated to v10.

Didn't quite get the subscription to work. Will get to that tomorrow.

### DIFF
--- a/examples/fastify-server/.gitignore
+++ b/examples/fastify-server/.gitignore
@@ -1,0 +1,2 @@
+node_modules
+dist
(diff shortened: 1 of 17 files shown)


juliusmarminge:
> Didn't quite get the subscription to work. Will get to that tomorrow.

Just spotted #2114 and that you are looking into changing the subscription API so maybe I hold off on this until then.

KATT (changes requested):
Same as comment on other PR, just re-open it on origin so I can add CI-checks :) 🙏 

Thanks!

PR #2135, opened 2022-07-08
Use the OnErrorFunction type for AWS Lambda adapter's onError method

pekala:
👋 Hey tRPC folks!
Currently the arguments to `onError` in AWS Lambda adapter are typed as a record of `unknown` which is cumbersome to work with. The more helpful type already exists, so I've just used it here. Let me know if I'm missing something.

PR #2157, opened 2022-07-11
Use the OnErrorFunction type for AWS Lambda adapter's onError method

pekala:
Re-creating #2135 

👋 Hey tRPC folks!
Currently the arguments to onError in AWS Lambda adapter are typed as a record of unknown which is cumbersome to work with. The more helpful type already exists, so I've just used it here.


pekala:
@KATT the failing test looks like a flaky one, right? I mean, the type changes would probably not cause a playwright test to fail.

Lilja (approved):

KATT:
Ahhhh fuck, now the history isn't linear on `main` anymore

sachinraja:
I think we should [only enable squash merges](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests).

KATT:
> I think we should [only enable squash merges](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests).

I've disabled that because of the `next` branch where I **need** merge to keep it up-to-date

sachinraja:
Maybe see [this](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-linear-history).

PR #2169, opened 2022-07-12
chore: upgrade next-minimal-starter to v10

juliusmarminge:
The old example wasn't named interop so renamed that as well to keep the consistency with the other examples.

Resolves (partially) #2152


KATT:
Wow, you're fast ❤️

juliusmarminge:
> Wow, you're fast 

Was checking in on the other ones and noticed I had 'missed' one ;)

juliusmarminge:
So I took another look at this and resolved the issue. But since this shouldn't be necessary and isn't if you build the project standalone, I believe there are some config further down the monorepo which is causing this?

PR #2198, opened 2022-07-14
Feat: Add `X-Powered-By` header

jlalmes:
> @KATT: I'd like to get a better picture of number of apps used in production, right now we can only do wild guesses based on npm downloads

## Feat: Add `X-Powered-By` header

```
X-Powered-By: tRPC <trpc.io>
```

This header will be useful for analytics, so that tools like [BuiltWith](https://builtwith.com), [Wappalyzer](https://www.wappalyzer.com/) etc. can track tRPC trends over time.

The `X-Powered-By` is enabled by default, it can be explicitly enabled/disabled with the `poweredByHeader`. 

```ts
const t = initTRPC()({
  ...,
  poweredByHeader: false,
});
```

We should add a reference on how to disable this into `API Reference` page of the new docs.

### DIFF
--- a/packages/server/src/core/initTRPC.ts
+++ b/packages/server/src/core/initTRPC.ts
@@ -10,6 +10,7 @@ import {
   CombinedDataTransformer,
   DataTransformerOptions,
   DefaultDataTransformer,
+  defaultPoweredByHeader,
   defaultTransformer,
   getDataTransformer,
 } from '../transformer';
@@ -47,21 +48,25 @@ export function initTRPC<TParams extends Partial<InitGenerics> = {}>() {
         : DefaultDataTransformer
       : DefaultDataTransformer;
     type $ErrorShape = ErrorFormatterShape<$Formatter>;
+    type $PoweredByHeader = NonNullable<$Options['poweredByHeader']>;
 
     type $Config = CreateRootConfig<{
       ctx: $Context;
       meta: $Meta;
       errorShape: $ErrorShape;
       transformer: $Transformer;
+      poweredByHeader: $PoweredByHeader;
     }>;
 
     const errorFormatter = options?.errorFormatter ?? defaultFormatter;
     const transformer = getDataTransformer(
       options?.transformer ?? defaultTransformer,
     ) as $Transformer;
+    const poweredByHeader = options?.poweredByHeader ?? defaultPoweredByHeader;
     const _config: $Config = {
       transformer,
       errorShape: null as any,
+      poweredByHeader,
       ctx: null as any,
       meta: null as any,
     };
@@ -85,6 +90,7 @@ export function initTRPC<TParams extends Partial<InitGenerics> = {}>() {
       router: createRouterFactory<$Config>({
         errorFormatter,
         transformer,
+        poweredByHeader,
       }),
       /**
        * Merge Routers
--- a/packages/server/src/core/internals/config.ts
+++ b/packages/server/src/core/internals/config.ts
@@ -2,6 +2,7 @@ import { ErrorFormatter } from '../../error/formatter';
 import {
   CombinedDataTransformer,
   DataTransformerOptions,
+  PoweredByHeader,
 } from '../../transformer';
 
 /**
@@ -19,6 +20,7 @@ export interface InitGenerics {
 export interface InitOptions<T extends InitGenerics> {
   transformer: DataTransformerOptions;
   errorFormatter: ErrorFormatter<T['ctx'], any>;
+  poweredByHeader: PoweredByHeader;
 }
 
 /**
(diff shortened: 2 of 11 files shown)


Ankcorn:
> This might not be 100% related, but I'll give it a shot. Could this be part of sensitive information and paint a target on your API's back if your attacker knows what framework is being used?

The discussion around this for express is pretty helpful and would lean in favor of adding the header.

https://github.com/expressjs/express/pull/2813#issuecomment-159270428

sir-mr-bean:
> > This might not be 100% related, but I'll give it a shot. Could this be part of sensitive information and paint a target on your API's back if your attacker knows what framework is being used?
> 
> 
> 
> The discussion around this for express is pretty helpful and would lean in favor of adding the header.
> 
> 
> 
> https://github.com/expressjs/express/pull/2813#issuecomment-159270428

> Does removing them help with security? No, only increases obscurity, which is not a valid form of security by any means.

Nail on the head. All for it 👍

mmkal:
> > > This might not be 100% related, but I'll give it a shot. Could this be part of sensitive information and paint a target on your API's back if your attacker knows what framework is being used?
> > 
> > 
> > The discussion around this for express is pretty helpful and would lean in favor of adding the header.
> > [expressjs/express#2813 (comment)](https://github.com/expressjs/express/pull/2813#issuecomment-159270428)
> 
> > Does removing them help with security? No, only increases obscurity, which is not a valid form of security by any means.

I personally don't really agree with that quote from https://github.com/expressjs/express/pull/2813#issuecomment-159270428. It's a bit of a developer-centric view - and in my experience it isn't shared by folks who work specifically in InfoSec depts. I suspect there was some proto "security through obscurity is insufficient" type quote that got bastardised into "security through obscurity is worthless". Increasing obscurity _does_ help with security. Almost always, security vulnerabilities are time-based. If there is a remote-code execution flaw in trpc which is exposed, hackers will likely immediately get to work looking for vulnerable servers. At the same time, trpc maintainers will look for a mitigation or fix, publish it, and notify end-users. The end-users will then have to updated their trpc version, and deploy their servers. Nobody can say how long any of those things will take, so the easier we make it for bad actors to probe servers and find out information about them, the worse. The fact that trpc has _somewhat_ unique response shapes is a much less clear signal than a header which clearly and unambiguously says "I am using software which is vulnerable to XYZ". Attackers will always go for the low hanging fruit first. There's just more bang for buck that way.

>If someone wanted to attack you via trpc it'd be easy to detect anyway. All Next.js trpc apis are on the /api/trpc route.

That's a convention, not a requirement. And maybe, one trpc should move away from for that reason.

I don't think this is a _huge_ deal, to be clear. But if we want to merge this, I think we should at least include a note in the docs saying that it's the default behaviour, why you _might_ not want that, and provide an easy way to disable it.

I'm also going to review the code itself but separating that out since this message ended up longer than I'd expected!

mmkal (commented):
As mentioned in the other comment, we should include something somewhat prominently in the documentation if we want to do this.

Also, I don't know exactly the right way to address this, but it feels really, really wrong to take 116 lines of code in eleven files to add a single, hard-coded header to default responses. The value for money for this tiny feature is just way off, and maybe a sign there's something missing in the internal design. A couple of options that occur to me, but I haven't explored in practice:

1. Replaces all the instances of `PoweredByHeader` with `Headers` - a `Record<string, string>` which is spread into the response headers instead:

```diff
-  ...(router._def.poweredByHeader && { 'X-Powered-By': 'tRPC <trpc.io>' }),
+  ...router._def.headers, 
```

That would have the advantage of not requiring _another_ 116 lines of code next time someone wants to add another optional hard-coded header to HTTP responses by default

2. Make the implementation live in specific http adapters, maybe using a common helper. That might be better because as far as I'm aware, this would be the first thing that couples trpc generally to the HTTP protocol - it makes sense for `expressAdapter`, `nextjsAdapter` or `awsLambdaAdapter` to know about the word "header" - but what about a future adapter which doesn't use HTTP at all (I've idly thought about a CLI adapter for invoking trpc procedures from the command line - what does a header mean there)?

jlalmes:
#### Update:

Since opening this PR I've had some second thoughts.

The purpose for adding this header was purely for analytics so that a service like _BuiltWith_ could track tRPC usage in the wild (_npm_ downloads don't really mean much).

* Typically `X-Powered-By` header is set by a server (i.e. `express`/`nextjs`/`lambda`,etc.) not a framework (i.e. `tRPC`)
* Additionally `nextjs` will strip all [custom response headers](https://vercel.com/docs/concepts/edge-network/headers#custom-response-headers) unless you whitelist them in your `next.config.js`.

Ultimately, this does not seem like an effective solution to our analytics problem - and hence I will close this issue. 

I have opened a discussion for any ideas 💡 (https://github.com/trpc/trpc/discussions/2244).

PR #2207, opened 2022-07-15
refactor: split react.test.tsx in multiple files

### LINKED ISSUE #1661: Refactor `react.test.tsx`

joaosamouco:
This MR just splits `react.test.tsx` into multiple logical files.
Maybe more improvements could be made but those changes would be hard to track in this PR. I believe this is a good first step and we can work from there.

Closes #1661


packages/server/test/interop/react/__testHelpers.tsx
```diff
@@ -0,0 +1,265 @@
...
+});
+
+type Context = {};
+export type Post = {
+  id: string;
+  title: string;
+  createdAt: number;
+};
+
+export function createAppRouter() {
```
KATT, line 39: This is the v9 way of doing stuff.

```suggestion
/** @deprecated **/
export function createLegacyAppRouter() {
```

KATT (commented):
Great!

We can use this to have a foundation for v10 test suite. As you probably noticed, this is all in the `/interop`-folder which is legacy (current v9) stuff

packages/server/test/interop/react/__testHelpers.tsx
joaosamouco: Added your suggestion and renamed everything accordingly :+1:

KATT:
@all-contributors add @joaosamouco for tests

PR #2209, opened 2022-07-15
server side calls

### LINKED ISSUE #2124: [RFC] Server-side calls & metadata
When calling tRPC procedures for instance integration tests or in pure backend calls that aren't over the wire.

### Given this backend

```ts
type Context = {
  user?: { id: string },
}
type Meta = {
  openapi: { 
   enabled: true, 
   method: 'GET', 
   path: '/say-hello' ,
  }
}
const t = initTRPC<{ 
  ctx: Context; 
  meta?: Meta; // meta is set to be optional
}>();

const appRouter = t.router({
  postById: procedure
  .input(
    z.object({
      id: z.string(),
    }),
  )
  .query(({ input }) => {
     //
  })
});
```

**We want to allow calling it like this:**

```ts
const result = await appRouter.post.byId({ 
  input: { id: '1' },
  ctx: {},
});
const result = await appRouter.post.byId({ 
  input: { id: '1' },
  ctx: {
    user: {
      id: 'KATT'
    }
  }
});
```

<details><summary>Old example ahead of @sachinraja's suggestion</summary>
```ts
const result = await appRouter.post.byId.call({ 
  input: { id: '1' },
  ctx: {},
});
const result = await appRouter.post.byId.call({ 
  input: { id: '1' },
  ctx: {
    user: {
      id: 'KATT'
    }
  }
});
```
</details>

**And we want to be able to access metadata like this**:

```ts
const postDef = appRouter.post.byId._def;
postDef.input // -> z.ZodType<...>
postDef.output //  ->undefined
postDef.meta // -> Meta | undefined
```

### Stretch goal

Related to this: we should delete the "internal procedure builder" and replace it by a procedure exposing an interface like this when `.query` or `mutation` is called:

```ts
interface Procedure<TParams> {
  _def: ProcedureDef;
  // input should be `input?:` if it input is optional
  call(opts: { input: TParams['input_in'], ctx: TParams['_config']['ctx']): Promise<TParams['_output_out']>;
}
```

This means we need to update the client and react lib as well to remap to their own implementation.

sachinraja:
Closes #2124

- call procedures directly server-side
  ```ts
  const caller = createCaller(appRouter)
  const result = await caller.post.byId({ 
    input: { id: '1' },
    ctx: {},
  });
  const result = await caller.post.byId({ 
    input: { id: '1' },
    ctx: {
      user: {
        id: 'KATT'
      }
    }
  });
  ```
  `ctx` is not set in the `createCaller` because I'm pretty sure we don't want people to spam proxy callers everywhere they need to change the context
- remove separate internal procedure builder


KATT (approved):
This is insane, can you add some tests to `smoke.test.ts` or a new file to show it off? :) ❤️

packages/server/test/caller.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  const router = t.router({
+    secret: protectedProcedure.query(({ ctx }) => ctx.foo),
+  });
+
+  const caller = createCaller(router);
+  expect(caller.secret({ ctx: {} })).rejects.toThrow(TRPCError);
+
+  const result = await caller.secret({ ctx: { foo: 'bar' } });
+  expect(result).toBe('bar');
+});
```
KATT, line 84: What do you think about adding something like this?

```suggestion
});


test('raw call', async () => {
  const router = t.router({
    hello: procedure.input(z.string()).query(() => 'world'),
  });
  router.hello({
    // @ts-expect-error this should be a string
    rawInput: undefined,
    ctx: {
      // @ts-expect-error this should bar
      foo: 'not-bar',
    },
    path: 'hello',
    // @ts-expect-error this should be 'query'
    type: 'mutation',
  });
});

```

Everything but the `path` should be able to be strict right?

sachinraja: ah correct. What's the use case though?

KATT (approved):

mmkal:
@sachinraja 

>ctx is not set in the createCaller because I'm pretty sure we don't want people to spam proxy callers everywhere they need to change the context

The tradeoffs here seem like they might be off to me. 

First: this feels like it makes the wrong user have to be concerned with `ctx`. In my current team's set up, we rely heavily on `createCaller` in integration tests, and in the vast majority of cases, we _don't_ want or need to change the context. And especially, to very loosely think in terms of domains, the person who's calling `caller.whateverProcedure(...)` has no interest in `ctx`, and often no knowledge about it. It will be inconvenient to force everybody who uses a caller to have to pass in `ctx` as well as the input.
I can't say for sure whether my team's codebase is representative of most others, but I strongly suspect that contexts like `{ foo: 'not-bar' }` are very rare, so the terseness/readability of the tests here might be a bit misleading. There will be lots of fiddly-to-construct real-world stuff in there for most people. Only a small portion of which is not relevant to any given procedure. We already have some helpers which make heavy use of typescript wizardry for use in tests, so we'll be able to add to those to work around this and effectively give ourselves back the old API, but we were hoping we could do less typescript wizardry in our application's tests when v10 lands, not more.

Second: what do you mean by "spam proxy callers" everywhere? Are we worried about the perf impact of using `new Proxy(...)` too much? I ask this as someone who's feeling a bit uneasy about how big of a bet trpc v10 seems to be taking on `Proxy` in general. I don't particularly think there should be much perf impact, but I don't like how magical it is - it makes it harder to intuitively reason about things like performance, and generally harder for people to eyeball and just roughly know what's happening. That second part might not be relevant to this PR - but maybe that proves my point - when `Proxy` is used liberally, you can't look at _any_ piece of code and not be worried it's a `Proxy` which is doing _way_ more than it looks like it's doing!

(Sorry am catching up from being many months behind on a lot of v10-facing stuff, so please let me know if this would be better on Discord, or if it's actively unhelpful for me to being asking these questions at all)

KATT:
@mmkal I always appreciate your input, but beware that commenting on closed PRs and issues usually gets lost in the ether, I just saw this randomly.

Please open a discussion or thread on discord if you think the approach taken here (or in the follow-up) is wrong

sachinraja:
@mmkal I've updated it in #2233 to passing the context to `createCaller` instead of each procedure call

Issue #2203, opened 2022-07-14
Make a banner on top of the website


KATT:
I'd like a top banner on the top of the Docusaurus page where we can have like _"hey you're watching V10, read the documentation guide"_

[reactjs.org](https://reactjs.org/) has a top banner about Ukraine and I'd like something similar, their source code is at https://github.com/reactjs/reactjs.org

PR #2210, opened 2022-07-15
Add docs announcement bar

OlegGulevskyy:
As per issue #2203 , adding support to docausurus config file, that will render pretty minimal shy announcement bar.
Now the default announce bar is, like I said, pretty shy, so I'd be up for customizing it a bit to make it a bit more fancy and dancy, but that's of course only if you think it'd be cool.
For this I can swizzle the component and customize it further.
Without customization it looks like so
image

Background and font color are cusotmizable, and for this PR are taken from tailwind config.
You can render `html` inside the announcement bar `content` property.

### DIFF
--- a/www/docusaurus.config.js
+++ b/www/docusaurus.config.js
@@ -23,6 +23,11 @@ module.exports = {
       // contextualSearch: true,
       // searchParameters: {},
     },
+		announcementBar: {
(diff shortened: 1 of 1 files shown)


KATT:
@all-contributors add @OlegGulevskyy for docs

KATT:
Thank you!

PR #2224, opened 2022-07-17
feat(docs): add versioning

### LINKED ISSUE #2179: Can we leverage Docusaurus versioning?
Can we leverage [Docusaurus versioning](https://docusaurus.io/docs/versioning) for V10?

I **don't** want to do versioning for each minor/major, but it would be nice to have a `9.x` and `10.x` checked in.

Otherwise, it's also perfectly fine if we have to independently deployed apps as well.

juliusmarminge:
Resolves #2179 

Added versioning of docs that will place the v9 docs under a subpath `docs/9/*` and the v10 as default `/docs/*`.


KATT (commented):
Looks great, did you copy all the docs from `main` for version 9?

juliusmarminge:
> Looks great, did you copy all the docs from main for version 9?

Yes. Were unsure what way I should have done it (copy v10 to main or v9 to next) but ended up copying [main/docs](https://github.com/trpc/trpc/tree/main/www/docs) into [next/docs/v9](https://github.com/trpc/trpc/tree/docs/versioning/www/versioned_docs/version-9.x).

KATT:
Nice. I did a bunch of updates to it now :) does it look good to you still?

juliusmarminge:
Nice. I did a bunch of updates to it now :) does it look good to you still?

Yeah having it to the right looks good.

Reason I didn't use current was I couldn't get the paths to resolve correctly without having v10 in versioned (so I ended up having v9, v10 and current which I don't think was the goal in the original issue) but looks like you figured it out :)

KATT:
There are a lot of warnings in the console and the build, is there anything we could do about those, @juliusmarminge? (They were there before my changes fwiw)

juliusmarminge:
> (They were there before my changes fwiw)

I noticed. Will continue to investigate tonight 👍

KATT:
@all-contributors add @juliusmarminge for docs

juliusmarminge:
Looks like it is some issues with babel dependencies. Builds without warnings now but still warns in dev-mode:

CleanShot 2022-07-17 at 14 20 09

CleanShot 2022-07-17 at 14 21 27

juliusmarminge:
So I investigated some more into this and the base `next` branch also has these warnings. I recreated the versioned docs based on the main branch and there it worked without warnings, so the issue is most likely outside of the `www`-scope.

I checked the diff between main and next in regards to build/setup and noticed you have changed some stuff and included rollup with babel plugins etc. Maybe there are dependencies here that fails to resolve and waterfalls down to Docusaurus?

ixahmedxi:
you guys sure there are babel errors? I'm not getting any for me @KATT @juliusmarminge I ran both build & start and everything works fine

PR #2233, opened 2022-07-17
use `router.createCaller` for server-side calls

### LINKED ISSUE #2225: [RFC] Server-side calls, second iteration
Now that we actually need the abstraction of the `createCaller()`, I think we'd be better off making by either:

- making this a part of the existing `appRouter.createCaller()` but make the old fns deprecated
- adding `ctx` to `createCaller()` like `createCaller({ router, ctx })` 

this is my bad, i forgot that `path` and `type` so I thought we could just do `appRouter.post.byId({ ctx, input})`

## Current

```ts
const caller = createCaller(router);
await caller.post.delete({
  input: 0,
  ctx: {},
});
```

### A)

```ts
const caller = createCaller({ router, ctx: {} });
await caller.post.delete(0);
```

### B)

```ts
const caller = router.createCaller({} /* ctx here */ );
await caller.post.delete(0 /* input here */);
```

### C) better suggestion?

sachinraja:
closes #2225

I went with `router.createCaller`:
```ts
const router = t.router({
  greeting: t.procedure
    .input(z.object({ name: z.string() }))
    .query(({ input }) => `Hello ${input.name}`),
});

const caller = router.createCaller({});
const result = await caller.greeting({ name: 'Sachin' });
```


packages/server/src/core/router.ts
```diff
@@ -291,54 +297,39 @@ export function createRouterFactory<TConfig extends RootConfig>(
...
-          ),
-          subscriptions: createRouterProxy((path, rawInput) =>
-            callProcedure({
-              path,
-              rawInput,
-              ctx,
-              type: 'subscription',
-            }),
-          ),
-        };
```
sachinraja, line 341: I don't know what these were for and they were untested so I removed them.

KATT (approved):
🐐

KATT:
Cc @mmkal see this and open a discussion if you have ideas or questions

Issue #1599, opened 2022-03-06
[TRP-65] 🤑 Get paid for contributing to tRPC


KATT:
Hey all 👋 

We've started getting in some non-trivial amounts of sponsorship money and we will finally be able to pay contributors for work.

1. Find **any issue** on GitHub _(or suggest an improvement - we're open to ideas!)_
1. Create a PR with the improvement
1. Set up [GitHub sponsors](https://github.com/sponsors/accounts)
1. We'll send you $$$

Usually in the realm of $10-$50 per PR but happy to send significantly more for non-trivial work. 

While we realize this isn't meaningful payment for qualified labor, maybe it can nudge someone over the edge if they want to get into contributing to OSS?

> If someone has questions on what they can get their hands dirty on or ideas for other things, my DMs are open both [on Twitter](https://twitter.com/alexdotjs) & [Discord](https://trpc.io/discord) 📬 

---

Massive thanks to all of [the sponsors](https://trpc.io/docs/sponsors) who enable us to do this! ❤️ 

<sub>[TRP-65](https://linear.app/trpc/issue/TRP-65/🤑-get-paid-for-contributing-to-trpc)</sub>

KATT:
I will send $500 once it's merged if someone solves this issue: https://github.com/microsoft/TypeScript/issues/49033

PR #2240, opened 2022-07-17
chore: Update react query version

### LINKED ISSUE #2352: ~React~ TanStack Query 4 support

dyaa:
Closes #2352

### DIFF
--- a/packages/client/src/internals/TRPCClient.ts
+++ b/packages/client/src/internals/TRPCClient.ts
@@ -82,7 +82,7 @@ export interface TRPCRequestOptions {
   /**
    * Pass additional context to links
    */
-  context?: OperationContext;
+  requestContext?: OperationContext;
 }
 
 export interface TRPCSubscriptionObserver<TValue, TError> {
@@ -203,7 +203,7 @@ export class TRPCClient<TRouter extends AnyRouter> {
     path: TPath,
     ...args: [...inferHandlerInput<TQueries[TPath]>, TRPCRequestOptions?]
   ) {
-    const context = (args[1] as TRPCRequestOptions | undefined)?.context;
+    const context = (args[1] as TRPCRequestOptions | undefined)?.requestContext;
     return this.requestAsPromise<
       inferHandlerInput<TQueries[TPath]>,
       inferProcedureOutput<TQueries[TPath]>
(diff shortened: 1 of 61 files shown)


KATT:
> 1. `setLogger` has been [removed](https://github.com/TanStack/query/blob/c5c4417f430a58b2dc7911cee99837b68f44a59c/docs/guides/migrating-to-react-query-4.md#setlogger-is-removed) in the latest beta version, and instead, a custom `logger` can be passed to `QueryClient`'s constructor.

Feel free to remove our use of `setLogger` everywhere.

> 2. There will be type mismatches with `@trpc/react` package since it's using an older version of `react-query`.

What sort of stuff? Will it be impossible to support both?

> Next:
> 
> 1. I'll make a full upgrade to the `react-query` package to their beta version and update all internal packages to use the beta version.
> 2. I'll remove the dependency on the env variable `REACT_QUERY_VERSION`.

Will it be impossible to support both versions?

dyaa:
I don't think that's possible without having `@trpc/react` using the same version for `react-query`.

KATT:
cc @dyaa update to `main` which doesn't try to make this cross-compatible anymore. Will be a lot easier to fix the issues then.

KATT:
also please get github sponsors working so I can send you money for it :) #1599

KATT:
I'll send some more money if you fix it, just sent you a fiver for a beer or a coffee now 🍻

KATT:
Oops, I closed the wrong PR!

packages/client/src/internals/TRPCClient.ts
```diff
@@ -82,7 +82,7 @@ export interface TRPCRequestOptions {
   /**
    * Pass additional context to links
    */
-  context?: OperationContext;
+  requestContext?: OperationContext;
```
dyaa, line 85: @KATT looks like we'll need to change our `OperationContext` name to avoid conflicts with Context Provider, wdyt?

dyaa:
Failing deployments, Do not have access to investigate 
- Vercel – next-prisma-starter
- Vercel – todomvc

CC\ @KATT would you please check?

packages/client/src/internals/TRPCClient.ts
KATT, line 85: Could you add a note of this in https://alpha.trpc.io/docs/migrate-from-v9-to-v10 anywhere? I'll clean it up later.

@sachinraja @jlalmes what do you think we should do here now that the `context` keyword is taken?

Some options:

- Put all trpc option under `trpc` - quite annoying for the 99% use-case but also nice to have a distinct place for "our" versus "theirs" in options
- Call it `trpcContext`? Potentially prefix all our vars with `trpc`?
- `operationContext`?
- `opContext`?
- `callContext`?

KATT (commented):
Lovely, I just want to get some feedback from @trpc/trpc-core in regards to the `requestContext` property

packages/client/src/internals/TRPCClient.ts
KATT: Moved to #2394.

Will merge this one and we can iterate as it's still alpha.

KATT:
@all-contributors add @dyaa for code, test

PR #2262, opened 2022-07-19
docs: add version-tag to react-query in installation snippets

juliusmarminge:
Update installation snippets since react-query have tagged v4 as stable and it is not currently compatible.


KATT (commented):
Should it be `react-query@^3` or `react-query@3`?

juliusmarminge:
> Should it be `react-query@^3` or `react-query@3`?

Not sure if that matters? Tried both and they both installed `3.39.2`, only difference is the symbol displayed in package.json. Not sure how this works when you dont specify an exact version...

```json
  "react-query": "3",
  // vs
  "react-query": "^3",
```

KATT (approved):
Thanks!

PR #2271, opened 2022-07-20
fix(www): sync docusaurus and tailwind dark mode 

ixahmedxi:
This pull request changes the following:

- Syncs both docusaurus and tailwind to be on dark/light mode based on system preferences and disables manual docusaurus theme toggling that way both docusaurus and tailwind will always be on light/dark mode together.
- This allows the ability to use the `dark:` class name prefixes of tailwind and for them to work as expected again.


ixahmedxi:
don't worry about the `index.tsx` class name shuffle around it's just a vscode extension I have that organises class names of tailwind.

KATT (approved):

PR #2273, opened 2022-07-20
feat(www): reorganises & styles the hompage navbar

ixahmedxi:


www/docusaurus.config.js
```diff
@@ -39,24 +39,13 @@ module.exports = {
...
-          label: 'Quickstart',
-        },
-        {
-          to: 'docs/awesome-trpc',
-          label: 'Awesome tRPC Collection',
-        },
-        {
-          to: 'docs/nextjs',
-          label: 'Usage with Next.js',
-        },
```
KATT, line 55: Hmm why do you wanna delete these?

ixahmedxi: It's based on the new proposed design
CleanShot 2022-07-21 at 13 36 27@2x

The two deleted links are also in the first section of the docs immediately, if necessary we could probably have a dropdown button that would link some important sections of the docs because now that the navbar has the container class of tailwind it won't fit many items. As for the twitter and discord links we will have those underneath the main introductory content in the landing page so they aren't gone they will just be moved.

KATT (changes requested):
Hmmmm... I would be more comfortable if we started with the rest of the layout before we make the decision of hiding stuff in the top bar.

In short: let's not tamper with the top bar _for now_


Also, on your proposed designs, we have previously mentioned that we want to show the client and the server at once in order to showcase the inference, so please bear in mind that that is still the case.

If you want "safe" sections to work on in terms of things we "know" I'd have a look at everything that was _below_ the hero and the top bar area.

KATT:
Closing this for now

PR #2277, opened 2022-07-21
Added uWebSockets adapter

romanzy-1612:
Adapter for [uWebSockets.js](https://github.com/uNetworking/uWebSockets.js)

Everything works, except the res object in the context (see notes in adapter code). Also, there are some questions in code regarding potential issues/weirdness.

BTW, need a better name for registration function. Currently its `registerTRPCasUWebSocketsEndpoint` 🥴

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -66,6 +66,7 @@
     "node-fetch": "^2.6.1",
     "superstruct": "^0.15.2",
     "typescript": "4.4.4",
+    "uWebSockets.js": "uNetworking/uWebSockets.js#v20.10.0",
     "ws": "^8.0.0",
     "yup": "^0.32.8",
     "zod": "^3.0.0"
--- a/packages/server/src/adapters/uWebSockets.ts
+++ b/packages/server/src/adapters/uWebSockets.ts
@@ -3,13 +3,12 @@ import { TRPCError } from '../TRPCError';
 import { HTTPRequest } from '../http/internals/types';
 import { resolveHTTPResponse } from '../http/resolveHTTPResponse';
 import { AnyRouter, inferRouterContext } from '../router';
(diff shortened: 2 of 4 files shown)


jlalmes:
Thanks for the PR @romanzy-1612!

I think this adapter should probably be released as it's own package (something like `trpc-uwebsockets`) so that the tRPC team aren't left to maintain. You're probably already aware, but we support already supports native WebSockets out of the box.

I've never actually seen [`uWebSockets.js`](https://github.com/uNetworking/uWebSockets.js) before, it seems to be a super impressive node WebSocket implementation from the performance benchmarks.

KATT (changes requested):
Hey! Thank you so much for your contribution!

We aren't currently accepting big changes into the `main` branch as we're hard at work with the `next`-branch (new major coming soon! 🥳)

Unfortunately, we haven't _properly_ fixed the support for subscriptions on the next major yet (#2114), but contributions are welcome.

Feel free to change your suggested changes `main` to `next` and we'll happily take a look at it. 

Furthermore:

1. Close this PR


KATT:
Also, I'm not sure if we want to support more adapters, we have one for `ws` that works, but you could release a `trpc-uwebsockets` package? Is there any internals that you need to access?

romanzy-1612:
Only internals needed are types, which can be easily imported. As per your suggestions, I have published it as [trpc-uwebsockets](https://www.npmjs.com/package/trpc-uwebsockets). Only 2 new files are added, and they should be independent from the rest of the code.

uWebsockets, despite its name, is also a really good HTTP server with scalable pub-sub. Static file and SSL performance is phenomenal. It is not very popular because it exposes a much lower level API then express. But with tRPC it is easy to use!

KATT:
Nice! Add it to the awesome trpc collection in `/www` ☺ 🙏 ♥

PR #2286, opened 2022-07-21
docs: rewrite docs for type inference

juliusmarminge:
Referencing this comment on a pull request on trpc/zart:
https://github.com/trpc/zart/pull/24#discussion_r926082681

We should probably rewrite the inferance-helper section as well. I began drafting this out but quickly realized I didn't have enough knowledge of how this new way worked and how to document this so I think I'l leave it to someone else to pick up on, or if you can point some stuff out that I can continue upon.


www/docs/server/infer-types.md
```diff
@@ -15,81 +15,37 @@ It is often useful to wrap functionality of your `@trpc/client` or `@trpc/react`
 - `inferProcedureInput<TProcedure>`
 - `inferSubscriptionOutput<TRouter, TPath>`
```
sachinraja, line 16: I think this should be removed

KATT: Agreed!

juliusmarminge: Ok. Wasn't sure about the API of the subscription so I just left it there for now. Can remove it. Anything else we should add?

KATT:
shippp iiit

PR #2285, opened 2022-07-21
feat(www): implements new introduction section

ixahmedxi:

PR #2287, opened 2022-07-21
feat(www): implements new features section

ixahmedxi:
needs https://github.com/trpc/trpc/pull/2285

### DIFF
--- a/www/src/components/SectionTitle.tsx
+++ b/www/src/components/SectionTitle.tsx
@@ -1,0 +1,17 @@
+import React, { FC } from 'react';
+
+type SectionTitleProps = {
+  readonly title: string;
+  readonly description: string;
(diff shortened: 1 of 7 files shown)


www/src/components/SectionTitle.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { FC } from 'react';
+
+type SectionTitleProps = {
+  readonly title: string;
+  readonly description: string;
+};
+
+export const SectionTitle: FC<SectionTitleProps> = ({ title, description }) => {
```
ixahmedxi, line 8: There are other sections in the design that will also have a title and a description with the same styling, hence why I opted into making a component for that. It will be used in the future with other sections as well.

Issue #2284, opened 2022-07-21
Add server proxy wrapper helper to @trpc/server v10


IgnusG:
I'm experimenting with a new stack and using v10 of trpc for it (the new proxy calls are awesome btw.). During SSR, I want to call my router procedures directly from the server (so without the client) - the adapters seem to still use the previous `.createCaller` API (maybe there's a newer one available?). I've noticed the types do actually include the correct (procedure) paths but those values are all undefined during runtime:

```ts
const UserRouter = trpc.router({
	getAll: trpc.procedure.query(async () => {
		return (await prisma.user.findMany()) as UserRecord[];
	}),
});

console.log(UserRouter.getAll.query()); // <- Typescript is happy, but it doesn't work during runtime (.getAll is undefined)
```

Based on the client proxy wrapper (which gave me an error when I wrapped the server router with it), I wrote this little helper:

```ts
import type { AnyRouter } from "@trpc/server";

export function createTRPCServerProxy<T extends AnyRouter>(router: T, path: string[] = []): T {
	return new Proxy(router, {
		get(target, p: string) {
                        // You can extend this for mutations too - 'tis just an example
			if (p === "query") {
				return (input: any, opts: any) => {
					const caller = target.createCaller(opts.context);

					return caller.query(path.join("."), input);
				};
			} else {
				return createTRPCServerProxy(target, [...path, p]);
			}
		},
	});
}
```

```ts
console.log(createTRPCServerProxy(UserRouter).getAll.query()); // <- Works and typechecks correctly
```

The types for the router are already there (as I mentioned earlier) so we don't have to adjust those. I just wrap the property accesses and chain them into the query after the `createCaller` has been finally called.

This worked for me, it would be nice though if trpc@10 had this method built-in 🙂  
Alternatively it would also work if the procedure properties were actually defined during runtime (I don't know if this is planned since the types are there but it might require more changes if it is not).

KATT:
See #2233, does that look like what you want?

KATT:
Maybe I didn't release a new alpha with that 🤔

IgnusG:
Oh yeah, this is perfect! I think the version I have (`10.0.0-alpha.38`) doesn't have that yet (neither during runtime, nor the typings)

JonParton:
I can't seem to figure out how to get this `caller` to work either (runtime or types) so hopefully just not included in the release!

KATT:
[...]

PR #2303, opened 2022-07-23
Use chainable proxy in `appRouter.createCaller()`

KATT:


examples/.interop/next-prisma-starter/src/server/routers/post.test.ts
```diff
@@ -13,10 +13,8 @@ test('add and get post', async () => {
     text: 'hello test',
     title: 'hello test',
   };
-  const post = await caller.mutation('post.add', input);
-  const byId = await caller.query('post.byId', {
-    id: post.id,
-  });
+  const post = await caller['post.add'](input);
+  const byId = await caller['post.byId']({ id: post.id });
```
KATT, line 17: cc @sachinraja 


Shouldn't this be?:

```suggestion
  const post = await caller.post.add(input);
  const byId = await caller.post.byId({ id: post.id });
```

❓

KATT: Related: #2284

cc @IgnusG

sachinraja: Yes but I'm not sure it's worth it to implement. Type perf is already worsened on the interop routers. We'd have to replicate the structure of a nested router, derived from an interop router.

KATT: Alright, let's take this as a task on the roadmap

PR #2304, opened 2022-07-23
fix(docs): double slash comment was misaligned

juliusmarminge:
Sorry, OCD was killing me about this..

CleanShot 2022-07-23 at 12 58 17@2x

so i fixed the indentation...

CleanShot 2022-07-23 at 13 03 39@2x


KATT (approved):
Love the ocd, thanks

PR #2307, opened 2022-07-24
react-query utils proxy

### LINKED ISSUE #2113: Create proxy methods for `react-query`'s `useContext`
We might want to attack https://github.com/trpc/trpc/issues/2124 ahead of this

```ts
function MyComponent() {
  trpc.proxy.post.list.useQuery(/* input */)

  const utils = trpc.proxy.useContext()

  // in a callback:
  utils.post.list.invalidate(/* optional input */)
  utils.post.list.prefetchQuery(/* input */)
  utils.post.list.setQueryData(/* optional input */, /* updater */)

  return <>...</>
}
```

sachinraja:
closes #2113

```ts
const Component = () => {
  const utils = trpc.proxy.useContext()
  
  const deletePostMutation = trpc.proxy.post.delete.useMutation({
    onSuccess(_, { id }) {
      await utils.post.all.invalidate()
    }
  })
}
```
I've omitted `query` and `queries` from the util method names because I think they read better without. The query key is already set for the user, so it's more like you're calling a method off the query, rather than calling a function with the query. This isn't a strong opinion though.


KATT:
Very coooool

examples/next-prisma-todomvc/src/pages/[filter].tsx
```diff
@@ -70,13 +70,12 @@ function ListItem({ task }: { task: Task }) {
 
   const editTask = trpc.useMutation('todo.edit', {
     async onMutate({ id, data }) {
-      await utils.cancelQuery(['todo.all']);
-      const allTasks = utils.getQueryData(['todo.all']);
+      await utils['todo.all'].cancel();
```
Comment (KATT) [inline examples/next-prisma-todomvc/src/pages/[filter].tsx:73]: I really think this ought to be

```suggestion
      await utils.todo.all.cancel();
```

Any reason why you opted for not doing it that way?

packages/react/src/createReactQueryHooksProxy.tsx
```diff
@@ -107,6 +115,12 @@ export function createReactQueryHooksProxy<
 >(trpc: CreateReactQueryHooks<TRouter, TSSRContext>) {
   const proxy = createProxy((opts) => {
     const args = opts.args;
+
+    if (opts.path[0] === 'useContext') {
+      const context = trpc.useContext();
+      return createReactQueryUtilsProxy(context);
+    }
```
KATT, line 122: Nice

packages/react/src/internals/utilsProxy.ts
```diff
@@ -0,0 +1,233 @@
...
+};
+
+/**
+ * @internal
+ */
+export type DecoratedProcedureUtilsRecord<TRouter extends AnyRouter> =
+  OmitNeverKeys<{
+    [TKey in keyof TRouter['_def']['record']]: TRouter['_def']['record'][TKey] extends AnyRouter
+      ? DecoratedProcedureUtilsRecord<TRouter['_def']['record'][TKey]>
+      : // utils only apply to queries
```
KATT, line 153: Oh, do they? I actually didn't know!

packages/react/src/internals/utilsProxy.ts
```diff
@@ -0,0 +1,233 @@
...
+      }
+      case 'getInfiniteData': {
+        const [input] = args;
+        const queryKey = getQueryKey(fullPath, input);
+        return context.getInfiniteQueryData(queryKey);
+      }
+    }
+  });
+  return proxy as DecoratedProcedureUtilsRecord<TRouter>;
+}
```
KATT, line 233: This file is impressive, well done 🙏 😻

KATT (commented):
Main thing is that we should probably not use the string literal :) Looks awesome

KATT:
Okay, seems to work great! Should we update `/www` too? 🤔

sachinraja:
I think I already updated the relevant docs

sachinraja:
all docs are updated now @KATT

PR #2308, opened 2022-07-24
Sponsors section for landing page

bautistaaa:
### Description
Adds sponsored section "inspired" by tanstack

image


www/src/components/sponsors/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+    children: sponsors,
+    name: 'root',
+    radius: 0,
+    distance: 0,
+  };
+
+  const root = React.useMemo(
+    () =>
+      hierarchy(pack)
+        .sum((d) => 1 + d?.monthlyPriceInDollars)
```
bautistaaa, line 18: types for `hierarchy` seem broken

verified they were also throwing errors in tanners repo.

I can `as unknown as MyType` to kill them if desired

www/src/components/sponsors/data.ts
```diff
@@ -0,0 +1,204 @@
+export const sponsors = [
```
bautistaaa, line 1: hardcoded data until we replace with job.

I didn't fully verify `individuals` in case you wanted to take this over and have the data be fed from the source. If we want to hold on that I can manually verify the whole list

Issue #2311, opened 2022-07-24
[v10] Make interop non-proxy string literals & and new proxy calls mutually exclusive


KATT:
## Problem

Currently, it's possible to use the old client APIs to call the new ways of defining routers and procedures and the other way around. This creates quite a heavy load on TypeScript performance as we flatten all procedures on the router, whether it's used or not.

Seeing that we are stepping away from string literals, this is an undesirable bottleneck & makes tRPC slower.

## Proposed solution

- Procedures defined with old way (i.e. `trpc.router().query('oldProc', { resolve: () => '..' )`) should **not** be callable through a proxy, and **only** with the legacy string literals
- Procedures defined with new way, (i.e. `t.router({ newProc: t.query(() => '..') })`), should not be callable **except through a proxy**

## Implications

- Proxy will be the only way to call things in the default tRPC clients
- People who want to support environments [which can't support `Proxy`](https://caniuse.com/#search=Proxy) will need another tRPC client *(TBD if we make one of those ourselves)*
- We need to change the way we define interop mode on the client
- We will deprecate `createTRPCClient` & `createReactQueryHooks` in favor of proxy-based APIs
- We will need to define interop clients or make the current clients automatically expose a `.proxy` object
- This **should not** make incremental migration from v9->v10 harder

## Proposed APIs

Given the following backend:

```ts
const t = initTRPC()();
const legacyRouter = trpc
  .router()
  .query('legacyProc', {
    input: z.string().optional(),
    resolve({ input }) {
      return `hello ${input ?? 'world'}`;
    },
  })
const newAppRouter = t.router({
  newProc: t.procedure.query(() => "I am just a test, I don't know! "),
});
const appRouter = t.mergeRouters(legacyRouter.interop(), newAppRouter);
```

### `@trpc/client`

**New way of creating a client:**
```ts
// Create new client
const client = createTRPCProxyClient<AppRouter>({ 
  url: 'http://localhost:3001',
})

// Usage:
client.newProc.query();
```

**Interop:**

```ts
// Will show deprecation error
const client = createTRPCClient<AppRouter>({
  url: 'http://localhost:3001',
})

// Expose a proxy prop automatically:
const proxy = client.proxy;

// ✅ These calls will work
proxy.newProc.query();
client.query('legacyProc');

// ❌ These will now show TypeScript error
proxy.legacyProc.query();
client.query('newProc');
```

### `@trpc/react`

**New default way of creating a client:**

```tsx
[...]

PR #2316, opened 2022-07-25
[idea] path proxy

mmkal:
This _isn't_ ready to be merged as-is. It's just the easiest way I could come up with for communicating an idea.

Basically, this is an alternative to the Proxy-is-love-Proxy-is-life API that's upcoming in v10 - if it were up to me, it would be a replacement, but there's no reason they can't both co-exist.

Basically, this more-or-less keeps the "v9" way of invoking clients, but instead of typing `'` then writing what _looks_ like a plain string (although of course it's type-checked), you can type `trpc.paths.` and build a string dynamically, using a similar `Proxy` method that `next` currently does for invocable functions.

Demo using an existing example:

https://user-images.githubusercontent.com/15040698/180673363-ad694102-7ec4-44f3-93e4-abd1ab707b74.mov

Implementation wise, it's done too "late" - I semi-arbitrarily put it in the `createReactQueryHooks.tsx` file, meaning it's taking a string in the format `a.b.c` (i.e. which has already been "glued together",  so has to be split apart again). This is unnecessary, but I want to show the POC to see what people thought.

Here are the benefits, in my mind:

1. Go to definition in IDEs, like the existing Proxy API
2. Easy-to-understand usage - it's closer to being "honest". `proxy.mutations.deleteById(...)` feels to me like it's _lying_ to the user, by pretending that you're actually calling a delete function. `client.query(...)`, `trpc.useQuery(...)` etc. much more closely correspond to what happens in reality. trpc now just provides a helper for constructor paths.
3. Reusable for _all_ clients, including ones not in this repo. React and Next.js will be dethroned some day, maybe even soon. This technique doesn't require a new `Proxy`-based implementation for every client. They'll all need strings!
4. Clients that can't support proxies could much more easily write a code transformer plugin which swaps these path-builder-proxies for literal strings, instead of being totally out in the cold.

Downside(s) vs the existing Proxy API:

1. Slightly more verbose. `trpc.proxy.queries.getById({ id: 1 })` vs `trpc.query(trpc.paths.getById, { id: 1 })`: 37 vs 41 (although obviously both forms can be "golfed" plenty).
2. Less magical-seeming. I think this is a good thing, but some people seem to want to make it feel like you can literally call server-side functions from the client. I guess the clue's in the name!
3. Might be killed by https://github.com/trpc/trpc/issues/2311 (?)

Also related #2270 

I wanted to open this up early to enable discussion, and because with #2311 it seems time-sensitive.

### DIFF
--- a/examples/.interop/next-prisma-starter/src/pages/index.tsx
+++ b/examples/.interop/next-prisma-starter/src/pages/index.tsx
@@ -4,7 +4,8 @@ import Link from 'next/link';
 
 const IndexPage: NextPageWithLayout = () => {
   const utils = trpc.useContext();
-  const postsQuery = trpc.useQuery(['post.all']);
+  const postsQuery = trpc.useQuery([trpc.paths.post.all]);
+  const singlePostQuery = trpc.useQuery([trpc.paths.post.byId, {id: 'foo'}])
   const addPost = trpc.useMutation('post.add', {
     async onSuccess() {
       // refetches posts after a post is added
--- a/packages/react/src/createReactQueryHooks.tsx
+++ b/packages/react/src/createReactQueryHooks.tsx
@@ -112,7 +112,7 @@ type Stringable<S> = S | { toString: () => S };
 /**
  * e.g. `StringBuilder<['a', 'b', 'c']`
  *
- * => `{ a: { b: { c: {toString: () => 'c'} } } }`
+ * => `{ a: { b: { c: {toString: () => 'a.b.c'} } } }`
  */
 type StringBuilder<
   Segments extends string[],
@@ -341,7 +341,7 @@ export function createReactQueryHooks<
   >(
     _pathAndInput: [
       path: Stringable<TPath>,
-      ...args: inferHandlerInput<TQueries[TPath]>
+      ...args: inferHandlerInput<TQueries[TPath]>,
     ],
     opts?: UseTRPCQueryOptions<
       TPath,
@@ -412,7 +412,7 @@ export function createReactQueryHooks<
   >(
     pathAndInput: [
       path: TPath,
-      ...args: inferHandlerInput<TSubscriptions[TPath]>
+      ...args: inferHandlerInput<TSubscriptions[TPath]>,
     ],
     opts: {
       enabled?: boolean;
--- a/packages/react/src/createReactQueryHooks.tsx
+++ b/packages/react/src/createReactQueryHooks.tsx
@@ -98,6 +98,73 @@ type inferProcedures<TObj extends ProcedureRecord> = {
   };
 };
 
+/** e.g. `a.b.c` => `['a', 'b', 'c'] */
+export type Split<
+  S extends string,
+  Separator extends string = '.',
+> = S extends `${infer Head}${Separator}${infer Tail}`
+  ? [Head, ...Split<Tail>]
+  : [S];
+
+/** Proxies mean we might need to explicitly call `toString()` to get an actual string */
+type Stringable<S> = S | { toString: () => S };
+
+/**
+ * e.g. `StringBuilder<['a', 'b', 'c']`
+ *
+ * => `{ a: { b: { c: {toString: () => 'c'} } } }`
+ */
+type StringBuilder<
+  Segments extends string[],
+  Value extends string = '',
+> = Segments extends []
+  ? Stringable<Value>
+  : Segments extends [string, ...infer Tail]
+  ? {
+      [K in Segments[0]]: StringBuilder<
+        Extract<Tail, string[]>,
+        Value extends '' ? Segments[0] : `${Value}.${Segments[0]}`
+      >;
+    }
+  : never;
+
+/** e.g. `{a: 1} | {b: 2}` => `{a: 1; b: 2}` (more or less) */
+type UnionToIntersection<T> = (T extends any ? (x: T) => any : never) extends (
+  x: infer R,
+) => any
+  ? R
+  : never;
+
+/**
+ * create a proxy object that can be used to build string paths dynamically, but in
+ * a strongly typed way:
+ * @example
+ * ```
+ * const paths = createStringProxy<'a.b.c' | 'd.e'>()
+ *
+ * paths.a.b.c // `{toString: () => 'a.b.c.'}` at compile time and runtime
+ * paths.d.e.zzzz // compile time error, 'zzz' not found
+ * ```
+ */
+function createStringProxy<S extends string>(
+  path: string[] = [],
+): UnionToIntersection<StringBuilder<Split<S>>> {
+  return new Proxy(
+    {},
+    {
+      get(_, key) {
+        if (typeof key !== 'string') {
+          return;
+        }
+        if (key === 'toString') {
+          return () => path.join('.');
+        }
+        return createStringProxy(path.concat([key as string]));
+      },
+    },
+  ) as any;
+}
+
 function createHookProxy(callback: (...args: [string, ...unknown[]]) => any) {
   return new Proxy({} as any, {
     get(_, path: string) {
@@ -272,7 +339,10 @@ export function createReactQueryHooks<
     TQueryFnData = TQueryValues[TPath]['output'],
     TData = TQueryValues[TPath]['output'],
   >(
-    pathAndInput: [path: TPath, ...args: inferHandlerInput<TQueries[TPath]>],
+    _pathAndInput: [
+      path: Stringable<TPath>,
+      ...args: inferHandlerInput<TQueries[TPath]>
+    ],
     opts?: UseTRPCQueryOptions<
       TPath,
       TQueryValues[TPath]['input'],
@@ -281,6 +351,10 @@ export function createReactQueryHooks<
       TError
     >,
   ): UseQueryResult<TData, TError> {
+    const pathAndInput: typeof _pathAndInput = [
+      _pathAndInput[0].toString(),
+      ..._pathAndInput.slice(1),
+    ] as any;
     const { client, ssrState, queryClient, prefetchQuery } = useContext();
 
     if (
@@ -305,7 +379,7 @@ export function createReactQueryHooks<
     TPath extends keyof TMutationValues & string,
     TContext = unknown,
   >(
-    path: TPath | [TPath],
+    path: Stringable<TPath> | [Stringable<TPath>],
     opts?: UseTRPCMutationOptions<
       TMutationValues[TPath]['input'],
       TError,
@@ -322,7 +396,7 @@ export function createReactQueryHooks<
 
     return __useMutation((input) => {
       const actualPath = Array.isArray(path) ? path[0] : path;
-      return (client.mutation as any)(actualPath, input, opts);
+      return (client.mutation as any)(actualPath.toString(), input, opts);
     }, opts);
   }
 
@@ -338,7 +412,7 @@ export function createReactQueryHooks<
   >(
     pathAndInput: [
       path: TPath,
-      ...args: inferHandlerInput<TSubscriptions[TPath]>,
+      ...args: inferHandlerInput<TSubscriptions[TPath]>
     ],
     opts: {
       enabled?: boolean;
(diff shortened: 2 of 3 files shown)


KATT:
I think the (seemingly) proper "call" in contrast to "invocation" is one of the most awe-inspiring things we have achieved with the new API. I don't really mind that it's magical; it's _a lot less_ magical than any alternative that is based on a precompile-step (i.e. blitz et al)

I am aware you have hesitated about the bet on proxies before; are there more reasons you are concerned about that is left out above?

I do understand that it raises the bar of entry for third-party packages a bit, and this is a valid point, but I think we can provide assistance where/if needed. Most things that are non-React will be able to use the raw client straight off though.


FWIW, nothing is really very "time-sensitive". tRPC v9 works and is stable & it's good to have consensus rather than ppl to forcing people to rewrite their apps _twice_. 

@TheoBr -- I would love your take on this, would you have preferred something like `useQuery([trpc.post.byId, { id: '1' }])` over `trpc.post.byId.useQuery({ id: '1' })`?

TheoBr:
This has me thinking a lot. The over-use of the `trpc` keyword is the big thing that hurts. I _slightly_ lean towards the non-proxy'd approach, but I think I'd opt to rename _almost everything_ if I adopted it. I.e...

```ts
// in utils/trpc/client.ts for client
export const {useQuery: useOurQuery, useMutation: useOurMutation, api: ourApi} = setupTRPC<....>(...);

// In some random file
const {data, isLoading} = useOurQuery(ourApi.post.byId);
```

"our" stands in for whatever you'd want to call it for your codebase, in our case would probably do something like `usePingQuery` and `pingApi`

sachinraja:
Impressive work on the types for this.

> if it were up to me, it would be a replacement, but there's no reason they can't both co-exist

I don't think we should offer two different APIs, it'll lead to some unnecessary discussion on which way to use in other places and be bad for type-checking perf.

> Easy-to-understand usage - it's closer to being "honest". proxy.mutations.deleteById(...) feels to me like it's lying to the user, by pretending that you're actually calling a delete function. client.query(...), trpc.useQuery(...) etc. much more closely correspond to what happens in reality. trpc now just provides a helper for constructor paths.

Keep in mind that for React it looks like `proxy.deleteById.useQuery()` and I think that maps pretty well to vanilla React Query in your head. IMO I think it's ok to lie to the user a bit here to offer a nice API. Using proxies doesn't make the code itself flaky.

TheoBr:
@sachinraja I prefer it for a handful of pretty subjective reasons
- Nested/"keyed" hooks (i.e. `trpc.useQuery`) always felt a bit weird
- The conflation of `trpc` the "server thing" and `trpc` the "client thing" has always made the import story a little awkward
  - the new `t` pattern helped but not enough - tbh it's what started this thought rabbit hole for me
- The path for working w/ things like optimistic updates is much simpler
  - `client.setQueryData(ourApi.friends, {friends: [...]})`
- More alignment w/ React Query (in particular their docs)

Digging the [not-too-dissimilar EdgeDB TS query builder stuff](https://www.edgedb.com/docs/clients/js/index#the-query-builder) - could be worth some 👀 for inspiration

thelinuxlich:
I like this alternative API a lot, much more intuitive!

mmkal:
@KATT @sachinraja will reply to "concerns with Proxy" and "shouldn't offer alternative API" together.

My concerns with the current Proxy API (sorry, this got long!) (I ended up using "proxy-path" and "proxy-api" to refer to the two approaches, but pls let me know if this was unclear):

#### Intuitiveness

One of the huge strengths of trpc is that it's natural and intuitive. Someone who's never seen it before can see some trpc client code and instantly "get it". I think this is what gives it the viral factor. With proxies, it's more _impressive_, but that's not necessarily a good thing. Trpc has been a great way for people to understand at a glance how a system works. What the client does vs what the server does is easy to eyeball. This makes it harder, and I worry will make both frontend and backend developers more afraid of their own codebase. Especially in a fully-typescript codebase, knowing for sure that your code isn't going to try to use a database client in the frontend is very valuable. Trpc maintainers know, of course, but I think that's because they're intimately familiar to trpc. Most people aren't, though, and they will waste time reverse-engineering trpc's implementation so they can stop worrying about importing server-side code on the client. This is what ended up happening when I happened to share a v10 snippet with a couple of people on my team.

#### If it ain't broke, don't fix it.

image

Trpc is just fine as it is! People love it. I love it. A huge change like this is also a huge risk. This isn't a good enough reason on its own, but it should be taken into consideration. My team are using trpc in production, for pretty much everything, it's going great, and other than fear of being left behind, we have no reason to _want_ to change our client code. We'll have to do it incrementally, it will take up headspace for weeks or months, and we won't really gain anything from it (the syntax change, that is - we will gain plenty from all the other improvements coming in v10).

#### Either inconsistent, or un-idiomatic syntax

In React:

If we go with `proxy.account.users.getById.useQuery(...)` - it will be the first instance I'll have ever seen of a hook that looks like that. Every other hook I've come across gives a hint to the developer _immediately_ what kind of hook it is: the difference between `trpc.useQuery(...)` vs `trpc.useMutation(...)` is instantly visible.

The alternative, `proxy.useQuery.account.users.getById(...)` is also very unusual. Every other hook is a function itself, not a parent of a function.

(This can be taken with a pinch of salt. Maybe I'm just a nay-sayer who's trying to prevent you breaking new ground! Maybe more and more react hooks will look like this some day. I don't _think_ so - trpc is already "new ground" - but maybe it could be new new ground)

In a vanilla client it's different: `client.queries.account.users.getById(...)` or `client.account.users.getById(...)`, and likely for every client for every platform it'll be subtly different in ways that will be tough to document. Some proxies return promises, some return query/mutation objects, etc. etc.

With proxy strings they're just... strings. Plug them into whatever you want. Hey presto, we have Proxy support for react-query v179, for futureframeworkthatdoesntevenexistyet, etc. etc.

#### Too big a change

Do we need to change everything about the server and everything about the client? What if we just did the server change and sat on the big client change for a while? I don't know how representative I am, but even if you don't agree with my other points here, I do think it's worth worrying about causing another "python 2 vs python 3" type situation.

#### Not solving a big problem

This huge change (Proxies) is, IMO, too strong a reaction to a small problem - missing F12 support. If I'm searching for the `getById` query I just do a find-in-files for `getById'` (with the quote) and it works. I get that that's an awkward workflow, but making such a drastic change just doesn't feel commensurate with the problem size. This PR is intended to provide an alternative, **much** less impactful way of getting pretty much the same benefits. (Other benefits like creating a client that only makes request to a subrouter can be easily achieved by explicit helper methods)

#### One-way door

Based on #2311, it seems like trpc the repo is going to need make a call - Proxy or string. If we go Proxy and it turns out it was a bad decision - as you say, we'll have to ask people to rewrite their apps twice. They won't do that - they'll churn. (And with my end-user hat on, I think asking them to rewrite their apps _once_ isn't really reasonable tbh).

With this proposal, no rewrite required. This _isn't_ really an alternative API. It's just a helper for building strongly-typed string paths (@sachinraja I guess that's my response to "not sure we should offer an alternative API"). The door to go to proxy-api is still open and wouldn't require any back-tracking.

___

The decision making process should be different now than it was when trpc was brand-new. Change should be approached really really carefully. IMO the proxy-path-builder approach can get roughly identical benefits to proxy-client without fundamentally altering the makeup of a beautiful thing. We could do this, release v10, wait for feedback, try it out for a few months, then see if there's really a need for a full-blown proxy client API.

___

@sachinraja to reply to some of your questions directly (sorry that some of this is repeating what I wrote above):

>I don't think we should offer two different APIs, it'll lead to some unnecessary discussion on which way to use in other places and be bad for type-checking perf.

This isn't really an alternative API. It's the v9 API, with a helper for building strings. We could very easily even drop it in v9 as a minor change.

>Keep in mind that for React it looks like proxy.deleteById.useQuery() and I think that maps pretty well to vanilla React Query in your head. IMO I think it's ok to lie to the user a bit here to offer a nice API. Using proxies doesn't make the code itself flaky.

`proxy.deleteById.useQuery()` isn't so bad, and I noticed that that's what most examples in this repo look like, because they're just that... examples. In real-life apps, it'll be more like `proxy.patient.accounts.user.fhir.patient.encounters.deleteById.useQuery()` which maps much less well to vanilla React query (probably a refactor opportunity to use a sub-router proxy like `proxy.patient.accounts.user.fhir.patient` or something - but in reality that won't always get done). 

>Like Alex said, we can provide assistance and from my experience working with the new proxy API for server-side callers, the vanilla client, and the React client, it's not too difficult to implement. We'll also offer as many helpers (codewise) as we can for this.

That's great, but every human being has limited bandwidth, and IMO it'd just be better to not require a bespoke implementation per client-type. The quality of some will be lower than others, apart from anything else.

>IE is EOL so we do have to think about how far we're willing to go to support it. I think it'd be good to find out how many users really care about supporting IE. We'll still have the interop client APIs for v10 which don't depend on proxies.

Fair enough - I don't think trpc maintainers should have to test on IE, but they should be empathetic to their consumers. Some trpc users will have end-users who depend on locations like public libraries, many of which rely on IE still. Again I'm not suggesting trpc explicitly support it, but proxy-path leaves the door totally open to consumers just using raw strings (trading off F12 support for IE support). Proxy-API plus #2311 means any such consumers will be on the doomed-sounding "legacy/interop" mode forever (which I imagine would be EOL'd eventually too). (Also, proxy-path would be easy to write a babel/swc plugin for which converts the `paths.foo.bar` into `'foo.bar'` in production-mode, for IE - much harder to do for proxy-api).

>>Slightly more verbose
>I think this is the most difficult downside to justify, especially seeing how this was one of the things brought up in https://github.com/trpc/trpc/discussions/2270.

Again, people **love** trpc the way it is. This is not making it more verbose, it's just adding an optional helper that people can choose to use or not. This level of engagement and love for a repo isn't normal! So I'm not too worried about what is essentially the status quo being considered problematic.

Also, there are decent options that could be bikeshedded (or left to userland) for those worried about typing `trpc.` too much:

```ts
import {trpc, paths} from '~/utils/trpc'

export default () => {
  const myQuery = trpc.useQuery(paths.foo.bar.baz)
  return <>myQuery.data</>
}
```

>Keep in mind both ways use proxies. Also can you elaborate on what makes you prefer it? Is it how it more closely mirrors vanilla React Query?

proxy-path has opt-in proxies, with a much lower surface-area, for teams that choose to bet on them. proxy-api is proxy-or-die.

alavkx:
A pretty well put together set of considerations. I’ll +1 just how excellent it was to start working with v9 trpc.useQuery. Coming from react-query there was exactly 0 new api to learn. It ONLY made my life significantly better with 0 discernible trade off. A pretty rare scenario in software. I’ve never thought that part of trpc needed much iteration, it feels almost perfect. The rough around the edges bit of the library is almost entirely in the rear, which y’all have made excellent progress on in v10. 
Certainly the folks that have worked so hard on the proxy API will be hard pressed to look past the sunk cost of their tremendous efforts, but this issue presents some proper wisdom.

mmkal:
I don't think it'd be a sunk cost. Proxies give F12 support - which this PR doesn't have, but with a little reshuffling, I think proxy-path can just borrow the exact implementation of proxy-api. The end result should be the same but the "leaf" is a `{toString: () => TPath}` type instead of a `{ query: (input: TInput) => ...; mutate: (input: TInput) => ... }` type). 99% of the hard work done on Proxies would be needed for proxy-path to be ready to go live - when I get a little more time I will try to do that reshuffle in this PR.

KATT:
I have spent a lot of time thinking about this & will give a proper write-up of what I think and stronger motivations for the API decisions taken. I truly appreciate your feedback and the time taken to do this draft PR. 🙏

KATT (commented):
Thanks so much for all the thought and work that went into this.

I've decided **not** to go further with this proposal. We've had alpha/beta for v10 for a while now and once people get used to it, no one seems to have a problem with it.

I personally really love the magical feeling of _calling functions_ and I don't want to lose that, even if I realize your proposal here is more flexible.

Feel free to give opinions over at #3042 for the last remaining open question on API design.

PR #2317, opened 2022-07-25
Return errors to links correctly

goknsh:
A fix for `TRPCError`s thrown on the server arriving in `next` instead of `error` in links.

I think this is clearer that the text above:
```ts
const customLink: TRPCLink<AppRouter> = (runtime) => {
  return ({ next, op }) => {
    return observable((observer) => {
      const unsubscribe = next(op).subscribe({
        next(value) {
          // `TRPCError`s from the server are sent here
          observer.next(value);
        },
        error(value) {
          // instead of here
          observer.error(value);
        },
        complete() {
          observer.complete();
        }
      });
      return unsubscribe;
    });
  };
};

export const client = createTRPCClient<AppRouter>({
  fetch(url, options) {
    return fetch(url, { ...options, credentials: 'include' });
  },
  transformer: superjson,
  links: [
    customLink,
    httpBatchLink({
      url: '/trpc'
    })
  ]
});

export const proxy = createTRPCClientProxy(client);
```

I'm pretty sure I got the fix working for `httpBatchLink` and `httpLink`, but not 100% sure what I did for `wsLink` actually works.


packages/client/src/links/wsLink.ts
```diff
@@ -283,7 +298,7 @@ export function wsLink<TRouter extends AnyRouter>(
     return ({ op }) => {
       return observable((observer) => {
         const { type, path, id, context } = op;
-
+        context.runtime = runtime;
```
sachinraja, line 301: is this needed?

sachinraja (commented):
Looks good, but we should really add some tests, especially if you;re unsure that it actually works.

packages/client/src/links/wsLink.ts
goknsh: That's the easiest way I found to get `runtime` into `createWSClient` without changing any of its `opts`.

KATT (commented):
- Some tests would be great indeed! Nice to see its passing though!
- It looks like it's added at the right place, but I think if it's added somewhere I think it needs to be removed somewhere else haha

KATT:
Will send you some money for the trouble if you setup github sponsors @goknsh

goknsh:
> Will send you some money for the trouble if you setup github sponsors @goknsh

Sure, I'll try and set it up. fb5dbee378794e799c2910f449f3eba3e46c9bfa adds tests for `httpLink` and `httpBatchLink` with an attempt at tests for `wsLink`. I can't figure out how to get the `wsLink` to work. The `regular` query runs before the subscription for `onQuery` has time to setup...

goknsh:
Got it figured out, @KATT I think it's ready to merge, just give the tests a once-over. Also, I thought that the `complete` function on the client subscription would get called after unsubscribing, but turns out that's not the case; i.e. I thought [this line](https://github.com/trpc/trpc/pull/2317/files#diff-4a2c6559832bd6e8afcbedfd1eca51563250bd6a6a86aa33a2c828e20e6e25b1R248) should have been `expect(subCompleteFunction).toHaveBeenCalledTimes(1);`, but that fails. Is this correct behavior?

packages/client/src/links/httpLink.ts
```diff
@@ -20,13 +21,22 @@ export function httpLink<TRouter extends AnyRouter>(
...
               context: res.meta,
-              data: res.json as any,
-            });
+              data: res.json,
+            } as OperationResult<TRouter, any>;
+            const transformed = transformOperationResult(result, runtime);
+            if (transformed.ok) {
+              observer.next({
+                context: res.meta,
+                data: res.json as any,
```
KATT, line 32:
```suggestion
                data: res.data as any,
```

This should probably return `transformed.data` but that requires a bunch of refactoring which I think is a symptom of something larger we should fix before we have a public beta. I have previously written a note about revisiting observables a bit closer. 

Currently, we have this 

```ts
type OperationResultData<TRouter extends AnyRouter, TOutput> =
  | TRPCResponse<TOutput, inferRouterError<TRouter>>
  | TRPCResponseMessage<TOutput, inferRouterError<TRouter>>;

export interface OperationResult<TRouter extends AnyRouter, TOutput> {
  data: OperationResultData<TRouter, TOutput>;
  context?: OperationContext;
}
```

& I think `data` should only be the success state here.


I don't have any time to code in the next few days as I'm traveling & going to a wedding tomorrow, but maybe @sachinraja or @jlalmes can help out

This is a bit of a remain from how the behavior was previous to #1855 and we missed doing this refactor.

goknsh: Regarding the suggested change, there is no `res.data`, only `res.json` and `res.meta`. Also, I had the same concern as you in the Discord thread, thanks for mentioning it here as well. And I agree that having consistency of returning transformed data would be better here. I think that will be a bit too much for me to tackle, so help is appreciated.

KATT: Yeah, I think only me, @sachinraja, or @jlalmes are the right ppl for this job 😅

KATT:
This should now be fixed by #2402

PR #2320, opened 2022-07-25
specify parameter names for React hooks

### LINKED ISSUE #2315: [V10] make type inlays for queries and mutations more descriptive
When entering the `input` object and then `options` object into a `useQuery` or `useMutation` the type inlays are not particularly helpful. 

image

It would be better if these were 
- `input` instead of `args_0`
- `options` or something similar, instead of `args_1`

This in particular would help with the case of when providing `undefined` for a method that does not have an `input` but you do want to use `options`. 

p.s. There may be other areas where these inlays could be improved?...

sachinraja:
closes #2315


KATT (approved):
🚀

JonParton:
Awesome - I'll keep an eye out for a new release on NPM and give it a whirl! Thanks @sachinraja ! 💖🚀

PR #2322, opened 2022-07-25
feat(www): implements twitter wall of love

ixahmedxi:


www/src/components/TwitterWall/tweets.ts
```diff
@@ -0,0 +1,98 @@
+export const tweets = [
+  [
+    {
+      id: 1,
```
sachinraja, line 4: What is the `id` for here? If it's to provide a unique `key`, `handle` should be unique unless we have multiple tweets from the same author (which seems odd to have on a tweet wall)

KATT: I think this needs to be the actual id of the tweet so we can link it

KATT (commented):
This looks awesome! Here's some feedback:


- Retain link to tweet behavior (add `<a` and _maybe_ `onClick`):
  - Analyze click/hover here: https://www-git-tweet-wall-trpc.vercel.app
  - Clicking avatar/handle should link to their profile
  - Clicking the tweet as a whole should open the twwet

PR #2326, opened 2022-07-26
Fix support for `Set-Cookie` header in fetch adapter

### LINKED ISSUE #2312: Cannot set `Set-Cookie` response header
I'm working with implementing authentication using `trpc`. I use the `fetch` adapter. I need to set the `Set-Cookie` response header. However in trying so I discovered that `fetch`'s `Response` objects do not allow setting this header, as per the spec [here](https://fetch.spec.whatwg.org/#forbidden-response-header-name).

> A forbidden response-header name is a [header name](https://fetch.spec.whatwg.org/#header-name) that is a [byte-case-insensitive](https://infra.spec.whatwg.org/#byte-case-insensitive) match for one of:
>
> - `Set-Cookie`
> - `Set-Cookie2`

After some debugging, I was able to come to the conclusion that the problem lies with the `res.headers.set(...)` call on line `52`. The `Headers` object's implementation silently ignores calls to this (and `.append` I think) with the `key` set to one of the above alternative from the spec.

https://github.com/trpc/trpc/blob/2af7a0070f3bc01b81fe6f14f591e77c549520a7/packages/server/src/adapters/fetch/fetchRequestHandler.ts#L42-L59

Fortunately there is a potential solution in my eyes. This would be to extract the header keys and values from `result.headers` before constructing the `Response` object. Passing the new `headers` record into the `headers` option.

```ts
const headers = {};

// collect headers...

const res= new Response(result.body, {
  status: result.status,
  headers,
});
```

I am a little worried this solution might cause problems considering I don't know if this approach would bypass all of `Headers` addition checks. Maybe if this is the case, only extract and inject the "forbidden response-header names" before constructing the `Response` object and continue using the current approach for all rest.

If this behaviour is intentional, or if you require more information please let me know.

Thanks.

### Info

- `trpc`: `10.0.0-alpha.40`

emme1444:
Fix support for `Set-Cookie` header in fetch adapter.

Closes #2312


emme1444:
Huh? I screwed that up somehow?

emme1444:
Nevermind I know what I did wrong. WIll fix

PR #2329, opened 2022-07-26
docs: Fix Github Star on some i18n

juliusmarminge:
Discussed on Discord: https://discord.com/channels/867764511159091230/1001593966237585459

On some i18n systems the `react-animated-numbers` has a bug and renders the wrong number.

CleanShot 2022-07-27 at 00 06 05

I cut an issue out for this: https://github.com/heyman333/react-animated-numbers/issues/34, but until they fixed that we should probably disable the includeComma flag since that seem to be the culprit both from debugging and from reading the internal source of react-animated-numbers.

CleanShot 2022-07-27 at 00 08 54

---

_Note: I also removed the BrowserOnly stuff, didn't seem to do anything on my side and it is working as expected on my end without it (at least locally, we'll see when it deploys...)_

### DIFF
--- a/www/package.json
+++ b/www/package.json
@@ -22,7 +22,6 @@
     "@visx/responsive": "^2.10.0",
     "clsx": "^1.2.1",
     "react": "^18.2.0",
-    "react-animated-numbers": "^0.13.0",
     "react-dom": "^18.2.0",
     "react-github-btn": "^1.3.0",
     "react-icons": "^4.4.0",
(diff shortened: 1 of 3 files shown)


juliusmarminge:
Filed a PR (https://github.com/heyman333/react-animated-numbers/pull/35) to fix this on their end, guess we'll see what happens there

ixahmedxi (approved):
looks good to me!

KATT (commented):
My suggestion would be the following:

1. Delete the AnimatedNumbers and it's dependency
1. Only have the "reveal" animation for now
1. Use [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) instead of animated numbers
1. If we want, introduce some cool animation

PR #2331, opened 2022-07-27
feat(www): tweak twitterwall and link stuff

juliusmarminge:
Implements tweaks suggested in #2322

- [x] Clicking the tweet links to the tweet
- [x] Clicking the profile-part of the tweet links to the profile

Seems like there is some weird link behavior (children of link get underlined) set in some config so just gotta fix that

### DIFF
--- a/www/src/components/TwitterWall/script.ts
+++ b/www/src/components/TwitterWall/script.ts
@@ -8,13 +8,20 @@ import fetch from 'node-fetch';
  * Script has shitty typesafety but hey, it works :)
  */
 
+const {
+  OAUTH_CONSUMER_KEY,
+  OAUTH_NONCE,
+  OAUTH_SIGNATURE,
+  OAUTH_TIMESTAMP,
+  OAUTH_TOKEN,
+} = process.env;
+
 const url =
   'https://api.twitter.com/1.1/collections/entries.json?id=custom-1441435105910796291';
 const options = {
   headers: {
     accept: '*/*',
-    Authorization:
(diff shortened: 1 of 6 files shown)


ixahmedxi:
Tweets that have `&` render as `&amp;` on the website, can we somehow parse that text?

www/src/components/TwitterWall/tweets.ts
```diff
@@ -1,98 +1,213 @@
...
+    },
+    {
+      id: '1548607105934004227',
+      name: 'Ben 🏳️‍🌈',
+      handle: 'nurodev',
+      date: 'Jul 17',
+      avatar:
+        'https://pbs.twimg.com/profile_images/1542996607083462659/sTlPB-kO_normal.jpg',
+      content:
+        '@alexdotjs @trpcio @github Amazing &amp; much well deserved!\nPeople are starting to see how awesome tRPC is &amp; I love it! 👏',
```
juliusmarminge, line 62: Interesting. That is actually coming from the API🤔

juliusmarminge (commented):
Can look into this tonight

www/src/components/TwitterWall/script.ts
```diff
@@ -0,0 +1,84 @@
...
+    const userId = tweetData.user.id_str;
+
+    const content = tweetData.text;
+
+    // this removes the trailing line at the end of the tweet's content
+    const contentWithoutTrailingLink = content.replace(
+      /\b(http(s|):\/\/.+)\W*$/,
+      '',
+    );
+
```
juliusmarminge, line 46:
```suggestion
const escapedContent = content
    .replace(/&amp;/, '&')
    .replace(/&lt;/, '<')
    .replace(/&gt;/, '>')
    .replace(/&quot;/, '\"')
    .replace(/&#39;/, '\'');      

```

@ixahmedxi too hacky?

juliusmarminge:
Not sure if we need to un-escape more than the ones I did now (like all the hundreds of special characters. if so we should probably find a better more scalable solution)? 

Only problem is if the tweet contains a link in the middle (which twitter shortens to their domain btw), that doesn't render as a link:

CleanShot 2022-07-28 at 15 57 00

KATT (approved):

KATT:
Awesome! Thanks so much for this.

For the next iteration, I think a good idea would be to

- Feature specific tweets (probably the ones in the tweetsOld) and then be able to just use those, otherwise, we gotta keep this up-to-date and re-run the script all the time :)
- Unsure if we should have it dynamic at all based on the twitter love, but we should definitely link it in the bottom

PR #2332, opened 2022-07-27
Support Set-Cookie header in fetch adapter (3)

### LINKED ISSUE #2312: Cannot set `Set-Cookie` response header

emme1444:
Fix support for Set-Cookie header in fetch adapter.

Closes #2312

@KATT

### DIFF
--- a/packages/server/src/adapters/fetch/fetchRequestHandler.ts
+++ b/packages/server/src/adapters/fetch/fetchRequestHandler.ts
@@ -39,8 +39,31 @@ export async function fetchRequestHandler<TRouter extends AnyRouter>(
     },
   });
 
+  const headers: Record<string, string> = {};
+  if ('Set-Cookie' in (result.headers ?? {})) {
+    const value = result.headers?.['Set-Cookie'];
+
+    switch (typeof value) {
+      case 'undefined':
+        break;
+      case 'string':
+        headers['Set-Cookie'] = value;
+        break;
+      default:
+        if (value.length > 0) {
+          const firstValue = value[0];
+
+          if (typeof firstValue === 'string') {
+            headers['Set-Cookie'] = firstValue;
+          }
+        }
+        break;
+    }
+  }
+
   const res = new Response(result.body, {
     status: result.status,
+    headers,
   });
 
   for (const [key, value] of Object.entries(result.headers ?? {})) {


sachinraja:
does this work?

emme1444:
@sachinraja from my tests this works. I can't speak for if this is unstable as in a bug or something though. It is a little weird. From what I understand the problem is using the `fetch` interfaces on the server. It seems they were really only meant for the client. Maybe, or something...

The only, annoying, limitation I couldn't get working, is that not more than one cookie can be set per response. This is, I'm guessing, because the `Response` object isn't handling the `Set-Cookie` header differently. Since, it wasn't made for it, I guess? It will just concatenate multiple values with a comma. Which is not what the browser expects for multiple `Set-Cookies`. As I understand it. I'm not an expert.

KATT (changes requested):
Needs tests, see https://github.com/trpc/trpc/blob/next/packages/server/test/interop/adapters/fetch.test.ts

You might want to do the client request raw, without the tRPC client, as the trpc client's query function don't currently expose the returned headers

KATT:
Closing as I haven't heard back on this for over a month, feel free to re-open if this functionality is needed.

PR #2345, opened 2022-08-01
examples: Add a more minimal express example

juliusmarminge:
Related discussion: #2325 

As pointed out in that discussion, there isn't really a minimal example that "follows the docs" (For usage with Next there is #2169). This implements just that:
`A simple router without any of the fancy functions exposed by a single express endpoint and consumed by the vanilla trpc client.`


KATT (approved):
Kung 👑

PR #2350, opened 2022-08-02
Update caching.md

colinhacks:
Tiny fixes. 

I still think the content column is in desperate need of a max-width - it would dramatically improve readability imo. I added one at some point but it was rolled back.


KATT (approved):
Thanks!

PR #2351, opened 2022-08-02
Content types

### LINKED ISSUE #1937: Support for additional `Content-Type`s
If we go down this route what we're really talking about is straying away from JSON as the defacto mode of transport. If we do this we should also have support for some sort of binary format when doing `POST`s that could take the compression further and let us have something else than `content-type: application/json`.

Furthermore, I think the server should be able to receive different types of request types and be informed on what to do. I have thought about this a bit before in the context of transformers, so outlining below.

## Some ideas for communicating transformers/serializer

> Rough ideas below, I think best by writing code, but don't see any of this as a final design.

### `contentType`-param

Goals:

- Allow for multiple content-types
- Allow for deserializing from a `Response` with a `body`
- Allow for serializing into a `Fetch.body`
- Allow for serializing into a string for query params

**When creating a server:**

```ts
import { jsonContentType, createContentType } from '@trpc/server/content-type';

// JSON crush could be a community plugin that looks something like this:
import jsoncrush from 'jsoncrush';
const crush = createContentType({
  key: 'crush',
  contentType: 'application/json',
  // a smarter content type would use something that could be streamed, i.e ReadableStream from `Response`
  // used for `POST`:
  fromResponse: async (res) => jsoncrush.uncrush(await res.text()),
  fromString: (str) => jsoncrush.uncrush(str),
  toBody: data => jsoncrush.crush(data), // this could allow for the types that the `Fetch API` allows - ArrayBuffer, string, [...]
  // used for `GET`
  toString: data => jsoncrush.data(data),
})

const trpc = initTRPC({
  contentTypes: [
    jsonContentType,
    crush,
  }
});
```

**HTTP level**

```
// inform server that we are sending `crush` content
GET http://localhost?content=crush&input=[...]
GET http://localhost?content=json&input=[...]

POST http://localhost?content=crush
content-type: "application/json"
// [..]
```

### Similarly, `transformer`-param could be used

```ts
import { defaultTransform } from '@trpc/server/transform';
// JSON crush could be a community plugin
import { crush } from 'trpc-json-crush';

const trpc = initTRPC({
  transforms: [
      defaultTransform, // our "no-op" trnasform
      { key: "superjson", value: superjson },
   ],
});
```

### The client story

Maybe only needs to allow for one content type.

```ts
import { crush } from "~/shared/contentType"

const client = createTRPCClient({
  links: [ /* ... */ ],
  contentType: crush,
});

```

_Originally posted by @KATT in https://github.com/trpc/trpc/discussions/1936#discussioncomment-2818345_

sachinraja:
closes #1937

In its current state, **this PR is a breaking change**. I completely removed support for `transformer` because I don't think maintaining interoperability is useful here. It comes at a high code complexity with little gain for users. The majority of users can just switch to use the `superjsonContentType` and we won't have to support two ways of transforming the data everywhere.

```ts
import { superjsonContentType } from '@trpc/server/content-type/superjson'

// router.ts
const t = initTRPC()({
  contentTypes: [superjsonContentType]
})

// client.ts
const client = createTRPCClient({ contentType: superjsonContentType })
```

Replace `transformer` with `contentType` *everywhere*. If you don't pass in a content type, the default `jsonContentType` is used.


KATT (commented):
This is some impressive stuff! Some thoughts in random order below:

- Not sure I like having a breaking change - I was pretty happy with forcing users to only add `.interop()` as part of v10. 
- The client should be able to support more than 1 content-type for this to be useful
- One of the key goals is to allow file uploads in a more sane way where a mutation resolver could receive a stream of the content type and for instance be able to to `.pipe()` that to an s3 bucket.
- Another nice thing would be if we could somehow enable multipart form data to be sent so we could handle forms in SSR and with one field being a file that is streamed to the resolver
- I'm pretty sure there are cases where we need transformers as-is. One case I can think of is `wsLink` where we encode/decode each message
- I'm thinking we might to do a more in depth RFC, starting from the usage of what we want to achieve from the server & client perspectives and work backwards from there

jlalmes:
Top work @sachinraja - this simplifies lots of the existing code which is always great! I think this PR actually has created so many open-ended questions for me. We first need a list of all use-cases first then we can work to establish an agreed API design. 

> Not sure I like having a breaking change - I was pretty happy with forcing users to only add .interop() as part of v10.

(kinda unrelated, but) I've raised this question on Discord - are we at the point where benefits of breaking changes outweigh the complexity of backward-compat?

> The client should be able to support more than 1 content-type for this to be useful.

What would this API look like? (I don't like this proposal - just an example).

```ts
// server.ts
const t = initTRPC()({
  contentTypes: [
    superjsonContentType, 
    messagePackContentType,
  ],
});

t.router({
  getUserById: t.procedure
    .input(z.string())
    .query(({ input }) => {
      const user = await db.users.getById(input)
      return user;
    }),
});

// client.ts

...setup `trpc` with contentTypes

trpc.useQuery('getUserById', 'user-id', { contentType: 'superjson' }) // with superjson content-type
trpc.useQuery('getUserById', 'user-id', { contentType: 'messagepack' }) // with messagepack content-type
```

Additionally, should we be able to specify a different content-type for the request & response payloads.

> One of the key goals is to allow file uploads in a more sane way where a mutation resolver could receive a stream of the content type and for instance be able to to .pipe() that to an s3 bucket.

Again, I've got so many questions regarding what this API could/should look like?

```ts
// server.ts
const t = initTRPC()({
  contentTypes: [
    jsonContentType, 
    octectStreamContentType,
  ],
});

t.router({
  uploadVideo: t.procedure
    .input(z.stream()) // ??
});
```

> I'm thinking we might to do a more in depth RFC, starting from the usage of what we want to achieve from the server & client perspectives and work backwards from there

Think this is the best plan of action moving forward.

juls0730:
Amazing work, although having it be a breaking change is probably one of my only problems with it. Instead of removing the `transformer` attribute and just assuming we can use the superjson content type, why not just have the options of content type be json and multipart-form and custom content types and if its not Json the transformer attribute just does nothing?

sachinraja:
RFC at https://github.com/trpc/trpc/issues/2387

KATT:
Closing this for now as it likely won't be picked up very soon, but it got the icebox label :)

PR #2365, opened 2022-08-03
fix(doc): Change env for vercel url for nextjs

Kubessandra:
In the frontend the env is provided by VERCEL with `NEXT_PUBLIC_`

Source: https://vercel.com/docs/concepts/projects/environment-variables#system-environment-variables


sachinraja:
It returns a blank string for a relative url on the client (hence the `typeof window !== 'undefined'` check). The checks after that are only for the server.

PR #2376, opened 2022-08-05
[WIP] simplify rollup and project structure

### LINKED ISSUE #2108: Simplify rollup config and submodules

juliusmarminge:
Aims to remove barrel files and improve rollup config. Would also like to remove need to specify `exports` in `package.json`.

Post suggestions here.

Closes #2108


KATT (commented):
I love this direction

rollup.config.ts
```diff
@@ -0,0 +1,156 @@
...
+
+function esm({ external, input, packageDir }: Options): RollupOptions {
+  return {
+    external,
+    input,
+    output: {
+      format: 'esm',
+      preserveModules: true,
+      exports: 'named',
+      dir: `${packageDir}/dist/esm`,
```
sachinraja, line 111: `esm` files should have a `.mjs` extension. With this, we can emit all of the files into the same `dist` directory instead of `dist/esm`, `dist/cjs`, and `dist/types`.

packages/client/links/httpBatchLink/index.d.ts
```diff
@@ -1 +0,0 @@
-export * from '../../dist/./links/httpBatchLink';
```
sachinraja, line 1: Unfortunately we need all these files that copy the structure of our `exports` because some bundlers (like React Native's Metro) do not support the `exports` field.

sachinraja (commented):
Awesome work! I know it's still draft, but just left a few comments. Generally though I love the config hoisting, one config makes more sense than each package having an individual config file.

rollup.config.ts
```diff
@@ -0,0 +1,156 @@
+import { babel } from '@rollup/plugin-babel';
+import nodeResolve from '@rollup/plugin-node-resolve';
+import path from 'path';
+import { RollupOptions } from 'rollup';
+import dts from 'rollup-plugin-dts';
```
sachinraja, line 5: I would've used `rollup-plugin-dts` for our config before, but it broke on some types. I love what it does and it might work now but I'm reluctant to use it in tRPC because of how crazy the types can get.

juliusmarminge:
> Awesome work! I know it's still draft, but just left a few comments. Generally though I love the config hoisting, one config makes more sense than each package having an individual config file.

Thanks, new to this whole bundle-stuff but learning as I go :)

Regarding the barrel-files, I believe there should be a way to get the imports working through the dist-folder (which keeps the same structure as src) without explicitly having barrel-files in root?

sachinraja:
> I believe there should be a way to get the imports working through the dist-folder 

Unfortunately, I don't think so. This is only possible if you have one, `main` entry point, but we have a lot of entry points. What we could do is generate those files based on our inputs though.

rollup.config.ts
juliusmarminge: Oh didnt know. Will look into that and make sure its correct 👍

juliusmarminge:
> > I believe there should be a way to get the imports working through the dist-folder 
> 
> 
> 
> Unfortunately, I don't think so but if you know of a way that'd be amazing. I think this is only possible if you have one, `main` entry point, but we have a lot of entry points. What we could do is generate those files based on our inputs though. 

There are definetely ways, e.g. you can do some prepublish stuff and publish from `dist`, dont know if thats any "cleaner" though. Will look into this further and see what alternatives exists.

juliusmarminge:
Moved to #2385

PR #2385, opened 2022-08-06
simplify rollup setup

### LINKED ISSUE #2108: Simplify rollup config and submodules

juliusmarminge:
Redo of #2376 

Took two steps back for now and only simplified the rollup config.

Also rewrote the entrypoints script. It's now way simpler and works for all packages. It syncs the `exports` field in the package.json and creates the necessary barrel-files.

Personally, I wouldn't mind adding the barrels to a .gitignore so that the repo is cleaner to browse? (yes, I go to the barrelfile instead of the sourcefile way too often...)

Closes #2108


packages/server/adapters/aws-lambda/index.d.ts
```diff
@@ -1 +1 @@
-export * from '../../dist/adapters/aws-lambda';
+export * from '../../dist/adapters/aws-lambda/index';
```
juliusmarminge, line 1: could escape the index part but it doesn't really hurt so keeping the script as minimal as possible

juliusmarminge:
It should be possible to automate this even further, and use the script for the exports field in package.json too from the inputs in the rollup config.

packages/client/README.md
```diff
@@ -1,60 +0,0 @@
...
-  // Querying the greeting
-  const helloResponse = await proxy.greeting.query({
-    name: 'world',
-  });
-
-  console.log('helloResponse', helloResponse); // Hello world
-}
-
-main();
-```
```
KATT, line 60: Is this intended..?

KATT (commented):
Questions & comments!

- Overall looks awesome and makes the whole thing of adding a new entrypoint a lot less error prone
- Happy to move barell files to gitignore
- Will `yarn dev` just work or does it require a `build` first`? If so, please make the `postinstall` script make it "just work"
- How come `README.md` and `LICENSE` files were removed?
- @sachinraja WDYT?

juliusmarminge:
> * Will `yarn dev` just work or does it require a `build` first`? If so, please make the `postinstall` script make it "just work"

Will look into the scripts tomorrow. I believe there are a few that needs modification.

> * How come `README.md` and `LICENSE` files were removed?

Sorry bout that. Removed some stuff as I got things working. Will restore those 👍🏼

juliusmarminge:
Just checked and following the contributing guide works, all you need is a `yarn` command which will install deps and do initial build and codegen, then `yarn dev` will startup watch-mode 👍🏼 

@sachinraja one thing I don't really understand is it includes `test` and `examples` on build, even though they are excluded in the tsconfigs, any idea why?
CleanShot 2022-08-09 at 13 35 22@2x

juliusmarminge:
Whoops, looks like the CI didn't like moving jest-config to root (the examples are picking it up it seems like. Should I move it back or modify it so that only the packages uses it? Even though all tests are in server right now, to me, it feels more intuitive to have a single config for all packages if we were to add tests to the other ones as well down the road.

juliusmarminge:
Rewrote the script again, now it is fully automated.

Put a new entry in the rollup.config.ts INPUTS, and the barrelfile will be created, the `exports` defined in package.json as well as the "top-level-barrel-directory" will be added to `files`.

.gitignore
```diff
@@ -12,4 +12,13 @@ yalc.lock
...
+*.tsbuildinfo
+
+# Barrel files
+packages/client/links
+packages/react/ssg
+packages/server/adapters
+packages/server/observable
+packages/server/rpc
+packages/server/shared
+packages/server/subscription
```
sachinraja, line 24: I imagine these will get annoying to ignore as it's not automated, I'm fine with just committing them anyway.

package.json
```diff
@@ -37,23 +37,26 @@
   "scripts": {
     "lerna": "lerna",
     "start": "yarn watch",
+    "clean:packages": "lerna run clean --scope=@trpc/*",
+    "clean:examples": "lerna run clean --scope=@examples/*",
+    "clean:root": "rm -rf node_modules",
+    "clean": "run-p clean:*",
```
sachinraja, line 43:
```suggestion
    "clean:packages": "rm -rf **/node_modules **/dist",
    "clean:examples": "lerna run clean --scope=@examples/*",
    "clean": "run-p clean:*",
```
This way we won't need the `clean` scripts in each `package.json` and also won't need to have `node_modules` in each of the clean scripts for the examples.

packages/client/README.md
```diff
@@ -1,60 +1,52 @@
 <p align="center">
```
sachinraja, line 1: Can we revert the changes here? Unsure what the large diff is for.

packages/client/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
sachinraja, line 1: Should this be here?

packages/next/LICENSE
```diff
@@ -1,21 +1,21 @@
 MIT License
```
sachinraja, line 1: We should probably delete all these individual licenses and just have one in the root.

package.json
```diff
@@ -80,11 +83,14 @@
     "@changesets/cli": "^2.17.0",
     "@manypkg/cli": "^0.19.0",
     "@rollup/plugin-babel": "^5.3.0",
+    "@rollup/plugin-multi-entry": "^4.1.0",
     "@rollup/plugin-node-resolve": "^13.1.3",
+    "@rollup/plugin-typescript": "^8.3.4",
```
sachinraja, line 88: Since we're using `rollup-plugin-typescript2`, we should be able to uninstall this.

rollup.config.ts
```diff
@@ -0,0 +1,155 @@
...
+const extensions = ['.ts', '.tsx'];
+
+const babelPlugin = babel({
+  babelHelpers: 'runtime',
+  extensions,
+});
+
+// Exporting this for generating barrel-files in scripts/entrypoints.ts
+export const PACKAGES = ['server', 'client', 'react', 'next'] as const;
+export const INPUTS: Record<typeof PACKAGES[number], string[]> = {
```
sachinraja, line 24: Love that this is strictly typed :smile:

rollup.config.ts
```diff
@@ -0,0 +1,155 @@
...
+  return [types(options), esm(options), cjs(options)];
+}
+
+function types({ input, packageDir }: Options): RollupOptions {
+  return {
+    input,
+    output: {
+      dir: `${packageDir}/dist`,
+      entryFileNames: '[name].ts.js',
+      chunkFileNames: '[name]-[hash].ts.js',
```
sachinraja, line 95: Just as a note if you're confused as to why I had this here, it's because `rollup-plugin-typescript2` produces JS files that'll overwrite the files from the ESM build... If you have a better idea or know of a better rollup plugin/TS integration that would be awesome.

rollup.config.ts
```diff
@@ -0,0 +1,155 @@
...
+      externals({
+        packagePath: path.resolve(packageDir, 'package.json'),
+      }),
+      nodeResolve({
+        extensions,
+      }),
+      babelPlugin,
+    ],
+  };
+}
```
sachinraja, line 155: Nice!

scripts/entrypoints.ts
```diff
@@ -1,71 +1,103 @@
...
 
-function getEntrypoint(filename: string) {
-  const parts = filename.split('/');
-  const name = parts.pop();
+// create directories on the way if they dont exist
+function writeFileSyncRecursive(filePath: string, content: string) {
+  const dir = path.dirname(filePath);
+  if (!fs.existsSync(dir)) {
+    fs.mkdirSync(dir, { recursive: true });
+  }
```
sachinraja, line 18:
```suggestion
fs.mkdirSync(dir, { recursive: true, force: true });
```
nit

sachinraja (changes requested):
> it feels more intuitive to have a single config for all packages if we were to add tests to the other ones as well down the road.

I agree, but let's leave it in `packages/server` for this PR at least.

Nice job on this! This makes changing entry points so much easier.

.gitignore
juliusmarminge: Yea I initially wanted a glob pattern here to say like, ignore subdirectory of `packages/*` but `src` and `test` but couldn't get that to work. You know what pattern to use?

package.json
juliusmarminge: Yep I believe I had something like that earlier. However, I quite like having a clean command for everyone so that when I am working on something and want to make sure nothing is "left stale", I run a `clean` for that project only. Can definitely change to your suggestion if you don't like this.

packages/client/README.md
juliusmarminge: Absolutely, not sure why it's diffing entire file, the content is the exact same...

packages/client/LICENSE
juliusmarminge: There were `LICENSE` in some of the packages earlier so I just added it to all of them. I also think it's enough to have a single one in root if @KATT is fine with that too?

scripts/entrypoints.ts
sachinraja: nevermind, apparently I made this option up, it doesn't exist :laughing:

rollup.config.ts
juliusmarminge: Yea we need to overwrite the defaults for these fields so that they don't overwrite eachother... tried to get the typescript one to only emit the declaration files but that didn't seem to do it. Can take a second look and see if there isn't a way 👍

package.json
juliusmarminge: Actually, I think we do need this. This plugin is the one which allows us to have a typescript config.

rollup.config.ts
juliusmarminge: > tried to get the typescript one to only emit the declaration files but that didn't seem to do it

Looks like it did do it... no more empty files :)

We do need these options for the ESM output though, could remove for the CJS since that is using the defaults but better to be explicit about the differences there imo.

package.json
sachinraja: Ah I see

.gitignore
```diff
@@ -12,4 +12,10 @@ yalc.lock
...
 
-*.tsbuildinfo
\ No newline at end of file
+*.tsbuildinfo
+
+# Barrel files
+packages/*/*
+!packages/*/package.json
+!packages/*/src
+!pacakges/*/README.md
```
KATT, line 21:
```suggestion
!packages/*/README.md
```

KATT: Could this work?

```suggestion
# Barrel files
## Ignore all folders
packages/*/*
## Don't ignore files
!packages/*/*.*
## Don't ignore `src` folder
!packages/*/src
```

examples/cloudflare-workers/package.json
```diff
@@ -5,6 +5,7 @@
   "type": "module",
   "scripts": {
     "dev": "wrangler dev",
+    "clean": "rm -rf dist node_modules",
```
KATT, line 8: Should we get rid of these clean scripts? They only work on windows and I don't know why we need them


Instead, we should have a `clean` file in the root that does this if/when it's needed

juliusmarminge: Could absolutely remove them if you don't like them. I like having clean scripts if I want to test something from a clean sleight.

.gitignore
juliusmarminge: Yep that works. Also included `test` which was missing 👍🏼

examples/cloudflare-workers/package.json
KATT: I like having them too but the only example I really care about being representative of a "production environment" is the `next-prisma-starter`, the otherwise shouldn't contain unnecessary noise

juliusmarminge: Alright, I'll remove them

sachinraja (approved):

PR #2394, opened 2022-08-08
remove double declarations of `useSubscription` tests

KATT:
Delete double declaration of `useSubscription` test


jlalmes:
Whoops - thanks @KATT

PR #2403, opened 2022-08-08
feat(www): restyles navbar

ixahmedxi:
This PR does the following changes:

- restricts the homepage navbar to the same width as the content and as per new design specs.
- Changes the background color of the application and navbar to be the same and to be the same as the new design specs.

## Old navbar
CleanShot 2022-08-08 at 22 28 20@2x

## New navbar
CleanShot 2022-08-08 at 22 29 33@2x


KATT (commented):
I like the width restriction but I want to be able to search straight away when I go to the site. Believe it or not, but I often go into trpc.io and search for something myself :D

KATT (approved):

PR #2408, opened 2022-08-09
restrict old and new procedures to their respective clients

### LINKED ISSUE #2311: [v10] Make interop non-proxy string literals & and new proxy calls mutually exclusive

sachinraja:
closes #2311

## 🎯 Changes
- procedures defined using the new API are only callable using the new proxy client API
- procedures defined using the old API are only callable using the old client API
- new proxy methods are exposed for the client - `createTRPCProxyClient` and `createTRPCReact`

## ✅ Checklist

- [x] I have followed the steps listed in the [Contributing guide](/CONTRIBUTING.md).


packages/server/src/deprecated/interop.ts
```diff
@@ -58,11 +58,15 @@ type MigrateProcedure<
   TProcedure extends AnyOldProcedure,
   TType extends ProcedureType,
 > = TType extends 'query'
-  ? QueryProcedure<convertProcedureParams<TConfig, TProcedure>>
+  ? QueryProcedure<convertProcedureParams<TConfig, TProcedure>> & { _old: true }
```
KATT, line 61: good shout

PR #2421, opened 2022-08-11
add minimal example

Crazytieguy:
See https://github.com/trpc/trpc/pull/1948


examples/minimal/client/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@examples/minimal-client",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@trpc/client": "^10.0.0-alpha.45",
+    "node-fetch": "^2.6.7"
```
KATT, line 7: We can get rid of this by using node 18

KATT (commented):
We can make it even more minimal ☺

Crazytieguy:
@KATT Looks like there's a failing test that's not related to my code.
Regarding dependencies - I can downgrade wait-port, but I'm not sure it's a good idea to downgrade @types/node since the example uses node 18. Is there a way to bypass this lint?

Crazytieguy:
@KATT So should I roll back the node-18 change or is there a way to bypass the lint?

Crazytieguy:
@sachinraja What do you think - should I roll back to node 17 or is there a way to bypass the 'common range' lint?

sachinraja:
Ah sorry will look at this sometime today.

sachinraja:
Let's just rollback for now @Crazytieguy

Crazytieguy:
Well, it looks like now the common range is 18... changing back to 18

Crazytieguy:
@sachinraja Done 😃

sachinraja (approved):

sachinraja (approved):

KATT (approved):

PR #2422, opened 2022-08-11
switch to swc and fix bundle warnings

sachinraja:
## 🎯 Changes

- switched from Babel to SWC, simplifying config and speeding up a few things
- changed the target version to `es2020`, which is supported by Node v14
  - we don't need any helpers with this change
- switched `ts-node` to use SWC
- I imagine in the future we could also use `@swc/jest` instead of `ts-jest` (tests are excruciatingly slow right now because they're doing typechecking too)
- fixed the new warnings that Rollup was emitting
- remove Terser, we were never using it and I don't think it's our job to run a minifier anyway


tsconfig.json
```diff
@@ -14,6 +14,7 @@
     }
   },
   "ts-node": {
+    "swc": true,
```
KATT, line 17: Nice, been meaning to do this

KATT (approved):

package.json
```diff
@@ -46,7 +46,7 @@
     "codegen:generateMergeRouters": "ts-node scripts/generateMergeRouters.ts",
     "codegen": "run-p codegen:*",
     "postinstall": "yarn codegen",
-    "build": "yarn codegen:entrypoints && rollup --config rollup.config.ts --configPlugin typescript",
+    "build": "yarn codegen:entrypoints && rollup --config rollup.config.ts --configPlugin rollup-plugin-swc3",
```
juliusmarminge, line 49: I noticed this got rid of the rough cold starts when building. Nice!

juliusmarminge (approved):

juliusmarminge:
> I imagine in the future we could also use @swc/jest instead of ts-jest (tests are excruciatingly slow right now because they're doing typechecking too)

Yea this is also causing some flaky tests failing in CI cause they aren't finished in time, could override the timeout but probably not necessary if using swc instead.

package.json
sachinraja: Yeah the `typescript` config plugin was type checking the whole codebase

PR #2425, opened 2022-08-11
feat: Move client context to trpc object

### LINKED ISSUE #2395: TanStack query `context` versus tRPC's `context`
From @dyaa in https://github.com/trpc/trpc/issues/2352#issuecomment-1208122411

> It seems like we'll need to make a breaking naming change to the `OperationContext` property name from `context` to something else _ex. `requestContext`_ to avoid naming colliding with `@tanstack/react-query`.
> 
> More context can be found here [#2240 (comment)](https://github.com/trpc/trpc/pull/2240#discussion_r940177822)

_From @KATT in https://github.com/trpc/trpc/pull/2240#discussion_r940286482_

> @sachinraja @jlalmes what do you think we should do here now that the `context` keyword is taken?
> 
> Some options:
> 
> - Put all trpc option under `trpc` - quite annoying for the 99% use-case but also nice to have a distinct place for "our" versus "theirs" in options
> - Call it `trpcContext`? Potentially prefix all our vars with `trpc`?
> - `operationContext`?
> - `opContext`?
> - `callContext`?
>

dyaa:
Closes #2395 

## 🎯 Changes

- Moved client `context` to own `trpc` object.
- Made the `requestId` counter be a property on the `TRPCClient` instead of a global

## ✅ Checklist

- [x] I have followed the steps listed in the [Contributing guide](/CONTRIBUTING.md).


KATT (commented):
This is great!! how come it's still a draft? :)

packages/client/src/internals/TRPCClient.ts
```diff
@@ -32,10 +32,6 @@ type CancellablePromise<T = unknown> = Promise<T> & {
   cancel: CancelFn;
 };
 
-let idCounter = 0;
-function getRequestId() {
-  return ++idCounter;
-}
```
KATT, line 38: This broke idempotency of the tests so I moved it to a prop of the `TRPCClient`

packages/client/src/internals/TRPCClient.ts
```diff
@@ -215,7 +218,7 @@ export class TRPCClient<TRouter extends AnyRouter> {
       TRPCRequestOptions?,
     ]
   ) {
-    const context = (args[1] as TRPCRequestOptions | undefined)?.requestContext;
```
juliusmarminge, line 218: Why are we spreading the inferHandlerInput and then typecasting TRPCRequestOptions as the 2nd element? Wouldn't this be cleaner, or is there something going on under the hood that makes this mandatory?
```ts
public query<
    TQueries extends AssertLegacyDef<TRouter>['queries'],
    TPath extends string & keyof TQueries,
  >(
    path: TPath,
    ...args: [
       inferHandlerInput<AssertType<TQueries, ProcedureRecord>[TPath]>[0],
       TRPCRequestOptions?,
     ]
   ) {
    const context = args[1]?.requestContext;
    //    ^? TRPCRequestOptions | undefined
```

KATT: A lot of this is very old stuff so might not be a good reason

packages/server/test/interop/react/useQuery.test.tsx
```diff
@@ -6,6 +6,8 @@ import { render, waitFor } from '@testing-library/react';
 import { expectTypeOf } from 'expect-type';
 import React, { useState } from 'react';
 
+jest.retryTimes(3);
```
sachinraja, line 9: Did you mean to add these?

sachinraja (approved):
Nice!

packages/server/test/interop/react/useQuery.test.tsx
dyaa: I think it sometimes get flaky in Github Actions

KATT: Yes. I did intend to. I can't be arsed doing a proper fix 😬

packages/client/src/internals/TRPCClient.ts
sachinraja: Good catch, I think this is actually a mistake. It allows you to do this:
```ts
client.query(
  "greeting",
  { text: "hello" },
  {
    context: {
      batch: false,
    },
  },
  {
    context: {
      batch: false,
    },
  }
);
```
pass two options objects and the last one will just be ignored.

packages/server/test/interop/react/useQuery.test.tsx
sachinraja: Ah ok, we can look at them later

packages/client/src/internals/TRPCClient.ts
juliusmarminge: I can fix this in a separate PR

KATT (approved):

PR #2449, opened 2022-08-13
Update SSG helpers docs

bennettdams:
## 🎯 Changes

Here are three updates for the SSG helpers docs.

- Use a real example of filling the `ctx` field via `await createContext()`.
- Remove `await` when calling `createSSGHelpers`, as it is not async.
- Replace `fetchQuery` with `prefetchQuery` and add a comment to mention their use cases.


KATT (approved):
Thanks!

PR #2463, opened 2022-08-15
[RFC] Call procedures w/o ending `.query()`

KATT:
Client calls deciding if a procedure is a query or a mutation based on **naming conventions**.

### How the code changes

```diff
- await proxy.postById.query('1')
+ await proxy.postById('1')
- await proxy.editPost.mutation({ id: '1', data: { title: 'Hello tRPC' }})
+ await proxy.editPost({ id: '1', data: { title: 'Hello tRPC' }})
```

### What it requires

Uses a mapper & **naming conventions** to decide if something is a query or mutation:

https://github.com/trpc/trpc/blob/76f69e96c2c015c96589dbc69740a8506427b231/packages/client/src/internals/defaultProcedureTypeSwitch.ts#L3-L13

This mapper is configurable.

### Notes

- This approach _could_ also be used on the backend to get replace of `.query()`/`.mutation()` at the end of a procedure builder with a `.resolve()`
- A lint plugin would be preferable to help devs write things according to a specific structure
- Related conversations:
  - #2270
  - https://github.com/trpc/trpc/issues/887#issuecomment-1200828262
  - #2316 
- Could allow environments where one doesn't ever care about HTTP caching `GET` to map everything as a `POST` procedure

## I want your input

- Is it worth exploring this direction or does it make the behavior too implicit?

### DIFF
--- a/packages/client/src/internals/defaultProcedureTypeSwitch.ts
+++ b/packages/client/src/internals/defaultProcedureTypeSwitch.ts
@@ -1,0 +1,13 @@
+import { ProcedureTypeSwitch } from '..';
+
+const subscriptionPaths = /\b(on)/;
+const mutationPaths = /\b(edit|create|add|delete|remove|do)/;
+export const defaultProcedureTypeSwitch: ProcedureTypeSwitch = (op) => {
+  if (mutationPaths.test(op.path)) {
+    return 'mutation';
+  }
+  if (subscriptionPaths.test(op.path)) {
+    return 'subscription';
+  }
+  return 'query';
+};
--- a/packages/client/src/internals/defaultProcedureTypeSwitch.ts
+++ b/packages/client/src/internals/defaultProcedureTypeSwitch.ts
@@ -1,7 +1,7 @@
 import { ProcedureTypeSwitch } from '..';
 
 const subscriptionPaths = /\b(on)/;
-const mutationPaths = /\b(edit|create|add|delete|remove|do)/;
+const mutationPaths = /(edit|create|add|delete|remove|mutation)|\b(do)/i;
 export const defaultProcedureTypeSwitch: ProcedureTypeSwitch = (op) => {
   if (mutationPaths.test(op.path)) {
     return 'mutation';
--- a/packages/server/test/smoke.test.ts
+++ b/packages/server/test/smoke.test.ts
@@ -73,7 +73,7 @@ test('very happy path', async () => {
     expectTypeOf<TError['data']['foo']>().toMatchTypeOf<'bar'>();
   }
   const { proxy, close } = routerToServerAndClientNew(router);
-  expect(await proxy.greeting.query('KATT')).toBe('hello KATT');
+  expect(await proxy.greeting('KATT')).toBe('hello KATT');
   close();
 });
 
@@ -97,7 +97,7 @@ test('middleware', async () => {
       .query(({ ctx }) => `${ctx.prefix} ${ctx.user}`),
   });
   const { proxy, close } = routerToServerAndClientNew(router);
-  expect(await proxy.greeting.query()).toBe('hello KATT');
+  expect(await proxy.greeting()).toBe('hello KATT');
   close();
 });
 
@@ -108,7 +108,7 @@ test('sad path', async () => {
   const { proxy, close } = routerToServerAndClientNew(router);
 
   // @ts-expect-error this procedure does not exist
-  const result = await waitError(proxy.not.found.query(), TRPCClientError);
+  const result = await waitError(proxy.not.found(), TRPCClientError);
   expect(result).toMatchInlineSnapshot(
     `[TRPCClientError: No "query"-procedure on path "not.found"]`,
   );
@@ -117,12 +117,12 @@ test('sad path', async () => {
 
 test('call a mutation as a query', async () => {
   const router = t.router({
-    hello: procedure.query(() => 'world'),
+    editWorld: procedure.query(() => 'world'),
   });
   const { proxy, close } = routerToServerAndClientNew(router);
 
-  await expect((proxy.hello as any).mutate()).rejects.toMatchInlineSnapshot(
-    `[TRPCClientError: No "mutation"-procedure on path "hello"]`,
+  await expect((proxy.editWorld as any)()).rejects.toMatchInlineSnapshot(
+    `[TRPCClientError: No "mutation"-procedure on path "editWorld"]`,
   );
 
   close();
--- a/packages/server/test/smoke.test.ts
+++ b/packages/server/test/smoke.test.ts
@@ -184,7 +184,7 @@ test('subscriptions', async () => {
     }),
   });
 
-  const subscription = proxy.onEvent.subscribe(10, {
+  const subscription = proxy.onEvent(10, {
     onStarted: onStartedMock,
     onData: onDataMock,
     onComplete: onCompleteMock,
--- a/packages/client/src/createTRPCClientProxy.ts
+++ b/packages/client/src/createTRPCClientProxy.ts
@@ -7,7 +7,6 @@ import type {
   Procedure,
   ProcedureArgs,
   ProcedureRouterRecord,
-  ProcedureType,
   inferProcedureOutput,
 } from '@trpc/server';
 import type {
@@ -27,10 +26,7 @@ type Resolver<TProcedure extends Procedure<any>> = (
   ...args: ProcedureArgs<TProcedure['_def']>
 ) => Promise<inferProcedureOutput<TProcedure>>;
 
-type SubscriptionResolver<
-  TProcedure extends Procedure<any>,
-  TRouter extends AnyRouter,
-> = (
+type SubscriptionResolver<TProcedure extends Procedure<any>> = (
   ...args: [
     input: ProcedureArgs<TProcedure['_def']>[0],
     opts: ProcedureArgs<TProcedure['_def']>[1] &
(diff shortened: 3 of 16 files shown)


packages/client/src/internals/defaultProcedureTypeSwitch.ts
```diff
@@ -0,0 +1,13 @@
...
+const mutationPaths = /(edit|create|add|delete|remove|mutation)|\b(do)/i;
+export const defaultProcedureTypeSwitch: ProcedureTypeSwitch = (op) => {
+  if (mutationPaths.test(op.path)) {
+    return 'mutation';
+  }
+  if (subscriptionPaths.test(op.path)) {
+    return 'subscription';
+  }
+  return 'query';
+};
```
KATT, line 13: cc @sachinraja

packages/server/test/smoke.test.ts
```diff
@@ -97,7 +97,7 @@ test('middleware', async () => {
       .query(({ ctx }) => `${ctx.prefix} ${ctx.user}`),
   });
   const { proxy, close } = routerToServerAndClientNew(router);
-  expect(await proxy.greeting.query()).toBe('hello KATT');
+  expect(await proxy.greeting()).toBe('hello KATT');
```
KATT, line 100: cc @sachinraja

austins:
I like this. I think mutation naming could include "update" and "rename" based on [this](https://cloud.google.com/apis/design/naming_convention#method_names).

packages/client/src/internals/defaultProcedureTypeSwitch.ts
mmkal, line 13: I think it's scary/dangerous for there to be a default, especially one that defaults to `return 'query'` - queries can be cached. Here's a procedure that could go pretty badly wrong with this setup:

```ts
const proxy = createProxyTrpcClient({ /* ...defaults... */ })

const inc1 = await proxy.account.incrementBalance({ usd: 100 }) // whoops, this is a query, since we forgot "increment"
const inc2 = await proxy.account.incrementBalance({ usd: 100 }) // might be cached! but would you notice?

assert.ok(inc1.success && inc2.success) // succeed, but shouldn't
```

Other problems with the regexes above:

```ts
await proxy.oneTimeCode.send({ email: ... }) // whoops, subscription!
``` 

This is a pretty big footgun - I think trpc should, from my least favourite option to most:

1. Default to `query` like this
2. Default to `mutation` unless keywords like `get`, `list`, etc. are used
3. Do 2. and force users to explicitly pass in the `defaultProcedureTypeSwitch` value (maybe rename to `recommendedProcedureTypeSwitch`
4. Don't provide an implementation of `ProxedureTypeSwitch` at all
5. Switch to path-proxies #2316 (🙃 😬 ) (not sure this issue would have come up if the convention was `await client.mutate(paths.account.incrementBalance, { usd: 100 })`)

jlalmes: I second the above 👆 - really not keen on this RFC. 

> defaults to return 'query'

Yeah (if we continue with this RFC) it should really default to `mutation` unless `get/read/find/search` etc.

thelinuxlich:
I love this direction!

packages/client/src/internals/defaultProcedureTypeSwitch.ts
KATT: Agreed, that was an implementation detail.

I'm very skeptical about this direction in general, but didn't wanna poison other people with my opinion.

I think a good thing to do is to move the `.query` etc from the end to beginning though.

Unsure about #2316. Makes stuff simpler to maintain but extra stuff to use - doesn't feel "as RPC"-esque.

austins: Moving the `.query` etc to the beginning would be more intuitive to me than at the end.

juliusmarminge: Not the biggest fan of this tbf, I'd rather have it explicit what action you are doing. Relying on naming conventions on the procedures seems sketchy. Would it even work with `interop` mode, or would it be a completely breaking change?

I would either keep it as is: `trpc.path.to.procedure.query()`, or do something similar to #2316: `trpc.query([trpc.path.to.procedure])`, the only odd thing here being the double use of `trpc`, maybe the path could be inferred so the second one isn't required: `trpc.query([path.to.procedure])`.

lkj4:
Just found this RFC and LOVE it! ❤️ Also @mmkal's [additions](https://github.com/trpc/trpc/pull/2463#discussion_r946859905) feel great. @KATT Thanks for putting the time into this 👍

packages/client/src/internals/defaultProcedureTypeSwitch.ts
lkj4: I still think we should default to mutation because of no unwanted caching and switch to query if `.match(/get/)` which can be extended by the user. Not going to repeat all the pros and cons but rather would like to stress that we debate this topic for two months now. 

I find also this always brought up argument, 'oh implicit behavior is so dangerous`, very misleading and poisoning the discussion. The worst which can happen that requests won't be cached but not the way around which is good and far from being "dangerous". So, pls hesitate from using this non-reason.

I know that there are a gazillion other issues but I think this is the most important because it's the API which will be most talked about, discussed, presented outside of our bubble and win new users.

I find adding explicit keywords ugly, redundant/not DRY and again, there's no reason to do so. An abstraction abstracts things away a user doesn't need to know. And if an algorithm can easily infer stuff, then it should do so.

But before we start again, let's just decide, implement or not and if not I am perfectly fine and don't have to come back every week and see if there's some progress.

Apologies for the harsh words but reading the same arguments ("dangerous" lol) over and over again is no fun either.

mmkal: You suggested defaulting to mutation and using the regex `/get/` to indicate a query.

```ts
const foodBudget = await trpc.createBudget()
const travelBudget = await trpc.createBudget()
```

Since trpc treated these as queries, the second call is cached and now we are going to start using the wrong budget all over the place.

It's dangerous. Come up with a better regex and I'll come up with more examples of how it can go badly wrong. And imagine how impossible it'd be to come up with a regex that doesn't trip up on something like `getLatestFooAndCreateNewBar`. Granted, that probably isn't the best design, but it's totally not ok for trpc to break users' codebases when they use English the "wrong" way.

lkj4: I wrote (100x in various threads) that the keywords are user-configurable, the user can make regexes, where you require the keyword, e.g. get to be at the beginning. TS can use literal types can help there as well. You edge case can be easily solved.  

But my actual point is that you exaggerate with terms like __dangerous__, __scary__, __pretty badly wrong__, __pretty big footgun_,_ __Scunthorpe Problem__ (this was all in ONE post!) and create a notion which doesn't reflect reality and again, poison any potential creativity and fuel a crucial discussion with unnecessary FUD.

Even worse if a contributor does this b/c it's harder for @KATT to push back and keep a balanced discussion culture where innovation happens (which doesn't because it's instantly shutdown with FUD).

KATT: @ikj4 WDYT about #2663? If we enabled that behavior, you could probably roll your on ending HTTP link that works with query or mutation depending how you like it.

lkj4: @KATT optional is better than nothing, so yes. The only con is less focus and more distraction bc ofc we need to think through a regex-based approach and if we allow everything we might get avg api design with both options.

re your twitter poll: the thing is that  most got educated by graphlql which distinguishes between query and mutation, so IDK if that or any majority is a good benchmark. Also remember that one HN thread where an army of gRPC fanbois came around and questioned your efforts and actually didn't get at all why we need something like TS RPC.

This debate isn't easy and this is the reason it gets prolonged without any actual result. Optional is good idea just to move on and maybe we find better ideas on our way.

KATT: There are pros and cons with every solution. Feel free to have a stab at implementing #2663 if it removes your pain, but I'm not interested in having long debates with someone who has not contributed to tRPC in any shape or form apart from complaining about API-design choices.

I don't really follow what you're trying to say with the above. I don't care what gRPC-people say and no one asked me to build tRPC either. At the end of the day, it's popularity has gotten to where it is because of my work and my API-design choices so I'll have to trust my own judgment & people who are active contributors or have been around for a long time here.

lkj4: @KATT not sure why your tone got harsher, I mean I just said, yes optional is a way, not sure if it might distract + great design isn't always driven by the majority of users; at least your one contributor stopped creating continuous fear; in order to move on, I'd happy if you decide on and announce a final API design; as I said multiple times, I'm happy with _any_ decision

ps, re my contributions: feedback is the first step of contributing (and one of my suggestions has been initially picked up by others and yourself) and why are you so sure I can't/won't contribute more in future? and if I were you I'd keep a healthy balance of opposing opinions, struggle is where greatness comes from

KATT: Apologies if it came across as hostile. I'm doing my best to balance the API decisions & at some point, we just gotta make a decision and commit. I think this sort of API is still possible through #2663 whilst allowing people who _do_ want a separation between queries and mutations still possible.

PR #2475, opened 2022-08-17
feat: Client-side cancellation

### LINKED ISSUE #2283: How do you perform client request cancelation in v10?
### Discussed in https://github.com/trpc/trpc/discussions/2282

<sup>Originally posted by **IgnusG** July 21, 2022</sup>
I noticed that v10 abandoned the cancelable promise design returning instead a plain promise. How can you cancel a request that is already in progress?</div>

We should do something with an `AbortController` signal in the extra parameters we can send. We can look at react-query for inspiration

juliusmarminge:
Closes #2283

## 🎯 Changes

Changes how queries are aborted

```ts
// Vanilla
//                                         v input    v options
const query = trpc.proxy.post.getAll.query(undefined, { signal: ac.signal });

ac.abort();

// React alt 1 - keep similar API to @tanstack/react-query
const query = trpc.proxy.post.getAll.useQuery();
const utils = trpc.proxy.useContext();
utils.post.getAll.cancel();
```

## ✅ Checklist


juliusmarminge:
cc @sachinraja

packages/react/src/internals/utilsProxy.ts
```diff
@@ -98,6 +98,7 @@ type DecorateProcedure<
   ): Promise<void>;
 
   /**
+   * @deprecated - pass an AbortSignal to the query instead
```
KATT, line 101: It's okay to just remove it too, no backwards compat needed

examples/next-minimal-starter/src/pages/index.tsx
```diff
@@ -1,18 +1,27 @@
...
-        <h1>Loading...</h1>
-      </div>
-    );
-  }
+  const ac = new AbortController();
+  const utils = trpc.proxy.useContext();
+
+  const longQuery = trpc.proxy.longQuery.useQuery('Julius', {
+    trpc: { signal: ac.signal },
+  });
```
sachinraja, line 10: This shouldn't be necessary for cancelling React Query queries, there's no difference between cancelling and aborting.

packages/client/src/internals/TRPCClient.ts
```diff
@@ -186,25 +187,32 @@ export class TRPCClient<TRouter extends AnyRouter> {
...
-    return cancellablePromise;
+      promise
+        .then((envelope) => {
+          resolve((envelope.result as any).data);
+        })
+        .catch((err) => {
+          reject(TRPCClientError.from(err));
+        });
+    }) as CancellablePromise<any>;
+    abortablePromise.cancel = abort;
```
sachinraja, line 213: Shouldn't need this since we don't need to support `.cancel` anymore

packages/react/src/createReactQueryHooks.tsx
```diff
@@ -309,7 +309,18 @@ export function createReactQueryHooks<
...
+
+    React.useEffect(() => {
+      const abortQuery = () => {
+        void cancelQuery([pathAndInput[0]]);
+      };
+      opts?.trpc?.signal?.addEventListener('abort', abortQuery);
+      return () => {
+        opts?.trpc?.signal?.removeEventListener('abort', abortQuery);
+      };
+    });
```
sachinraja, line 323: React Query [passes an abort signal to each query](https://tanstack.com/query/v4/docs/guides/query-cancellation) so we should use that instead. We also shouldn't allow passing a signal in the options here since you should just be calling `utils.post.all.cancel()`

sachinraja:
Nice work! Left a few comments, the vanilla client implementation looks good but I think we could improve the React Query one.

juliusmarminge:
> Nice work! Left a few comments, the vanilla client implementation looks good but I think we could improve the React Query one.

We need to agree on what API we should support. I left two alternatives for React at the top. Not sure which one I prefer the most.

sachinraja:
Ah sorry, I think we should go for this one:
```ts
const query = trpc.proxy.post.getAll.useQuery(undefined);
const utils = trpc.proxy.useContext();
utils.post.getAll.cancel();
```
because it's more standard React Query. @KATT do you agree?

KATT:
## Vanilla API

As you mentioned above, this is great:

```ts
// Vanilla
//                                   v input    v options
const query = trpc.proxy.post.getAll.query(undefined, { signal: ac.signal });

ac.abort();
```


## React API

- With the React-API, AFAIK we shouldn't need or be able to maybe to pass a `signal`. It should happen automatically as the component unmounts
- RQ automatically gives a `{ signal }` property in the `useQuery` callback - [example](https://tanstack.com/query/v4/docs/guides/query-cancellation#using-fetch)

packages/react/src/internals/utilsProxy.ts
KATT, line 101: This should actually be fine to use

```suggestion
```

juliusmarminge:
Maybe look into adding a few more tests and improving the docs, but the vast majority is done now.

Currently on travelling foot so I won't be putting in too many hours into coding over the weekend

packages/react/src/createReactQueryHooks.tsx
```diff
@@ -320,12 +320,17 @@ export function createReactQueryHooks<
...
     }
-    const actualOpts = useSSRQueryOptionsIfNeeded(pathAndInput, opts);
+    const ssrOpts = useSSRQueryOptionsIfNeeded(pathAndInput, opts);
 
     return __useQuery(
       pathAndInput as any,
-      () => (client as any).query(...getClientArgs(pathAndInput, actualOpts)),
-      actualOpts,
+      ({ signal }) => {
+        const actualOpts = { ...ssrOpts, trpc: { ...ssrOpts?.trpc, signal } };
```
sachinraja, line 328: I don't think we should allow passing in a signal at all for React Query. Can we not include `signal` for `TRPCReactRequestOptions`?

packages/react/src/createReactQueryHooks.tsx
```diff
@@ -320,12 +320,17 @@ export function createReactQueryHooks<
...
+        return (client as any).query(
+          ...getClientArgs(pathAndInput, actualOpts),
```
sachinraja, line 330: I think we made a mistake when changing trpc properties to be under the `trpc` object (this also applies to `useMutation`)
```suggestion
          ...getClientArgs(pathAndInput, actualOpts.trpc),
```

packages/server/src/core/procedure.ts
```diff
@@ -17,6 +17,7 @@ export interface ProcedureOptions {
    * Client-side context
    */
   context?: ClientContext;
+  signal?: AbortSignal; // FIXME: Only for query
```
sachinraja, line 20: Why? Can't mutations be aborted too (in the vanilla client, not react query)?

packages/server/test/abortQuery.test.ts
```diff
@@ -0,0 +1,56 @@
+import { routerToServerAndClientNew, waitMs } from './___testHelpers';
```
sachinraja, line 1: Might be worth adding a test for react query

examples/next-minimal-starter/src/pages/index.tsx
```diff
@@ -21,6 +26,7 @@ const styles = {
   width: '100vw',
   height: '100vh',
   display: 'flex',
+  flexDirection: 'column' as any, // wtf?
```
sachinraja, line 29: I think we need to `as const` the styles object

examples/express-minimal/src/client.ts
```diff
@@ -1,21 +1,28 @@
...
+import { createTRPCProxyClient } from '@trpc/client';
 import fetch from 'node-fetch';
 import type { AppRouter } from './router';
 
 // polyfill
 global.fetch = fetch as any;
 
 async function main() {
-  const client = createTRPCClient<AppRouter>({
+  const proxy = createTRPCProxyClient<AppRouter>({
```
sachinraja, line 9: nit: I think we should name this `client`, not `proxy`

sachinraja (changes requested):
Nice work! This looks mostly good, but we need to clean up a few more things

packages/react/src/createReactQueryHooks.tsx
juliusmarminge: why would we only want to pass in the trpc options here?

sachinraja: Those are the options meant to be passed to `client.query`. Otherwise, we'll just be passing React Query options to the vanilla client.

packages/react/src/createReactQueryHooks.tsx
sachinraja: I get that but right now it's possible to do this:
```ts
const abortController = new AbortController()
const signal = abortController.signal

// this aborts
trpc.post.byId.useQuery({ id: 1 }, { trpc: { signal } })

abortController.abort()
```
and I don't think this should be possible. I think we should only allow aborting through `.cancel`.

packages/server/test/react/useContext.test.tsx
```diff
@@ -285,3 +285,79 @@ test('getData', async () => {
...
+    );
+
+    await waitFor(() => {
+      expect(utils.getByTestId('data')).toHaveTextContent('undefined');
+      expect(utils.getByTestId('isFetching')).toHaveTextContent('idle');
+      expect(utils.getByTestId('isPaused')).toHaveTextContent('paused');
+    });
+  });
+
+  test('typeerrors and continues with signal', async () => {
```
juliusmarminge, line 328: cc @sachinraja

packages/react/src/createReactQueryHooks.tsx
juliusmarminge: That wouldnt work. I cc'ed you on the test that checka for this.

sachinraja: Ah nice, I didn't see that!

packages/server/src/core/procedure.ts
sachinraja: I think we should still do it for the vanilla client.

packages/react/src/createReactQueryHooks.tsx
juliusmarminge: Just confirmed, the `getClientArgs` flattens the options:

https://github.com/trpc/trpc/blob/98a774098a1b46ffd126ded905d53c26204a808e/packages/react/src/createReactQueryHooks.tsx#L95-L101

juliusmarminge:
Should be all good to go now. I will iterate on the docs and use shiki in my migration-guide PR if that's ok.

packages/server/src/core/procedure.ts
sachinraja, line 20: This fixme should be removed now

sachinraja (approved):
Awesome work and thanks for sticking with our feedback!

packages/react/src/createReactQueryHooks.tsx
```diff
@@ -47,7 +47,9 @@ export type OutputWithCursor<TData, TCursor extends any = any> = {
   data: TData;
 };
 
-export interface TRPCReactRequestOptions extends TRPCRequestOptions {
+export interface TRPCReactRequestOptions
+  // For RQ, we use their internal AbortSignals instead of letting the user pass their own
+  extends Omit<TRPCRequestOptions, 'signal'> {
```
KATT, line 52: Nice

KATT (approved):

KATT (approved):

PR #2498, opened 2022-08-22
add ability to have multiple input parsers

### LINKED ISSUE #2413: Multiple input parsers
Ability to do this:

```ts

type User = {
  id: string;
  name: string;
  roomIds: string[];
};
type Context = {
  user: User | null;
};


const isAuthed = t.middleware(({ next, ctx }) => {
  if (!ctx.user) {
    throw new TRPCError({
      code: 'UNAUTHORIZED',
    });
  }
  return next({
    ctx: {
      user: ctx.user,
    },
  });
});

const roomProcedure = t.procedure
  .use(isAuthed)
  .input(z.object({ roomId: z.string() }))
  .use(({ ctx, input, next }) => {
    if (!ctx.user.roomIds.includes(input.roomId)) {
      throw new TRPCError({
        code: 'UNAUTHORIZED',
      });
    }
    return next();
  });

const sendMessage = roomProcedure
  .input(z.object({ message: z.string() }))
  .mutation(({ input }) => {
    expectTypeOf(input).toMatchTypeOf<{
      message: string;
      roomId: string;
    }>();
    ///
  });
```

KATT:
Closes #2413 

## 🎯 Changes

Add ability to have multiple input parsers

Docs: https://www-git-multi-input-trpc.vercel.app/docs/router#multiple-input-parsers

## ✅ Checklist

### DIFF
--- a/packages/server/src/core/internals/procedureBuilder.ts
+++ b/packages/server/src/core/internals/procedureBuilder.ts
@@ -1,6 +1,6 @@
 import { TRPCError } from '../../error/TRPCError';
 import { getCauseFromUnknown, getErrorFromUnknown } from '../../error/utils';
-import { MaybePromise } from '../../types';
+import { FlatOverwrite, MaybePromise } from '../../types';
 import { MiddlewareFunction, MiddlewareResult } from '../middleware';
 import { Parser, inferParser } from '../parser';
 import {
@@ -31,7 +31,7 @@ type CreateProcedureReturnInput<
   _output_out: FallbackValue<TNext['_output_out'], TPrev['_output_out']>;
 }>;
 
-type OverwriteIfDefined<T, K> = UnsetMarker extends T ? K : Overwrite<T, K>;
+type OverwriteIfDefined<T, K> = UnsetMarker extends T ? K : FlatOverwrite<T, K>;
 
 export interface ProcedureBuilder<TParams extends ProcedureParams> {
   /**
--- a/packages/server/src/core/internals/procedureBuilder.ts
+++ b/packages/server/src/core/internals/procedureBuilder.ts
@@ -45,8 +45,8 @@ export interface ProcedureBuilder<TParams extends ProcedureParams> {
       : inferParser<$TParser>['out'] extends Record<string, unknown>
       ? TParams['_input_out'] extends Record<string, unknown>
         ? $TParser
-        : ErrorMessage<'Both input parsers did not resolve to an object'>
-      : ErrorMessage<'Both input parsers did not resolve to an object'>,
+        : ErrorMessage<'All input parsers did not resolve to an object'>
+      : ErrorMessage<'All input parsers did not resolve to an object'>,
   ): ProcedureBuilder<{
     _config: TParams['_config'];
     _meta: TParams['_meta'];
--- a/packages/server/src/core/internals/procedureBuilder.ts
+++ b/packages/server/src/core/internals/procedureBuilder.ts
@@ -250,6 +250,10 @@ export function createBuilder<TConfig extends RootConfig>(
   };
 }
 
+function isPlainObject(obj: unknown) {
+  return obj && typeof obj === 'object' && !Array.isArray(obj);
+}
+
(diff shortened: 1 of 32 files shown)


jlalmes (commented):
Hi @KATT, good job on this PR.

I've given it a test and it seems to only suitable for the happy path (merging objects). The result of using any other primitives (string/number/array etc.) as input types seems to give an invalid and/or indecipherable input typing in the resolver function.

Example: 
```ts
t.procedure
  .input(z.string())
  .input(z.string())
  .query(({ input }) => {
//           ^? junk
  });
```

Would be great to have this feature, but I wander if the sheer volume of edge cases would push this outside of the tRPC scope at this time?

Not sure what the expected output would be then? What keys would the different strings get?

I think it's fair to assume if you want the object merged you gotta explicit name each key, ie. passing an object should be mandatory for chained inputs?

jlalmes:
Additionally you lose the type-safety that you would expect from:

```ts
t.procedure
  .input(z.enum(['a', 'b']))
  .input(z.literal('a'))
  .query(({ input }) => {
//           ^? junk equivalent to `string`
  });
``` 

Easy alternatives: 

- Override the last `input` parser added as the source of truth.
- Allow only one `input` parser per procedure.

juliusmarminge:
> Hi @KATT, good job on this PR.
> 
> 
> 
> I've given it a test and it seems to only suitable for the happy path (merging objects). The result of using any other primitives (string/number/array etc.) as input types seems to give an invalid and/or indecipherable input typing in the resolver function.
> 
> 
> 
> Example: 
> 
> ```ts
> 
> t.procedure
> 
>   .input(z.string())
> 
>   .input(z.string())
> 
>   .query(({ input }) => {
> 
> //           ^? junk
> 
>   });
> 
> ```
> 
> 
> 
> Would be great to have this feature, but I wander if the sheer volume of edge cases would push this outside of the tRPC scope at this time?
> 

Sorry @jlalmes edited your comment instead of quoting it xD

Not sure what the expected output would be then? What keys would the different strings get?

I think it's fair to assume if you want the object merged you gotta explicit name each key, ie. passing an object should be mandatory for chained inputs?

KATT:
I think we should only allow multiple inputs for object-like input parsers, WDYT @jlalmes?

KATT:
> > I think we should only allow multiple inputs for object-like input parsers, WDYT @jlalmes?
> 
> I guess this could work - I would be keen to ensure that the DX story is really great if we go down this route though.
> 
> ```ts
> t.procedure
>   .input(z.object({ foo: z.string() }))
>   .input(z.string()) // 👈 TSError
>   
> t.procedure
>   .input(z.string())
>   .input(z.object({ foo: z.string() })) // 👈 TSError
> ```

yeah, that was my thought too, added some failing tests

KATT:
Good enough story @jlalmes?

- Screenshot 2022-08-22 at 19 02 51
- Screenshot 2022-08-22 at 19 03 00

juliusmarminge:
Nit: Are we limited to two parsers or can it be infinitely chained? In that case maybe adjust the message

packages/server/src/core/internals/procedureBuilder.ts
```diff
@@ -31,19 +31,35 @@ type CreateProcedureReturnInput<
...
    */
   input<$TParser extends Parser>(
-    schema: $TParser,
+    schema: TParams['_input_out'] extends UnsetMarker
+      ? $TParser
+      : inferParser<$TParser>['out'] extends Record<string, unknown>
+      ? TParams['_input_out'] extends Record<string, unknown>
+        ? $TParser
+        : ErrorMessage<'Both input parsers did not resolve to an object'>
+      : ErrorMessage<'Both input parsers did not resolve to an object'>,
```
KATT, line 49: Good catch @juliusmarminge 

```suggestion
        : ErrorMessage<'All input parsers did not resolve to an object'>
      : ErrorMessage<'All input parsers did not resolve to an object'>,
```

jlalmes (approved):
Ship it baby ✅ 🚀

PR #2500, opened 2022-08-22
hide noisy shiki

### LINKED ISSUE #2499: Less noisy shiki twoslash
Maybe we should disable hover completely?

---

cc @KATT Generally this looks very cool and helpful, however, is there any configuration option for this? IMO kind of annoying that every single keyword exposes a "big menu". 
CleanShot 2022-08-18 at 21 23 09

These stuff seems kinda useless for our documentation
CleanShot 2022-08-18 at 21 25 36@2x

Would prefer it only displaying types where we choose to display them, like 
CleanShot 2022-08-18 at 21 24 28@2x

and the autocomplete highlighting also looks really good and showcases useful functionality.
CleanShot 2022-08-18 at 21 24 44@2x

_Originally posted by @juliusmarminge in https://github.com/trpc/trpc/issues/2481#issuecomment-1219866316_

juliusmarminge:
Closes #2499 

## 🎯 Changes

"Disables" hover tooltips for Shiki Twoslash. Now only what we explicitly show is displayed.

CleanShot 2022-08-22 at 16 43 06

## ✅ Checklist

### DIFF
--- a/www/src/css/custom.css
+++ b/www/src/css/custom.css
@@ -214,3 +214,13 @@ h5 {
 .markdown ol {
   @apply list-decimal list-inside;
 }
+
+/** Hide noisy shiki stuff */
+
+data-lsp {
+  border-bottom: 0 !important;
+}
+
+data-lsp::before {
+  display: none;
+}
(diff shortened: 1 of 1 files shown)


juliusmarminge:
If we want to actually disable the functionality (and not just hide it with css), I think we need to find a different package (with only twoslash) since the "intellisense" is what shiki is actually for.

PR #2522, opened 2022-08-24
fix: Handle WebSocket errors to avoid Node.js crashes

frol:
## 🎯 Changes

We caught the following error on our production:

image

I believe those malformed requests come from some vulnerability scanners. In either case, such requests should not crash the instance. During the investigation I found this discussion: https://github.com/websockets/ws/issues/1354#issuecomment-774616962, and started to look into the way of handling the error. It seems there is no other way to inject this handler, so here is the PR. I have tested it already and it prevents the crash of the nodejs process.

## ✅ Checklist

- [ ] I have added or updated the tests related to the changes made. - I have no idea how to add the test here


jlalmes (approved):

sachinraja (approved):

KATT:
cc @frol I've updated it to use `opts.onError` instead - lmk if you see any problems with that.

sachinraja (approved):

KATT (approved):

frol:
That makes total sense, thanks for updating and merging the PR! 🎉

PR #2536, opened 2022-08-25
improve dx for type inference

juliusmarminge:
Closes #2482

## 🎯 Changes

Export a `GetInferanceHelpers` from core that can be used to more easily infer types without needing to import your router for every occasion.

Credits to @sachinraja for the actual type.

## ✅ Checklist

- [x] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [x] If necessary, I have added documentation related to the changes made.
- [ ] I have added or updated the tests related to the changes made.

Issue #2482, opened 2022-08-18
docs: include an inference helper in documentation


juliusmarminge:
### Area of Improvement

Currently in V10, we infer the types like this:
```ts
// sample router
const router = t.router({
  post: t.procedure.input(z.string()).mutation(() => {
    // ...
    return post.id;
  })
});
type AppRouter = typeof appRouter;

// usage
import type { inferProcedureInput } from "@trpc/server";
import type { AppRouter } from "routers/_app";

export type PostInput = inferProcedureInput<AppRouter['post']>;
```

the double import here is kinda annoying imo, would be nice to provide a wrapper in the documentation that would allow for something like this:
```ts
// utils/trpc.ts
import type{ inferProcedureInput } from "@trpc/server";
import type { AppRouter } from "routers/_app";

export type InferProcedureInput[path] = inferProcedureInput<AppRouter[path]>;

// usage
import type { InferProcedureInput } from "utils/trpc";

type PostInput = InferProcedureInput['post'];
```

### Link to related docs

https://alpha.trpc.io/docs/infer-types

### Additional information

Not a TS-Wizard so I don't know how to make such generic type, since routers can be arbitrary deep. Is it possible?

### 👨‍👧‍👦 Contributing

- [ ] 🙋‍♂️ Yes, I'd be down to file a PR implementing the suggested changes!

mmkal:
In V9, we do something like this:

server/trpc.ts:
```ts
export type RouterTypes = {
  queries: {
    [K in keyof AppRouter['_def']['queries']]: {
      input: trpc.inferProcedureInput<AppRouter['_def']['queries'][K]>
      output: trpc.inferProcedureOutput<AppRouter['_def']['queries'][K]>
    }
  }
  mutations: {
    [K in keyof AppRouter['_def']['mutations']]: {
      input: trpc.inferProcedureInput<AppRouter['_def']['mutations'][K]>
      output: trpc.inferProcedureOutput<AppRouter['_def']['mutations'][K]>
    }
  }
}
```

client/utils/trpc.ts:

```ts
import type {AppRouter, RouterTypes} from '@ourcompany/api/trpc'

export const trpc = createReactQueryHooks<AppRouter>()
export type trpc = RouterTypes
```

client/SomeComponent.tsx:
```tsx
import {trpc} from '~/utils/trpc'

export const SomeComponent = () => {
  const user = trpc.useQuery(['user/lookup', {id: 123}])
  
[...]

PR #2557, opened 2022-08-26
Experimental router shape helper and TypeScript perf changes

colelawrence:
Example related to #2482 and #2536 for @KATT, @sachinraja, @juliusmarminge 

**This PR is for conversation only. It's unlikely to merge.**

## 🎯 Changes

Anticipated reduced lookup costs associated with a few inference helpers.

## ✅ Checklist

### DIFF
--- a/packages/server/src/core/procedure.ts
+++ b/packages/server/src/core/procedure.ts
@@ -6,6 +6,7 @@ import {
   ProcedureCallOptions,
 } from './internals/procedureBuilder';
 import { UnsetMarker } from './internals/utils';
+import { ItemKind, TRPCRouterItem } from './types';
 
 type ClientContext = Record<string, unknown>;
 
@@ -85,7 +86,14 @@ export type ProcedureArgs<TParams extends ProcedureParams> =
  *
  * @internal
  */
-export interface ProcedureBase<TParams extends ProcedureParams> {
+export interface ProcedureBase<
+  TParams extends ProcedureParams,
+  Kind extends string,
+> extends TRPCRouterItem<ItemKind.Procedure> {
+  /**
+   * @internal
+   */
+  _procedureKind: Kind;
   _def: TParams & ProcedureBuilderDef;
   /**
    * @deprecated use `._def.meta` instead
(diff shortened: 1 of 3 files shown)


packages/server/src/core/types.ts
```diff
@@ -24,22 +35,38 @@ export const procedureTypes = ['query', 'mutation', 'subscription'] as const;
...
+
+type RouterRecordValueShape<
+  TRouterOrProcedure extends AnyRouter | AnyProcedure,
+> = TRouterOrProcedure extends TRPCRouterItem<ItemKind.Router>
+  ? RouterShape<TRouterOrProcedure['_def']['record']>
+  : TRouterOrProcedure extends TRPCRouterItem<ItemKind.Procedure>
+  ? {
+      kind: TRouterOrProcedure['_procedureKind'];
+      input: ProcedureArgs<TRouterOrProcedure['_def']>;
+      output: TRouterOrProcedure['_def']['_output_out'];
```
colelawrence, line 48: Notice the addition of the `kind`, which is required if you want to allow people to create types based on the procedure type.

packages/server/src/core/procedure.ts
```diff
@@ -85,7 +86,14 @@ export type ProcedureArgs<TParams extends ProcedureParams> =
  *
  * @internal
  */
-export interface ProcedureBase<TParams extends ProcedureParams> {
+export interface ProcedureBase<
+  TParams extends ProcedureParams,
+  Kind extends string,
```
juliusmarminge, line 91:
```suggestion
  Kind extends ProcedureType
```

sachinraja (commented):
Looks interesting. I'd be curious to know if it's actually better performance wise though

KATT (commented):
I don't have any strong feelings against this PR, looks good to me. I'm curious to understand why you think this is going to have a perf effect on lookups? I'd love to understand more about ts perf, if you know content that is good to read, pass it over please ☺

packages/server/src/core/procedure.ts
colelawrence: Seems like a good idea, but given this is an internal type, I was thinking in my mind that keeping the type constraints looser is actually going to improve perf, since it removes another look-up that actually doesn't affect the static checking on the user's side.

This is very hand-wavvy, though. I would love to find time for benchmarking large complex nested routers in the longer term, because without benchmarks everything I'm saying about performance should be very questionable 😂

juliusmarminge: Oh yea I don't know anything about TS-perf just thought it would perhaps prevent some mistakes if giving it a wrong Kind. I do love my squiggles and autocomplete 😏

colelawrence:
> I don't have any strong feelings against this PR, looks good to me. I'm curious to understand why you think this is going to have a perf effect on lookups? I'd love to understand more about ts perf, if you know content that is good to read, pass it over please ☺

The main idea, here, is that we're using a discriminated union tag at the start of the Router vs Procedure type to reduce the need for structural checking the overlaps between the two types.

e.g. it's faster to do `TProcedureOrRouter extends {$$kind: Kind.Procedure} ? true : false` than  `TProcedureOrRouter extends AnyProcedure ? true : false`, because we're shortcutting a structural type-check by using a discriminated tag, which can be the first property declared on the type.

This is a bit hand-wavvy, so I just wanted to get the PR on paper to start a conversation. We really shouldn't sacrifice maintainability for performance we cannot measure, though.

KATT:
Closing because of the number of conflicts & that it hasn't been touched in quite a while.

PR #2562, opened 2022-08-26
Fix subtle PromiseLike errors with `ProxyPromiseGuard`

### LINKED ISSUE #2521: bug: v10 resolving to return of `createTRPCProxyClient` throws `client[procedureType] is not a function` (path = `"then"`)

```
  System:
    OS: macOS 12.5
    CPU: (8) arm64 Apple M1
    Memory: 74.78 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.5.0 - ~/.nvm/versions/node/v18.5.0/bin/node
    Yarn: 1.22.17 - ~/.yarn/bin/yarn
    npm: 8.12.1 - ~/.nvm/versions/node/v18.5.0/bin/npm
  Browsers:
    Chrome: 104.0.5112.101
    Chrome Canary: 107.0.5257.0
    Firefox Developer Edition: 104.0
    Safari: 15.6
  npmPackages:
    @trpc/client: 10.0.0-alpha.50 => 10.0.0-alpha.50 
    @trpc/server: 10.0.0-alpha.50 => 10.0.0-alpha.50 
```

### Describe the bug

```js
async function normalLookingFunction() {
    return createTRPCProxyClient(anything)
}

// Throws `client[procedureType] is not a function` (path = `"then"`)
normalLookingFunction().then(client => console.log(client))
```

### To reproduce

See above

### Additional information

I understand there's some reason behind switching to `[operation].(mutate|query)` ordering, but I'm not exactly sure how to address this kind of issue where the user might have a `then` operation.

### 👨‍👧‍👦 Contributing

colelawrence:
Closes #2521

## 🎯 Changes

Add `ProxyPromiseGuard` type and defensive handling in TRPC Proxy Client.

## ✅ Checklist

### DIFF
--- a/packages/client/src/createTRPCClientProxy.ts
+++ b/packages/client/src/createTRPCClientProxy.ts
@@ -101,9 +101,11 @@ export function createTRPCClientProxy<TRouter extends AnyRouter>(
   const proxy = createProxy(({ path, args }) => {
     const pathCopy = [...path];
     const clientCallType = pathCopy.pop()! as keyof DecorateProcedure<any, any>;
+
     const procedureType = clientCallTypeMap[clientCallType];
 
     const fullPath = pathCopy.join('.');
+
     return (client as any)[procedureType](fullPath, ...args);
   });
   return proxy as inferRouterProxyClient<TRouter>;
--- a/packages/client/src/createTRPCClientProxy.ts
+++ b/packages/client/src/createTRPCClientProxy.ts
@@ -102,12 +102,7 @@ export function createTRPCClientProxy<TRouter extends AnyRouter>(
     const pathCopy = [...path];
     const clientCallType = pathCopy.pop()! as keyof DecorateProcedure<any, any>;
     const procedureType = clientCallTypeMap[clientCallType];
-    if (!procedureType) {
-      // pass it on if we're being treated like a promise or something else
-      throw new TypeError(
-        `Unexpected call to "${path.join('.')}()" on TRPCProxyClient.`,
-      );
-    }
+
     const fullPath = pathCopy.join('.');
     return (client as any)[procedureType](fullPath, ...args);
   });
--- a/packages/server/src/shared/createProxy/index.ts
+++ b/packages/server/src/shared/createProxy/index.ts
@@ -4,20 +4,6 @@ interface ProxyCallbackOptions {
 }
 type ProxyCallback = (opts: ProxyCallbackOptions) => unknown;
 
-export type ProxyPromiseGuard = {
-  /**
-   * We prevent our procedures from accidentally being treated as a PromiseLike
-   * by always ensuring that "then" resolves to `undefined`.
-   *
-   * This prevents errors where a proxy client is used in a promise callback or
-   * in a `Promise.resolve(proxy)`.
-   *
-   * This does mean that `then` is a reserved word and cannot be used to
-   * name any of your procedures.
-   */
-  then: undefined;
-};
-
 function createProxyInner(callback: ProxyCallback, ...path: string[]) {
   const proxy: unknown = new Proxy(
     () => {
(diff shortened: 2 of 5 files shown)


packages/client/src/createTRPCClientProxy.ts
```diff
@@ -93,11 +93,17 @@ const clientCallTypeMap: Record<
 export function createTRPCClientProxy<TRouter extends AnyRouter>(
   client: Client<TRouter>,
 ) {
-  const proxy = createProxy(({ path, args }) => {
+  const proxy = createProxy((path, args) => {
```
colelawrence, line 96: Spread out args are more performant, as we aren't allocating a new object every time.

packages/client/src/createTRPCClientProxy.ts
```diff
@@ -93,11 +93,17 @@ const clientCallTypeMap: Record<
...
-    const procedureType = clientCallTypeMap[clientCallType];
-
+    const clientCallType = pathCopy.pop()!;
+    const procedureType =
+      clientCallTypeMap[clientCallType as keyof DecorateProcedure<any, any>];
+    if (!procedureType) {
+      // pass it on if we're being treated like a promise or something else
+      throw new TypeError(
+        `Unexpected call to "${path.join('.')}()" on TRPC proxy client.`,
+      );
```
colelawrence, line 105: Ask: Copy/error okay?

KATT (changes requested):
We need a test otherwise I'll definitely break it in the future

packages/server/src/shared/createProxy/index.ts
```diff
@@ -4,6 +4,20 @@ interface ProxyCallbackOptions {
...
+   * We prevent our procedures from accidentally being treated as a PromiseLike
+   * by always ensuring that "then" resolves to `undefined`.
+   *
+   * This prevents errors where a proxy client is used in a promise callback or
+   * in a `Promise.resolve(proxy)`.
+   *
+   * This does mean that `then` is a reserved word and cannot be used to
+   * name any of your procedures.
+   */
+  then: undefined;
```
juliusmarminge, line 18: Also add `catch` and `finally` in here?

packages/server/src/shared/createProxy/index.ts
```diff
@@ -12,6 +26,11 @@ function createProxyInner(callback: ProxyCallback, ...path: string[]) {
     {
       get(_obj, name) {
         if (typeof name === 'string') {
+          if (name === 'then') {
```
juliusmarminge, line 29: And here

packages/server/src/shared/createProxy/index.ts
colelawrence: I don't believe it's necessary. `catch` and `finally` aren't important because they aren't actually checked in a Promise test if `then` is `undefined`.
Check
```js
Promise.resolve({ get catch() { console.log("catch") }, get then() { console.log("then") }, get finally() { console.log("finally") } })
```

colelawrence: I don't believe it's necessary. `catch` and `finally` aren't important because they aren't actually checked in a Promise test if `then` is `undefined`.
Check
```js
Promise.resolve({ get catch() { console.log("catch") }, get then() { console.log("then") }, get finally() { console.log("finally") } })
```

juliusmarminge: Oh nvm then. Learning something new everyday 😃

KATT (changes requested):
Needs a test case to assert that this works.

All tests are in https://github.com/trpc/trpc/tree/next/packages/server/test

KATT (approved):

KATT:
cc @colelawrence  -- I cleaned this up to what I think is the minimum needed and added tests

PR #2604, opened 2022-08-31
rm double `()()` on `initTRPC` in favour of builder pattern

### LINKED ISSUE #2595: remove initTRPC double function call

I have always the feeling while editing /reading / writing the docs, (syntax) pain points arise. I was reading the migration guide and found this section
```ts

// Beware of the double `()()`
export const t = initTRPC()();
```
The comment already calls out this syntax is unfavourable (not obvious/ naturally) . I also dislike the syntax, because the function says nothing about what the input is for the first and second function.  In the example below, why couldn't the error formatter in the first one?.

```ts
export const t = initTRPC<{
  ctx: Context;
}>()({
  // Optional:
  transformer: superjson,
  // Optional:
  errorFormatter({ shape }) {
    return {
      ...shape,
      data: {
        ...shape.data,
      },
    };
  },
});
```
*That you also need to put in the context type makes even more difficult* 

In general, I think tRPC is missing a config server side. I propose two solutions as alternative:

```ts
// I looked into the source code  and the first function has zero params (?)
// TRPC exported function
const defineConfig = (config) => initTRPC()(config)

import { defineConfig, initTRPC } from "@trpc/server"; // defineConfig is more from the vue / vite landscape.
export const t = defineConfig() 

// for TRPC  the following names are maybe more correct:  
export const t = defineAppRouter(); 
export const t = defineBaseRouter(); 

```
After inspecting the source code I believe this syntax is possible. This would also improve the docs/ learning curve (one documentation page could explain all configurations of TRPC) 

Option 2: change docs only (change trpcInner to a meaningful function name)
```ts
const trpcInner = initTRPC(); 
export const t =  trpcInner();
```
You could also take it one step further and create it as express app.
```ts
const trpc = initTRPC();

// custom methods to use on trpc 
trpc.onError();
trpc.use();

export const t =  trpc.create(); 
```

current syntax

#### Food for thought:
- would it be nice to create a file convention like  `trpc.config.ts` or `trpc.router.ts` ?

#### Question:  
- is the style of the new client, `client.queries.getUsers()` or `client.getUsers.query()`? 

### 👨‍👧‍👦 Contributing

mmkal:
ℹ️ Updated by @KATT
---

Closes #2595

Option 1 of 2, see #2605 for something that simplifies things a bit further. 

## 🎯 Changes

Replaces:

- `initTRPC()()` with `initTRPC.create()`
- `initTRPC<{ context: Context }>()()` with `initTRPC.context<Context>().create()`
- `initTRPC<{ meta: Meta }>()()` with `initTRPC.meta<Meta>().create()`

### Notes

- I'm keeping the naming `initTRPC` because I don't believe it doesn't need to be as "easy to grab" since it's only used once and `trpc` conflicts with the naming convention on frontend
- The deployments fails as I haven't made a new version - I want some approvals first :)

## ✅ Checklist

## Previous description, by @mmkal

<details>

Closes #2595

Option 1 of 2, see #2605 for something that simplifies things a bit further. 

## 🎯 Changes

Replaces:

- `initTRPC()()` with `trpc.create()`
- `initTRPC<{ context: Context }>()()` with `trpc.context<Context>().create()`
- `initTRPC<{ meta: Meta }>()()` with `trpc.meta<Meta>().create()`

## ✅ Checklist

Some thoughts:

This could go even further. If `.create()` was moved down to the inner trpc-thing (the object with `.router(...)` and `.procedure(...)` etc. hanging off it, we could think of it as the user getting a "trpc" for free, with `ctx = {}` and `meta = {}`. Maybe there's no need for the intermediate `TRPCBuilder` concept, and no need to call `.create()`. The trpc-thing just has `context` and `meta` methods which return another trpc-thing with a different type. Will play around with this a little to see if it actually makes sense.

We could probably also rename `initTRPC.ts`.

</details>


packages/server/src/core/initTRPC.ts
```diff
@@ -23,7 +23,48 @@ import { PickFirstDefined, ValidateShape } from './internals/utils';
...
+              ctx: TParams['ctx'] extends undefined
+                ? {}
+                : NonNullable<TParams['ctx']>;
+              meta: TParams['meta'] extends undefined
+                ? {}
+                : NonNullable<TParams['meta']>;
+            }>
+          >
+        >
+      | undefined,
```
mmkal, line 59: this is copy-pasted from the d.ts file which is why it's so ugly, working on cleaning it up

mmkal:
I'm going to close this in favour of #2605. Taking another look at both, #2605 is just a simpler, easier to understand, cleaner API without any downsides I've spotted so far other than having to create a couple of internal helper types, which doesn't affect users.

The diff between this one and #2605 is linked in its PR body, so if anyone would prefer this approach, we can debate in the comments of #2605 about why, and maybe reopen or recreate this if necessary.

www/docs/server/metadata.md
```diff
@@ -36,7 +36,7 @@ interface Meta {
   hasAuth: boolean;
 }
 
-export const t = initTRPC<{ ctx: Context; meta: Meta }>()();
+export const t = initTRPC.context<Context; meta: Meta>.create();
```
KATT, line 39:
```suggestion
export const t = initTRPC.context<Context>().meta<Meta>).create();
```

www/docs/server/metadata.md
```diff
@@ -18,7 +18,7 @@ interface Meta {
...
```
KATT, line 21:
```suggestion
export const t = initTRPC.context<Context>().meta<Meta>().create();
```

packages/server/src/core/initTRPC.ts
KATT: Can be done later np

juliusmarminge (commented):
didn't look at the technical stuff but the rest looks good.

don't have a strong opinion on the API change, I didn't have any problems with the previous one, n'or do I see any issues with this. You set it up once then you're done.

jlalmes (approved):
Very happy with this change - great job all involved! 🚀

mmkal:
Thanks for getting this in! (in general will be able to help out in short bursts so v glad that didn't slow anything down)

KATT:
Thank you for helping @mmkal!

PR #2605, opened 2022-08-31
`trpc` instance - rm `initTRPC()()`

### LINKED ISSUE #2595: remove initTRPC double function call

mmkal:
Closes #2595

## 🎯 Changes

Core change: https://github.com/trpc/trpc/pull/2605/files#diff-5fcc784e757060a955b57c434a2c094a47e81634caafe788bc37c304ac6f8003

Going a little further than #2604. Instead of having a `TRPCBuilder` class with a `.create()` method to get a t-thing, add a `TRPC` class which has the `.router(...)`, `.procedure(...)` methods on it. It's equivalent to the return value of `initTRPC()()` but it also has `.context<...>()` and `.meta<...>()` methods, which return new `TRPC` instances with the respective `ctx`/`meta` types. In practice this means this is all you need to do to make a vanilla router:

```ts
import { trpc } from '@trpc/server'

const appRouter = trpc.router({
  greeting: trpc
    .procedure
    .input(z.string())
    .query(({ input }) => `hello ${input}!`)
})
```

Or if you need `ctx`:

```ts
import { trpc } from '@trpc/server'

const t = trpc.context<{ greeting: string }>()

const appRouter = t.router({
  greeting: t
    .procedure
    .input(z.string())
    .query(({ input, ctx }) => `${ctx.greeting} ${input}!`)
})
```

Add options:

```ts
import { trpc } from '@trpc/server'

const t = trpc.context<{ greeting: string }>().options({
  transformer: superjson,
})

const appRouter = t.router({
  greeting: t
    .procedure
    .input(z.string())
    .query(({ input, ctx }) => `${ctx.greeting} ${input}!`)
})
```

___

### Going even further?

If we want to maintain the single-letter import, maybe `trpc` could be aliased to `t`. I didn't do that in this PR, I did `import { trpc as t } from '@trpc/server'` in a few places. Might be worth doing, but might be a bit controversial, so probably better in a follow-up PR.

### Summary

Comparison with #2604:https://github.com/trpc/trpc/compare/trpc.create...trpc.nocreate

Replaces:

- `initTRPC()()` with `import { trpc } from '@trpc/server'`
- `initTRPC<{ context: Context }>()()` with `trpc.context<Context>()`
- `initTRPC<{ meta: Meta }>()()` with `trpc.meta<Meta>()`
- `initTRPC()({ transformer: ... })` with `trpc.options({ transformer })`

## ✅ Checklist

### DIFF
--- a/.tmp/v10-docs.md
+++ b/.tmp/v10-docs.md
@@ -166,7 +166,7 @@ type Context = {
 
 export const t = trpc
   .context<Context>()
-  .create({
+  .options({
     /* optional */
     transformer: superjson,
     // errorFormatter: [...]
--- a/.tmp/v10-docs.md
+++ b/.tmp/v10-docs.md
@@ -164,13 +164,13 @@ type Context = {
   };
 };
 
-export const t = initTRPC<{
-  ctx: Context;
-}>()({
-  /* optional */
-  transformer: superjson,
-  // errorFormatter: [...]
-});
+export const t = trpc
+  .context<Context>()
+  .create({
+    /* optional */
+    transformer: superjson,
+    // errorFormatter: [...]
+  });
 
 const {
   /**
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,7 +61,7 @@ This package contains the core and server-side functionality. If something is sh
 
 #### Building a Router
 
-This is where tRPC has the most interaction with users, so it should be treated with a great deal of importance. We care about offering a simple, intuitive API in which the HTTP layer disappears for the user. Here the user's entry point is [`initTRPC`](packages/server/src/core/initTRPC.ts) where root configuration such as a [data transformer](https://trpc.io/docs/data-transformers) is set and factory functions for router, procedure, middleware, etc. creation are returned.
+This is where tRPC has the most interaction with users, so it should be treated with a great deal of importance. We care about offering a simple, intuitive API in which the HTTP layer disappears for the user. Here the user's entry point is [`trpc`](packages/server/src/core/initTRPC.ts) where root configuration such as a [data transformer](https://trpc.io/docs/data-transformers) is set and factory functions for router, procedure, middleware, etc. creation are returned.
 
 The most complex types are also in this area because we must keep track of the context, meta, middleware, and each procedure and its inputs and outputs. If you are ever struggling to understand a type, feel free to ask for help on [Discord](https://trpc.io/discord).
 
--- a/examples/cloudflare-workers/src/router.ts
+++ b/examples/cloudflare-workers/src/router.ts
@@ -1,4 +1,4 @@
-import { trpc } from '@trpc/server';
+import { trpc as t } from '@trpc/server';
 import { z } from 'zod';
 
 let id = 0;
@@ -11,8 +11,6 @@ const db = {
     },
   ],
 };
-
-const t = trpc.create();
 
 const postRouter = t.router({
   createPost: t.procedure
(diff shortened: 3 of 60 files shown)


KATT:
I think it might be bad to allow this:

```ts
import { trpc } from '@trpc/server'

const appRouter = trpc.router({})
```

It makes the behaviour needlessly implicit, and once someone decides discovers they do want a ctx they need to refactor their whole app.

I think I want us to enforce a workflow that creates a `t` object in 1 file that you `import` and use everywhere to avoid userland bugs and confusion.

mmkal:
> I think it might be bad to allow this:
> 
> ```ts
> import { trpc } from '@trpc/server'
> 
> const appRouter = trpc.router({})
> ```
> 
> It makes the behaviour needlessly implicit, and once someone decides/discovers they do want a ctx they need to refactor their whole app.
> 
> I think I want us to enforce a workflow that creates a `t` object in 1 file that you `import` and use everywhere to avoid userland bugs and confusion.


Devil's advocate - wouldn't the below be an acceptable two-line refactor?


```diff
-import { trpc } from '@trpc/server'
+import { trpc as _trpc } from '@trpc/server'

+export const trpc = _trpc.context<...>()

const appRouter = trpc.router({})
```

...although I guess you'd then need to import your `trpc` instead of the default `trpc` everywhere. So you'd be changing a bunch of imports. Fair enough. Do you think #2604 is better then?

sachinraja:
> maybe a hybrid could make sense. the class feels sensible to me so there's just one thing to learn - the trpc instance, which has all the relevant methods.

IMO the ambiguity of using `trpc.*` vs `t.*` would end up being more confusing. I do prefer #2604.

KATT:
I much prefer #2604 indeed. 

Just unsure about using the `trpc`-keyword as that is what we advise ppl to use on the front-end and we don't need something that ppl use *exactly once* to be that easily reachable.

KATT:
Closing in favor of #2604

PR #2624, opened 2022-09-02
fix: make v10 calls exclusive to v10 procedures

juliusmarminge:
showcase prisma starter with a new procedure from following migration guide


examples/.interop/next-prisma-starter-migration/src/pages/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const addPost = trpc.useMutation('post.add', {
+    async onSuccess() {
+      // refetches posts after a post is added
+      await utils.invalidateQueries(['post.all']);
+    },
+  });
+
+  // @ts-expect-error - should not be possible to call old procedures using proxy
+  const { data: yay } = proxy.healthz.useQuery();
+  //            ^?
```
juliusmarminge, line 17: this should error but it doesn't.

no need to make the proxy object include old procs?

examples/.interop/next-prisma-starter-migration/src/server/routers/_app.ts
```diff
@@ -0,0 +1,46 @@
...
+  })
+  /**
+   * Merge `postRouter` under `post.`
+   */
+  .merge('post.', postRouter)
+  .interop();
+
+const newRouter = t.router({
+  newQuery: t.procedure.query(() => 'Hello v10'),
+});
```
juliusmarminge, line 42: this should be the only thing callable using proxy

examples/.interop/next-prisma-starter-migration/.env
```diff
@@ -0,0 +1,2 @@
+# Make sure to override these in deployment
```
mshd, line 1: env should not be in git by default, maybe rename it to .env.example

juliusmarminge: Thanks for spotting.

Just threw this up as showcase for an internal discussion, probably more to do before it's ready.

Issue #2612, opened 2022-08-31
feat: AbortSignal


TkDodo:
### Describe the feature you'd like to request

right now, trpc defaults to sending the `AbortSignal` produced by react-query to the query function. This makes sure requests that become unused while they are still in-flight will be aborted.

This is nice, but not the default behaviour of react-query. We default to letting the promise finish so that it will be put into the cache, and it will stay there until `cacheTime` elapses. This makes sure that if you come back to the page, you already have a result.

It's also interesting that trpc doesn't have a way to customize this, either globally or per-query.

In RQ, you'd customize this by either using the provided `signal` or by _not_ using it:

```
// ⬇️ this request aborts
useQuery(key, ({ signal }) => axios.get(url, { signal })

// ⬇️ this request doesn't abort
useQuery(key, () => axios.get(url)
```

### Describe the solution you'd like to see

either keep the default in sync with react-query, and/or provide a way to opt out of the behaviour. Since we don't expose the queryFn, it would need to be done in a different way.

### Desribe alternate solutions

things can stay as they are, but I think letting data resolve would be the better default, and I would at least like the option to use this feature of react-query :)

### 👨‍👧‍👦 Contributing

- [ ] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

PR #2628, opened 2022-09-04
make react query cancellation configurable

bautistaaa:
Closes https://github.com/trpc/trpc/issues/2612

Wanted to get feedback before I commit to tests / documentation

## 🎯 Changes
Adds a global option to enable cancelling requests on unmount globally(defaults to false) with the option to override per request. I struggled to find a good place to put this so please let me know if you think theres somewhere else better.

```ts
// cancel at the global level
export const trpc = setupTRPC<AppRouter, SSRContext>({
  config() {
   
    return {
     // ...
     abortOnUnmount: true,
     // ...
    };
  },
});
// cancel at the request level (overrides global config)
trpc.todos.get({ id: 1}, { trpc: { ssr: false, abortOnUnmount: false } })

```

## ✅ Checklist

### DIFF
--- a/packages/next/src/withTRPC.tsx
+++ b/packages/next/src/withTRPC.tsx
@@ -54,6 +54,7 @@ function transformQueryOrMutationCacheErrors<
 export type WithTRPCConfig<TRouter extends AnyRouter> =
   CreateTRPCClientOptions<TRouter> & {
     queryClientConfig?: QueryClientConfig;
+    abortOnUnmount?: boolean;
   };
 
 interface WithTRPCOptions<TRouter extends AnyRouter> {
@@ -94,23 +95,36 @@ export function withTRPC<
         trpc?: TRPCPrepassProps;
       },
     ) => {
-      const [{ queryClient, trpcClient, ssrState, ssrContext }] = useState(
-        () => {
-          if (props.trpc) {
-            return props.trpc;
-          }
+      const [
+        /**
+         * this shouldn't need to be part of prepass props? below fails though
+         *  TRPCPressProps |  {
+         *    abortOnUnmount: boolean;
+         *    queryClient: QueryClient;
+         *    trpcClient: TRPCClient<TRouter>;
+         *    ssrState: false | "mounting";
+         *    ssrContext: null;
+         *  }
+         *  */
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        { abortOnUnmount, queryClient, trpcClient, ssrState, ssrContext },
+      ] = useState(() => {
+        if (props.trpc) {
+          return props.trpc;
+        }
 
-          const config = getClientConfig({});
-          const queryClient = new QueryClient(config.queryClientConfig);
-          const trpcClient = trpc.createClient(config);
-          return {
-            queryClient,
-            trpcClient,
-            ssrState: opts.ssr ? ('mounting' as const) : (false as const),
-            ssrContext: null,
-          };
-        },
-      );
+        const config = getClientConfig({});
+        const queryClient = new QueryClient(config.queryClientConfig);
+        const trpcClient = trpc.createClient(config);
+        return {
+          abortOnUnmount: config.abortOnUnmount ?? false,
+          queryClient,
+          trpcClient,
+          ssrState: opts.ssr ? ('mounting' as const) : (false as const),
+          ssrContext: null,
+        };
+      });
 
       const hydratedState = trpc.useDehydratedState(
         trpcClient,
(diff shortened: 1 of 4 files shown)


packages/next/src/withTRPC.tsx
```diff
@@ -94,31 +95,32 @@ export function withTRPC<
...
+        return {
+          abortOnUnmount: config.abortOnUnmount,
+          queryClient,
+          trpcClient,
+          ssrState: opts.ssr ? ('mounting' as const) : (false as const),
+          ssrContext: null,
+        };
+      });
 
+      const { queryClient, trpcClient, ssrState, ssrContext } = prepassProps;
```
bautistaaa, line 115: couldn't destructure `abortOnUnmount` above since it wasn't a common property in the union. move it down here so i could cast below (not the best, open to suggestions)

bautistaaa:
cc @TkDodo

packages/next/src/withTRPC.tsx
TkDodo: you could add `abortOnUnmount?: never` to the other side of the union, which would be effectively the same as it not being present, but it would allow for destructuring it...

packages/next/src/withTRPC.tsx
```diff
@@ -54,6 +54,7 @@ function transformQueryOrMutationCacheErrors<
 export type WithTRPCConfig<TRouter extends AnyRouter> =
   CreateTRPCClientOptions<TRouter> & {
     queryClientConfig?: QueryClientConfig;
+    abortOnUnmount?: boolean;
```
TkDodo, line 57: not sure if we should add `OnUnmount` to the name. Technically, requests will be aborted when all observers have unsubscribed. In future versions of react, this might happen not only on unmounting, but also e.g. with a proposed new offscreen api when the observer goes offscreen but still remains mounted.

I would say that `abort?: boolean` would be fine, too.

packages/react/src/createReactQueryHooks.tsx
```diff
@@ -165,13 +169,14 @@ export function createReactQueryHooks<
   }
 
   function TRPCProvider(props: {
+    abortOnUnmount: boolean;
```
juliusmarminge, line 172: should be optional here too

packages/react/src/createReactQueryHooks.tsx
```diff
@@ -181,6 +186,7 @@ export function createReactQueryHooks<
     return (
       <Context.Provider
         value={{
+          abortOnUnmount: abortOnUnmount,
```
juliusmarminge, line 189: and then default to false down here

packages/next/src/withTRPC.tsx
```diff
@@ -94,31 +95,32 @@ export function withTRPC<
...
 
+      const { queryClient, trpcClient, ssrState, ssrContext } = prepassProps;
       const hydratedState = trpc.useDehydratedState(
         trpcClient,
         props.pageProps?.trpcState,
       );
 
       return (
         <trpc.Provider
+          abortOnUnmount={(prepassProps as any).abortOnUnmount ?? false}
```
juliusmarminge, line 123: if we want Next and React to always behave the same, we should probably not handle the default value here, but let the nullish value fall through to createReactQueryHooks and default it there

packages/react/src/createReactQueryHooks.tsx
```diff
@@ -165,13 +169,14 @@ export function createReactQueryHooks<
...
   function TRPCProvider(props: {
+    abortOnUnmount: boolean;
     queryClient: QueryClient;
     client: TRPCClient<TRouter>;
     children: ReactNode;
     ssrContext?: TSSRContext | null;
     ssrState?: SSRState;
   }) {
-    const { client, queryClient, ssrContext } = props;
+    const { abortOnUnmount, client, queryClient, ssrContext } = props;
```
juliusmarminge, line 179: or here

packages/next/src/withTRPC.tsx
bautistaaa: great point I only was testing this with next. will implement all the suggestions

packages/next/src/withTRPC.tsx
KATT: I'll leave is as-is for now, `abort` is a bit unclear to me what it means and we can always rename and deprecate later.

If we figure out a better name ahead of `10.0.0` we can rename it too.

KATT (approved):
Great, I'll merge this now and create a ticket for docs/tests

jpemberton94:
FYI I'm in alpha.70 using trpc with nextJS and having some trouble with this feature? I may be mistaken, but I believe the intent was to set this at the query level?
image

sachinraja:
I'm unable to reproduce @jpemberton94. Make sure all your trpc packages are on the latest alpha. Also make sure your next `_app` is setup properly. The error seems to be because it can't find `TRPCProvider` which is added when you do `trpc.withTRPC`.

PR #2641, opened 2022-09-06
refactor(react): changes context to a record and simplifies logic of utilsProxy

ixahmedxi:


packages/react/src/shared/proxy/utilsProxy.ts
```diff
@@ -197,87 +196,52 @@ export function createReactQueryUtilsProxy<
...
+            >;
+            const queryKey = getQueryKey(fullPath, input);
+            return {
+              input,
+              queryKey,
+              rest,
+            };
+          };
+
+          const { queryKey, rest, updater, input } = opts(utilName);
```
sachinraja, line 224: Looks good! Can we just rename this function? We use `opts` as a variable for object parameters in a lot of places. Maybe `getOpts`?

KATT: Agreed

packages/react/src/shared/proxy/utilsProxy.ts
```diff
@@ -197,87 +196,52 @@ export function createReactQueryUtilsProxy<
...
-            }
-            case 'setData': {
-              const [updater, input, ...rest] = args as Parameters<
-                AnyDecoratedProcedure['setData']
-              >;
-              const queryKey = getQueryKey(fullPath, input);
-              return context.setQueryData(queryKey, updater, ...rest);
-            }
-            case 'setInfiniteData': {
+          const opts = (name: typeof utilName) => {
```
KATT, line 199:
```suggestion
          const getOpts = (name: typeof utilName) => {
```

KATT (changes requested):
Same feedback as @sachinraja

KATT (approved):

PR #2688, opened 2022-09-11
update to node 18

KATT:


juliusmarminge:
You "can't use" Next Auth with Node 18 https://github.com/nextauthjs/next-auth/issues/4575#issuecomment-1142239163

PR #2706, opened 2022-09-13
feat(www): update docusaurus deps, skip reload

KATT:
- Update docusarus
- Don't think I need to reload anymore


www/src/pages/index.tsx
```diff
@@ -46,11 +46,7 @@ const HomeContent: React.FC = () => {
     } else if (requestedVersion === '10' && !isV10) {
       savePreferredVersionName('current');
     }
-
-    // Reload page to apply changes
-    window.location.href = '/' + window.location.hash;
```
juliusmarminge, line 51: if you dont need this you can remove the early return above

PR #2713, opened 2022-09-14
feat(react): Change to use Array Path for Tanstack Query Key and add invalidate() at router level

### LINKED ISSUE #2611: individual array entries for query key path

JonParton:
Closes #2611

## 🎯 Changes

As per #2611 this pull request looks to change to using a Sub array for the `path` element of the tanstack querykey in react hooks.

This allows us to take action on related queries such as invalidating all queries on a router which is harder to do with the current use of `.` seperated path keys such as `["post.simple.getById", {id:10}]` rather than `[["post", "simple", "getById"], {id:10}]`.

This will then allow us to implement an extension to the `invalidateQueries` proxy api such as: 

```tsx
utils.post.invalidateQueries() // Notice this is not a full path but a partial one!
```

## 🔨 Implementation

In this PR I have implemented this at the internal hooks level so it reflects across both the proxy and legacy implementation without any breaking changes to those api's. Implementing it this way also allows us to keep code shared between both implementations which we may not be able to do if implemented further down the chain. 

Even though this seems to work well it doesn't feel 100% right so definitely open to comment if I am off the mark or anyone can see problems ... but it does make sense as it's only the tanstack keys we want to affect. 

## ⚠️ Breaking Change (kinda)

The only breaking change from this is if someone is delving in and using the query keys not through the tRPC provided API's but by using Tanstack Query's `queryClient` directly. In this case users will see keys change from something like
`["post.List"]` to `[["post","list"]]`

An example of the breaking change can be seen in the test I had to change in this PR. 

## ☑️ Things left to do: 

- [x] ~~[One locally failing test left to fix](https://github.com/trpc/trpc/blob/c88edc3b8453d949954e7b40300bfd58411e8873/packages/server/test/interop/react/regression/issue-1645-setErrorStatusSSR.test.tsx#L57) .... Any help on this one is appreciated as I can't see where I have mucked it up ... Might just be too late 😴 - It was too late!~~ ✅
- [x] ~~Add API to be able to use `invalidateQueries` at any proxy path level~~ 
     - ~~Any pointers on how to do this @sachinraja would be appreciated ... I havent quite nailed down where I can hook into a partial path on the proxy types 🤔~~
- [x] ~~Add tests of the above new api to invalidate at any path level.~~ 
- [x] ~~Add documentation on new invalidating capabilities~~
- [x] ~~Add a note about the breaking change to Query Keys in a migration page / docs?~~ 

Thanks for bearing with me on this one and all comments welcome; still got a long way to go on my opensource journey! 😜

## ✅ Checklist


JonParton:
Transferring this comment from @KATT  from [the Fork](https://github.com/trpc/trpc/pull/2711) to this PR: 

--------

> In terms of the actual code. There is an edge-cases that I'd like us to consider:
> 
> When people use dot-notation in their procedure names, I am not sure if they should be split.
> 
> **Example:**
> 
> ```ts
> const router = t.router({
>   foo: t.router({
>     'bar.qux': t.procedure.query(() => 'hello'),
>   }),
> })
> ```
> 
> Should a call to `trpc.foo['bar.qux'].useQuery()` become:
> 
> - `['foo', 'bar', 'qux']` or
> - `['foo', 'bar.qux']`
> 
> I have yet to decide - WDYT @TkDodo?

packages/react/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -177,6 +177,32 @@ export function createHooksInternal<
...
 
+  /**
+   * To allow easy interactions with groups of related queries, such as
+   * invalidating all queries of a router, we use an array as the path when
+   * storing in tanstack query. This function converts from the `.` separated
+   * path passed around internally by both the legacy and proxy implementation.
+   * https://github.com/trpc/trpc/issues/2611
+   */
+  const getArrayQueryKey = (
+    queryKey: string | [string] | [string, ...unknown[]],
```
KATT, line 188: We can only have 1 input arg

```suggestion
    queryKey: string | [string] | [string, unknown],
```

JonParton: Yeah ... I did think / try this too but this then doesn't play nice when the type of the `pathAndInput` passed to some of the implementations in the [`TRPCProvider` are `path: TPath, ...args: ProcedureArgs<inferProcedureParams<TProcedure>>`](https://github.com/trpc/trpc/blob/3741fc32d04ea8f6515e3df71234408170b4142a/packages/react/src/shared/hooks/createHooksInternal.tsx#L261) ... It won't happily take the `...args` type it has there and I can't allow that specific type as it is a generic ... 

Is this type actually wrong for the `pathAndInput` too? 

...This was my only escape hatch to allow the function to be used in all instances we are actually sneakily passing `queryKeys` in!

packages/server/test/interop/react/invalidateQueries.test.tsx
```diff
@@ -43,8 +43,8 @@ describe('invalidateQueries()', () => {
           <button
             data-testid="refetch"
             onClick={() => {
-              queryClient.invalidateQueries(['allPosts']);
-              queryClient.invalidateQueries(['postById']);
+              queryClient.invalidateQueries([['allPosts']]);
+              queryClient.invalidateQueries([['postById']]);
```
KATT, line 47: This doesn't look quite right? Shouldn't it be unchanged?

JonParton: > The only breaking change from this is if someone is delving in and using the query keys not through the tRPC provided API's but by using Tanstack Query's queryClient directly. In this case users will see keys change from something like
> ["post. List"] to [["post","list"]]
> 
> An example of the breaking change can be seen in the test I had to change in this PR.

👆This was what this comment in the PR description is about 🙂

----------

We are now always storing the `path` as a sub array as the first element of the `queryKey`. 

The initial conversation around whether it should always be a sub array or top level elements is [here in the issue](https://github.com/trpc/trpc/issues/2611#issuecomment-1237490730), but we decided to go with the sub array direction so that the `path` is always kept neatly as the "first element" of the key! @TkDodo [confirmed there is no functional difference](https://github.com/trpc/trpc/issues/2611#issuecomment-1244181242) in terms of invalidation. 

This doesn't effect usage any where if you are using the tRPC provided API's but in this test we are delving in and getting the tanstack `queryClient` directly... And so the keys would be changed from what they were before! 

For a deeper defined procedure this is a bit more evident and would break this test even if it was a sub array or not!: `queryClient.invalidateQueries(['posts.all']);` becomes `queryClient.invalidateQueries([['posts','all']]);`

----------
Does that make sense? what do you think?

TkDodo: if users have access to the queryClient directly as well, it might be better to not do the nested arrays.

TkDodo:
> Should a call to trpc.foo['bar.qux'].useQuery() become:
> 
> ['foo', 'bar', 'qux'] or
> ['foo', 'bar.qux']

I would say `['foo', 'bar.qux']`

packages/server/test/interop/react/invalidateQueries.test.tsx
JonParton: I'm still personally a fan of the nested array (can you guess? 😂) just so that array keys are always a nice touple of `[path[], input]` which if we do them at the top level they won't be it would be more `[path1, path2, path3, input]`

If we did make them top level this test would technically not have to change to pass... however, its a simplified edge case example as it only has one level of the path... `queryClient.invalidateQueries(['posts. All']);` would still become `queryClient.invalidateQueries(['posts.','all']);` and break when working directly with the `queryClient`! 

🤔💭

JonParton:
so @KATT  / @TkDodo  ... 

I've done a bit more digging on the whole `[['foo', 'bar', 'qux']]` vs `[['foo', 'bar.qux']]` keys question

So given a router of 
```tsx
const appRouter = t.router({
  foo: t.router({
    "bar.qux": t.procedure.query(() => {
      return {
        log:
          "the`foo`.`bar.qux` procedure responded (IE the one with a . in the property name)"
      };
    }),
    bar: t.router({
      qux: t.procedure.query(() => {
        return {
          log:
            "the`foo`.`bar`.`qux` procedure responded (IE the one defined as sub routers)"
        };
      })
    })
  })
});
```

both of the below: 

```tsx
  const query1 = trpc.foo["bar.qux"].useQuery();
  const query2 = trpc.foo.bar.qux.useQuery();
```

will hit the second procedure defined using sub routers. 

[Code Sandbox to play around and see this in action](https://codesandbox.io/p/github/JonParton/trpc-example-dot-defined-procedures-vs-sub/draft/long-wind?file=%2Fsrc%2Fpages%2F_app.tsx&selection=%5B%7B%22endColumn%22%3A27%2C%22endLineNumber%22%3A13%2C%22startColumn%22%3A27%2C%22startLineNumber%22%3A13%7D%5D&workspace=%257B%2522activeFileId%2522%253A%2522cl81rb5l30005lof6ad7o77mk%2522%252C%2522openFiles%2522%253A%255B%2522%252FREADME.md%2522%252C%2522%252Fsrc%252Fpages%252F_app.tsx%2522%255D%252C%2522sidebarPanel%2522%253A%2522GIT%2522%252C%2522gitSidebarPanel%2522%253A%2522COMMIT%2522%252C%2522sidekickItems%2522%253A%255B%257B%2522type%2522%253A%2522PREVIEW%2522%252C%2522taskId%2522%253A%2522dev%2522%252C%2522port%2522%253A3000%252C%2522key%2522%253A%2522cl81rkgg904qb356ihnbtrlp0%2522%252C%2522isMinimized%2522%253Afalse%257D%252C%257B%2522type%2522%253A%2522TASK_LOG%2522%252C%2522taskId%2522%253A%2522dev%2522%252C%2522key%2522%253A%2522cl81rbfp40061356i6vtqdjxb%2522%252C%2522isMinimized%2522%253Afalse%257D%255D%257D)  --> Hit fork in top right to get out of read only ... 

Basically the current tRPC implementation cannot distinguish between procedures defined using `.` procedure names or sub routers... it just takes the latest one defined! 

So I think it is more honest and congruent with that if we were to have both `useQuery`'s have the same key: `[['foo','bar','qux']]`. If we don't do that you could also get two queries in the cache actually feeding from the same getter function!

In addition it also means invalidating the `[['foo','bar']]` key would correctly invalidate all procedures defined as either `bar.xi: t.procedure...` and `bar: t.router({ za: t.procedure(....)})` ... which wouldn't happen if we had some `'bar.xi'` keys in the mix.  

What do you think???

_[p.s. No bias here even though this way is much easier to implement ... honest 😂]_

JonParton:
So I've now managed to add to the `utils` types to allow invalidation at partial paths

When it's not so late I'll knock up an example but basically: 

given the queries: 
1. `trpc.foo.bar.useQuery()`
2. `trpc.foo.qux.useQuery()`
3. `trpc.foo.getByIDAndOrg.useQuery({ id: 10 , org:"google")`
4. `trpc.foo.getByIDAndOrg.useQuery({ id: 8 , org:"google")`
5. `trpc.foo.getByIDAndOrg.useQuery({ id: 4, org:"microsoft")`
6. `trpc.zar.beta.useQuery()`

we can do:

```tsx
utils.foo.invalidate() // will invalidate 1,2,3,4,5 but not 6
utils.foo.invalidate({org:google}) // Filtered by partial input so will invalidate 3,4
```

🥳

### Extra TODO's ✅ & questions ❓ after these changes: 

- [ ] At the moment you can only do `.invalidate` after selecting the initial router ie `utils.invalidate()` is not possible to invalidate all queries. Do we want to be able to do this? If so I will figure out what I need to do in the types... I kind of though what I did would add this but it didn't!... 
- [ ]  At the moment I have had to type the input on invalidate at a partial path as `unknown` ... Do we want to put effort in here to try and extract properties available across all queries it will invalidate?? 🤯 an optional intersection of all properties? ... may be too hard for initial implementation and may have TS perf implications?...

packages/react/src/shared/proxy/utilsProxy.ts
```diff
@@ -146,6 +146,25 @@ type DecorateProcedure<
...
+ * NOTE: All generics have been removed from this type at the moment ... But we
+ * may want to add these back in if we can be smarter about inferring input
+ * type...
+ */
+type DecoratePartialPathProcedure = {
+  /**
+   * @link https://react-query.tanstack.com/guides/query-invalidation
+   */
+  invalidate(
+    input?: unknown, // We can't infer the input type as it may be multiple queries!?
```
KATT, line 162: Should be possible to do by grabbing the intersection of all the types & if there's an impossible one we can enforce ut to be `undefined`

KATT (commented):
It's _cool_ that you can do like `.invalidate(input)` on a router but I'm not sure how useful it is? And I don't really like having parts of tRPC's API that is `unknown` in types

Unsure about the array-in-array-thing, WDYT @jlalmes @sachinraja ?

JonParton:
@KATT --> Just wanted to confirm regarding [this comment](https://github.com/trpc/trpc/pull/2713#issuecomment-1246956147) are we happy that a procedure defined as 

```tsx
const appRouter = t.router({
  foo: t.router({
    "bar.qux": t.procedure.query(() => {
      return "dave"
    }),
  }),
})
```

does actually get a query key of `[['foo','bar','qux']]` instead of `[['foo','bar.qux']]`. Taking into account the points made in that comment 😜

_p.s. ignoring the sub array pending above convo but more the splitting of the `.` key_

JonParton:
Righty then - After waaaaaayyyyyy too much time learning advanced TS Types and a load of research I have managed to add possible types of the `router.invalidate()` function! 🥳🎊

I've managed to grab all the possible input properties across all queries on the invalidated router and all sub routers 🤯. The extracted types are then turned into a big union that I have added to the input as a `Partial<>` as the input filtering could be very sparse (Only one property like `userId` for example). 

This seems to work pretty well and even handles if we have multiple properties with the same name but different types ... showing you that it could be a union of the different options 👍

Right bed time! 🛌

## ✅ Extra Todo's

- [ ] Merge in recent Turbo Type changes and fix conflicts
- [ ] Review function name .... Does `invalidateQueries()` make more sense than just `invalidate()` as it is working on a whole router tree? ... 🤔This then also [aligns with Tanstack Query](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientinvalidatequeries) too...

packages/react/src/shared/proxy/utilsProxy.ts
JonParton: Done! 👍

Probably needs a bit more testing though ... next jobs!

packages/react/src/shared/proxy/utilsProxy.ts
```diff
@@ -146,6 +146,63 @@ type DecorateProcedure<
...
+          ? RecursivelyInferAllQueryInputTypesUnderRouter<SubProcedureOrRouter>
+          : never
+        : never;
+    };
+  };
+
+type InferAllRouterQueryInputTypes<TRouter extends AnyRouter> =
+  ExtractUnionTypesFromRouterNest<
+    RecursivelyInferAllQueryInputTypesUnderRouter<TRouter>
+  >;
```
sachinraja, line 189: Awesome work on these types! I'm unsure about the recursion, seems like it could get messy, but here's an alternative type:
```ts
type InferAllRouterQueryInputTypes<TRouter extends AnyRouter> = {
  [TKey in keyof TRouter['_def']['record']]: TRouter['_def']['record'][TKey] extends infer ProcedureOrRouter
    ? ProcedureOrRouter extends AnyQueryProcedure
      ? ProcedureArgs<TRouter['_def']['record'][TKey]>[0]
      : ProcedureOrRouter extends AnyRouter
      ? InferAllRouterQueryInputTypes<ProcedureOrRouter>
      : never
    : never;
} & DecorateRouterProcedure<TRouter>;
```

KATT: Maybe we can compare build times on https://github.com/trpc/trpc/tree/next/examples/next-big-router on this branch vs `next`

JonParton: Fantastic! Thanks @sachinraja !

I couldn't quite get the exact syntax you wrote above to work _(Possibly `AnyQuery` is in new type changes I haven't merged yet and it didn't like the types of `ProcedureArgs<ProcedureOrRouter['_def']>[0]`)_  but with some tweaking taking your intent this has the same result: 

```ts
/**
 * A type that will traverse all procedures and sub routers of a given router to create a union of
 * their possible input types
 */
type InferAllRouterQueryInputTypes<TRouter extends AnyRouter> = {
  [TKey in keyof TRouter['_def']['record']]: TRouter['_def']['record'][TKey] extends infer ProcedureOrRouter
    ? ProcedureOrRouter extends QueryProcedure<any>
      ? inferProcedureInput<ProcedureOrRouter>
      : ProcedureOrRouter extends AnyRouter
      ? InferAllRouterQueryInputTypes<ProcedureOrRouter> // Recurse as we have a sub router!
      : never
    : never;
}[keyof TRouter['_def']['record']]; // This flattens results into a big union and ignores never keys
```

This is an awful lot more elegant as it removes the need for my Router nesting business, I had to resort to 👍👍 However, it is doing the same traverse across all sub routers that I also had concerns about performance wise ... I think the new improved version is saving the type compiler doing it twice compared to my old solution though so it may be 2x faster at least! 🚀.

Performance wise I guess I'll give it a shot on the new `bigBoi` example and if OK we see how it goes with use? 

I know we were keen to have types on this rather then an `unknown` so this seems the way!!

JonParton: Great minds @KATT... (maybe being too generous on my mind here!)... I didn't see your comment before I posted but my brain went to the new big router example too!

JonParton: I'll merge in the turbo type changes from `next` then we can do this comparison and see how it goes!

JonParton:
Soo - I think this is ready for a proper review / merging @KATT  / @sachinraja 🥳🎉. 

I've now implemented a good amount of tests (review welcome!) and also added to the documentation explaining the new capability and also the changes to Query Keys which I have included in the V9 ➡️ V10 migration page. 

The only things outstanding are: 
- Discuss if we change the name of `invalidate` when used on a router rather then a query. 
  - I think becuase it is acting on the router and possibly multiple queries it could be 
    - `invalidateQueries()` (Matching Tanstack) 
    - `invalidateRouter()`
    - `invalidateRouterQueries()`
    - keep as `invalidate()`
- Springing from your enquiries @KATT around procedure names with `.` dots in I also thought of another, I think rare, edge case where our `input` types may be missing possibilities. 
  - There is [a test for this that I have disabled](https://github.com/trpc/trpc/blob/fe8517ffbedb5c8f5c5a9aa64fe20757e51e7468/packages/server/test/react/invalidateRouters.test.tsx#L414)
  - As I think the chance of this happening is very low... I've made an issue for this (#2759) to be picked up alongside (#2737). 
    - Also if it does happen, we could just tell people the easy fix of moving the query inside the sub router...

After ~30 hours (😵) learning the implementation, brushing up my TS skills and implementing this we may even get it into the beta release! 🤞🎉🎉🥳🥳

JonParton:
> .... we may even get it into the beta release! 🤞🎉🎉🥳🥳

Oops ... Seems I missed the Beta Boat without even realising 😣

Awesome for getting it out the door though! 🙂🎉..

sachinraja (approved):

juliusmarminge (approved):

sachinraja:
Forgot to leave a comment. Just wanted to say awesome work on this and thanks for sticking with it!

JonParton:
Thanks @sachinraja - glad to see it got merged and used 🙂🎊🎉 (I was a little worried there for a while...😂) 

Thanks for the idea, input and help all!! 🙂

TkDodo:
would it be possible to also add other methods next to `invalidate` where fuzzy matching would be good? I'm specifically looking at `setQueryDefaults` and `setMutationDefaults` at the moment.

For example, we could do:

```
utils.post.setQueryDefaults({ staleTime: Infinity })
```

which would set this default value for all matching query keys of that router.

you can read more here:
- https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientsetquerydefaults
- https://tkdodo.eu/blog/testing-react-query#setquerydefaults

I'm currently missing this feature completely. Let me know if I should create a separate issue for that.

JonParton:
Absolutely possible to add these extra methods and I did set it up in the code to allow this in the future! 

Probably a quick OK from @KATT  / @sachinraja  / Others to agree we are happy adding these extra methods from a TS Perf perspective etc but totally doable. 

The full list of possible fuzzy methods that can act on a router that I found are: 

- [queryClient.invalidateQueries](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientinvalidatequeries)✅
- [getQueriesData](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientgetqueriesdata)
  - Returnes the data for only queries that are currently in the Tanstack Query cache ...not trigger TRPC calls 
- [queryClient.refetchQueries](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientrefetchqueries)
  - TS types could be interesting ... but I think possible. 
- [queryClient.cancelQueries](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientcancelqueries)
- [queryClient.removeQueries](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientremovequeries)
- [queryClient.resetQueries](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientresetqueries)
- [queryClient.getQueryDefaults](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientgetquerydefaults)
- [queryClient.setQueryDefaults](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientsetquerydefaults)
- [queryClient.setQueriesData](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientsetqueriesdata)
  - For the "updater" function to be trueley type safe when operating on multiple queries we would need to know which particular key the updater was currently working on.... I know this [was on me to go look at adding to the tabstack API](https://twitter.com/TkDodo/status/1572261676434575360) but I haven't managed to prioritise. 
- Are Mutiation defaults relevant in trpc? Do these mutation defaults get added to the TanstackQuery Client when we extract a mutation function from the trpc proxy object? ie `const mutation = trpc.login.useMutation();`
  - [queryClient.getMutationDefaults](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientgetmutationdefaults) 
  - [queryClient.setMutationDefaults](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientsetmutationdefaults)

Would all of these actually be valuable @TkDodo ?...

JonParton:
Note there are also a few that only operate at the root of all routers: 

- [queryClient.getLogger](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientgetlogger)
- [queryClient.getDefaultOptions](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientgetdefaultoptions)
- [queryClient.setDefaultOptions](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientsetdefaultoptions)
- [queryClient.getQueryCache](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientgetquerycache)
- [queryClient.getMutationCache](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientgetmutationcache)
- [queryClient.clear](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientclear)
- [queryClient.resumePausedMutations](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientresumepausedmutations)

However, as no `keys` are involved it might be better to leave these to be accessed through the tanstack query [`useQueryClient`](https://tanstack.com/query/v4/docs/reference/useQueryClient) API ...  `const queryClient = useQueryClient()`

juliusmarminge:
> Note there are also a few that only operate at the root of all routers: 
> 
> 
> 
> - [queryClient.getLogger](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientgetlogger)
> 
> - [queryClient.getDefaultOptions](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientgetdefaultoptions)
> 
> - [queryClient.setDefaultOptions](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientsetdefaultoptions)
> 
> - [queryClient.getQueryCache](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientgetquerycache)
> 
> - [queryClient.getMutationCache](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientgetmutationcache)
> 
> - [queryClient.clear](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientclear)
> 
> - [queryClient.resumePausedMutations](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientresumepausedmutations)
> 
> 
> 
> However, as no `keys` are involved it might be better to leave these to be accessed through the tanstack query [`useQueryClient`](https://tanstack.com/query/v4/docs/reference/useQueryClient) API ...  `const queryClient = useQueryClient()`

Yes we removed the queryClient from context to encourage using it directly from tanstack, so this would align with that.

TkDodo:
That's a pretty good list @JonParton 👍 

> For the "updater" function to be trueley type safe when operating on multiple queries we would need to know which particular key the updater was currently working on

This is one limitation of `setQueriesData` - we expect that people only use this function if they also update caches of the same shape.

PR #2738, opened 2022-09-17
add big router example

KATT:
Adds a Next.js-example with a big router

Might be useful for people migrating to v10 to see that it's worth the effort.


sachinraja (approved):

.gitignore
```diff
@@ -23,3 +23,6 @@ packages/*/*
 ## Dont ignore src or test
 !packages/*/src
 !packages/*/test
+
+
+examples/next-big-router/src/server/routers
```
KATT, line 28:
```suggestion
examples/next-big-router/src/server/routers

```

PR #2746, opened 2022-09-18
refactor(www): more consistent and minimal design

ixahmedxi:
This commit brings the following changes:

- Updates preview to new v10 landscape
- Remove light mode preview
- Updates features descriptions to all fit 2 lines max.
- Updates section title to not have # as that in most cases it results in the title text looking not centered.
- Updates the styling of the alex quote.
- Updates the twitter feed.
- More minimalistic twitter wall design with no scrolling.

## ✅ Checklist

### DIFF
--- a/www/docs/landing-intro/Step3.md
+++ b/www/docs/landing-intro/Step3.md
@@ -20,7 +20,7 @@ const appRouter = t.router({
 export type AppRouter = typeof appRouter;
 ```
 
-```ts twoslash title='client.ts'
+```ts twoslash
 // @module: esnext
 // @target: esnext
 // @include: server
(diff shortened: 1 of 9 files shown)


KATT (commented):
We still want portrait on mobile, right? :)

Screenshot 2022-09-18 at 22 08 38

ixahmedxi:
> We still want portrait on mobile, right? :)
> 
> 
> 
> Screenshot 2022-09-18 at 22 08 38
> 
> 

We could definitely do that or as @jlalmes suggested before most trpc website visitors will be on desktop so we can focus on that. Obviously perks of not needing to switch preview videos is that we would only be loading 2 (v9 and v10) as opposed to 3 or even 4 which should help a lot with perf of the website. I'm open to suggestions, don't really have any  preferences here.

KATT:
> We could definitely do that or as @jlalmes suggested before most trpc website visitors will be on desktop so we can focus on that. Obviously perks of not needing to switch preview videos is that we would only be loading 2 (v9 and v10) as opposed to 3 or even 4 which should help a lot with perf of the website. I'm open to suggestions, don't really have any preferences here.

There's definitely hacks around to make sure it's not loaded regardless on size, but I'm okay with 2 loaded for now -- we def want the portrait on mobile as the other one becomes unredable

KATT (approved):

PR #2758, opened 2022-09-19
bug: utils.setData updaterFn has implicit any

juliusmarminge:
Closes #

## 🎯 Changes

- Updates `@tanstack/react-query` peerDependency to `^4.3.0` and resolves a private import to fix `utils.setData((prev) => ...)` so that `prev` no longer is typed as implicit any

## ✅ Checklist


juliusmarminge:
Updated @tanstack/react-query peered to 4.3.0, this version includes a public export of `Updater` which seems to have been the culprit.

Release: https://github.com/TanStack/query/releases/tag/v4.3.0
PR: https://github.com/TanStack/query/pull/4092/files
https://github.com/TanStack/query/pull/4092/files#diff-6efa3784e21b3ef82db58f25096616adb7333dbdf46ced0c09f706115705e69c

Unfortunatly, I wasn't able to reproduce the behavior in any test so I have no way of telling for sure that this works, but when I `patch-package`d locally it did.

I'll leave the live-test in the next-prisma-starter in for now, and we can see if it works postpublish.

examples/next-prisma-starter/src/pages/index.tsx
```diff
@@ -18,6 +18,12 @@ const IndexPage: NextPageWithLayout = () => {
     },
   );
 
+  utils.post.list.setData((old) => {
+    old;
+    //^?
+    return old!;
+  });
+
```
juliusmarminge, line 26: TODO: Remove after publish

juliusmarminge:
Here is a link to a patched repro: https://github.com/juliusmarminge/optimisticUpdateTest

sachinraja (approved):

KATT (approved):

examples/next-prisma-starter/src/pages/index.tsx
```diff
@@ -18,6 +18,13 @@ const IndexPage: NextPageWithLayout = () => {
     },
   );
 
+  utils.post.list.setData((old) => {
+    old;
+    //^?
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    return old!;
```
TkDodo, line 25: you should be able to return `old` without the non-null assertion, because returning `undefined` from `setQueryData` is allowed in v4. We'll only create the cache entry if you return something that is not `undefined`, which is also the reason why `undefined` is disallowed a value for a successful query.

```suggestion
    return old;
```

juliusmarminge: Interesting, might have to look into the types then cause I get typeerrors if i return undefined 🤔 

CleanShot 2022-09-21 at 16 02 43@2x

juliusmarminge: Should we no longer have the `| undefined` then?
CleanShot 2022-09-21 at 16 06 22@2x

KATT (approved):

examples/next-prisma-starter/src/pages/index.tsx
```diff
@@ -18,6 +18,12 @@ const IndexPage: NextPageWithLayout = () => {
     },
   );
 
+  utils.post.list.setData((old) => {
+    old;
+    //^?
+    return old;
+  });
+
```
KATT, line 26:
```suggestion
```

examples/next-prisma-starter/src/pages/index.tsx
TkDodo: agree, it should be:

`Updater<TOutput | undefined, TOutput | undefined>`

see: https://github.com/TanStack/query/blob/454c58481f6f5d5c3e182e6e418d65fdb4a9ec1f/packages/query-core/src/queryClient.ts#L129-L133

PR #2763, opened 2022-09-20
fix(examples): pageProps is `{}` by default

### LINKED ISSUE #2754: bug(example): websockets example has type error

v10

### Describe the bug

https://github.com/trpc/trpc/tree/next/examples/next-prisma-starter-websockets has a type error when using latest version of `next-auth` [as shown here](https://github.com/trpc/examples-next-prisma-starter-websockets)

### To reproduce

See failing build at https://github.com/trpc/examples-next-prisma-starter-websockets

### Additional information

Probably fixable by upgrading `next-auth` and fixing type error

### 👨‍👧‍👦 Contributing

minsk-dev:
Closes #2754

## 🎯 Changes

Fixes Next type error (`pageProps` now is `{}` by default).

## ✅ Checklist

- [ ] ~~If necessary, I have added documentation related to the changes made.~~
- [ ] ~~I have added or updated the tests related to the changes made.~~

### DIFF
--- a/examples/next-prisma-starter-websockets/src/pages/_app.tsx
+++ b/examples/next-prisma-starter-websockets/src/pages/_app.tsx
@@ -5,7 +5,7 @@ import { trpc } from 'utils/trpc';
 
 const MyApp: AppType = ({ Component, pageProps }) => {
   return (
-    <SessionProvider session={pageProps.session}>
+    <SessionProvider session={(pageProps as any).session}>
       <Component {...pageProps} />
     </SessionProvider>
   );


minsk-dev:
It's this the best way to do this or creating a props type would be more useful (as it is a scaffold, too)?

examples/next-prisma-starter-websockets/src/pages/_app.tsx
```diff
@@ -5,7 +5,7 @@ import { trpc } from 'utils/trpc';
 
 const MyApp: AppType = ({ Component, pageProps }) => {
   return (
-    <SessionProvider session={pageProps.session}>
+    <SessionProvider session={(pageProps as any).session}>
```
KATT, line 8: We should be able to fix this without `any`

KATT (changes requested):

examples/next-prisma-starter-websockets/src/pages/_app.tsx
minsk-dev: @KATT Only if we create a type for it. But this is not intended behavior. `next@12.3.1` [fixes this](https://github.com/vercel/next.js/pull/40391). Do I just bump the version?

KATT: Yes we likely need to bump the version on `next` and `next-auth`

juliusmarminge: It should be 

const MyApp: AppType<{ session: Session | null }> = () => ...

minsk-dev: I will try bumping the version and if it doesn't work already I will add the type. Is that ok?

KATT: What Julius says above is probably right -- I didn't look into _why_ it started failing, I just noted the issue down

PR #2771, opened 2022-09-21
fix(example): fix websocket build

KATT:


examples/next-prisma-starter-websockets/src/pages/_app.tsx
```diff
@@ -5,7 +5,8 @@ import { trpc } from 'utils/trpc';
 
 const MyApp: AppType = ({ Component, pageProps }) => {
   return (
-    <SessionProvider session={pageProps.session}>
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    <SessionProvider session={(pageProps as any).session}>
```
KATT, line 9: This was too annoying to fix the type of 🤷

PR #2783, opened 2022-09-22
fix: resubscribe on reconnect

### LINKED ISSUE #2776: bug: Subscriptions are not re-registered on socket reconnect

```
  System:
    OS: Linux 5.4 Ubuntu 20.04.4 LTS (Focal Fossa)
    CPU: (16) x64 Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
    Memory: 28.23 GB / 31.32 GB
    Container: Yes
    Shell: 3.2.2 - /usr/bin/fish
  Binaries:
    Node: 14.18.3 - ~/.nvm/versions/node/v14.18.3/bin/node
    Yarn: 1.22.19 - /usr/bin/yarn
    npm: 8.18.0 - ~/.nvm/versions/node/v14.18.3/bin/npm
  Browsers:
    Chrome: 102.0.5005.61
```

WSL2 though...

### Describe the bug

When the socket connect is lost to the tRPC server and then reconnect subscriptions requests are not automatically sent out again. So subscriptions will be lost.

### To reproduce

```bash
clone https://github.com/aphex/trpc-socket-issue
cd server
npm i
cd ../client
npm i
```

Terminal 1 (in server)
```bash
npm start
```

Terminal 2 (in client)
```bash
npm run dev
```

Vue page and see uptime coming in from socket
See Terminal 1 log for subscription endpoint hit and uptime emitting

Kill Server in Terminal 1
Check Console, see socket disconnected
Restart Server
```bash
npm start
```

See console shows socket is reconnected, however subscription is never receievd
See Terminal 1 subscription endpoint is never hit again, emit never happens

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

aphex:
Initial POC of how reconnecting could work, likely has some bad side-effects though

Closes #2776 

## 🎯 Changes

First we don't queue a subscription.stop if we are not connected to the socket. If it is queued this line here https://github.com/trpc/trpc/blob/next/packages/client/src/links/wsLink.ts#L124 will keep the request to subscribe from being sent as they both share the same id

Second we prevent the error callback if we have not closed "normally". This will prevent this method from being called here https://github.com/trpc/trpc/blob/next/packages/client/src/links/wsLink.ts#L319 which keep things from being unsubscribed while the socket tried to reconnect.

## ✅ Checklist


aphex:
fixed up linting and trying to sneak in a minor TS change here https://github.com/trpc/trpc/pull/2783/files#diff-d5eb300ee573b9c1ced9ca84f2cb9be7746b78404935ac11c96c3bfa7d42a205R47

I would like to have access to this type in my plugin, so can we kindly export it :)

juliusmarminge:
> fixed up linting and trying to sneak in a minor TS change here https://github.com/trpc/trpc/pull/2783/files#diff-d5eb300ee573b9c1ced9ca84f2cb9be7746b78404935ac11c96c3bfa7d42a205R47
> 
> 
> 
> I would like to have access to this type in my plugin, so can we kindly export it :)

I think we are trying to keep these sort of types (I.e types needed for plugins and shouldn't necessarily be used in apps) exported from /shared (but maybe /integrations) would be a better name 🤔

aphex:
@juliusmarminge While you're considering that I wonder if you all would be up for bringing these types over from my composable project into tRPC proper? They let you flatten a router into a dot notation string and then get procedure info back from the flatten version like `namespace1.namespace2.procedure`.

This lets other implementations reference procedures my a single string but also keep type safety and auto-complete.

Maybe the naming is slightly wrong as its called `inferProcedureNames` but its both procedures and subscriptions. Whatever name needs to be used though would be nice not to have to repeat these in different projects.

https://github.com/aphex/use-trpc/blob/main/src/types.ts#L6

KATT (commented):
It looks good to me and I'm happy no existing tests fail, but I would like some tests to proves the new behavior is correct

It'd go in here: https://github.com/trpc/trpc/blob/next/packages/server/test/interop/websockets.test.ts

(We haven't fully migrated all tests to v10 but thanks to interop we can still test the behavior)

aphex:
Hey all bringing my thoughts over from discord to work through here. The focus now is not only getting resubscribe to work but to clean up the `stopped` and `complete` states.

As James mentioned `stopped` and `complete` are very similar. It is possible we could just have one of them however I am guessing they are supposed to describe two different situations. 

`Stopped` - When a subscription is not actively monitored anymore. This can happen when it is unsubscribed from or if the socket server loses connection to the client or the client isn't pinging and drops due to inactivity.

`Complete` - When the backing observable for the subscription has completed. Essentially saying there is no more data here to send. This is a pretty rare situation as I doubt most subscriptions "complete" like this, but I don't see why it shouldn't be possible. 

So technically I think a subscription could be completed and stopped. Though it does seem fair to say if a subscription is completed, it should just be unsubscribed automatically anyway. 

Right now completed and stopped are mixed into a few things. Particularly when it comes to errors and disconnects. 

Also since disconnecting is mixed up in this we cannot count on the server to push state to the client. So both need to be aware of these concepts. 

So this is how I could see it playing out.

Client side subscribe -> `onStarted`
Client side unsubscribe -> `onStopped`
Client side detects disconnect that was not intentional -> error thrown -> `onStopped`
Client side detects disconnect that was intentional by the server -> error thrown -> `onStopped`
Client side disconnects and it was that was intentional by the client -> `onStopped`
Server side observable completes -> `onComplete` -> `onStopped` (maybe? :man_shrugging: )

For non-intentional disconnects the client will retry connecting, once reconnected is will subscribe again -> `onStarted`
For intentional disconnects it will not try to reconnect (maybe it should? :man_shrugging: )

That seems to separate start/stop and complete as complete is now only an indication that the observable has no more data. 

However I could also see an argument that a subscription is just a single observable anyway so in the case of unsubscribe, disconnect, or complete it is always just a `onComplete` to the client. Subscribe technically does create a new observable. The downside to this is there is no sure way to know if an observable has actually completed, in the sense that there is no more data to be sent. 

Perhaps someone is using this observable like a timer, they are going to `next` every second but then `complete` after a minute. This could be for more standard stateful servers like a hosted nodeJS express server.  For the serverless version of this observables are pretty much ignored and subscriptions are fired by other events (dynamo stream, sns, sqs, etc) however it could be done with a step function or cron and still have the concept of `complete`

Let me know your thoughts, we can get this all cleared up. also for the record there is a websocket test alright to test disconnect and reconnect/resubscribe. It was just marked `skip` as its noted to be flakey. So we will make sure that tests is cleaned up and add a few others around this start/stopped/complete state once we land on a direction.

KATT:
@jlalmes, have you had a chance to look at this yet? :)

willosof:
Where would be the best place to throw money to get this resolved? 🤑

KATT (dismissed):

packages/client/src/links/loggerLink.ts
```diff
@@ -44,7 +44,7 @@ const palette = {
   mutation: ['c5a3fc', '904dfc'],
   subscription: ['ff49e1', 'd83fbe'],
 };
-type LoggerLinkOptions<TRouter extends AnyRouter> = {
+export type LoggerLinkOptions<TRouter extends AnyRouter> = {
```
KATT, line 47: Seems unrelated and causes conflicts
```suggestion
type LoggerLinkOptions<TRouter extends AnyRouter> = {
```

packages/client/src/links/wsLink.ts
```diff
@@ -212,11 +212,15 @@ export function createWSClient(opts: WebSocketClientOptions) {
         if (req.ws !== conn) {
           continue;
         }
-        req.callbacks.error?.(
-          TRPCClientError.from(
-            new TRPCWebSocketClosedError('WebSocket closed prematurely'),
-          ),
-        );
+
+        if (state === 'closed') {
```
KATT, line 216: This should be the _desired_ state, so we should log here if it's _not_ closed

```suggestion
        if (state !== 'closed') {
```

KATT: Actually I think we should always wait here.... we shouldn't try to close the connection if there are pending requests

KATT (dismissed):
Sorry for the insane wait-time, I thought someone else would look into this. My bad.


I have tweaked the logic and added comments. Feel free to setup GitHub Sponsors @aphex and I'll send you a thank you lunch or beer :)

aphex:
@KATT no worries on that, but thanks for the offer! Happy to help this project get better. Where is this now? Does it need testing or we all good to go?

Also from that long comment I had above, where did this land? What are the expected results for all the different `stop`,`start`, and `complete` states. I can pull this PR and test to confirm if needed.

KATT:
> @KATT no worries on that, but thanks for the offer! Happy to help this project get better. Where is this now? Does it need testing or we all good to go?

I'll merge it. The tests are pretty extensive so pretty sure it doesn't break any _existing behavior_.

> Also from that long comment I had above, where did this land? What are the expected results for all the different `stop`,`start`, and `complete` states. I can pull this PR and test to confirm if needed.

- `open` - WS conn is open and we can send messages
- `connecting` - we're waiting to connect
- `closed` we've _explicitly_ told the connection to close (but it won't be actually closed until it has finished the pending requests)

KATT (approved):

PR #2789, opened 2022-09-23
v10 migration doc cleanup.

anthonyshew:
Closes #

## 🎯 Changes

A bunch of cleanup!

### Most significant change

Wanted to see how doing the `Summary of Changes` as a collection of `<details>` tags felt.

My reasoning behind doing this is we want to get folks to the part where they actually perform the migration as quickly as we can. While it's important to talk about the changes that have been made, I suppose I could argue that that's what the whole of the docs is suppose to be doing. The `Migrating from v9` header is where our reader begins to take action with this particular page of our documentation so I like the focus being on that piece of the puzzle for this specific page.

If we make our reader take a long time reading the summary of changes, we might make the v10 migration look more daunting and scary than it actually is. `.interop()` makes this migration easy for 95% of our users so I would like to make sure that we make the migration look as easy as it really is.

By making the change summaries "opt-in" (opening the `<details>` tags), we are shortening the distance from the moment our reader opens the page to when they start taking action. But we still get to have the summary of changes on the page still!

## ✅ Checklist

### DIFF
--- a/www/docs/migration/migrate-from-v9-to-v10.mdx
+++ b/www/docs/migration/migrate-from-v9-to-v10.mdx
@@ -6,8 +6,10 @@ slug: /migrate-from-v9-to-v10
 ---
 
 :::info Is v10 ready for production?
-tRPC v10 is **ready for production** & can safely be used today. However, we will do changes to the API before we do a final release where you'll have to figure out the migration path between the releases on your own.
+tRPC v10 is **ready for production** & can safely be used today. There will be minor changes to the API comparing v10-beta to v10. You will need to figure out the migration path between the beta and general availability releases on your own.
 :::
+
+Welcome to tRPC v10! We're excited to bring you a new major version to continue the journey towards perfect end-to-end type safety with excellent DX.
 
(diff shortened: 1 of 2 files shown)


www/docs/migration/migrate-from-v9-to-v10.mdx
```diff
@@ -171,21 +138,58 @@ const appRouter = t.router({
...
+            error.code === 'BAD_REQUEST' && error.cause instanceof ZodError
+              ? error.cause.flatten()
+              : null,
+        },
+      };
+    },
+    transformer: superjson,
+  });
+```
+
```
KATT, line 185: Do we want all of this ina disclosure too maybe?

KATT (approved):
Very nice, agreed the disclosures are good, left one little question. 

Also added you as a contributor so if you push stuff to origin we'll get CI and stuff running :)

www/docs/migration/migrate-from-v9-to-v10.mdx
anthonyshew: I'm in for it!

KATT:
Superseded by #2816

PR #2809, opened 2022-09-25
refactor: modernize examples with Vitest and tsx

juliusmarminge:
Closes #

## 🎯 Changes

- updates all examples to use esbuild's `tsx` instead of `ts-node`. It's faster and overall simpler to use (no need to worry bout esm and shit).
  - I changed in root too but no perf diff there since we were using swc there
- nuke's Jest where possible in favor of Vitest.
- remove initialization of Playwright container. Necessary deps are instead installed via `@playwright/test` which can be cached.

## ✅ Checklist


KATT (commented):
Beware that both the nextjs example and the websockets one are also live so you can't do migration changes on them

- https://nextjs.trpc.io/
- https://websockets.trpc.io/

juliusmarminge:
Pretty much done. Must decide what to do with examples.

- Do we try making Postgres faster and keep it as we had, or swap to sqlite? If we do the latter we prob need some adjustments since we want live examples and sqlite not the best for that :/

- Some examples were using only Playwright (no unit/integration) so there is no need to have Vitest in those. Should I add it anyways so that they are ready to go and add tests when users need to and for demonstration purposes? (See next-prisma-starter [config](https://github.com/trpc/trpc/blob/optimize-ci-2/examples/next-prisma-starter/vitest.config.ts), [test](https://github.com/trpc/trpc/blob/optimize-ci-2/examples/next-prisma-starter/src/server/routers/post.test.ts))

sachinraja (dismissed):

juliusmarminge:
Leaving this here so it doesn't get lost in the discord too 

![image](https://user-images.githubusercontent.com/51714798/192169122-1aa3e1d1-2851-476e-a711-0cabef855022.jpeg

examples/next-prisma-todomvc/prisma/schema.prisma
```diff
@@ -2,8 +2,10 @@
 // learn more about it in the docs: https://pris.ly/d/prisma-schema
 
 datasource db {
-  provider = "postgres"
-  url      = env("DATABASE_URL")
+  // provider = "postgres"
+  // url      = env("DATABASE_URL")
+  provider = "sqlite"
+  url      = "file:./dev.db"
```
KATT, line 8: Breaks https://todomvc.trpc.io/, please reset

examples/next-prisma-starter/.github/workflows/main.yml
```diff
@@ -3,41 +3,50 @@ on: [push]
...
 
-      # - name: Install deps and build (with cache)
-      #   uses: bahmutov/npm-install@v1
-      - run: yarn install
+      - name: Install deps (with cache)
+        run: pnpm install
 
       - name: Next.js cache
-        uses: actions/cache@v2
+        uses: actions/cache@v3
```
KATT, line 43: Be extra careful when editing this -- we won't know if it breaks until it's subtree'd into https://github.com/trpc/examples-next-prisma-starter

examples/next-prisma-starter-sqlite/.github/workflows/main.yml
```diff
@@ -3,41 +3,41 @@ on: [push]
...
-          key: ${{ runner.os }}-${{ runner.node }}-${{ hashFiles('**/yarn.lock') }}-nextjs
+          key: ${{ runner.os }}-${{ runner.node }}-${{ hashFiles('**/pnpm-lock.yaml') }}-nextjs
 
-      - run: yarn playwright install --with-deps
-      - run: yarn lint
-      - run: yarn build
-      - run: yarn test-start
-      - run: yarn test-dev
+      - name: Build and test
+        run: pnpm build && pnpm test-start && pnpm test-dev
```
KATT, line 40: Be careful, see other ocmments

.github/workflows/examples.yml
```diff
@@ -72,20 +61,16 @@ jobs:
...
-        uses: actions/cache@v3
-        with:
-          key: ${{ runner.os }}-playwright-bin-v1.26.0
-          path: ${{ env.PLAYWRIGHT_BROWSERS_PATH }}
-
       - run: node -v
       - name: Install deps and build (with cache)
         run: pnpm --filter ./examples/${{ matrix.dir }}  --filter "@trpc/*" --filter root install
 
+      - run: echo ${{ github.workspace }}/${{ matrix.dir }}/.next/cache
```
juliusmarminge, line 68: the nextjs cache is never hitting, any ideas @sachinraja. I added this to try see if the path was wrong but seems to be good 🤔

examples/next-prisma-starter/.github/workflows/main.yml
juliusmarminge: Tested it here: https://github.com/juliusmarminge/next-prisma-starter/actions

examples/next-prisma-starter-sqlite/.github/workflows/main.yml
juliusmarminge: Tested it here: https://github.com/juliusmarminge/next-prisma-starter-sqlite/actions

examples/next-prisma-starter-websockets/.github/workflows/main.yml
```diff
@@ -21,20 +22,36 @@ jobs:
...
-      - run: yarn build
-      - run: yarn test-start
-      - run: yarn test-dev
+      - name: Upload test results
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: test results
+          path: |
+            playwright/test-results
```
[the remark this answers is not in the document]
juliusmarminge: Tested it here: https://github.com/juliusmarminge/next-prisma-starter-websockets/actions

juliusmarminge:
Should be all done now. I created mirror-repos for the examples that had "internal workflows" to make sure they'll work when they're subtree'd:

- https://github.com/juliusmarminge/next-prisma-starter
- https://github.com/juliusmarminge/next-prisma-starter-sqlite
- https://github.com/juliusmarminge/next-prisma-starter-websockets

However seems like the migrations on the `next-prisma-starter` has gone corrupt and I'm not sure how to fix it. Would you mind taking a look @KATT?

examples/.interop/next-prisma-starter-websockets/package.json
```diff
@@ -47,19 +50,18 @@
     "@trpc/server": "^10.0.0-proxy-beta.8",
     "clsx": "^1.1.1",
     "next": "^12.3.1",
-    "next-auth": "^4.10.3",
+    "next-auth": "~4.10.3",
```
juliusmarminge, line 53: next-auth@^4.11.0 requires next@12.2.5 as peerDep so locking this at 4.10 for now

KATT (changes requested):
How come we're double-wielding sqlite and Postgres in all the examples?

KATT:
> > How come we're double-wielding sqlite and Postgres in all the examples?
> 
> It's cause starting a Postgres container in CI takes ~1min so by doing this we can run sqlite in CI and Postgres for the live ones and save time

I like the thinking but I don't want to maintain two schemas and two migration sets for every example... also it's confusing if examples includes migrations etc for _two_ databases

juliusmarminge:
> > > How come we're double-wielding sqlite and Postgres in all the examples?
> > 
> > 
> > It's cause starting a Postgres container in CI takes ~1min so by doing this we can run sqlite in CI and Postgres for the live ones and save time
> 
> I like the thinking but I don't want to maintain two schemas and two migration sets for every example... also it's confusing if examples includes migrations etc for _two_ databases

We don't need to include migrations for the sqlite if we don't want to. Just the schema is fine

KATT:
> > > > How come we're double-wielding sqlite and Postgres in all the examples?
> > > 
> > > 
> > > It's cause starting a Postgres container in CI takes ~1min so by doing this we can run sqlite in CI and Postgres for the live ones and save time
> > 
> > 
> > I like the thinking but I don't want to maintain two schemas and two migration sets for every example... also it's confusing if examples includes migrations etc for _two_ databases
> 
> We don't need to include migrations for the sqlite if we don't want to. Just the schema is fine

Can we temporarily create the migration files as well maybe as part of CI? Should be a search & replace of `datasource db { .. }`

KATT:
I'm happy to merge this as-is and we can do those that as follow-ups! :)

examples/next-prisma-starter-websockets/.github/workflows/main.yml
```diff
@@ -3,38 +3,45 @@ on: [push]
...
+        node: ['16.x']
         os: [ubuntu-latest]
-    services:
-      postgres:
-        image: postgres:12.1
-        env:
-          POSTGRES_USER: postgres
-          POSTGRES_DB: trpcdb
-        ports:
-          - 5432:5432
```
sachinraja, line 21: for an actual app, I would want to test against the same database. we could merge and fix this in a follow-up but might be easier to just revert in this PR

sachinraja (approved):

PR #2811, opened 2022-09-25
docs(ssg-helpers): Improve SSG helpers with `prefetch` vs. `fetch` and real context creation

bennettdams:
## 🎯 Changes

- Replace `fetch` with `prefetch` and add a comment to mention their use cases.
We now use `prefetch` instead and also show a small explanation for the difference.

- Import statement for the context creation function
- Use a real example of filling the `ctx` field via `await createContext()`.

  This makes sure to await the context creation when using the SSG helpers in `v10`.
  I'm not sure if this is really needed regarding the implementation, but without it, there's a TypeScript error:
  
  image
  
  
  image
  
  
  For reference, this is how my `createContext` function looks like:
  
  
  ```ts
  import * as trpcNext from '@trpc/server/adapters/next'
  import { prisma } from '../prisma/prisma'
  
  /**
   * Creates context for each incoming request.
   * Will be available in all resolvers.
   *
   * @link https://trpc.io/docs/context
   */
  export async function createContext(opts?: trpcNext.CreateNextContextOptions) {
    const req = opts?.req
    const res = opts?.res
  
    return {
      req,
      res,
      prisma,
      // userIdAuth,
    }
  }
  
  export type ContextTRPC = trpc.inferAsyncReturnType<typeof createContext>
  ```
---
### Out of scope

By the way: When using `createNextApiHandler`, just passing through the context creation function is fine. Not sure if different, but maybe a good idea to unify the expected types:

```ts
import { createContextTRPC } from '../../../server/context-trpc'

import * as trpcNext from '@trpc/server/adapters/next'
import { appRouter } from '../../../server/routers/_app'

  router: appRouter,
  // #####  🟦 this is fine!
  createContext: createContextTRPC,
  ...
})

```

## ✅ Checklist

### DIFF
--- a/www/docs/nextjs/ssg-helpers.md
+++ b/www/docs/nextjs/ssg-helpers.md
@@ -40,8 +40,11 @@ export async function getServerSideProps(
   });
   const id = context.params?.id as string;
 
-  // Prefetch `post.byId`
-  await ssg.post.byId.fetch({ id });
+  /*
+   * Prefetching the `post.byId` query here.
+   * `prefetch` does not return the result - if you need that, use `fetch` instead.
+   */
+  await ssg.post.byId.prefetch({ id });
 
   // Make sure to return { props: { trpcState: ssg.dehydrate() } }
   return {
--- a/www/docs/nextjs/ssg-helpers.md
+++ b/www/docs/nextjs/ssg-helpers.md
@@ -9,10 +9,11 @@ slug: /ssg-helpers
 
 ```ts
 import { createProxySSGHelpers } from '@trpc/react/ssg';
+import { createContext } from 'server/context';
 
 const ssg = createProxySSGHelpers({
   router: appRouter,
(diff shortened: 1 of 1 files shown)


www/docs/nextjs/ssg-helpers.md
```diff
@@ -39,8 +40,11 @@ export async function getServerSideProps(
   });
   const id = context.params?.id as string;
 
-  // Prefetch `post.byId`
-  await ssg.post.byId.fetch({ id });
+  /*
+   * Prefetching the `post.byId` query here.
+   * `prefetch` does not return the result - if you need that, use `fetch` instead.
```
KATT, line 45:
```suggestion
   * `prefetch` does not return the result and never throws - if you need that behavior, use `fetch` instead.
```

KATT (approved):

PR #2871, opened 2022-10-01
change from pnpmfile to pnpm overrides

sachinraja:
fixes Renovate, it doesn't like our postinstall scripts :(


KATT (approved):
Haha, the amount of hacks we have in postinstall is getting a bit out of hands 🤦 😄

PR #2884, opened 2022-10-01
ability to pass in a custom query client

### LINKED ISSUE #2843: feat(trpc/next): allow passing complete QueryClient instead of queryClientConfig

When using `createTRPCNext`, we can pass in a `queryClientConfig` in the [config-callback](https://trpc.io/docs/v10/nextjs#config-callback), for example:

```
export const trpc = createTRPCNext<AppRouter>({
    config({ ctx }) {
        return {
            queryClientConfig: { defaultOptions: { queries: { staleTime: 20 * 1000 } } },
        }
    },
    ssr: true,
})
```

trpc then creates the queryClient for us:

https://github.com/trpc/trpc/blob/18a2757a996187f4637f3bcd7ba3080e3c0198a1/packages/next/src/withTRPC.tsx#L105

https://github.com/trpc/trpc/blob/18a2757a996187f4637f3bcd7ba3080e3c0198a1/packages/next/src/withTRPC.tsx#L174

However, the config alone is not enough to define all defaults. If you want to use [queryClient.setQueryDefaults](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientsetquerydefaults) to set default values for some queries only, we often do this:

```
const [queryClient] = React.useState(() => {
  const client = new QueryClient({ defaultOptions: { queries: { staleTime: 20 * 1000 } } })

  client.setQueryDefaults(['todos'], { staleTime: Infinity })

  return client
})
```

this will make sure that:

- all queries have a default staleTime of 20 seconds
- queries that match the `['todos']` key have a default staleTime of Infinity

This is currently not possible to achieve in trpc because we don't have access to the queryClient after creation, but before it is put into the `QueryClientProvider`.

I'd like to pass the complete `QueryClient` to `createTRPCNext` instead of just the config. If we create the QueryClient on consumer side, the problem would not exist.

## alternative 1

`trpc.useContext()` could grant access to type-safe wrappers of `setQueryDefaults` / `setMutationDefaults`, like it does for `invalidate`. However:

- this access must exist on the top level (each nested router level) because we likely want fuzzy matching, so it depends on:
  - https://github.com/trpc/trpc/pull/2713
- we would technically need to read-and-set this in an effect because the operation itself is not pure, but if we render our app in the meantime, the queries would pick up the wrong defaults, and also, it wouldn't work on the server.

## alternative 2

we could amend the api of the `QueryClient` constructor in react-query so that multiple, imperative calls to `setQueryDefaults` become unnecessary. I am considering this (open for api proposals, too), but I think it would be an orthogonal change to trpc because calling `queryClient.setQueryDefaults` is still something you can do at any time in react-query, but not really in trpc at the moment.

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

KATT:
Closes #2843

## 🎯 Changes

Ability to pass in custom query client

- [x] Code
- [x] Test
- [x] Docs


packages/next/src/withTRPC.tsx
```diff
@@ -102,7 +110,8 @@ export function withTRPC<
         }
 
         const config = getClientConfig({});
-        const queryClient = new QueryClient(config.queryClientConfig);
+        const queryClient =
+          config.queryClient ?? new QueryClient(config.queryClientConfig);
```
TkDodo, line 114: just for my understanding: there's a second mention of `new QueryClient` in this file:

https://github.com/trpc/trpc/blob/09f41cda414ca9615e783da0ce9bffee81fc6b2e/packages/next/src/withTRPC.tsx#L183

wouldn't we need to make that change there as well?

sachinraja: It's also in a few other places, like the ssg helpers.

TkDodo (dismissed):

sachinraja (approved):

KATT:
@TkDodo I'm happy to merge this. Where do you think the docs should live? Feel free to just to do some modifications to `/www` on this branch if you want

TkDodo:
@KATT I added documentation for the `queryClient` config option.

- [1d5028c](https://github.com/trpc/trpc/commit/1d5028c4084934750a9137a948a1d85913cafed1)

PR #2885, opened 2022-10-01
Custom overrides

KATT:
Closes #

## 🎯 Changes

## ✅ Checklist

### DIFF
--- a/packages/react/src/shared/types.ts
+++ b/packages/react/src/shared/types.ts
@@ -1,0 +1,24 @@
+import { QueryClient } from '@tanstack/react-query';
+import { AnyRouter, MaybePromise } from '@trpc/server';
+
+export interface UseMutationOverride {
+  onSuccess: (opts: {
+    /**
+     * Calls the original function that was defined in the query's `onSuccess` option
+     */
+    originalFn: () => MaybePromise<unknown>;
+    queryClient: QueryClient;
+  }) => MaybePromise<unknown>;
+}
+
+/**
+ * @internal
+ */
+export interface CreateTRPCReactOptions<_TRouter extends AnyRouter> {
+  /**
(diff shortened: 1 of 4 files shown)


packages/react/src/shared/types.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @internal
+ */
+export interface UseMutationOverride {
+  onSuccess: (opts: {
+    /**
+     * Calls the original function that was defined in the query's `onSuccess` option
+     */
+    originalFn: () => MaybePromise<unknown>;
+    queryClient: QueryClient;
+  }) => MaybePromise<unknown>;
```
juliusmarminge, line 14: allow overriding `onError`, `onMutate` etc too?

packages/react/src/shared/types.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * @internal
+ */
+export interface CreateTRPCReactOptions<_TRouter extends AnyRouter> {
+  /**
+   * Override behaviors of the built-in hooks
+   */
+  unstable_overrides?: {
+    useMutation?: Partial<UseMutationOverride>;
+  };
```
juliusmarminge, line 26: allow overriding of other hooks too?

juliusmarminge: not sure what the use case of overriding useQuery might be though.

packages/react/src/shared/types.ts
KATT: Meh. I don't use it

packages/react/src/shared/types.ts
KATT: Meh. Maybe later! :D

sachinraja (approved):

Issue #2568, opened 2022-08-28
[v10] bug: TypeScript not willing to serialize the type of my client


anthonyshew:
### Provide environment information

  System:
    OS: macOS 12.5
    CPU: (10) arm64 Apple M1 Pro
    Memory: 152.64 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.16.0/bin/yarn
    npm: 8.11.0 - ~/.nvm/versions/node/v16.16.0/bin/npm
  Browsers:
    Chrome: 104.0.5112.101
    Safari: 15.6

### Describe the bug

I am seeing an error in my editor saying: `The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.` on my `setupTRPC` for my Next.js project.

This is coming after a migration from v9 to v10 where I was hoping to see this problem resolve. However, it seems to have moved from my root router to the `setupTRPC`.

This leads me to believe that the type annotation is getting too long for the TS interpreter in VSCode so it's punting the effort/complaining. However, I'll let the real TS genies figure it out. 😄 

### To reproduce

Providing a minimal reproduction might not be applicable here? This bug occurs because of lots of code, not little code. 😄 

Rather,[ here is a link to my source code](https://github.com/doggo-sports/coregi/tree/111979699addbf6983aebbdc31f92c02e0f5b58a).

You'll find the TS error on [this line](https://github.com/doggo-sports/coregi/blob/111979699addbf6983aebbdc31f92c02e0f5b58a/packages/trpc/nextjs/client/appWrapper.ts#L14).

Removing [these two falsy declaration values](https://github.com/doggo-sports/coregi/blob/111979699addbf6983aebbdc31f92c02e0f5b58a/packages/trpc/tsconfig.json#L7) from tsconfig will reveal the error!

### 👨‍👧‍👦 Contributing

- [ ] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

KATT:
Thanks for having the repo, helps a lot to have a repo to play with! Would you be able to add me as a contributor? Then I can play around with different things as soon as I get the time

sachinraja:
Thanks for creating this! It makes sense as our proxy types are huge to emit. But users shouldn't need to emit types for the client.

[...]

PR #2890, opened 2022-10-02
reduce declaration size

sachinraja:
#2568

### DIFF
--- a/packages/server/src/core/internals/procedureBuilder.ts
+++ b/packages/server/src/core/internals/procedureBuilder.ts
@@ -68,16 +68,18 @@ type OverwriteIfDefined<TType, TWith> = UnsetMarker extends TType
 
 type ErrorMessage<TMessage extends string> = TMessage;
 
-export type ProcedureBuilderDef = {
+export type ProcedureBuilderDef<TParams extends ProcedureParams> = {
   inputs: Parser[];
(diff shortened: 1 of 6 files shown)


packages/server/src/core/router.ts
```diff
@@ -188,25 +190,30 @@ const reservedWords = [
 ];
 
 /**
- *
+ * @internal
+ */
+export type CreateRouterInner<
```
sachinraja, line 195: Switching to an interface here makes TypeScript emit the named type instead of actually resolving the type and emitting that. This reduces the size of the server declarations by half.

packages/server/src/core/internals/procedureBuilder.ts
```diff
@@ -44,12 +42,43 @@ type CreateProcedureReturnInput<
   _output_out: FallbackValue<TNext['_output_out'], TPrev['_output_out']>;
 }>;
 
+/**
+ * @internal
+ */
+export interface BuildProcedure<
```
sachinraja, line 48: This doesn't actually help but I did it and it makes our source code more concise anyway :shrug:

KATT (approved):
Lil' type mismatch around Meta in the test but this looks sold. Thanks for sharing _why_ it works

PR #2895, opened 2022-10-02
import package mocks in test helpers

sachinraja:
`___packages` was imported in the `__testHelpers` before but I removed it for some reason. This fixes an error when running Jest in watch mode


KATT (approved):
Oh shit I've noticed this

PR #2896, opened 2022-10-02
feat(ci): test sqlite swapper

### LINKED ISSUE #2835: Use proper dbs in CI
... or find a way to dynamically swapping between sqlite/pg

Pros:

- Easier to grok examples
- Tests represent reality better

Cons:

- Slower CI

>        for an actual app, I would want to test against the same database. we could merge and fix this in a follow-up but might be easier to just revert in this PR

_Originally posted by @sachinraja in https://github.com/trpc/trpc/pull/2809#discussion_r983043748_

juliusmarminge:
Closes #2835

## 🎯 Changes

[Made this](https://github.com/juliusmarminge/prisma-sqlite-ci) the other day, testing if it works for our use-case.

## ✅ Checklist


juliusmarminge:
Leave your thoughts before I do the rest of the examples.

I might tidy up [the action](https://github.com/juliusmarminge/prisma-sqlite-ci) and do a publish, in which case we could use that instead of keeping the source.

KATT (commented):
I think we should have a stab at #2937 ahead of going further down the route of using SQLite

juliusmarminge:
> I think we should have a stab at #2937 ahead of going further down the route of using SQLite

Agree if we can get snappy Postgres that's better!

PR #2904, opened 2022-10-03
add `isDev` variable

KATT:
## 🎯 Changes

- Programmatically supply `isDev: boolean` rather than relying on `process.env.NODE_ENV`

## ✅ Checklist

### DIFF
--- a/packages/server/src/core/router.ts
+++ b/packages/server/src/core/router.ts
@@ -79,6 +79,7 @@ export interface RouterDef<
    * @deprecated
    */
   subscriptions: {};
+  isDev: boolean;
 }
 
 export type AnyRouterDef<TContext = any> = RouterDef<TContext, any, any, any>;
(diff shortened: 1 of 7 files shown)


packages/server/src/core/router.ts
```diff
@@ -209,8 +212,12 @@ export type CreateRouterInner<
  * @internal
  */
 export function createRouterFactory<TConfig extends RootConfig>(
-  defaults?: RouterDefaultOptions<TConfig['ctx']>,
+  defaults: RouterDefaultOptions<TConfig['ctx']>,
 ) {
+  const transformer = defaults?.transformer ?? defaultTransformer;
+  const errorFormatter = defaults?.errorFormatter ?? defaultFormatter;
```
iduuck, line 218: The optional chain operator is not needed here, since the `?` flag was removed from `defaults`.

iduuck: However, is it really "safe" to make it non-optional?

iduuck (changes requested):

KATT:
Superseded by #2907

PR #2907, opened 2022-10-03
feat: rm `process.env` deps &  refactor root config

KATT:
Closes #

## 🎯 Changes

- Add `allowOutsideOfServer` & `isServer` so we can run tRPC in non-server environments (like the browser)
- Add `isDev: boolean` to signal if we're running our tRPC server in dev mode 
  - Only affects whether or not to send the server's stack trace on tRPC errors right now
- Refactor RootConfig to make a bit more sense:
  - It now consists of two parts: the Root Config **Types** (the generics) & the Runtime **Config** (stuff that are actually used (which can also affect the types))
  - Generics are stored at `$types` and wrapped in a proxy in case they're accessed at runtime
- ~I commented-out some tests that are `// Maybe *` that might be nice to fix for interop safety?~ fixed
- Remove `_ctx_in` on procedures as that can be derived from the config

## ✅ Checklist


packages/server/src/core/router.ts
```diff
@@ -41,48 +36,32 @@ export interface ProcedureStructure {
...
    */
-  mutations: {};
+  mutations: any;
   /**
    * V9 subscriptions
    * @deprecated
    */
-  subscriptions: {};
-  isDev: boolean;
+  subscriptions: any;
```
sachinraja, line 60: I think the example is erroring because of these

KATT: yeah, but not having it broke other stuff

KATT: i wanted the type `AnyRouter` and it's "children" to be _somewhat_ typesafe so we got autocompletion/typesafety when using 

```ts
const router: AnyRouter = null as any;

router._def._config
//                          ^  i want autocompletion etc here
```

.eslintrc
```diff
@@ -53,7 +53,7 @@
         "format": ["PascalCase"],
         "leadingUnderscore": "allow",
         "custom": {
-          "regex": "^T[A-Z][a-zA-Z]+[0-9]*$",
+          "regex": "^(\\$|T)[A-Z][a-zA-Z]+[0-9]*$",
```
KATT, line 56: I like `$MyType` for things that are derived from fn arguments

packages/server/src/core/internals/config.ts
```diff
@@ -11,6 +7,9 @@ import {
 export interface InitGenerics {
   ctx: Record<string, unknown>;
   meta: Record<string, unknown>;
+  // FIXME this should be typed
```
KATT, line 10: maybe

.eslintrc
```diff
@@ -53,7 +53,7 @@
         "format": ["PascalCase"],
         "leadingUnderscore": "allow",
         "custom": {
-          "regex": "^T[A-Z][a-zA-Z]+[0-9]*$",
+          "regex": "^(T|\\$)[A-Z][a-zA-Z]+[0-9]*$",
```
KATT, line 56: I like `$` for types that are derived from arguments

packages/server/src/core/internals/config.ts
```diff
@@ -23,40 +21,62 @@ export const isServerDefault: boolean =
...
+  $types: TGenerics;
 }
 
 /**
  * @internal
  */
-export type CreateRootConfig<TConfig extends RootConfig> = TConfig;
+export type AnyRootConfig = RootConfig<{
+  ctx: any;
+  meta: any;
```
KATT, line 79: Would be slightly nicer if these were


```suggestion
  ctx: Record<any, any>;
  meta: Record<any, any>;
```

sachinraja (approved):
Awesome work!

packages/server/src/core/initTRPC.ts
```diff
@@ -40,22 +43,22 @@ type CreateInitGenericsFromPartial<TType extends PartialInitGenerics> =
  * - Doesn't need to be a class but it doesn't really hurt either
  */
 
-class TRPCBuilder<TParams extends Partial<InitGenerics> = {}> {
-  context<TNewContext extends InitGenerics['ctx']>() {
+class TRPCBuilder<TParams extends Partial<RootConfigTypes> = {}> {
```
bautistaaa, line 46: can you reuse `PartialRootConfigTypes` like on line 31

packages/server/src/core/initTRPC.ts
```diff
@@ -68,15 +71,15 @@ class TRPCBuilder<TParams extends Partial<InitGenerics> = {}> {
  */
 export const initTRPC = new TRPCBuilder();
 
-function createTRPCInner<TParams extends Partial<InitGenerics>>() {
-  type $Generics = CreateInitGenericsFromPartial<TParams>;
+function createTRPCInner<TParams extends Partial<RootConfigTypes>>() {
```
bautistaaa, line 74: can you reuse `PartialRootConfigTypes`?

KATT: Wanna do a PR and tweak these? ☺

bautistaaa: Yep!

PR #2956, opened 2022-10-10
docs: add example of `unstable_overrides`

juliusmarminge:
Ref: #2954

Thought it would be nice with a simple example


KATT (commented):
This is not enough? :) https://trpc.io/docs/v10/query-invalidation#invalidate-full-cache-on-every-mutation

juliusmarminge:
> This is not enough? :) https://trpc.io/docs/v10/query-invalidation#invalidate-full-cache-on-every-mutation

Oh I must've misread that hard lol 🤦🏼‍♂️

PR #2968, opened 2022-10-10
feat: make false link in splitLink optional

### LINKED ISSUE #2929: feat: make `false` link in `splitLink` optional

While I was doing the Link documentation overhaul, I noticed in the [Apollo link docs](https://www.apollographql.com/docs/react/api/link/introduction/#directional-composition) that their `split` implementation has the `false` link as optional. I thought it would be cool to add that functionality to tRPC's own `splitLink` so that in the case the user just wants a conditional link they can use it without supplying a fallback.

In simpler terms, right now `splitLink` just provides an "if else" flow, and I would like it if it could also offer the "if" flow, without the else.

Basic implementation would be something like:

```ts
export function splitLink<TRouter extends AnyRouter = AnyRouter>(opts: {
  condition: (op: Operation) => boolean;
  /**
   * The link to execute next if the test function returns `true`.
   */
  true: TRPCLink<TRouter> | TRPCLink<TRouter>[];
  /**
   * The link to execute next if the test function returns `false`.
   */
  false?: TRPCLink<TRouter> | TRPCLink<TRouter>[];
}): TRPCLink<TRouter> {
  return (runtime) => {
    const yes = asArray(opts.true).map((link) => link(runtime));
    const no = opts.false
      ? asArray(opts.false).map((link) => link(runtime))
      : [];
    return (props) => {
      return observable((observer) => {
        const passesCondition = opts.condition(props.op);
        if (passesCondition === false && !opts.false) {
          return props.next(props.op).subscribe(observer);
        }
        const links = passesCondition ? yes : no;
        return createChain({ op: props.op, links }).subscribe(observer);
      });
    };
  };
}
```

Not add it and leave it as is.

### Additional information

Tests would need to be added to validate that it works properly.
Documentation would need to be modified

### 👨‍👧‍👦 Contributing

rhenriquez28:
Closes #2929 

## 🎯 Changes

I modified `splitLink` so that `false` is optional in the case the user doesn't want to provide fallback when the condition isn't met. I've also added clarification on how `splitLink` actually works in the docs.

## ✅ Checklist

### DIFF
--- a/packages/client/src/links/splitLink.test.ts
+++ b/packages/client/src/links/splitLink.test.ts
@@ -1,7 +1,6 @@
 /* eslint-disable @typescript-eslint/no-empty-function */
 import { observable } from '@trpc/server/observable';
-import { splitLink } from '../';
-import { OperationLink, TRPCLink } from '../';
+import { OperationLink, TRPCLink, splitLink } from '../';
 import { AnyRouter } from '../../../server/src';
 import { createChain } from '../links/internals/createChain';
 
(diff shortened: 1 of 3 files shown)


KATT (changes requested):
Sorry, I missed your feature request the other day. 

I think this is a really solid contribution and you did a great job, but I don't think we should have this functionality. 

I think it's better to explicitly define a link that throws than it is to allow optionality. It's one of those features that might cause unexpected behavior because the user simply forgot the `false`-link

Issue #2923, opened 2022-10-05
bug: Prefetching with SSG helpers is broken since beta.14 (`v10`)


bennettdams:
### Provide environment information

```
  System:
    OS: Linux 5.10 Ubuntu 20.04.5 LTS (Focal Fossa)
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 16.17.0 - /usr/bin/node
    npm: 8.15.0 - /usr/bin/npm
  npmPackages:
    @tanstack/react-query: ^4.9.0 => 4.9.0 
    @trpc/client: ^10.0.0-proxy-beta.13 => 10.0.0-proxy-beta.13 
    @trpc/next: ^10.0.0-proxy-beta.13 => 10.0.0-proxy-beta.13 
    @trpc/react: ^10.0.0-proxy-beta.13 => 10.0.0-proxy-beta.13 
    @trpc/server: ^10.0.0-proxy-beta.13 => 10.0.0-proxy-beta.13 
    next: 12.3.1 => 12.3.1 
    react: ^17.0.2 => 17.0.2 
    typescript: 4.8.4 => 4.8.4
```

### Describe the bug

Prefetching via SSG Helpers is broken since `10.0.0-proxy-beta.14`.
No matter if you prefetch on the server or not, the query will be refetched on the client.

Some particular change that maybe is responsible: https://github.com/trpc/trpc/commit/1d187b11f83dade3a0e5643a66b3b4fe5974070c

This change happened for beta.14, and the prefetching was still working in beta.13.

### Link to reproduction

https://github.com/bennettdams/trpc-10-unmounted-bug

## Via reproduction repo :

(the repo was created for another issue, but works for this issue as well)

Steps:

1. `npm i`
2. `npm run dev`
3. Go to <http://localhost:3000/user/123>
4. See in the devtools that the prefetched query is **not** fetched on the client.
5. `npm install-14` (to install beta.14)
6. Restart, `npm run dev`
3. Go to <http://localhost:3000/user/123>
4. See in the devtools that the prefetched query **is** fetched on the client.

![image](https://user-images.githubusercontent.com/29319414/194181230-cc35211f-6e40-4cce-b167-6cccd2c4346a.png)

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

KATT:
This depends on your query client config in react-query and isn't related to tRPC as such.

PR #2977, opened 2022-10-11
fix: apply array query key normalization in SSG helpers

freeatnet:
Related to #2923, https://github.com/trpc/trpc/discussions/2925#discussioncomment-3851754.

## 🎯 Changes

This PR updates SSG helpers to use the same `pathAndInput` normalization as was added to the hooks in https://github.com/trpc/trpc/pull/2713.

## ✅ Checklist


packages/react/src/internals/getArrayQueryKey.ts
```diff
@@ -0,0 +1,17 @@
...
+ * path passed around internally by both the legacy and proxy implementation.
+ * https://github.com/trpc/trpc/issues/2611
+ */
+export function getArrayQueryKey(
+  queryKey: string | [string] | [string, ...unknown[]] | unknown[],
+): [string[]] | [string[], ...unknown[]] | [] {
+  const queryKeyArrayed = Array.isArray(queryKey) ? queryKey : [queryKey];
+  const [path, ...input] = queryKeyArrayed;
+
+  const arrayPath = path ? path.split('.') : [];
```
freeatnet, line 14: Heads-up: argument types technically allow an array in the first position (via `unknown[]`), but an array can't be `.split`.

freeatnet, line 14: @KATT Heads-up: argument types technically allow an array in the first position (via `unknown[]`), but an array can't be `.split`.

KATT (approved):

packages/react/src/internals/getArrayQueryKey.ts
KATT: I think I've simplified it now... it's a bit of a mess still but it seems to work assuming that the tests are right for invalidation

Issue #2914, opened 2022-10-05
bug: multiple input parsers with optional params


KATT:
### Provide environment information

v10 latest stuff

### Describe the bug

When combining two inputs where one has an optional field, the procedure shouldn't have to be called with the optional one

### Link to reproduction

https://stackblitz.com/edit/trpc-trpc-cs7n2b?file=src%2Fpages%2Fapi%2Ftrpc%2F[trpc].ts&file=src%2Fpages%2Findex.tsx

#### Backend

```ts

const roomProcedure = t.procedure.input(
  z.object({
    roomId: z.string(),
  })
);

const appRouter = t.router({
  roomQuery: roomProcedure
    .input(
      z.object({
        optionalParam: z.string().optional(),
      })
    )
    .query(({ input }) => {
      return { input };
    }),
});

```

#### Usage

```tsx
export default function IndexPage() {
  const result = trpc.roomQuery.useQuery({
    roomId: 'asd',
    // This should below should not be required
    // optionalParam: 'x'
  });
})
```

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

PR #2980, opened 2022-10-11
bug: optional input types are lying

juliusmarminge:
Related : #2914

Works on type-level but fails in runtime

### DIFF
--- a/packages/server/test/input.test.ts
+++ b/packages/server/test/input.test.ts
@@ -124,6 +124,33 @@ describe('double input validator', () => {
   });
 });
 
+test('happy-ish', async () => {
+  const t = initTRPC.create();
+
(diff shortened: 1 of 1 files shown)


KATT (commented):
I suggest we do something that prevents users from doing this

Similar:

- https://github.com/trpc/trpc/blob/d9914b26e761f2a02a5cc7b53b5675e19b8dc409/packages/server/test/input.test.ts#L127-L154
- https://github.com/trpc/trpc/blob/d9914b26e761f2a02a5cc7b53b5675e19b8dc409/packages/server/src/core/internals/procedureBuilder.ts#L94

It could be something like _"Impossible merge you can't override a defined property with an optional one"_

Issue #2919, opened 2022-10-05
bug: Hydration logic error when prefetching existing query in subroute (`v10` / Next.js)


bennettdams:
### Provide environment information

```
  System:
    OS: Linux 5.10 Ubuntu 20.04.5 LTS (Focal Fossa)
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 16.17.0 - /usr/bin/node
    npm: 8.15.0 - /usr/bin/npm
  npmPackages:
    @tanstack/react-query: ^4.9.0 => 4.9.0 
    @trpc/client: ^10.0.0-proxy-beta.13 => 10.0.0-proxy-beta.13 
    @trpc/next: ^10.0.0-proxy-beta.13 => 10.0.0-proxy-beta.13 
    @trpc/react: ^10.0.0-proxy-beta.13 => 10.0.0-proxy-beta.13 
    @trpc/server: ^10.0.0-proxy-beta.13 => 10.0.0-proxy-beta.13 
    next: 12.3.1 => 12.3.1 
    react: ^17.0.2 => 17.0.2 
    typescript: 4.8.4 => 4.8.4
```

### Describe the bug

Under specific circumstances, there is some error in the hydration logic.

The setup is the following:
 - Index route that fetches data on the client
 - Navigate to a dynamic subroute that prefetches the same query on the server
 - -> Error!

The result is the generic, well-known error message: "Can't perform a React state update on an unmounted component.":

![image](https://user-images.githubusercontent.com/29319414/194131159-2fca6dea-5739-40b2-a434-8367a7165b43.png)

I know that this is not a huge issue in itself, but maybe this still relevant for other parts of the logic and worth investigating, especially because of some weird quirks I'm describing below.

### Link to reproduction

https://github.com/bennettdams/trpc-10-unmounted-bug

## Via reproduction repo:

Steps:

1. `npm i`
2. `npm run dev`
3. Go to <http://localhost:3000>
4. Click on "Dynamic Subroute"
5. Error in console:
   ```
   Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
   at Home
   ```

### Additional information

It took me ages to even being able to create a reproduction repo, because the setup for the error is so weird.

You can make small changes in the repo that will make the error go away, changing only **one** of the following is enough:

- see `pages/_app.tsx`
[...]

PR #2993, opened 2022-10-12
Issues/2919

KATT:
#2919

## 🎯 Changes

Looking into ways to repro #2919.

Unsure what's causing the underlying issue.

## ✅ Checklist


TkDodo:
@KATT calling `Hydrate` multiple times is generally fine. What will happen is that we look at the `dataUpdatedAt` timestamp of each query, and we'll only put those queries in the cache that have newer data than the data in the cache:

https://github.com/TanStack/query/blob/357ec041a6fcc4a550f3df02c12ecc7bcdefbc05/packages/query-core/src/hydration.ts#L145-L151

PR #1958, opened 2022-06-06
fix(types): expose Procedure

jgoux:
In ESM context, when generating types for my own package `@acme/api` and exposing the type `AppRouter`, TypeScript is unable to resolve the internal type `Procedure` because it's not explicitly exported in `exports` map:

```typescript
import * as trpc from "@trpc/server";
export declare const appRouter: trpc.Router<{}, {}, {}, Record<"hello", import("@trpc/server/dist/internals/procedure.js").Procedure<{}, {}, {}, {
    name?: string | undefined;
}, {
    name?: string | undefined;
}, {
    greeting: string;
}, unknown, {
    greeting: string;
}>>, {}, {}, trpc.DefaultErrorShape>;
export declare type AppRouter = typeof appRouter;
//# sourceMappingURL=router.d.ts.map
```

It results in `AppRouter` being incomplete, `Procedure` being replaced with `any`:
image

The fix is to export `Procedure` as part of `@trpc/server` public types.

If I didn't export it from the proper place, don't hesitate to move the statement. 👍 

**edit**: I spotted another type not exported resulting in a failure to generate types declaration, this time in `@trpc/client`:
```bash
> tsc --project tsconfig.build.json --emitDeclarationOnly --declaration --declarationMap

src/config/trpc.ts:5:14 - error TS2742: The inferred type of 'client' cannot be named without a reference to '../../../../node_modules/@trpc/client/dist/internals/TRPCClient.js'. This is likely not portable. A type annotation is necessary.

5 export const client = createTRPCClient<AppRouter>({
               ~~~~~~


Found 1 error in src/config/trpc.ts:5
```
`TRPCClient` from `./internals/TRPCClient` should be exported instead of redefining it [here](https://github.com/trpc/trpc/blob/a3ecf56d83dd9774aaf885e83e9ccfe363fbab48/packages/client/src/createTRPCClient.ts#L14).

PR #2995, opened 2022-10-12
fix(router): allow naming of inferred type

### LINKED ISSUE #2994: bug: The inferred type of 'appRouter' cannot be named without a reference to ...

Local machine:
```
envinfo: command not found
```
¯\\\_(ツ)\_/¯

Stackblitz repro:
```
  System:
    OS: Linux 5.0 undefined
    CPU: (6) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 7.17.0 - /usr/local/bin/npm
  npmPackages:
    @trpc/server: 10.0.0-proxy-beta.18 => 10.0.0-proxy-beta.18 
    typescript: ^4.8 => 4.8.4 
```

### Describe the bug

Under `"moduleResolution": "Node16"` (or `"NodeNext"`), compilation of TypeScript types for the router can fail with error
```
TS2742: The inferred type of 'appRouter' cannot be named without a reference to '../node_modules/@trpc/server/dist/core/internals/config.js'. This is likely not portable. A type annotation is necessary.
```

### Link to reproduction

https://stackblitz.com/edit/trpc-server-inferred-type-cannot-be-named?file=src%2Findex.ts&view=editor

### To reproduce

See repro:

1. Set up package with Typescript
2. Set `"moduleResolution": "Node16"`
3. Try to compile types, e.g. via `tsc --declaration`

### Additional information

Related issues:
- #2717

Tried several suggested solutions (e.g. mentioned [here](https://github.com/microsoft/TypeScript/issues/42873#issuecomment-1140506416)); none worked, or had undesired side-effects.

The same issue exists in 9.x (after having manually worked around the module resolution).

### 👨‍👧‍👦 Contributing

echocrow:
Closes #2994

## 🎯 Changes

Export `RootConfig`, so TypeScript does not complain about not being able to name an inferred type.

Similar to (and builds on top of) https://github.com/trpc/trpc/pull/1958.

Unsure if this can be captured in a test somehow, as it seems to rely on a more modern `moduleResolution`?

### DIFF
--- a/packages/server/src/internals.ts
+++ b/packages/server/src/internals.ts
@@ -3,6 +3,7 @@
  */
 export type { DefaultErrorShape } from './error/formatter';
 export type { mergeRoutersGeneric } from './core/internals/__generated__/mergeRoutersGeneric';
+export type { RootConfig } from './core/internals/config';
 export type {
   ProcedureBuilder,
   BuildProcedure,


packages/server/src/core/index.ts
```diff
@@ -19,3 +19,5 @@ export { createInputMiddleware, createOutputMiddleware } from './middleware';
 
 export { initTRPC } from './initTRPC';
 export * from './types';
+
+export type { RootConfig } from './internals/config';
```
KATT, line 23: Could we rename this to `TRPCRootConfig` throughout?

KATT (changes requested):
Great change! I just think we should like us to smurf our public types.

packages/server/src/core/index.ts
echocrow: could do

but given this type is in a `@trpc/server` package, does this not implicitly scope it as a tRPC thing? i.e. I believe this is _technically_ a `TRPCServerRootConfig`, though that prefix feels redundant.

in the same vain, higher up, this file also exports
```ts
export type {
  Procedure,
  AnyProcedure,
  AnyQueryProcedure,
  // etc.
} from './procedure';
```
rather than 
```ts
export type {
  TRPCProcedure,
  TRPCAnyProcedure,
  TRPCAnyQueryProcedure,
  // etc.
} from './procedure';
```

if the idea and direction is to give these internal but exposed & inferred types that tRPC-context for later, we could also just
```ts
export type { RootConfig as TRPCRootConfig } from './internals/config';
```

happy to leave as-is or apply either changes

KATT (approved):
True. We've not been consistent on the smurfing. We _should_ probably be more consistent and only export `TRPC`-named things publicly #2131

sachinraja:
Should we export from `packages/server/src/internals.ts` instead? It's re-exported in `index.ts` but we've already exported other types that have to be named through `internals.ts`. We could also easily export them through a different `@trpc/server/internals` entry point if we wanted to.

echocrow:
Closing in favor of #2999

PR #3009, opened 2022-10-14
docs: add server code to cors page

juliusmarminge:
not sure if this is the best place to document this (seems appropriate since CORS is a useful application) but maybe it should be somewhere on the server documentation 🤔 

seems like many (myself included until Sachin pointed it out) don't connect that it's just a function that takes req/res.

Not sure how many goes to this page: https://www-git-cors-server-docs-trpc.vercel.app/docs/v10/cors#enabling-on-the-server


KATT (commented):
Unsure how useful this is - I don't think many people are using the base HTTP server and most people would need help with how you wrap the `/api/trpc/[trpc].tsx` with a CORS-helper

neo773:
@KATT 
It's quite useful I asked the same question the other day in the discord help channel, not an issue if you use the Express or Fastify Adapter because you can just plug in the cors npm package.

If I'm starting a new tRPC project it'd make sense to use the built-in server instead of an Adapter since I'm not writing any node web framework specific code it's all just tRPC routers and middlewares.

bautistaaa:
could put on FAQ vs explicitly in the docs?

KATT:
More docs doesn't hurt, we can move it later!

PR #3015, opened 2022-10-15
callable procedures

sachinraja:
## 🎯 Changes

Just an idea I had so you can do this:
```ts
const postById = t.procedure
  .input(z.object({ id: z.number() }))
  .query(({ input }) => {
    return posts.find((post) => post.id === input.id);
  });

postById({ id: 1 }, await createContextInner({}));
```

This could also help with what we've been talking about for RSC.

Don't worry about tests failing. I can fix that if we want to move forward with this.

## ✅ Checklist

### DIFF
--- a/packages/server/src/core/internals/procedureBuilder.ts
+++ b/packages/server/src/core/internals/procedureBuilder.ts
@@ -1,9 +1,6 @@
-import { TRPCError } from '../../error/TRPCError';
-import { getTRPCErrorFromUnknown } from '../../error/utils';
 import { FlatOverwrite, MaybePromise, Simplify } from '../../types';
 import {
   MiddlewareFunction,
-  MiddlewareResult,
   createInputMiddleware,
   createOutputMiddleware,
 } from '../middleware';
@@ -15,6 +12,7 @@ import {
   AnySubscriptionProcedure,
   Procedure,
   ProcedureParams,
+  callProcedureInner,
 } from '../procedure';
 import { ProcedureType } from '../types';
 import { AnyRootConfig } from './config';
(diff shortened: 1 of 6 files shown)


packages/server/src/core/internals/procedureBuilder.ts
```diff
@@ -295,83 +293,25 @@ export interface ProcedureCallOptions {
...
+    type = 'subscription';
+  }
 
-    // there's always at least one "next" since we wrap this.resolver in a middleware
-    const result = await callRecursive();
+  const procedure = ((input: any, ctx: any) => {
+    return callProcedureInner(procedure, {
+      ctx,
+      rawInput: input,
+      path: '',
```
sachinraja, line 308: I just passed an empty string here for the `path` and exposed a new `procedure` argument for middleware. This way if you want to check if a certain procedure is being called you can do this:
```ts
const publicRouter = t.router({
 /* ... */
})
t.middleware(({ procedure }) => {
  if (procedure in Object.values(publicRouter._def.procedures)) {
    // procedure is public
  }
})
```

KATT (commented):
Great that you're exploring this!! Let's figure out if importing the full router in RSC is even a problem first or if having procedures or routers will feel the best when using RSC.

When the `createCaller` isn't used anymore, the context feels a bit unnatural.

Maybe a `{ input: x; ctx: y}`- struct would feel better? Or to inverse the calls so context is first (as it'll likely be curried by something)?

sachinraja:
> Let's figure out if importing the full router in RSC is even a problem first or if having procedures or routers will feel the best when using RSC.

Agreed, but should be noted people might prefer this method outside of RSC too.

> Maybe a { input: x; ctx: y}- struct would feel better? Or to inverse the calls so context is first (as it'll likely be curried by something)?

I thought about this but it's sort of working around the proxy API. If we wanted to do that we could also do something like this:
```ts
const callProcedure = getProcedureCaller(/* context */{ session: /* next auth */ })

callProcedure(trpc.post.byId, /* input */ { id: 0 })
```
You noted a similar thing here: https://twitter.com/alexdotjs/status/1581933764011515905?s=20&t=OrNETJQawpPsEPmjmK-9ww

juliusmarminge:
> > Let's figure out if importing the full router in RSC is even a problem first or if having procedures or routers will feel the best when using RSC.
> 
> 
> 
> Agreed, but should be noted people might prefer this method outside of RSC too.
> 
> 
> 
> > Maybe a { input: x; ctx: y}- struct would feel better? Or to inverse the calls so context is first (as it'll likely be curried by something)?
> 
> 
> 
> I thought about this but it's sort of working around the proxy API. If we wanted to do that we could also do something like this:
> 
> ```ts
> 
> const callProcedure = getProcedureCaller(/* context */{ session: /* next auth */ })
> 
> 
> 
> callProcedure(trpc.post.byId, /* input */ { id: 0 })
> 
> ```
> 
> You noted a similar thing here: https://twitter.com/alexdotjs/status/1581933764011515905

I don't love this, prefer the initial proposal.

Might just be me who prefers a single import (trpc) than two (getProcedureCaller, trpc) though 🤷‍♂️

KATT:
Closing as I'm cleaning up PRs.

PR #3037, opened 2022-10-19
feat: add unstable_overrides to next

juliusmarminge:
let these options be defined in `createTRPCNext` too

ref: https://discord.com/channels/867764511159091230/1032031459852156938

### DIFF
--- a/packages/next/src/createTRPCNext.tsx
+++ b/packages/next/src/createTRPCNext.tsx
@@ -15,7 +15,10 @@ export function createTRPCNext<
   TRouter extends AnyRouter,
   TSSRContext extends NextPageContext = NextPageContext,
 >(opts: WithTRPCNoSSROptions<TRouter> | WithTRPCSSROptions<TRouter>) {
-  const hooks = createHooksInternal<TRouter, TSSRContext>();
+  const unstable_overrides = opts.config({}).unstable_overrides;
+  const hooks = createHooksInternal<TRouter, TSSRContext>({
+    unstable_overrides,
+  });
 
   // TODO: maybe set TSSRContext to `never` when using `WithTRPCNoSSROptions`
   const _withTRPC = withTRPC<TRouter, TSSRContext>(opts);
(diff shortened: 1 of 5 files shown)


packages/next/src/withTRPC.tsx
```diff
@@ -90,7 +92,9 @@ export function withTRPC<
     ssrContext: TSSRContext;
   };
   return (AppOrPage: NextComponentType<any, any, any>): NextComponentType => {
-    const trpc = createReactQueryHooks<TRouter, TSSRContext>();
+    const trpc = createReactQueryHooks<TRouter, TSSRContext>({
+      unstable_overrides: opts.config({}).unstable_overrides,
```
juliusmarminge, line 96: this is a bit hacky, not sure what the best approach is to pass the options down?

packages/next/src/createTRPCNext.tsx
```diff
@@ -15,7 +15,10 @@ export function createTRPCNext<
   TRouter extends AnyRouter,
   TSSRContext extends NextPageContext = NextPageContext,
 >(opts: WithTRPCNoSSROptions<TRouter> | WithTRPCSSROptions<TRouter>) {
-  const hooks = createHooksInternal<TRouter, TSSRContext>();
+  const unstable_overrides = opts.config({}).unstable_overrides;
+  const hooks = createHooksInternal<TRouter, TSSRContext>({
+    unstable_overrides,
+  });
```
juliusmarminge, line 21: bit hacky. maybe these options should be outside the config-callback?

packages/next/src/withTRPC.tsx
juliusmarminge: maybe they aren't even needed here? is the hooks created in `createTRPCNext`(see above) the important ones?

KATT: This is fine 🤷

KATT (approved):

PR #3046, opened 2022-10-20
www: modified navbar active link classes

rhenriquez28:
Closes #

## 🎯 Changes

When you're reading the docs all the links in the navbar turn blue. This also happens in the sidebar when reading on mobile. Links are now white while reading docs and only turn blue when you hover over them.

Screen Shot 2022-10-20 at 5 10 09 PM

## ✅ Checklist


juliusmarminge:
if only Docusaurus could give my pr some attention that would fix this in core... https://github.com/facebook/docusaurus/pull/8079

rhenriquez28:
Yeah, I was trying that at first and then realized that you were using navbar doc links instead of regular links :(

juliusmarminge:
> Yeah, I was trying that at first and then realized that you were using navbar doc links instead of regular links :(

made that pr 1,5 months ago haven't gotten a single response... such a minor change to allow regex matching on docnavbar items in the same way you can on any other navbar item...

rhenriquez28:
Sucks tbh. But what do we do about this now then? I figured you wouldn't want to change the doc links for the regular ones, hence why I came up with this approach

juliusmarminge:
> Sucks tbh. But what do we do about this now then? I figured you wouldn't want to change the doc links for the regular ones, hence why I came up with this approach

I think regular ones doesnt support versioned docs, which we want.

I pinged and bumped my pr in Docu's discord, hope someone notices

www/src/css/custom.css
```diff
@@ -40,6 +40,14 @@ div[class^='announcementBar_'] {
 .navbar {
   border-bottom: 1px solid var(--ifm-toc-border-color);
 }
+
```
juliusmarminge, line 43:
```suggestion

/* FIXME: Revert later. Blocked by https://github.com/facebook/docusaurus/pull/8079 */
```

juliusmarminge (dismissed):

juliusmarminge (approved):

PR #2986, opened 2022-10-12
add SOA example

KATT:
Closes #2985

## 🎯 Changes

- SOA example
- The "gateway" doesn't actually create a server; the _client_ decides at runtime which service to call depending on the prefix of the call
- Caveat: think all services need the same context to be merged

## ✅ Checklist

- [ ] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [ ] If necessary, I have added documentation related to the changes made.
- [ ] I have added or updated the tests related to the changes made.

PR #3049, opened 2022-10-21
ability to have multiple TRPC Providers

### LINKED ISSUE #2121: Using multiple TRPC clients in the same app

KATT:
Closes #2121

## 🎯 Changes

- Adds the ability to have multiple TRPC providers in the same app
- **Note:** #2986 is another approach to the same problem that should work for all monorepo-style setups, this approach here is more if you do a tRPC-based React-sdk that you want to share
- (Docs needed? feel free to add)

## ✅ Checklist


packages/server/test/react/multiple-trpc-providers.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+                  client={C.queryClient}
+                  context={C.reactQueryContext}
+                >
+                  <MyComponent />
+                </QueryClientProvider>
+              </C.trpc.Provider>
+            </QueryClientProvider>
+          </B.trpc.Provider>
+        </QueryClientProvider>
+      </A.trpc.Provider>
```
KATT, line 114: Kheops-Pyramid

sachinraja (approved):

KATT:
Would love @mmkal's input who I know was curious about this

packages/server/test/react/multiple-trpc-providers.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+  }
+  function App() {
+    return (
+      <A.trpc.Provider queryClient={A.queryClient} client={ctx.A.client}>
+        <QueryClientProvider
+          client={A.queryClient}
+          context={A.reactQueryContext}
+        >
+          <B.trpc.Provider queryClient={B.queryClient} client={ctx.B.client}>
+            <QueryClientProvider
```
mmkal, line 96: Would it be possible to use one r-q provider but many trpc providers? If so, maybe a test for that scenario (with a slightly smaller 🔺) might be helpful.

mmkal (commented):
Exciting, looks great! Left a couple of minor qs inline, only other comment I'd make is that maybe it'd be clearer why this feature could be useful by giving semi-realistic contexts in the test rather than A/B/C. Like a data API, a CMS API, and a feature flag service or something? Those last two are things we're (extremely slowly) circling towards having a full-stack trpc based implementation of, but it doesn't have to be them specifically. It's really just naming in tests.

Anyway, I'm happy with this as is, I could maybe try to find some time to update the tests as a follow up.

KATT:
> Exciting, looks great! Left a couple of minor qs inline, only other comment I'd make is that maybe it'd be clearer why this feature could be useful by giving semi-realistic contexts in the test rather than A/B/C. Like a data API, a CMS API, and a feature flag service or something? Those last two are things we're (extremely slowly) circling towards having a full-stack trpc based implementation of, but it doesn't have to be them specifically. It's really just naming in tests.

@mmkal, have you looked at using something like #2986?

mmkal:
> @mmkal, have you looked at using something like #2986?

Yes, and that's more or less what we do now. I like this better as a long term/scalable solution though because the servers can be truly independent, don't have to have the same or compatible contexts, or share a client splitting based on a link. And even if there are other ways to achieve, it just seems like a nice design to have trpc not assume it's the Only One. Basically we want to build open-sourceable server+client sdks which you can just npm install and deploy to aws/vercel and use directly, without having to touch any existing trpc config you might have.

PR #3061, opened 2022-10-23
fix: use optional chaining for `process.env`

jsonnull:
## 🎯 Changes

When using trpc v10 outside of a server environment, `process` may be undefined. We can pass `isServer: false` in the RuntimeConfig, but the default `isServer` value is created in the module scope regardless, so the `process.env` check will throw.

This simply changes `process.env.VAR` to `globalThis.process?.env?.VAR` where applicable to support non-server environments.

## ✅ Checklist

### DIFF
--- a/examples/next-prisma-starter-websockets/src/server/context.ts
+++ b/examples/next-prisma-starter-websockets/src/server/context.ts
@@ -1,4 +1,3 @@
-import { PrismaClient } from '@prisma/client';
 import * as trpc from '@trpc/server';
 import * as trpcNext from '@trpc/server/adapters/next';
 import { NodeHTTPCreateContextFnOptions } from '@trpc/server/adapters/node-http';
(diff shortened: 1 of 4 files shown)


sachinraja:
Should we just check if `process` exists instead?

jsonnull:
>

jsonnull:
> Should we just check if `process` exists instead?

@sachinraja I can do that, want me to just optionally chain the `process.env.` parts?

jsonnull:
@sachinraja Updated ❤️

sachinraja:
We can't do that either because `process` won't even be a variable. So we'll get some error like `unresolved variable "process"`. I think we need to optionally chain it off of `globalThis`: `globalThis.process?.env`

jsonnull:
Of course! Updated again 😄

sachinraja (approved):

sachinraja:
Thanks for bearing with me 😄

jsonnull:
Thanks for thinking this through where I just didn't 😂

examples/next-prisma-starter-websockets/src/server/context.ts
```diff
@@ -1,4 +1,3 @@
-import { PrismaClient } from '@prisma/client';
```
jsonnull, line 1: Drive-by lint fix

sachinraja (approved):

PR #3113, opened 2022-11-04
fix: typo from testing www cache

juliusmarminge:
didn't revert my cache test

updated output folder on vercel so this should hit


juliusmarminge:
Nice

PR #3120, opened 2022-11-05
fix transformer types

OrJDev:
Closes #

## 🎯 Changes

## ✅ Checklist

### DIFF
--- a/packages/client/src/createTRPCClient.ts
+++ b/packages/client/src/createTRPCClient.ts
@@ -1,4 +1,10 @@
-import type { AnyRouter } from '@trpc/server';
+/* eslint-disable @typescript-eslint/naming-convention */
+import type {
+  AnyRouter,
+  AsTransformer,
+  ClientDataTransformerOptions,
+  DataTransformer,
+} from '@trpc/server';
 import {
   TRPCClient as Client,
   CreateTRPCClientOptions,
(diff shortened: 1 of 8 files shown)


OrJDev:
Some tests failed as you see but i think i figured it out

## Solution

Instead of making the baseType reject promises we create a new optional function that will reject promises,
so you can still use transformers like:

### No Error

```ts
{
  transformer: superjson;
}
```

### No Error

```ts
{
  transformer: withTransformer(superjson);
}
```

### Error

```ts
{
  transformer: withTransformer({
    serialize: (data) => {
      // not async so its fine
    },
    deserialize: async (data) => {
      /// cannot be async so error is thrown
    },
  });
}
```

Screenshot_1

OrJDev:
works for react, gotta fix for next

### This will return never so it cannot be used as a valid client
```ts
  const _example = trpc.createClient({
    links: [],
    transformer: {
      serialize: async () => {
        return {};
      },
      deserialize: () => {
        return {};
      },
    },
  });
```
### This is fine and will return a valid trpc client, can be used as a client
```ts
  const _example = trpc.createClient({
    links: [],
    transformer: {
      serialize: () => {
        return {};
      },
      deserialize: () => {
        return {};
      },
    },
  });
```

KATT (commented):
I'm starting to feel that the juice might not be worth the squeeze here?

OrJDev:
It is more complicated than it should've been, but after 13 commits I rather just fix it for `next` than throw it away

OrJDev:
Will reopen when i have time

PR #3125, opened 2022-11-06
fix versions in `packages/tests`

KATT:
Closes #

## 🎯 Changes

fix versions in `packages/tests`

@juliusmarminge can you tell why the versions are not automatically bumped in the test? 🤔 

## ✅ Checklist

- [ ] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/next/CONTRIBUTING.md).


packages/tests/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "tests",
+  "name": "@trpc/tests",
```
juliusmarminge, line 2: if we can get away with not naming this @trpc it makes our lives a lot easier.

all the install filters use @trpc/* and that would make this package match that too => unnecessary package installs in all examples and www

KATT: I'm unsure if this even helps tbh.... but won't turborepo just cache it anyway?

juliusmarminge: not the dependency installation. we can add another filter that removes this

```bash
pnpm --filter="@trpc/*" --filter="\!trpc@tests" install
```

KATT: Cool! Feel free to commit on this branch if you think this is the way :)

juliusmarminge:
I've updated the Vercel setting to use this filter too.

PR #3137, opened 2022-11-07
add experimental `useSuspenseQuery()`

### LINKED ISSUE #1354: Make `useQuery` with `suspense: true` to return non-nullable `data`

`{ suspense: true }` should let us use `data` without checking its integrity.

We can use `suspense: true` in SSR now, see #3130

Given

```ts
const postQuery = trpc.post.byId.useQuery({ id }, { suspense: true });
```

`postQuery.data` should be **not** be nullable.

Considerations: if `enabled: true`, we should not infer it as non-nullable

- Default options to set `suspense: true`
- Add a `useSuspenseQuery` / `useSuspenseInfiniteQuery`

Another approach would be to enable this behavior globally when creating the react-query hooks.

```tsx
export trpc = createTRPCReact<Context>({suspense: true})
```

### Additional information

- `enabled:` `true`/`false` should still work
- We need to narrow the generic so we still can't pass extra params to `opts`

KATT:
Closes #1354

## 🎯 Changes

- Adds experimental `useSuspenseQuery()` 
- It is **feature-flagged** as the exact implementation might change & we're waiting to see how the `react-query` team on how they're planning to deal with it
- Live demo: https://github.com/KATT/useSuspense-next-prisma-starter

### Enable experimental suspense

```tsx
const trpc = createTRPCReact<
  AppRouter,
  unknown,
  'ExperimentalSuspense' // <------------- This is the magical feature flag
>();
```

### Using it

The above will give you access to a new fresh hook:

```ts
const [data] = trpc.post.byId.useSuspenseQuery({ id: "1" })
//      ^? will be typeeeed and no loading state stuff
```

## ✅ Checklist

### DIFF
--- a/packages/next/src/createTRPCNext.tsx
+++ b/packages/next/src/createTRPCNext.tsx
@@ -1,6 +1,7 @@
 import {
   CreateReactUtilsProxy,
   DecoratedProcedureRecord,
+  ExperimentalFlags,
   createHooksInternal,
   createReactProxyDecoration,
   createReactQueryUtilsProxy,
@@ -14,6 +15,7 @@ import { WithTRPCNoSSROptions, WithTRPCSSROptions, withTRPC } from './withTRPC';
 export function createTRPCNext<
   TRouter extends AnyRouter,
   TSSRContext extends NextPageContext = NextPageContext,
+  TFlags extends ExperimentalFlags = never,
 >(opts: WithTRPCNoSSROptions<TRouter> | WithTRPCSSROptions<TRouter>) {
   const hooks = createHooksInternal<TRouter, TSSRContext>({
     unstable_overrides: opts.unstable_overrides,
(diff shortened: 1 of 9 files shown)


packages/tests/server/react/__reactHelpers.tsx
```diff
@@ -36,7 +36,11 @@ export function getServerAndReactClient<TRouter extends AnyRouter>(
   });
 
   const queryClient = createQueryClient();
-  const proxy = createTRPCReact<typeof appRouter>();
+  const proxy = createTRPCReact<
+    typeof appRouter,
+    unknown,
+    'ExperimentalSuspense'
+  >();
```
KATT, line 43: This is how you enable it

sachinraja: should feature flags be an array?

KATT: I'd also want autocomplete on them. Types became annoying. Feel free to have a stab at improving it

KATT: It doesn't need to be an array, it could work like the second arg of `Pick` etc

sachinraja: Ahh I read the `'ExperimentalSuspense'` argument as runtime code the first time but it's passed as a generic

packages/next/src/createTRPCNext.tsx
```diff
@@ -14,6 +14,7 @@ import { WithTRPCNoSSROptions, WithTRPCSSROptions, withTRPC } from './withTRPC';
 export function createTRPCNext<
   TRouter extends AnyRouter,
   TSSRContext extends NextPageContext = NextPageContext,
+  TFlags = null,
```
sachinraja, line 17:
```suggestion
  TFlags = never,
```

PR #3124, opened 2022-11-06
fix types for invalidating routers

KATT:
- closes #3122
- closes #3123 

## 🎯 Changes

Bug fixes

## ✅ Checklist

- [x] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/next/CONTRIBUTING.md).
- [x] If necessary, I have added documentation related to the changes made.
- [x] I have added or updated the tests related to the changes made.

PR #3165, opened 2022-11-13
feat: Add type to react-query query key - Attempt round 2!

### LINKED ISSUE #3128: feat: prefix infinite queries differently to "normal" queries?

Currently, the cache gets a bit messed up if you use an endpoint the with `.useInfiniteQuery()` in one part of the app and `.useQuery()` in another.

Currently, we store cache keys as `[pathArray, input]`-tuple

Maybe, we should store it as `['query' | 'infinite', pathArray, input]`-tuple?

Or a `[pathArray, 'query' | 'infinite', input]`-tuple?

Not doing it

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

JonParton:
Closes #3128

## 🎯 Changes

This pull request alters the [react-query](https://tanstack.com/query/v4/docs/adapters/react-query) `queryKey` to also include the `type` of the query (`query` or `infinite`) as well as placing the input within an object alongside it. 

New key Structure: 
```ts
[ 
   string[], //arrayPath!
   { 
      input?: any; 
      type: 'query' | 'infinite' 
   }
]
```
Example:
image

This change will mean that an `infinite` query and a `query` of the same endpoint will have different entries within the cache, which is required as the data structures are different!!

This key structure also maintains the ability to invalidate specifically `query`'s or `infinite`'s across entire routers as well as filtering queries using their input when interacting with the react-query `queryClient` directly (Escape Hatch! 👨‍🚒).

## 🔨 Possible improvements

Currently the exposed `invalidate()` function on a router or on a leaf node will invalidate both `queries` and `infiniteQueries` with no way to filter. The ability to specify only `queries` or `infiniteQueries` could be added at a later date but I think for now this should be fine and in line with the philosophy of being indiscriminate added in #3124  

## ✅ Checklist

### DIFF
--- a/packages/react-query/src/internals/getArrayQueryKey.ts
+++ b/packages/react-query/src/internals/getArrayQueryKey.ts
@@ -10,7 +10,7 @@ export type QueryType = 'query' | 'infinite' | 'any';
 export function getArrayQueryKey(
   queryKey: string | [string] | [string, ...unknown[]] | unknown[],
   type: QueryType,
-): [{ path: string[]; input?: unknown; type?: Omit<QueryType, 'any'> }] {
+): [{ arrayPath: string[]; input?: unknown; type?: Omit<QueryType, 'any'> }] {
   const queryKeyArrayed = Array.isArray(queryKey) ? queryKey : [queryKey];
   const [path, input] = queryKeyArrayed;
 
@@ -22,7 +22,7 @@ export function getArrayQueryKey(
   // https://github.com/trpc/trpc/issues/3128
   return [
     {
-      path: arrayPath,
+      arrayPath,
       ...(input && { input: input }),
       ...(type && type !== 'any' && { type: type }),
     },
--- a/packages/react-query/src/internals/getArrayQueryKey.ts
+++ b/packages/react-query/src/internals/getArrayQueryKey.ts
@@ -7,12 +7,22 @@
  */
 export function getArrayQueryKey(
   queryKey: string | [string] | [string, ...unknown[]] | unknown[],
-): [string[]] | [string[], ...unknown[]] | [] {
+  type: 'query' | 'infinite' | 'any',
+): [{ path: string[]; input?: unknown; type?: 'query' | 'infinite' }] {
   const queryKeyArrayed = Array.isArray(queryKey) ? queryKey : [queryKey];
   const [path, ...input] = queryKeyArrayed;
 
   const arrayPath =
     typeof path !== 'string' || path === '' ? [] : path.split('.');
 
-  return [arrayPath, ...input];
+  // Construct a query key that is easy to destructure and flexible for
+  // partial selecting etc.
+  // https://github.com/trpc/trpc/issues/3128
+  return [
+    {
+      path: arrayPath,
+      ...(input && { input: input }),
+      ...(type && type !== 'any' && { type: type }),
+    },
+  ];
 }
--- a/packages/react-query/src/internals/getArrayQueryKey.ts
+++ b/packages/react-query/src/internals/getArrayQueryKey.ts
@@ -10,7 +10,7 @@ export function getArrayQueryKey(
   type: 'query' | 'infinite' | 'any',
 ): [{ path: string[]; input?: unknown; type?: 'query' | 'infinite' }] {
   const queryKeyArrayed = Array.isArray(queryKey) ? queryKey : [queryKey];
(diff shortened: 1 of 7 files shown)


packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -301,7 +301,7 @@ export function createHooksInternal<
             (...args: any[]) => {
               const [queryKey, ...rest] = args;
               return queryClient.invalidateQueries(
-                getArrayQueryKey(queryKey),
+                getArrayQueryKey(queryKey, 'any'),
```
JonParton, line 304: No option to specify either `query` or `infinite` here ... Will invalidate indiscriminately!

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -312,23 +312,25 @@ export function createHooksInternal<
               const [queryKey, ...rest] = args;
 
               return queryClient.refetchQueries(
-                getArrayQueryKey(queryKey),
+                getArrayQueryKey(queryKey, 'any'),
```
JonParton, line 315: refetch currently only has an `any` mode. Without adding a separate type param or separating into two methods we can't tell if we are targeting `query` or `infinite` I don't think?

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -312,23 +312,25 @@ export function createHooksInternal<
...
               );
             },
             [queryClient],
           ),
           cancelQuery: useCallback(
             (pathAndInput) => {
-              return queryClient.cancelQueries(getArrayQueryKey(pathAndInput));
+              return queryClient.cancelQueries(
+                getArrayQueryKey(pathAndInput, 'any'),
+              );
```
JonParton, line 325: `cancelQuery` currently only has an `any` mode. Without adding a separate type param or separating into two methods we can't tell if we are targeting `query` or `infinite` I don't think?

packages/tests/server/react/useContext.test.tsx
```diff
@@ -375,7 +375,7 @@ test('setData', async () => {
 test('setInfiniteData', async () => {
   const { proxy, App } = ctx;
   function MyComponent() {
-    const listPosts = proxy.post.list.useQuery({}, { enabled: false });
+    const listPosts = proxy.post.list.useInfiniteQuery({}, { enabled: false });
```
JonParton, line 378: This was previously a bit of a bug in the tests that went unnoticed ... Before we were updating a `useQuery` with an `infinite` data structure ... With the new distinction this test broke! 

I've changed this over and fixed.

KATT (changes requested):
Can you do this?

1. Close this PR

This is in order for CI to run properly so we can get all those nice ✅ ahead of merging.

packages/react-query/src/shared/hooks/createHooksInternal.tsx
KATT: Can we add a test to make sure this invalidates both queries and infinite queries?

JonParton: Yep! I have tested it and it deffinitely does so but I never added a test 🙈

packages/react-query/src/shared/hooks/createHooksInternal.tsx
KATT: Naaah, I think it's fine to keep it `any` for now

packages/react-query/src/shared/hooks/createHooksInternal.tsx
KATT: All good

packages/react-query/src/internals/getArrayQueryKey.ts
```diff
@@ -7,12 +9,22 @@
...
 
-  return [arrayPath, ...input];
+  // Construct a query key that is easy to destructure and flexible for
+  // partial selecting etc.
+  // https://github.com/trpc/trpc/issues/3128
+  return [
+    arrayPath,
+    {
+      ...(input && { input: input }),
+      ...(type && type !== 'any' && { type: type }),
```
KATT, line 27: Nice

JonParton:
closed as transferred to #3170

PR #3170, opened 2022-11-14
feat: Add type to react-query query key - Attempt round 2!

### LINKED ISSUE #3128: feat: prefix infinite queries differently to "normal" queries?

JonParton:
Closes #3128

transferred from #3165

## 🎯 Changes

This pull request alters the [react-query](https://tanstack.com/query/v4/docs/adapters/react-query) `queryKey` to also include the `type` of the query (`query` or `infinite`) as well as placing the input within an object alongside it. 

New key Structure: 
```ts
[ 
   string[], //arrayPath!
   { 
      input?: any; 
      type: 'query' | 'infinite' 
   }
]
```
Example:
image

This change will mean that an `infinite` query and a `query` of the same endpoint will have different entries within the cache, which is required as the data structures are different!!

This key structure also maintains the ability to invalidate specifically `query`'s or `infinite`'s across entire routers as well as filtering queries using their input when interacting with the react-query `queryClient` directly (Escape Hatch! 👨‍🚒).

## 🔨 Possible improvements

Currently the exposed `invalidate()` function on a router or on a leaf node will invalidate both `queries` and `infiniteQueries` with no way to filter. The ability to specify only `queries` or `infiniteQueries` could be added at a later date but I think for now this should be fine and in line with the philosophy of being indiscriminate added in #3124  

## ✅ Checklist


JonParton:
review comments from @KATT can be found on the closed fork #3165 (My bad! 🤦)

JonParton:
- [ ] remaining action is to add a test for `utils.invalidate()` to show it works on both `infinite` and `query` => I can do this later but may get held up with my work day; feel free to add if it is holding things up!

www/docs/migration/migrate-from-v9-to-v10.mdx
```diff
@@ -499,7 +502,7 @@ export const appRouter = router({
 The query key used for `trpc.user.byId.useQuery({ id: 10 })` would change:
 
 - Key in V9: `["user.byId", { id: 10 }]`
-- Key in v10:`[["user", "byId"],{ id:10 }]`
+- Key in v10:`[["user", "byId"],{ input: { id:10 }, type: 'query' }]`
```
KATT, line 505: Oh nice catch

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -301,7 +301,7 @@ export function createHooksInternal<
             (...args: any[]) => {
               const [queryKey, ...rest] = args;
               return queryClient.invalidateQueries(
-                getArrayQueryKey(queryKey),
+                getArrayQueryKey(queryKey, 'any'),
```
KATT, line 304: I think this is prefect and that we don't need to allow granular invalidation for now, I don't see why anyone would _not_ want to invalidate the same query across both methods of fetching it.

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -312,23 +312,25 @@ export function createHooksInternal<
               const [queryKey, ...rest] = args;
 
               return queryClient.refetchQueries(
-                getArrayQueryKey(queryKey),
+                getArrayQueryKey(queryKey, 'any'),
```
KATT, line 315: I think this is prefect and that we don't need to allow granular invalidation for now, I don't see why anyone would _not_ want to refetch the same query across both methods of fetching it.

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -312,23 +312,25 @@ export function createHooksInternal<
...
                 ...rest,
               );
             },
             [queryClient],
           ),
           cancelQuery: useCallback(
             (pathAndInput) => {
-              return queryClient.cancelQueries(getArrayQueryKey(pathAndInput));
+              return queryClient.cancelQueries(
```
KATT, line 323: I think this is prefect and that we don't need to allow granular invalidation for now, I don't see why anyone would _not_ want to cancel the same query across both methods of fetching it.

KATT (approved):
I think this is great, but I would love some more testing. Happy to merge as-is for now though!

- [ ] Make sure that `invalidate()` or `reset` etc will invalidate both queries and infinite queries. Only need to test 1 I think, they are exactly the same

PR #3171, opened 2022-11-14
fix(types): export AnyRootConfig from internals

jgoux:
## 🎯 Changes

Since https://github.com/trpc/trpc/pull/2031, I went back to trpc and noticed that my build is broken because `AnyRootConfig` needs to be exported in order to generate type definition files for my package.

As a reminder, this is necessary to avoid these errors:

```bash
src/config/trpc/index.ts:5:14 - error TS2742: The inferred type of 't' cannot be named without a reference to '../../../../../node_modules/@trpc/server/dist/core/internals/config.js'. This is likely not portable. A type annotation is necessary.

5 export const t = initTRPC.context<Context>().create();
               ~


Found 1 error in src/config/trpc/index.ts:5
```

## ✅ Checklist

### DIFF
--- a/packages/server/src/internals.ts
+++ b/packages/server/src/internals.ts
@@ -3,7 +3,7 @@
  */
 export type { DefaultErrorShape } from './error/formatter';
 export type { mergeRoutersGeneric } from './core/internals/__generated__/mergeRoutersGeneric';
-export type { RootConfig } from './core/internals/config';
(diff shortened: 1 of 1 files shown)


KATT (approved):

KATT:
Can we add some testing to make sure these are included in the build? By definition, anything in `/internal` shouldn't be `export`ed so it's easy to accidentally remove these when doing janitor work

PR #3184, opened 2022-11-17
Simplify and clean up Next.js usage page.

anthonyshew:
## 🎯 Changes

Title!

One major change that I will call out is that I've removed all of the authenticated procedure from the "hello world" procedure example.

The motivation for doing this is that folks are most likely reading this page as newcomers to tRPC (and maybe the combination of tRPC and Next.js). Because of this, we want to meet folks where they are and introduce one concept at a time. This page is introductory material so throwing in more things to figure out makes it harder to make the initial leap to fall in love with tRPC.

Once they see the simplicity and ease of use that tRPC brings, we can add more layers of complexity and features after that.

## ✅ Checklist

- [ ] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/next/CONTRIBUTING.md).

### DIFF
--- a/www/docs/nextjs/introduction.mdx
+++ b/www/docs/nextjs/introduction.mdx
@@ -5,17 +5,18 @@ sidebar_label: Usage with Next.js
 slug: /nextjs
 ---
 
+import TabItem from '@theme/TabItem';
+import Tabs from '@theme/Tabs';
+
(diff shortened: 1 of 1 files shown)


KATT (approved):
❤️

PR #3195, opened 2022-11-18
Add query and infinite invalidation and cancel tests

### LINKED ISSUE #3173: Improve query invalidation testing
I think this is great, but I would love some more testing. Happy to merge as-is for now though!

- [ ] Make sure that `invalidate()` or `reset` etc will invalidate both queries and infinite queries. Only need to test 1 I think, they are exactly the same

_Originally posted by @KATT in https://github.com/trpc/trpc/pull/3170#pullrequestreview-1178732319_

JonParton:
Closes #3173 

## 🎯 Changes

This PR introduces tests for `.cancel()` on a leaf node and `.invalidate()` on both leaf and router nodes. 

These tests ensure that these functions operate on both queries and infinite queries. This is important due to the distinction added in the cache between `query`'s and `infinite`'s in #3165.   

## ✅ Checklist

### DIFF
--- a/packages/tests/server/react/useContext.test.tsx
+++ b/packages/tests/server/react/useContext.test.tsx
@@ -1,4 +1,5 @@
 /* eslint-disable react-hooks/exhaustive-deps */
+import { waitMs } from '../___testHelpers';
 import { getServerAndReactClient } from './__reactHelpers';
 import { render, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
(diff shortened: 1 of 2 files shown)


packages/tests/server/react/useContext.test.tsx
```diff
@@ -289,6 +290,87 @@ test('invalidate', async () => {
...
+
+  await waitFor(() => {
+    expect(utils.container).toHaveTextContent('new post');
+  });
+
+  // I've tried everything to avoid this but for some reason, even though
+  // infinite query is successful and data is on the component it still shows
+  // no calls to the proxy passed to "onSuccess" unless i do this wait ... even
+  // await new Promise(process.nextTick); doesn't work 🤷🤯
+  await waitMs(1);
```
JonParton, line 354: I'm not happy with this but I cannot for the life of me find out why this test will not work without the wait?! ... 

It seems like something weird is happening with the calling of the `onSuccess` callback on `.useInfiniteQuery()`

@TkDodo - Am I doing anything stupid here you can see straight off?? 

If we don't want to introduce this, we could remove this leaf node test and just keep the router test also introduced in this PR ... but they are slightly different routes, so I ideally wanted to test both in case they get implemented differently in the future...

TkDodo: a bit strange because it works in TanStack/query:

https://github.com/TanStack/query/blob/933cd410ae0c8e69f9375eac2d5fb38cc940e6eb/packages/react-query/src/__tests__/useQuery.test.tsx#L434-L458

However, I can't see how the queryFn is implemented in the test. There is lots of batching going on inside RQ, so if you queryFn is synchronous, it could be that there is a render that has `isFetching:false` immediately while _not_ having called `onSuccess`. As you can see, in our tests, we made the `queryFn` async and waited a bit in there.

JonParton: The query function being used in this `infiniteQuery` is horribly simple and not async / how an infinite query function would normally work so maybe it is something to do with that! 

```ts
....
type Post = {
  id: number;
  text: string;
};

const defaultPost = { id: 0, text: 'new post' };
const ctx = konn()
  .beforeEach(() => {
    const t = initTRPC.create({
      errorFormatter({ shape }) {
        return {
          ...shape,
          data: {
            ...shape.data,
            foo: 'bar' as const,
          },
        };
      },
    });

    const posts: Post[] = [defaultPost];

    const appRouter = t.router({
      post: t.router({
        list: t.procedure
          .input(
            z.object({
              cursor: z.string().optional(),
            }),
          )
          .query(() => posts),
```

I'll have a quick look if making the query function async fixes it but seems odd! 

Thanks for the rapid response though 🙂🤯.

JonParton: Making the query function async and adding a 10ms wait still didn't seems to solve it without waiting specifically in the test 🤔

It must be something to do with the order that things happen between `isFetching` being set to false and the `onSuccess` callback being called 🤷 ... Don't understand why `await new Promise(process.nextTick);` doesn't work though when a 1ms wait does!

packages/tests/server/react/useContext.test.tsx
```diff
@@ -296,6 +297,88 @@ test('invalidate', async () => {
...
+  await waitFor(() => {
+    expect(utils.container).toHaveTextContent('new post');
+  });
+
+  // I've tried everything to avoid this but for some reason, even though
+  // infinite query is successful and data is on the component it still shows
+  // no calls to the proxy passed to "onSuccess" unless i do this wait ... even
+  // await new Promise(process.nextTick); doesn't work 🤷🤯
+  // https://github.com/trpc/trpc/pull/3195#discussion_r1026952381
+  await waitMs(1);
```
KATT, line 362: We could just use `waitFor`, this LGTM

KATT (approved):

packages/tests/server/react/useContext.test.tsx
JonParton: Problem was what do you wait for? 

I was already waiting for the state of the `query` and `infinite` query to not be fetching but at this point the jest proxy function attached to the `onSuccess` of the `infinite` query was still only called 0 (first time) or 1 (after invalidation) times 😵‍💫... However waiting for `1ms` made them read what they should be! (1 then 2) 🤷

As above conversation with TkDodo it seems there is some odd batching going on and the `onSuccess`callback is not qued when fetching is set to false!

JonParton: Ohhh could you `waitFor` the actual proxy count? 🤦

JonParton: That worked and is much neater ... Thanks for pointing out the obvious answer I missed 🤦... Still odd behaviour!...

KATT (approved):

PR #3196, opened 2022-11-18
change invalidate test to call proxy spy

JonParton:
Closes #

## 🎯 Changes

Not sure if this one is actually a problem as the test was working in the old format by virtue of it testing for the value added by the mutation that is called after render, but the `stableProxySpy` didn't seem to be doing anything by just being called passing in the `proxy` in the `useEffect ` with no dependencies. 

This PR changes it to be called on running of the queryFunction but I may be missing the intent where the `stableProxySpy` was just checking that the whole component hadn't some how been thrown away and recreated and so that's why the text from the mutation appeared?.... Just close if that's the case but I wouldn't mind a schooling on why this may occur in a test! 🙂 (Think this was you @sachinraja if you can give me the insight!)

## ✅ Checklist

### DIFF
--- a/packages/tests/server/react/useContext.test.tsx
+++ b/packages/tests/server/react/useContext.test.tsx
@@ -237,14 +237,14 @@ test('invalidate', async () => {
   const stableProxySpy = jest.fn();
 
   function MyComponent() {
-    const allPosts = proxy.post.all.useQuery();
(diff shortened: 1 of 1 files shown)


sachinraja (approved):
I think this is good!

PR #3244, opened 2022-11-24
test: attemps to fix flakey v10 tests

nilskj:
We have some flakey tests that passed at first but sometimes fails.

## 🎯 Changes
This is one thing that I found in the original v9 tests that was missed in my v10 port.
Trying solve this step by step so we know the cause.

## ✅ Checklist

### DIFF
--- a/packages/tests/server/adapters/fastify.test.ts
+++ b/packages/tests/server/adapters/fastify.test.ts
@@ -23,7 +23,7 @@ import fetch from 'node-fetch';
 import { z } from 'zod';
 
 const config = {
-  port: 2022,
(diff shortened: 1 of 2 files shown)


nilskj:
Flakey bacons! This looks good on ci, but now these fails locally for me:
```
FAIL server/interop/react/withTRPC.test.tsx (11.381 s)
FAIL server/react/regression/issue-1645-setErrorStatusSSR.test.tsx
```

KATT (approved):
Hmm.. I feel like we should always be listening to port `0` like we do in the other tests, but this might do it too!

PR #3261, opened 2022-11-25
feat: automatically serialize JSON to equivalent type

### LINKED ISSUE #1694: feat: if no transformer is defined - serialize types as JSON would
## Goal

Automatically make a safe data transfer object to corresponds to what `JSON.parse(JSON.stringify(obj))` would do.

These 2 are likely two separate PRs/implementations:

- For _outputs_ it's pretty straight-forward
- For _inputs_ we somehow need to disable complex types that are not serializable? Or do some other wild stuff.

**Let's start with outputs.**

<details><summary>Example</summary>
## Example

**Convert this...**

```ts
interface Post {
  author: string
  title: string
  datePublished: Date
}
```

**... into this:**

```ts
// to:
type Post = {
  author: string
  title: string
  datePublished: string // <-- automatic transform according to `JSON.parse()`
}
```

Using only TypeScript generics!

</details>

## Bounty 🤑 

$200

## Overview of how to do it

- Should be done on the `@trpc/client` / `@trpc/react`-level. 
- `router.createCaller()` should not be affected
- We have a `DefaultDataTransformer` that can be used to check if a router or procedure is using a transformer (see #3254 for some initial hacks)
- Might be some tests that have the types assertions (i.e. using `undefined` instead of `null` - update those). If the tests are in `interop/*`, do whatever hack you need to do to make them work

### References

Both MIT, so we can vendor one of these:

- https://github.com/sindresorhus/type-fest/blob/main/source/jsonify.d.ts
- https://github.com/remix-run/remix/blob/main/packages/remix-server-runtime/serialize.ts

KATT:
Closes #1694

## 🎯 Changes

This PR adds support to automatically serialize the output as the JSON-representation of the same object.

- ~~It's currently using `type-fest` as a dependency, which should probably be changed~~ Using what Remix does
- This PR has the assumption that any `@trpc/client` or `@trpc/react-query` calls will happen over the HTTP-boundary and that the response will be JSON. This is a fair assumption right now but if/when we add for instance an "SSR-link" you'll get unexpected results
- See the test for an example: 
  - https://github.com/trpc/trpc/blob/74f27672977368e07731cbfaf42d549a377947b7/packages/tests/server/jsonify.test.ts#L15-L25
  - https://github.com/trpc/trpc/blob/74f27672977368e07731cbfaf42d549a377947b7/packages/tests/server/jsonify.test.ts#L35-L47

## ✅ Checklist

- [x] Remove `type-fest` somehow


packages/server/src/shared/internal/serialize.ts
```diff
@@ -0,0 +1,59 @@
...
+type UndefinedToOptional<T extends object> = {
+  // Property is not a union with `undefined`, keep as-is
+  [k in keyof T as undefined extends T[k] ? never : k]: T[k];
+} & {
+  // Property _is_ a union with `defined`. Set as optional (via `?`) and remove `undefined` from the union
+  [k in keyof T as undefined extends T[k] ? k : never]?: Exclude<
+    T[k],
+    undefined
+  >;
+};
```
sachinraja, line 59: This breaks jump to definition for properties

sachinraja: pushed a fix

sachinraja: yeah it does :(

sachinraja: i fixed it for real now

packages/server/src/shared/internal/serialize.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/** JSON serialize [tuples](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) */
+type SerializeTuple<T extends [unknown, ...unknown[]]> = {
+  [k in keyof T]: T[k] extends NonJsonPrimitive ? null : Serialize<T[k]>;
+};
+
+/** JSON serialize objects (not including arrays) and classes */
+type SerializeObject<T extends object> = {
+  [k in keyof T]: T[k] extends NonJsonPrimitive ? never : Serialize<T[k]>;
+};
```
sachinraja, line 43: right now any keys of type `NonJsonPrimitive` still show up as `never` because when I tried applying the filter keys strategy, TS broke

sachinraja (approved):
good for now, jump to definition works

sam3d:
This is an amazing quality of life improvement! Thank you ❤️🥳

PR #3264, opened 2022-11-25
hide never values in intellisense for jsonify

sachinraja:
continuation of #3261

### DIFF
--- a/packages/server/src/shared/internal/serialize.ts
+++ b/packages/server/src/shared/internal/serialize.ts
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/naming-convention */
+import { FilterKeys } from '../../types';
 
 /**
  * @link https://github.com/remix-run/remix/blob/2248669ed59fd716e267ea41df5d665d4781f4a9/packages/remix-server-runtime/serialize.ts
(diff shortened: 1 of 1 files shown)


KATT (approved):
Can we add some tests here? https://github.com/trpc/trpc/blob/main/packages/tests/server/jsonify.test.ts

packages/server/src/shared/internal/serialize.ts
```diff
@@ -49,14 +50,16 @@ type SerializeObject<T extends object> = {
...
   // Property is not a union with `undefined`, keep as-is
-  [k in keyof Pick<T, FilterDefinedKeys<T>>]: T[k];
-} & {
-  // Property _is_ a union with `defined`. Set as optional (via `?`) and remove `undefined` from the union
-  [k in keyof Omit<T, FilterDefinedKeys<T>>]?: Exclude<T[k], undefined>;
-};
+  Pick<T, FilterDefinedKeys<T>> & {
+    // Property _is_ a union with `defined`. Set as optional (via `?`) and remove `undefined` from the union
+    [k in keyof Omit<T, FilterDefinedKeys<T>>]?: Exclude<T[k], undefined>;
+  };
```
KATT, line 65: This looks quite "heavy" but before looked heavy too... reminds me of our `OmitNeverKeys` that ended up being slow.

FWIW, I think it's fine to expose never keys.... maybe an even better thing would be some DX like this:

```ts
const output = {
  bigint: 0n,
}

// ... then
serializedOutput.bigint
//                            ^? Error<"This type is not JSON-serializable">
```

sachinraja: yeah this is the same as before, I just refactored it a bit. This part doesn't change with the `never` keys. It just has to be done unless we somehow come up with something more optimized.

PR #3273, opened 2022-11-26
Make CI checks to run on all pull requests

### LINKED ISSUE #3272: feat: make the CI work in with forks

Right now the CI doesn't work with fork, leading us to add a lot of people with write access. This is annoying and should be unnecessary

Make our PRs work with third party 

.

### Additional information

@Andarist looked at this and thinks it's because our push trigger in our yaml workflows.

Example: https://github.com/trpc/trpc/blob/22f268d1a16daa9154db2a456323366471fe30d7/.github/workflows/main.yml#L3

We should change this to only run on pull requests. Running stuff on `main` is unnecessary as our branch protection rules prevent us from merging if it doesn't pass anyway.

I guess it's good to run at least the _"Build & Test"_ one on `main`.

### 👨‍👧‍👦 Contributing

Andarist:
Closes #3272

## 🎯 Changes

I think that this should suffice your needs but if anything is missing - let me know, I can tweak this further.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,10 @@
 name: Build & Test
 
-on: [push]
+on:
+  push:
+    branches: main
+  pull_request:
+    branches: '**'
 
 jobs:
   test:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,5 +58,5 @@ jobs:
 
       - uses: codecov/codecov-action@v1
         with:
-          token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
+          # token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
           directory: ./packages
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,6 @@
 name: Lint
 
-on: [push]
+on: [pull_request]
 
 concurrency:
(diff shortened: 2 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -1,6 +1,10 @@
 name: Build & Test
 
-on: [push]
+on:
+  push:
+    branches: main
+  pull_request:
+    branches: '**'
```
QuiiBz, line 7: We can remove this line if we want to run this workflow on every PRs
```suggestion
```

Andarist: That's what I did at first but a value-less field/property looked odd visually to me so I've added the value~ here. This should also execute for all branches

I don't mind adjusting it to the proposed version if that's what you prefer

.github/workflows/main.yml
```diff
@@ -1,6 +1,10 @@
...
 
 jobs:
   test:
```
Andarist, line 10: Actually, I've now noticed that this workflow depends on some secrets and those are not available to pull requests created by outside collaborators.

I think that Turbo should be able to fall back to just executing everything without using its remote cache (that's acceptable to me) and Codecov doesn't require the token for public repos... but you still provide it:
https://github.com/trpc/trpc/blob/22f268d1a16daa9154db2a456323366471fe30d7/.github/workflows/main.yml#L57

Can that explicit token just be removed?

KATT:
Moving this to #3275

PR #3289, opened 2022-11-27
feat: require transformers on frontend & backend when present

### LINKED ISSUE #3276: feat: improve DX when setting up transformers

A very common pitfall is that we get bug reports that "tRPC is not working" because the user has defined the transformer on only the backend or only the frontend.

Leverage the type `DefaultDataTransformer` similarly to how we did on #3261 to force the user to define a frontend transformer if they've defined a backend one.

- Not doing it
- Some dev only warning? We can probably detect this?
- FAQ page (#2740)

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

QuiiBz:
Closes #3276

## 🎯 Changes

Require the same transformer type on frontend and backend:

- When a transformer is set on the backend but not on the frontend, the frontend will throw an error and tell us that `transformer` is required
- When a transformer is set on the frontend but not on the backend, the frontend will throw an error and tell us that `transformer` is not assignable to `You must set a transformer on the backend router` (note: I'm not sure if this is the best way to inform the user about the issue)

https://user-images.githubusercontent.com/43268759/208251565-5997d330-df84-4759-ba2c-17cbf63e47fb.mov

## ✅ Checklist


KATT (commented):
Needs a test :)

You can create a regression test or just a file

Use `// @ts-expect-error` to assert it fails

Ideally we'd test React & Next too but I think they all use `CreateTRPCClientBaseOptions`

QuiiBz:
Yeah, still working on it to also throw an error when we provide a transformer on the frontend but not on the backend. Wil add tests after that!

KATT:
> [...] when we provide a transformer on the frontend but not on the backend. Wil add tests after that!

That seems impossible I think, feels like a circular dep 😅

QuiiBz:
> That seems impossible I think, feels like a circular dep 😅

Just found a trick, by throwing the error on the frontend instead of the backend. But not sure if that's really a good way to do it:

https://github.com/trpc/trpc/pull/3289/files#diff-17d6fa12f247458a34d34288610bbd74c9879d55a28b6c972aba8fe69af1528fR27-R29

KATT:
Did this die? :)

QuiiBz:
Sorry, I forgot to fix the last types errors. Will check asap

QuiiBz:
I've just updated the PR description with explanations and a video, it's working pretty well. All the errors are thrown on the frontend since we use the type of the router here.

KATT (approved):

PR #3294, opened 2022-11-27
feat: better handling for bad responses

### LINKED ISSUE #3085: transformResult generates error when called as result of 500 response from server

```
 System:
    OS: macOS 12.5
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 219.13 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.18.0 - /usr/local/bin/node
    npm: 8.19.2 - /usr/local/bin/npm
  Browsers:
    Chrome: 106.0.5249.119
    Firefox: 91.0.2
    Safari: 15.6
  npmPackages:
    typescript: * => 4.8.2 

```

### Describe the bug

If the response received by TRPCClient has not been generated by TRPCServer (e.g. it is an unexpected server error, in my case due to an unresponsive database), then the error is passed through to transformResult, which produces another error at [this line](https://github.com/trpc/trpc/blob/040e0fa7fbd1941c9d3ba11faf5d2194faaf64cd/packages/client/src/links/internals/transformResult.ts#L34) ...
```
    ...((!response.result.type || response.result.type === 'data') && {
``` 
...because response.result does not exist. 

This appears to be actually caused by [resolving a promise which should actually be rejected](https://github.com/trpc/trpc/blob/040e0fa7fbd1941c9d3ba11faf5d2194faaf64cd/packages/client/src/links/httpBatchLink.ts#L88), which causes transformResult to be called when it shouldn't be.

The resultant TRPCClientError now contains no details of the actual error cause, but instead shows the line pasted above as the cause of the error.

### Link to reproduction

https://stackblitz.com/edit/github-ujrsmm

### To reproduce

Make the server return a response with error code 500 and non-TRPC standard error format. I have done this by replacing the TRPC handler with a function that mimics the error I am receiving from my server. 

At time of writing StackBlitz is having a fit and not rendering the repro app... hopefully it works shortly

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

KATT:
Closes #3085

## 🎯 Changes

Detect better responses look off and break early.


KATT:
cc @andrewbateman do you think this would do it for you?

andrewbateman:
@KATT yep that should do it nicely!

KATT:
Waiting for your review @sachinraja :)

sachinraja (approved):

PR #3298, opened 2022-11-27
feat(reactQuery): add inference helpers for query/mutation parameters

### LINKED ISSUE #3293: feat: infer the `options` type of `useQuery`?
Add helpers for inferring options from `useQuery`, `useInfiniteQuery`, & `useMutation`.

### Additional context

- Simply using `Parameters<typeof trpc.post.byId.useQuery>` does not work because it is a generic function.
- We should probably `Omit<Options, 'select'>` as it won't work here

### Discussed in https://github.com/trpc/trpc/discussions/3287

<sup>Originally posted by **nkalpak** November 27, 2022</sup>
To set the stage: I would like to avoid sprinkling `trpc.something.useQuery` calls inside the codebase, because this makes it difficult to setup some common options across API resources (for example, `trpc.note.getOne` needs to always have `staleTime: Infinity`).

The way I've historically done this with React Query is by using custom hooks which utilize a helper type to infer the options of queries and mutations through the query/mutation function (ex. `function useNote(options: InferQueryOptions<typeof getNoteApi>) { ... }`). Now that I'm using TRPC, I would like to do the same thing, and I've managed to get it working for [mutations](https://codesandbox.io/s/nervous-sanderson-m9vuvw?file=/src/mutation.ts).

The issue that I ran into is queries. The same approach which does work for mutations [doesn't seem to work for queries](https://codesandbox.io/s/nervous-sanderson-m9vuvw?file=/src/query.ts). So my question is really about how to approach this, what would be the idiomatic way of inferring query options for a particular procedure?

> I should note that I know that by doing this inference on the query itself I lose the type information for (ex.) `select`, and that if there is a `select` inside of the custom hook the return type of the custom hook will not reflect the return type of the `select`. I am OK with this limitation and I don't expect the TRPC solution to support this use-case.</div>

nkalpak:
Closes #3293 

## 🎯 Changes

This implements inference helpers for:
- Options and input when used with a query
- Options when used with a mutation

The greater context for this feature can be found in #3287 

## ✅ Checklist


KATT:
cc @juliusmarminge do you wanna review? you did do the other inference helpers :)

juliusmarminge:
> cc @juliusmarminge do you wanna review? you did do the other inference helpers :)

Sure, I'll get on it tomorrow PM

www/docs/server/infer-types.md
```diff
@@ -115,3 +115,22 @@ async function main() {
...
+import type { InferReactQueryProcedureOptions } from '@trpc/react-query';
+import type { AppRouter } from './server';
+
+type ReactQueryProcedure = InferReactQueryProcedureOptions<AppRouter>;
+// Infers the mutation options for `post.create.useMutation`
+type PostCreateOptions = ReactQueryProcedure['post']['create']['options'];
+// Infers the query options for `post.byId.useQuery`
+type PostByIdOptions = ReactQueryProcedure['post']['byId']['options'];
+// Infers the query input for `post.byId.useQuery`
+type PostByIdInput = ReactQueryProcedure['post']['byId']['input'];
```
juliusmarminge, line 135: Not sure about this API - can we make it more coherent with the current [inference helpers](https://trpc.io/docs/infer-types#inference-helpers)? Also, since we already have inference helpers for input/output, I'm not sure how much sense to have them on here too?

```ts
// utils file
import type { inferReactProcedureOptions } from "@trpc/react-query";
import type { inferRouterInputs } from "@trpc/server";
export type RouterOptions = inferReactProcedureOptions<AppRouter>;
export type RouterInputs = inferRouterInputs<AppRouter>;
---------
// usage file
type PostCreateInput = RouterInputs["post"]["create"];
type PostCreateOptions = RouterOptions["post"]["create"];
```

nkalpak: 100% agree on the lower camel case to keep the convention of `inferRouterInputs`.

I am not too sure about excluding "react query" in `inferReactProcedureOptions` and `RouterOptions` though, as doing so implies that the options belong to the router, while they are actually only relevant to react query.

My thinking about the `options`/`input` accessors is: it keeps the related utilities in one place, so if you want to make a custom "post create" mutation hook you won't need to import both the `PostCreateInput` and `PostCreateOptions`, you can just import `PostCreateOptions` and have the input there. But I also see your point about having the redundancy of having the same thing available in two different places, and my thinking here might be too driven by my own use case.

Summary:
1. `InferReactQueryProcedureOptions` -> `inferReactProcedureOptions` - unless we agree that we need to keep the "react query" specifier, in that case `inferReactQueryProcedureOptions`
2. Exclude the `options` key from the output of the inference helper, and instead make `options` the result itself - unless we agree that it is worth it to increase redundancy for the benefit of having both `input`/`option` in one inference helper
3. Update the docs according to the changes made

Let me know what you think, I would be happy to implement any requested changes!

juliusmarminge: 1. Yea we can keep the Query in the same

2. I still think having it return only options would make most sense, and would help keep them as similar as possible so that we have a uniform api for all inference helpers 👍👍

packages/react-query/src/utils/inferReactQueryProcedure.ts
```diff
@@ -0,0 +1,53 @@
...
+  TRouter extends AnyRouter,
+  TPath extends string = '',
+> = {
+  [TKey in keyof TRouter['_def']['record']]: TRouter['_def']['record'][TKey] extends infer TRouterOrProcedure
+    ? TRouterOrProcedure extends AnyRouter
+      ? inferReactQueryProcedureOptions<
+          TRouterOrProcedure,
+          `${TPath}${TKey & string}.`
+        >
+      : TRouterOrProcedure extends AnyProcedure
```
juliusmarminge, line 43: this extra check here shouldn't be necessary since we're checking above if its a router or procedure. not sure what effects on perf it might have but can simplify it down one level

packages/react-query/src/utils/inferReactQueryProcedure.ts
```diff
@@ -0,0 +1,53 @@
...
+      ? TRouterOrProcedure extends AnyMutationProcedure
+        ? InferMutationOptions<TRouterOrProcedure>
+        : TRouterOrProcedure extends AnyQueryProcedure
+        ? InferQueryOptions<TRouterOrProcedure, `${TPath}${TKey & string}`>
+        : never
+      : never
+    : never;
+};
+
+export type { inferReactQueryProcedureOptions };
```
juliusmarminge, line 53: nit but everywhere else on the repo we just export straight from the type 

```ts
export type T = ...
```

juliusmarminge (changes requested):
Found a bug regarding serialization. The types here aren't matching.

CleanShot 2022-12-01 at 11 37 24@2x

CleanShot 2022-12-01 at 11 43 27@2x

Since the tests aren't using superjson, the bottom one is correct and top one should fail I believe.

Needs to be confirmed

nkalpak:
> Found a bug regarding serialization. The types here aren't matching.
> 
> CleanShot 2022-12-01 at 11 37 24@2x
> 
> CleanShot 2022-12-01 at 11 43 27@2x
> 
> Since the tests aren't using superjson, the bottom one is correct and top one should fail I believe.
> 
> Needs to be confirmed
> 
> Below is with superjson, then they both match and types are serialized
> 
> CleanShot 2022-12-01 at 11 55 37@2x

Great catch!

I guess when you use `superjson` the type remains Date, while when you don't, it gets automatically serialized to JSON-compatible (string). I've updated the inference helpers to use the transformed procedure output, and now it correctly serializes the type when there is no transformer.

I left the `expectedTypeOf` tests to test for the serialized outputs, since the `useQuery` and `useMutation` tests originally didn't use a transformer, so I didn't want to add one.

juliusmarminge:
> Great catch!
>
> I guess when you use superjson the type remains Date, while when you don't, it gets automatically serialized to JSON-compatible (string). I've updated the inference helpers to use the transformed procedure output, and now it correctly serializes the type when there is no transformer.
>
> I left the expectedTypeOf tests to test for the serialized outputs, since the useQuery and useMutation tests originally didn't use a transformer, so I didn't want to add one.

Looks good now 👍🏼, however I do think a few extra tests would be valuable here so that we don't regress here again. I found the same bug here #3323 and added tests with both transformer and without

juliusmarminge (approved):
Looks good from my side. I'll let @KATT decide if he wants more tests or if this is enough

www/docs/server/infer-types.md
```diff
@@ -115,3 +115,27 @@ async function main() {
...
+// @include: server
+// ---cut---
+// @filename: client.ts
+import type { inferReactQueryProcedureOptions } from '@trpc/react-query';
+import type { inferRouterInputs } from '@trpc/server';
+import type { AppRouter } from './server';
+
+// utils file
+type ReactQueryOptions = inferReactQueryProcedureOptions<AppRouter>;
+type RouterInputs = inferRouterInputs<AppRouter>;
```
KATT, line 132: export

KATT:
```suggestion
export type ReactQueryOptions = inferReactQueryProcedureOptions<AppRouter>;
export type RouterInputs = inferRouterInputs<AppRouter>;
```

KATT (commented):
Will look properly tomorrow


- Use shiki twoslash to improve docs? That will also ensure we export this type and that the example is correct across multiple files
- A fuller example would be good in general

www/docs/server/infer-types.md
KATT, line 132:
```suggestion
export type ReactQueryOptions = inferReactQueryProcedureOptions<AppRouter>;
	export type RouterInputs = inferRouterInputs<AppRouter>;
```

www/docs/server/infer-types.md
```diff
@@ -115,3 +115,27 @@ async function main() {
...
 main();
 ```
+
+## Infer React Query options based on your router
+
+```ts twoslash title="client.ts"
+// @module: esnext
+// @include: server
+// ---cut---
+// @filename: client.ts
```
KATT, line 125:
```suggestion
// @filename: trpc.ts
```

www/docs/server/infer-types.md
```diff
@@ -115,3 +115,27 @@ async function main() {
...
+import type { inferReactQueryProcedureOptions } from '@trpc/react-query';
+import type { inferRouterInputs } from '@trpc/server';
+import type { AppRouter } from './server';
+
+// utils file
+type ReactQueryOptions = inferReactQueryProcedureOptions<AppRouter>;
+type RouterInputs = inferRouterInputs<AppRouter>;
+
+// usage file (ex. your custom useCreatePost hook)
```
KATT, line 134:
```suggestion
// @filename: useCreatePost.ts

import { ReactQueryOptions } from './trpc';



```

www/docs/server/infer-types.md
```diff
@@ -115,3 +115,27 @@ async function main() {
...
+type PostCreateOptions = ReactQueryOptions['post']['create'];
```
KATT, line 135:
```suggestion
type PostCreateOptions = ReactQueryOptions['post']['create'];
//       ^?
```

nkalpak: I tried this originally, but the output is just the expanded `UseTRPCMutationOptions` generic:
image

juliusmarminge: Maybe do it on the onSuccess's data or something to showcase a custom hook being typed?

www/docs/server/infer-types.md
```diff
@@ -133,12 +143,21 @@ export type RouterInputs = inferRouterInputs<AppRouter>;
...
+// @noErrors
+import { type ReactQueryOptions, trpc } from './trpc';
 
 type PostCreateOptions = ReactQueryOptions['post']['create'];
 
 function usePostCreate(options?: PostCreateOptions) {
-  return trpc.post.create.useMutation(options);
+  const utils = trpc.useContext();
+  return trpc.post.create.useMutation({
+    onSuccess(post) {
```
juliusmarminge, line 154: not sure why but couldn't get shake working here, got implicit any so had to smash a `@noErrors` on it too :/

juliusmarminge (dismissed):
added a bit more to the example @KATT.

Looks good on my part

CleanShot 2022-12-12 at 13 30 42@2x

juliusmarminge (approved):

KATT (approved):

Issue #3012, opened 2022-10-15
feat: Additional Router Level Tanstack Query Methods


JonParton:
### Describe the feature you'd like to request

> ℹ️ Transferring here so the conversation isn't left on a completed PR

As requested by @TkDodo [here](https://github.com/trpc/trpc/pull/2713#issuecomment-1278887209) currently we only expose the [`invalidate`](https://trpc.io/docs/v10/query-invalidation#example-code) method on Routers from the tRPC Proxy (Implemented in #2711); However Tanstack-Query exposes many other fuzzy key match methods which are below: 

-----
- [invalidateQueries](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientinvalidatequeries)✅
- [getQueriesData](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientgetqueriesdata)
  - Returns the data for only queries that are currently in the Tanstack Query cache ...not trigger TRPC calls
- [refetchQueries](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientrefetchqueries)
  - TS types could be interesting ... but I think possible.
- [cancelQueries](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientcancelqueries)
- [removeQueries](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientremovequeries)
- [resetQueries](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientresetqueries)
- [getQueryDefaults](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientgetquerydefaults)
- [setQueryDefaults](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientsetquerydefaults)
- [setQueriesData](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientsetqueriesdata)
  - For the "updater" function to be trueley type safe when operating on multiple queries we would need to know which particular key the updater was currently working on.... I know this [was on me to go look at adding to the tanstack API](https://twitter.com/TkDodo/status/1572261676434575360) but I haven't managed to prioritise.
- [getMutationDefaults](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientgetmutationdefaults)
- [setMutationDefaults](https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientsetmutationdefaults)
-----

@TkDodo  Would like to see if we can implement the other functionality as currently there is no way of doing these on our existing API. 

[...]

PR #3302, opened 2022-11-27
feat: expose querykey getter and fix some querykey stuff

### LINKED ISSUE #3295: feat: expose queryKey

not all methods available on the `queryClient` are available on `useContext()` from trpc. While it might be desirable to add more methods to it (see: https://github.com/trpc/trpc/issues/3012), it doesn't have to be done for _all_ methods - only those that profit from type-safety.

There are a couple of methods exposed from TanStack/query that do not require additional typesafety, but would need access to the `queryKey`. For example:

- `queryClient.setQueryDefaults(key, defaults)`
- `queryClient.isFetching({ queryKey })`
  - or even `useIsFetching({ queryKey })`

additional type-safety is not needed because for the first example, `defaults` work the same for all keys, and for the second example, only a _number_ is returned.

It would be good if we could use these methods with queries created from trpc.

One way to achieve this is to expose the queryKey that was created internally, for example:

```ts
const queryClient = useQueryClient()
const utils = trpc.useContext()

const isFetching = useIsFetching(utils.post.byId.getQueryKey()) // QueryKey => [["post","byId"]]
queryClient.setQueryDefaults(utils.post.getQueryKey(), { staleTime: Infinity }) // QueryKey => [["post"]]
```

we could expose _all_ methods via `trpc.useContext()`, but I think it's an uphill battle and not really needed. Some things (but not all of them) will be covered with:
- https://github.com/trpc/trpc/issues/3012

### Additional information

reference:
- https://github.com/trpc/trpc/issues/3012#issuecomment-1319653598

### 👨‍👧‍👦 Contributing

### LINKED ISSUE #3455: bug: query invalidation doesn't accept filters

```
  System:
    OS: macOS 12.5.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 77.77 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.13.0/bin/yarn
    npm: 8.1.0 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Brave Browser: 108.1.46.144
    Chrome: 108.0.5359.124
    Firefox Developer Edition: 108.0
    Safari: 15.6.1
  npmPackages:
    @tanstack/react-query: 4.20.4 => 4.20.4 
    @trpc/client: 10.6.0 => 10.6.0 
    @trpc/react-query: 10.6.0 => 10.6.0 
    @trpc/server: 10.6.0 => 10.6.0 
    next: 12.3.0 => 12.3.0 
    react: 17.0.2 => 17.0.2 
```

### Describe the bug

passing filters to `invalidate` doesn't seem to have an effect - no invalidation is happening

examples:

```
utils.invalidate({ predicate: () => true })
utils.invalidate({ refetchType: 'all' })
```

### Link to reproduction

https://codesandbox.io/s/fancy-pond-dznks2?file=/src/pages/index.tsx

### To reproduce

click the `invalidate` button.

## expected:

the query should be refetched

## actually:

no refetch happens

### Additional information

it works without filters, so this is fine:

```
utils.invalidate()
```

### 👨‍👧‍👦 Contributing

### LINKED ISSUE #3456: bug: query invalidation can only target trpc queries

```
  System:
    OS: macOS 12.5.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 77.77 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.13.0/bin/yarn
    npm: 8.1.0 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Brave Browser: 108.1.46.144
    Chrome: 108.0.5359.124
    Firefox Developer Edition: 108.0
    Safari: 15.6.1
  npmPackages:
    @tanstack/react-query: 4.20.4 => 4.20.4 
    @trpc/client: 10.6.0 => 10.6.0 
    @trpc/react-query: 10.6.0 => 10.6.0 
    @trpc/server: 10.6.0 => 10.6.0 
    next: 12.3.0 => 12.3.0 
    react: 17.0.2 => 17.0.2 
```

### Describe the bug

given that you have an environment where you're mixing `useQuery` from `react-query` and queries from `trpc`, `utils.invalidate()` _used to_ target all queries, but now, it only targets the queries created by trpc.

### Link to reproduction

https://codesandbox.io/s/festive-newton-rodx1p?file=/src/pages/index.tsx

### To reproduce

- click the `invalidate` button
- observe that only the query created by trpc is invalidated

### Additional information

This used to work with `10.0.0-rc.3`

from discord:

```
quick question: I think the change of query keys (that we now have { type: 'query'} etc. at the end) kind of introduced an issue when using trpc together with "raw" react-query queries. For reference, we are gradually upgrading to tprc and still have some "untyped" useQuery calls lying around.

Our issue is that we call utils.invalidate() to invalidate all queries in some cases. It used to target all trpc and react-query queries, but now, it only targets the trpc queries. I've debugged into the queryClient and what we get into the invalidation as a key is [[],{}] and this does match "all trpc queries", but not "all queries".

Our workaround is to use the real queryClient for invalidation now, just wanted to let you know about that - maybe it just worked coincidentally before
```

### 👨‍👧‍👦 Contributing

juliusmarminge:
- Closes #3295
- Closes #3455
- Closes #3456

## 🎯 Changes

- Adds a `getQueryKey` helper for routers and procedures

- fixes some other queryKey related bugs (see #3455 and #3456)

Docs link: https://www-git-julius-querykeys-trpc.vercel.app/docs/useContext#-the-function-i-want-isnt-here

### API

Since we can't distinguish routers and procedures in runtime, we can't dynamically decide the QueryType to call `getArrayQueryKey` with. Therefore, I made these explicit arguments that the user inputs:

```ts
getQueryKey(undefined, "any")   // for routers
getQueryKey(TInput, "infinite") // for infinite queries
getQueryKey(TInput, "query")    // for regular queries
```

The default QueryType is `any`.

Usage is as follows:

```ts
import { useQueryClient } from "@tanstack/react-query";
import { trpc } from "~/utils/trpc";

const MyComp = () => {
  const qc = useQueryClient();
  const qKey = trpc.post.getQueryKey(undefined, "any");
  setQueryDefaults(qKey, { staleTime: 30 * 60 * 1000 });
}
```

Another usage is for route prefetching:

https://github.com/trpc/trpc/blob/377e271b97d9392263595b44b2868795e75e818e/examples/tanstack-router/client/src/utils/router.tsx#L55-L61

## ✅ Checklist


packages/tests/server/react/useContext.test.tsx
```diff
@@ -798,4 +799,119 @@ describe('query keys are stored separtely', () => {
...
+        <App>
+          <MyComponent />
+        </App>,
+      );
+
+      // should be fetching initially, and then not
+      expect(utils.container).toHaveTextContent('1');
+      await waitFor(() => {
+        expect(utils.container).toHaveTextContent('0');
+      });
```
juliusmarminge, line 914: @TkDodo This should pass, right?

TkDodo: no, because you don't have a subscription. `qc.isFetching` just reads the value once. You'd want:

```
const isFetching = useIsFetching({ queryKey: key });
```

you can import it directly:

```
import { useIsFetching } from '@tanstack/react-query'
```

juliusmarminge: oh right.. thanks for clarifying 🙏🏼

packages/react-query/src/shared/proxy/utilsProxy.ts
```diff
@@ -158,6 +163,15 @@ type DecorateProcedure<
...
   ): InfiniteData<inferTransformedProcedureOutput<TProcedure>> | undefined;
+
+  /**
+   * Method to extract the query key for a procedure
+   * @link https://trpc.io/docs/useContext#-the-function-i-want-isnt-here
+   */
+  getQueryKey(
+    input: inferProcedureInput<TProcedure>,
+    type: QueryType,
+  ): QueryKey;
```
KATT, line 174: On my phone so I might be missing something, but wouldn't it make sense to have both arguments optional?

KATT: Cc @TkDodo wdyt?

juliusmarminge: Why would we default to? "query"?

The input would be optional if the procedure's input is optional, but we need the type (I wrote a bit about this up top or in discord with @sachinraja)

KATT: If it's getting the key for sale of invalidation or resetting, `any` is a good default

If its for getting or changing the cache `any` might not be the best

TkDodo: what is `"any"` going to return? does it match both `query` and `infinite` ? If so I think it would be a good default for `getQueryKey`

JonParton: `any` returns the query key without the type object in the key, so as long as fuzzy matching is enabled for the methods then it will match both!

Do all methods match fuzzy? Or for the "query" versions only match exact? This may need some documentation!

TkDodo: > Do all methods match fuzzy?

all methods _can_, but it's not the default everywhere. Generally speaking:
- everything that has a plural in its name is fuzzy, like `getQueriesData`, `invalidateQueries` or `setQueriesDefault`
- everything that is singular defaults to exact, like `getQueryData`

JonParton:
> ### Alternatives
> Since this method doesn't use the query client, it could go on the regular trpc object:
> 
> ```ts
> const qKey = trpc.post.getQueryKey(...);
> ```
> 
> but I prefer it being on the utils one, feels like it "belongs" there? (personal preference maybe?).
> 
> It does require an additional hook to be called though:
> 
> ```ts
> // this
> import { useQueryClient } from "@tanstack/react-query";
> import { trpc } from "~/utils/trpc";
> const MyComp = () => {
>   const qc = useQueryClient();
>   const utils = trpc.useContext();
>   const qKey = utils.post.getQueryKey(undefined, "any");
>   setQueryDefaults(qKey, { staleTime: 30 * 60 * 1000 });
> }
> 
> // over this
> import { useQueryClient } from "@tanstack/react-query";
> import { trpc } from "~/utils/trpc";
> const MyComp = () => {
>   const qc = useQueryClient();
>   const qKey = trpc.post.getQueryKey(undefined, "any");
>   setQueryDefaults(qKey, { staleTime: 30 * 60 * 1000 });
> }
> ```
> 
> What do others think?

First @juliusmarminge; awesome job!! 

On the above, I also see the point around it living on the `useContext()` object mainly because we tend to call the object returned `utils` in our examples. 

However, I think there is value in it being directly on the trpc object or as a seperate helper entirely (that could be instaitated with AppRouter) . The reason being that this then opens it up to use outside of a react component, and I have already come across some use cases ( [Example here ](https://github.com/trpc/trpc/issues/3012#issuecomment-1279960638)  ) for doing this. I also know there is always a request around interacting with the query client outside of react context as well in many issues that get posted here... So it would be allowing the wider use of the escape hatch in those cases too!

packages/react-query/src/shared/proxy/utilsProxy.ts
JonParton: The fuzzy matching behaviour queried above for `xxxxxQuery` type methods also effects if the `input` param should be optional!... Although I think it should be as I know some deffinitely fuzzy match!

JonParton: Hmmn this may mean making things optional / default to a query key of `any` for the main path I expect this being used for (we provide many of the tanstack methods for individual queries already so this is likely to be used in the fuzzy matching ones!) but then put some good documentation on this noting the gotcha between needing exact keys for methods that are named `xxxxxQuery`

juliusmarminge:
> The reason being that this then opens it up to use outside of a react component, and I have already come across some use cases ( Example here ) for doing this. I also know there have been several requests in issues posted on the repo around interacting with the tanstack query queryClient outside of react context... So this would allow that wider use of the escape hatch too!

That's a fair point. That would allow users to use tRPC with @tanstack/react-router too for data loaders I believe. This is probably the way. What's your thought here @KATT @TkDodo ?

sachinraja:
I would put it on the `trpc` object, I think the data loaders stuff is a strong argument in favor of that.

examples/tanstack-router/README.md
```diff
@@ -0,0 +1,13 @@
+# A minimal React tRPC example using @tanstack/react-router
```
juliusmarminge, line 1: Just a playground for seeing the usage - I don't think we should keep this but instead contribute to https://github.com/TanStack/router/tree/beta/examples/react/with-trpc-react-query

examples/tanstack-router/client/src/utils/router.tsx
```diff
@@ -0,0 +1,145 @@
...
+  path: 'posts',
+  loaderMaxAge: 0,
+  errorComponent: () => 'Oh crap!',
+  loader: async () => {
+    const postKey = trpc.post.all.getQueryKey();
+
+    queryClient.getQueryData(postKey) ??
+      (await queryClient.prefetchQuery(postKey, getPosts));
+    return {};
+  },
```
juliusmarminge: here we got the usage in the loaders

examples/tanstack-router/client/src/utils/router.tsx
```diff
@@ -0,0 +1,145 @@
...
+  loader: async () => {
+    const postKey = trpc.post.all.getQueryKey(undefined, 'query');
+
+    queryClient.getQueryData(postKey) ??
+      (await queryClient.prefetchQuery(postKey, getPosts));
+    return {};
+  },
+
+  component: () => {
+    const postsQuery = trpc.post.all.useQuery();
```
juliusmarminge, line 64: this is how i thought it'd be useful

juliusmarminge (commented):
added a temporary example - should revert before merge (i think?)

sachinraja (approved):

examples/tanstack-router/client/src/utils/router.tsx
```diff
@@ -0,0 +1,145 @@
+/* eslint-disable react-hooks/rules-of-hooks */
```
juliusmarminge, line 1:
```suggestion
/**
 * @see https://tanstack.com/router/v1/docs/examples/react/basic
 **/
/* eslint-disable react-hooks/rules-of-hooks */
```

aulneau:
this is great! hope it can be merged soon

www/docs/reactjs/useContext.mdx
```diff
@@ -95,6 +95,49 @@ These are the helpers you'll get access to via `useContext`. The table below wil
...
+### ❓ The function I want isn't here!
+
+`@tanstack/react-query` has a lot of functions that we haven't put in the tRPC context yet. If you need a function that isn't here, feel free to [open a feature request](https://github.com/trpc/trpc/issues/new/choose) requesting it.
+
+In the meantime, you can import and use the function directly from `@tanstack/react-query`. We provide a `getQueryKey` helper on routers and procedures so that you can easily provide the native function the correct query key.
+
+```tsx
+function getQueryKey(
+  input: TInput,
+  type?: QueryType;
```
TkDodo, line 107: can we document the default value of `type` here? Looking at the code, I think it defaults to `"any"` ?

juliusmarminge: Yes definetely should be documented

examples/tanstack-router/client/src/utils/router.tsx
```diff
@@ -0,0 +1,145 @@
...
+        <div>Select a post.</div>
+      </>
+    );
+  },
+});
+
+const postRoute = postsRoute.createRoute({
+  path: '$postId',
+  loader: async ({ params }) => {
+    const postKey = trpc.post.byId.getQueryKey({ id: params.postId }, 'query');
```
KATT, line 108: Shouldn't we be able to use `.prefetch` here? Or is it b/c it's only on the context? Is there a way to make that happen?

examples/tanstack-router/client/src/utils/router.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+    queryClient.getQueryData(postKey) ??
+      (await queryClient.prefetchQuery(postKey, () =>
+        getPostById(params.postId),
+      ));
```
KATT, line 113: Is this run on the frontend and/or the backend? `import`ing backend stuff on something that is "universal" always scares me

examples/tanstack-router/client/src/utils/router.tsx
```diff
@@ -0,0 +1,145 @@
+/* eslint-disable react-hooks/rules-of-hooks */
+import {
+  Link,
+  Outlet,
+  createReactRouter,
+  createRouteConfig,
+  useMatch,
+} from '@tanstack/react-router';
+import { getPostById, getPosts } from '../../../server/fetchers';
```
KATT, line 9: Importing stuff from the server here shouldn't be needed, right? See comments further down

examples/tanstack-router/README.md
KATT: Seems like they haven't solved the data loading either? https://github.com/TanStack/router/blob/484af7fd844c3d4b02d67846ec268bc7310b9aba/examples/react/with-trpc-react-query/client/main.tsx#L111-L114

juliusmarminge: Maybe just remove the example for now then and revisit later?

examples/tanstack-router/client/src/utils/router.tsx
juliusmarminge: That doesn't work since this is outside the react tree and we don't have our context stuff here. Might be possible to do something similar to the SSG helpers (or maybe they even work?)

juliusmarminge:
I removed the example for now @KATT, maybe we can revisit it later but I dont think it should block this PR.

examples/tanstack-router/client/src/utils/router.tsx
KATT: Hmmmmmm..... I don't like embedding this example when the DX is not optimal

www/docs/reactjs/useContext.mdx
```diff
@@ -95,6 +95,49 @@ These are the helpers you'll get access to via `useContext`. The table below wil
...
+
+/** @note signature on router doesn't take any parameters */
+```
+
+```tsx
+import { useIsFetching, useQueryClient } from '@tanstack/react-query';
+import { trpc } from '~/utils/trpc';
+
+function MyComponent() {
+  const qc = useQueryClient();
```
KATT, line 123: We should call the var just `queryClient`

KATT (approved):

KATT (dismissed):

pnpm-lock.yaml
```diff
@@ -119,7 +119,7 @@ importers:
       react: 18.2.0
       react-dom: 18.2.0_react@18.2.0
       superjson: 1.11.0
-      zod: 3.20.2
+      zod: 3.19.1
```
KATT, line 122: Bad

juliusmarminge: weird..........

fixed now though

KATT (dismissed):

packages/tests/server/react/useContext.test.tsx
```diff
@@ -363,6 +367,71 @@ test('invalidate procedure for both query and infinite', async () => {
...
+
+    return (
+      <>
+        <button
+          data-testid="invalidate"
+          onClick={() => {
+            utils.invalidate(undefined, {
+              predicate: (query) => {
+                return (query.queryKey[0] as string[])[0] === 'post';
+              },
```
juliusmarminge, line 393: would love a bit nicer test here

packages/tests/server/interop/react/invalidateQueries.test.tsx
```diff
@@ -227,7 +227,7 @@ describe('invalidateQueries()', () => {
       'invalidate-2-tuple',
       'invalidate-3-exact',
       'invalidate-4-all',
-      'invalidate-5-predicate',
+      // 'invalidate-5-predicate',
```
juliusmarminge, line 230: not sure why it's failing but cba fixing this right now - is it necessary to fix this for interop? I doubt anyone will notice it not working if they haven't already xd

KATT: I'd like not to have conscious regression but make an issue of it and we can do it later

KATT (approved):

packages/tests/server/interop/react/invalidateQueries.test.tsx
juliusmarminge: Done, #3460

packages/tests/server/interop/react/invalidateQueries.test.tsx
```diff
@@ -192,12 +192,15 @@ describe('invalidateQueries()', () => {
...
                 predicate(opts) {
                   const { queryKey } = opts;
-                  const [path, input] = queryKey;
+                  const [path, rest] = queryKey;
 
-                  return path === 'count' && input === 'test';
+                  return (
+                    JSON.stringify(path) === JSON.stringify(['count']) &&
+                    (rest as any)?.input === 'test'
+                  );
```
juliusmarminge, line 203: nvm I found it and fixed it here instead

packages/tests/server/react/invalidateQueries.test.tsx
```diff
@@ -186,7 +186,7 @@ describe('invalidateQueries()', () => {
           <button
             data-testid="invalidate-4-predicate"
             onClick={() => {
-              utils.invalidate({
+              utils.invalidate(undefined, {
```
KATT, line 189: Isn't this changing the behaviour? 🤔 

Looks like a breaking change?

packages/tests/server/interop/react/invalidateQueries.test.tsx
```diff
@@ -192,12 +192,15 @@ describe('invalidateQueries()', () => {
           <button
             data-testid="invalidate-5-predicate"
             onClick={() => {
-              utils.invalidateQueries({
+              utils.invalidateQueries(undefined, {
```
KATT, line 195: Breaking chaange?

KATT (commented):
Hmm, is this a breaking change?

packages/tests/server/react/invalidateQueries.test.tsx
juliusmarminge: I mean yea theoretically although it wasnt working before so its more of a fix?

juliusmarminge: Mentioned this here: https://discord.com/channels/867764511159091230/1055437438266392646/1055450208424706099

We always take the first arg as input no matter what the types said so any filters applied would actually end up being supressed

packages/tests/server/interop/react/invalidateQueries.test.tsx
juliusmarminge: See above

KATT (approved):

TkDodo:
awesome - thanks for this ❤️

PR #3305, opened 2022-11-28
feat: add `useQueries`-support

### LINKED ISSUE #3188: feat: support useQueries

TanStack query supports running multiple queries in parallel via the `useQueries` hook, but this option is missing in trpc.

Add support for the `useQueries` hook. As discussed with Alex on Discord, the api could be something like:

```ts
trpc.useQueries(t => [
  t.post.byId({ 1 }, staleTime: Infinity)
  // ...
], { context: x })
```

```ts
trpc.useQueries(t => postIds.map(id => t.post.byId({ id }), { context: x })
```

If the `queryKey` and `queryFn` from trpc were exposed as a `raw` object (with queryKey and queryFn)

we could achieve this with `useQueries` from TanStack/query:

```ts
const utils = trpc.useContext()

useQueries({
  queries: [
    utils.todos.list.rawQuery(),
    utils.post.list.rawQuery(),
    utils.post.detail(5).rawQuery()
  ]
})
```

but it wouldn't be typesafe.

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

sachinraja:
Closes #3188

## 🎯 Changes
support TanStack Query's [useQueries](https://tanstack.com/query/v4/docs/reference/useQueries)

need:
- more tests
- docs
- way to specify options

Example:
```ts
const Component = () => {
  const results = trpc.useQueries({
    queries: [
      utils.post.byId.getQueryOptions({ id: "1" }),
      utils.greeting.getQueryOptions({ text: "yo" }),
    ],
  });
};
```

cc @TkDodo


packages/react-query/src/shared/proxy/utilsProxy.ts
```diff
@@ -265,6 +292,18 @@ export function createReactQueryUtilsProxy<
           context.setInfiniteQueryData(queryKey, updater, ...rest),
         getData: () => context.getQueryData(queryKey),
         getInfiniteData: () => context.getInfiniteQueryData(queryKey),
+        getQueryOptions: () => {
```
TkDodo, line 295: this is nice 🙌 . I'm just wondering if this somehow overlaps with:

https://github.com/trpc/trpc/pull/3302/files#diff-ef4e1d08a351372bae0fff79abe60481bb5e00914f176984386eaa48c1e209bdR288

which exposes a function to get the `queryKey` only.

If we have a function that returns all options, wouldn't the one for the `queryKey` only be a bit unnecessary, as you could just as well do:

```
utils.post.byId.getQueryOptions({ id: '1' }).queryKey
```

or maybe the extra function is fine, but it could be implemented with that internally - or is this different?

packages/tests/server/react/useQueries.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+  })
+  .done();
+
+test('useQueries()', async () => {
+  const { proxy, App } = ctx;
+  function MyComponent() {
+    const utils = proxy.useContext();
+    const results = proxy.useQueries({
+      queries: [utils.post.byId.getQueryOptions({ id: '1' })],
+    });
```
KATT, line 46: WDYT about this as an alternative API? Feels more intuitive to me

```suggestion
const results = proxy.useQueries(q => [ q.post.byId({ id: "1" }) ])
```

packages/tests/server/react/useQueries.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+    await ctx?.close?.();
+  })
+  .done();
+
+test('useQueries()', async () => {
+  const { proxy, App } = ctx;
+  function MyComponent() {
+    const utils = proxy.useContext();
+    const results = proxy.useQueries({
+      queries: [utils.post.byId.getQueryOptions({ id: '1' })],
```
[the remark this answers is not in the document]
sachinraja: I know it needs more tests, I mentioned it in the PR body

packages/tests/server/react/useQueries.test.tsx
sachinraja: on second look, maybe it's not useful anywhere other than `useQueries`, cc @TkDodo

TkDodo: if we expose the options, I think it's weird that we expose the queryKey separately as I've mentioned here:
- https://github.com/trpc/trpc/pull/3305#discussion_r1033190006

I also like the alternative api, and yes, I also wouldn't know what we would need the other options for _apart from_ the queryKey

sachinraja: Ok yeah I think I'll go for both here then. I'll keep `getQueryOptions` and we probably not do `getQueryKey` then? And I'll also do the special proxy for `useQueries`

JonParton: Just my 2 cents... I actually think not exposing a query options object on the `useContext()` object and letting Julius get to the bottom of how best to implement a getQueryKey function is a better idea; especially as he is handling router level as well as queries at a leaf node (ie `post` router as a whole not just `post.getById` )

Also I think the alternative api of a seperate proxy provided as an input to the use Queries function is better too (the afore called alternative api 🙂)

juliusmarminge: We could do like

- on trpc object, expose getQueryKey. This would be callable outside of the react context so I could use it with a raw query client in e.g. route loaders
- for the trpc.useQueries, go for the callback api?

packages/tests/server/react/useQueries.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+  })
+  .afterEach(async (ctx) => {
+    await ctx?.close?.();
+  })
+  .done();
+
+test('useQueries()', async () => {
+  const { proxy, App } = ctx;
+  function MyComponent() {
+    const results = proxy.useQueries((t) => [t.post.byId({ id: '1' })]);
```
sachinraja, line 45: This test is failing because of this error:
```ts
Type 'TrpcQueryOptionsForUseQueries<"post.byId", { id: string; }, "__result", unknown, TRPCClientError<CreateRouterInner<$Config, { byId: BuildProcedure<"query", { _config: $Config; _meta: {}; _ctx_out: {}; _input_in: { id: string; }; _input_out: { id: string; }; _output_in: unique symbol; _output_out: unique symbol; }, "...' is not assignable to type 'UseQueryOptionsForUseQueries<"__result", unknown, "__result", ["post.byId", { id: string; }]>'.
Types of property 'onError' are incompatible.
Type '((err: TRPCClientError<CreateRouterInner<$Config, { byId: BuildProcedure<"query", { _config: $Config; _meta: {}; _ctx_out: {}; _input_in: { id: string; }; _input_out: { id: string; }; _output_in: unique symbol; _output_out: unique symbol; }, "__result">; }>>) => void) | undefined' is not assignable to type '((err: unknown) => void) | undefined'.
Type '(err: TRPCClientError<CreateRouterInner<$Config, { byId: BuildProcedure<"query", { _config: $Config; _meta: {}; _ctx_out: {}; _input_in: { id: string; }; _input_out: { id: string; }; _output_in: unique symbol; _output_out: unique symbol; }, "__result">; }>>) => void' is not assignable to type '(err: unknown) => void'.
Types of parameters 'err' and 'err' are incompatible.
Type 'unknown' is not assignable to type 'TRPCClientError<CreateRouterInner<$Config, { byId: BuildProcedure<"query", { _config: $Config; _meta: {}; _ctx_out: {}; _input_in: { id: string; }; _input_out: { id: string; }; _output_in: unique symbol; _output_out: unique symbol; }, "__result">; }>>'.ts(2322)
```
cc @TkDodo this feels like an error in the `QueriesOptions` type for React Query. `t.post.byId({ id: '1' })` is (essentially) typed as `UseQueryOptions<inferProcedureOutput<TProcedure>, TRPCClientError<TRouter>, TData, [TPath, inferProcedureInput<TProcedure>]>` so I don't know why this is happening unless the underlying inference is incorrect. I don't see `QueriesOptions` checking for the `UseQueryOptions` type at all anywhere so that could be the issue too.

TkDodo: The typings of `useQueries` are a total mystery to me 😅

sachinraja: 😱

TkDodo: @artysidorenko maybe 😅 ?

artysidorenko: Hi folks,

Just looking at the type def for trpc `UseQueries`

```typescript
export type UseQueries<TRouter extends AnyRouter> = <T extends any[]>(
  queriesCallback: (
    t: UseQueriesProcedureRecord<TRouter>,
  ) => readonly [...QueriesOptions<T>],
  context?: UseQueryOptions['context'],
) => QueriesResults<T>;
```

Am I right in thinking `t.post.byId({ id: '1' })` should be the result of the queryFn? If so, that won't fit into `QueriesOptions<T>` which is expecting an options object. I think in your test example the type would need to look something like `(t) => [{queryFn: t.post.byId({ id: '1' })}]` to fit, if I understood correctly.

I think getting the order of options and results inferred correctly (the `[Promise1, Promise2]` vs `Promise[]` piece mentioned [here](https://github.com/trpc/trpc/issues/3188#issuecomment-1327416433)) is going to be a challenge - for context, the exported `QueriesOptions` helper you're using is so bulky because it's manually unwrapping the array of options object types and inferring what each individual element looks like, then `QueriesResults` is manually wrapping it back into a new array of result types (to preserve the strict order). If you're now trying to infer it from the result of the callback `queriesCallback`, it's going to get really messy :flushed:

One alternative that might be worth exploring for the types is to use an overloaded function definition instead. You'd need to write out lots of overloads to handle longer tuples, but it might be a bit more readable/maintainable?

<details>

<summary> something like this (example only goes up to 2) </summary>

```typescript
// handle 1 element tuple
function useQueries<TRouter extends AnyRouter, TQueryFnData1 = unknown>(
  queriesCallback:(
    t: UseQueriesProcedureRecord<TRouter>,
  ) => readonly [UseQueryOptions<TQueryFnData1>],
  context?: UseQueryOptions['context']
): [UseQueryResult<TQueryFnData1>]
// handle 2 element tuple - preserve variadic types
function useQueries<TRouter extends AnyRouter, TQueryFnData1 = unknown, TQueryFnData2 = unknown>(
  queriesCallback:(
    t: UseQueriesProcedureRecord<TRouter>,
  ) => readonly [UseQueryOptions<TQueryFnData1>,UseQueryOptions<TQueryFnData2>],
  context?: UseQueryOptions['context']
): [UseQueryResult<TQueryFnData1>, UseQueryResult<TQueryFnData2>]
// ^ add more overloads to handle bigger tuples

// default case - homogenous array of any length
function useQueries<TRouter extends AnyRouter, TQueryFnData = unknown>(
  queriesCallback:(
    t: UseQueriesProcedureRecord<TRouter>,
  ) => readonly UseQueryOptions<TQueryFnData>[],
  context?: UseQueryOptions['context']
): UseQueryResult<TQueryFnData>[]
{
  // actual function implementation
}
```

</details>

sachinraja: Thanks for the detailed response!

> Am I right in thinking t.post.byId({ id: '1' }) should be the result of the queryFn?

No, that function returns a type matching `UseQueryOptions`:
https://github.com/trpc/trpc/blob/d5d5e5adf59aa7933bb12e6a37eab0d83e25bf64/packages/react-query/src/shared/proxy/useQueriesProxy.ts#L33-L39.

> I think getting the order of options and results inferred correctly (the [Promise1, Promise2] vs Promise[] piece mentioned here) is going to be a challenge - for context, the exported QueriesOptions helper you're using is so bulky because it's manually unwrapping the array of options object types and inferring what each individual element looks like, then QueriesResults is manually wrapping it back into a new array of result types (to preserve the strict order). If you're now trying to infer it from the result of the callback queriesCallback, it's going to get really messy 😳

That's true but I think that should be handled by React Query instead. Inferring what the individual elements look like is also what's not working here. The type returned from t.post.byId() is UseQueryOptions but it fails to infer it and expects `err` to be `unknown` instead of what I passed. And looking into the `GetOptions` type, I don't see it handling the inference for certain things properly either. It doesn't capture errors at all unless I pass an `error` property (which is not a valid option?) and even if I do have an error property, the TQueryKey and possibly TData inference doesn't work.

Ideally it could just do `UseQueryOptions<infer TQueryFnData, infer Error, infer TData, infer QueryKey> but it doesn't and I'm not sure why exactly.

artysidorenko: > No, that function returns a type matching UseQueryOptions

Ah ok, gotcha, I misread the API!

In that case yeah, your queryCallback has the explicit `TError = TRPCClientError<TRouter>`, but on the `UseQueries` end we struggle to infer `TError` from anywhere in the options so it's stuck with `unknown` (a bit more context at the end of the message).

If you need an mvp solution/to clean up the tests, for now - you can try the below diff screenshot which should at least maintain all the _results_ inference correctly, but it means your error type will be left as `unknown`.

I will have more of a think as well if there's a way to fix the error type when you _know_ that it's always gunna be `TRPCClientError<TRouter>`.

<details>

<summary> diff screenshot </summary>

Line 18 => leave TError as unknown
Line 24 (separate issue) => TData should default to TFnData when you don't have any selector in your query options

image

</details>


> Ideally it could just do `UseQueryOptions<infer TQueryFnData, infer Error, infer TData, infer QueryKey>` but it doesn't and I'm not sure why exactly

You mean, we unwrap the array and try to infer each element using `UseQueryOptions` directly? This was my first port of call back in the day, but because `UseQueryOptions` has these type params appearing multiple times as context-sensitive types - e.g. TData both a callback param and a func return type - TS runs into problems handling both the array-logic and the contextual inference.. why you might see it's inferring from object literal signatures instead (going slightly off-topic 😅 ... let me know if you find something different/better though, been looking for ways to improve it 🙏 )...

sachinraja:
@KATT types work for the majority of cases but I can't get SSR working for some reason

See this failing test:
https://github.com/trpc/trpc/pull/3305/files#diff-ab06003c3085762c35329ba486f6bf0b9a2d4e0cecd9e7ec010b802feebc382cR46

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -82,6 +86,13 @@ export interface UseTRPCQueryOptions<TPath, TInput, TOutput, TData, TError>
   extends UseQueryOptions<TOutput, TError, TData, [TPath, TInput]>,
     TRPCUseQueryBaseOptions {}
 
+export interface TRPCQueryOptions<TPath, TInput, TData, TError>
+  extends QueryOptions<TData, TError, TData, [TPath, TInput]> {
+  trpc?: {
+    ssr?: boolean;
+  };
+}
```
KATT, line 94: Can we use `TRPCUseQueryBaseOptions`?

KATT, line 94:
```suggestion
export interface TRPCQueryOptions<TPath, TInput, TData, TError>
  extends QueryOptions<TData, TError, TData, [TPath, TInput]>, TRPCUseQueryBaseOptions {
}
```

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -654,6 +665,37 @@ export function createHooksInternal<
...
+          void prefetchQuery(queryOption.queryKey as any, queryOption as any);
+        }
+      }
+    }
+
+    return __useQueries({
+      queries: queries.map((query) => ({
+        ...query,
+        queryKey: getArrayQueryKey(query.queryKey, 'query'),
+      })) as readonly UseQueryOptions[],
```
KATT, line 694: I hate this....

Maybe we can make sure that `getArrayQueryKey()` of a result of the same fn is stable, then we can do this further "up" and only once?

packages/react-query/src/shared/proxy/useQueriesProxy.ts
```diff
@@ -0,0 +1,83 @@
...
+ * @internal
+ */
+export function createUseQueriesProxy<TRouter extends AnyRouter>(
+  client: TRPCClient<TRouter>,
+) {
+  return createRecursiveProxy((opts) => {
+    const path = opts.path.join('.');
+    const [input, ...rest] = opts.args;
+
+    const queryKey = getQueryKey(path, input);
```
KATT, line 71: On further comment, that could enable using `getArrayQueryKey` here instead?

packages/tests/server/react/useQueries.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const { proxy, App } = ctx;
+  function MyComponent() {
+    const results = proxy.useQueries((t) => [t.foo(), t.bar()]);
+
+    const foo = results[0].data;
+    const bar = results[1].data;
+    if (foo && bar) {
+      expectTypeOf(results[0].data).toEqualTypeOf<'foo'>();
+      expectTypeOf(results[1].data).toEqualTypeOf<'bar'>();
+    }
```
KATT, line 81: Added this test to test the types :)

KATT (commented):
Amazingggggggg, that's some mad types

- [ ] Make sure the types work in `/examples`, maybe we should add it to one of the `/examples/.test`? 🤔 
- [ ] Maybe - see if we can simplify the query key stuff a bit, it's a bit 🤯

packages/react-query/src/shared/hooks/createHooksInternal.tsx
sachinraja: might be awkward because `prefetchQuery` expects the old style query key

packages/react-query/src/shared/proxy/useQueriesProxy.ts
sachinraja: yeahh we should do this instead

sachinraja: not possible since `prefetchQuery` converts to array query key internally 😅

packages/react-query/src/internals/useQueries.ts
```diff
@@ -0,0 +1,71 @@
...
+  UseTRPCQueryResult,
+} from '../shared';
+
+export type UseQueryOptionsForUseQueries<
+  TQueryFnData = unknown,
+  TError = unknown,
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+> = Omit<UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>, 'context'>;
+export type TrpcQueryOptionsForUseQueries<
```
KATT, line 15:
```suggestion
export type TrpcQueryOptionsForUseQueries<
```

packages/react-query/src/shared/index.ts
```diff
@@ -1,9 +1,11 @@
 export * from './proxy/decorationProxy';
 export * from './proxy/utilsProxy';
+export * from './proxy/useQueriesProxy';
 export type {
   DecoratedProcedureRecord,
   DecorateProcedure,
 } from '../createTRPCReact';
+export type { UseQueries } from '../internals/useQueries';
```
KATT, line 8: Smurf it

```suggestion
export type { TRPCUseQueries } from '../internals/useQueries';
```

packages/react-query/src/internals/useQueries.ts
```diff
@@ -0,0 +1,71 @@
...
+    ? UseTRPCQueryResult<unknown extends TData ? TQueryFnData : TData, TError>
+    : never;
+};
+
+type GetOptions<TQueryOptions> =
+  TQueryOptions extends UseQueryOptionsForUseQueries<any, any, any, any>
+    ? TQueryOptions
+    : never;
+
+export type QueriesOptions<
```
KATT, line 41: If they're exported, use `/** internal **/`

KATT (approved):
- [ ] Smurf
- [ ] `/** @internal **/`
- [ ] Please do a `p publish-canary` and test it on an external project before merging. I got some type of error stuff when trying to use it in `/examples` the other day

KATT (approved):

KATT (approved):
What about docs? 👀

PR #3323, opened 2022-12-01
fix: type inference fails when not using a transformer

juliusmarminge:
Regression from #3261 

Inferred types were serialized even when there wasn't a transformer defined

Same bug as found here: https://github.com/trpc/trpc/pull/3298#pullrequestreview-1200729879

### DIFF
--- a/packages/server/src/core/types.ts
+++ b/packages/server/src/core/types.ts
@@ -1,4 +1,5 @@
 import { inferObservableValue } from '../observable';
+import { inferTransformedProcedureOutput } from '../shared';
 import { AnyProcedure, ProcedureArgs } from './procedure';
 import { AnyRouter, AnyRouterDef, Router } from './router';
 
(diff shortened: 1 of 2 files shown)


packages/server/src/core/types.ts
```diff
@@ -59,7 +60,7 @@ type GetInferenceHelpers<
       : TRouterOrProcedure extends AnyProcedure
       ? TType extends 'input'
         ? inferProcedureInput<TRouterOrProcedure>
-        : inferProcedureOutput<TRouterOrProcedure>
+        : inferTransformedProcedureOutput<TRouterOrProcedure>
```
KATT, line 63: This is a breaking change in another way than the other issue.

If you have an integration test or use the ssg helper, the data is unchanged.

I don't have a better idea right now but I am not confident this is the right solution

KATT: This was actually intentional b/c I didn't want a breaking change

sachinraja: I agree - we should not change this. If someone wants the transformed type they can use `inferTransformedProcedureOutput` directly (for now, at least).

KATT (approved):
YOLO

I think it's very very unlikely that anyone is using the output inference helper and get headaches here.

KATT:
tl;dr - this **is** a minor breaking change but I rather see it as a bug fix since we're < 2 weeks post-release and I doubt it will cause issues for anyone

Issue #3306, opened 2022-11-28
feat: first class support for io-ts validators


Nsttt:
### Describe the feature you'd like to request

[io-ts](https://github.com/gcanti/io-ts) is a decoding/encoding library based on [fp-ts](https://github.com/gcanti/fp-ts) functional programming patterns.

Currently, the behaviour when using it with tRPC is a bit weird, it just doesn't work as it should, I believe it can be improved since I've seen a few issues already about this topic.

Its syntax is very similar to `Zod` but validators returns an `Either<Error, Data>`.

### Describe the solution you'd like to see

Currently. When using io-ts as a validator, it works like this.

```ts
import * as t from "io-ts";
import * as E from "fp-ts/Either";
import { pipe } from "fp-ts/function";

const exampleCodec = t.type({
  text: t.union([t.string, t.null]),
});

export const exampleRouter = router({
  hello: publicProcedure.input(exampleCodec.decode).query(({ input }) =>
    pipe(
      input,
      E.fold(
        (error) => {
          console.log(error);
        },
        (response) => {
          return { greeting: `Hello ${response.text ?? "world"}` };
        }
      )
    )
  ),
});

```

Here, we can see we pass the `decode` function to the input of the procedure and makes the `input` type `E.Left<t.Errors> | E.Right<{text: string | null}>`.
No problems here. This is 100% expected. Some minor improvements could be that if its an `E.Left` case, it fallback to a TRPC error.

The big issue comes when I want to use `useQuery()` on this procedure. Because the input it asks for is an `Either`. I would like to be able to pass the structure of the `E.Right` case. I can access that structure dynamically using `t.TypeOf<typeof exampleCodec>`

### Describe alternate solutions

If it seems too much effort, not do it.

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

santosmarco-caribou:
Hey @Nsttt please take a look at this PR.

https://github.com/trpc/trpc/pull/3324

I got `io-ts` working for us. First-class type inference and parsing.

lloydrichards:
It looks like it's a little hard to implement this directly into tRPC without the library depending on io-ts to unwrap the Either.  Your workaround is good, but I tidied it up a bit in this [comment](https://github.com/trpc/trpc/pull/3324#issuecomment-1422658244) and made it into a generic function that can take any io-ts Decoder and return the type for tRPC to infer:

```typescript
[...]

PR #3324, opened 2022-12-01
feat: Add io-ts

santosmarco-caribou:
Closes https://github.com/trpc/trpc/issues/3306

## 🎯 Changes

1. Added esques for io-ts for correct type inference
2. Implemented a custom parser that doesn't depend on any of their fp-related libraries.
3. Added a test.

## ✅ Checklist

## TODO

- [x] Make build pass


Nsttt:
Looking good! I'd add a few extra tests types just to be sure that this works as supposed to. Such as `DateFromISOString`.

Thanks a lot for the help :)

izakfilmalter:
Stoked for this!

KATT (changes requested):
Build is failing, on my phone so can't see in detail

santosmarco-caribou:
Can't see as well since I don't have creds.

Happy to solve the issue, though.

It might very well be because I'm exporting io-ts in the tests but I did not add it to the package.json.

I can fix this tomorrow.

KATT:
```

@trpc/tests:test-ci: Summary of all failing tests
@trpc/tests:test-ci: FAIL tests/server/outputParser.test.ts
@trpc/tests:test-ci:   ● Test suite failed to run
@trpc/tests:test-ci: 
@trpc/tests:test-ci:     server/outputParser.test.ts:5:23 - error TS2307: Cannot find module 'io-ts' or its corresponding type declarations.
@trpc/tests:test-ci: 
@trpc/tests:test-ci:     5 import * as iots from 'io-ts';
@trpc/tests:test-ci:                             ~~~~~~~
@trpc/tests:test-ci:     server/outputParser.test.ts:299:30 - error TS2554: Expected 1 arguments, but got 0.
@trpc/tests:test-ci: 
@trpc/tests:test-ci:     299   expectTypeOf(output.input).toBeString();
@trpc/tests:test-ci:                                      ~~~~~~~~~~~~
@trpc/tests:test-ci: 
@trpc/tests:test-ci:       ../../node_modules/.pnpm/expect-type@0.15.0/node_modules/expect-type/dist/index.d.ts:78:18
@trpc/tests:test-ci:         78     toBeString: (...MISMATCH: MismatchArgs<Extends<Actual, string>, B>) => true;
@trpc/tests:test-ci:                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@trpc/tests:test-ci:         Arguments for the rest parameter 'MISMATCH' were not provided.
@trpc/tests:test-ci: 

```

Nsttt:
Also, something else I wonder, if I recall correctly in that example since we set the output function we get a validated input in our query. It might be helpful getting an `Either` there to manage a usual workflow using a pipe operator and everything else. What do you think ?

santosmarco-caribou:
> ```
> 
> 
> 
> @trpc/tests:test-ci: Summary of all failing tests
> 
> @trpc/tests:test-ci: FAIL tests/server/outputParser.test.ts
> 
> @trpc/tests:test-ci:   ● Test suite failed to run
> 
> @trpc/tests:test-ci: 
> 
> @trpc/tests:test-ci:     server/outputParser.test.ts:5:23 - error TS2307: Cannot find module 'io-ts' or its corresponding type declarations.
> 
> @trpc/tests:test-ci: 
> 
> @trpc/tests:test-ci:     5 import * as iots from 'io-ts';
> 
> @trpc/tests:test-ci:                             ~~~~~~~
> 
> @trpc/tests:test-ci:     server/outputParser.test.ts:299:30 - error TS2554: Expected 1 arguments, but got 0.
> 
> @trpc/tests:test-ci: 
> 
> @trpc/tests:test-ci:     299   expectTypeOf(output.input).toBeString();
> 
> @trpc/tests:test-ci:                                      ~~~~~~~~~~~~
> 
> @trpc/tests:test-ci: 
> 
> @trpc/tests:test-ci:       ../../node_modules/.pnpm/expect-type@0.15.0/node_modules/expect-type/dist/index.d.ts:78:18
> 
> @trpc/tests:test-ci:         78     toBeString: (...MISMATCH: MismatchArgs<Extends<Actual, string>, B>) => true;
> 
> @trpc/tests:test-ci:                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> 
> @trpc/tests:test-ci:         Arguments for the rest parameter 'MISMATCH' were not provided.
> 
> @trpc/tests:test-ci: 
> 
> 
> 
> ```

Thank you Katt, I will push a commit tomorrow adding io-ts to the dev deps

santosmarco-caribou:
> Also, something else I wonder, if I recall correctly in that example since we set the output function we get a validated input in our query. It might be helpful getting an `Either` instead to manage a usual workflow using a pipe operator and everything else. What do you think ?

I'm not an io-ts user nor a functional programming lover so my argument will be always to have the raw types available, if you want to then pipe the output you are free to do so. But I also get what you are saying–if I'm using oi-ts I expect to work with Eithers. 

I will let other people comment so we get more opinions.

KATT (commented):
- [ ] Make build pass
- [ ] Add docs https://trpc.io/docs/procedures#input-validation

Nsttt:
Hey sorry for the late answer, busy day.


Essentially yes, us who work with io-ts actually expect to also work with fp-ts.

Traditionally, we build our codecs as variables, and the we access the .decode() method with out input as a parameter.

That usually result with a an structure such as `Either<Error, Result>`. Then, we usually pipe that value with the `pipe()` function from fp-ts and there we discriminate what to do in case it is an error.

Again, traditionally a Promise.reject with a log of the array of errors.

Some implementation I'd like to take a look at is returning a new tRPC error in case it is an Error case, but at the same time, since we can infer the parameters thanks to Typescript I'm not sure if it is helpful in this case. What do you think?

This is an example of an API call using s traditional approach.

```ts
const BuildingsCodec = t.intersection([
  t.type({
    _id: t.string,
    name: t.string,
    coordinates: t.tuple([t.number, t.number]),
    locations: t.Array,
  }),
  t.partial({ address: t.string }),
]);

export async function get() {
  try {
    const token = Cookies.get("token");
    const response = await axios.get("api", {
      headers: {
        "Access-Token": token!,
        "Content-Type": "application/json",
      },
    });

    return pipe(
      response.data,
      BuildingsCodec.decode,
      E.fold(
        (error) => Promise.reject(error),
        (resource) => {
          return Promise.resolve(
            resource._items.map((item) => {
              return {
                name: item.name,
                location: item.address,
                latitude: item.coordinates[1],
                longitude: item.coordinates[0],
                locations: item.locations + item.name,
                id: item._id,
              };
            }),
          );
        },
      ),
    );
  } catch (error) {
    return Promise.reject(error);
  }
}

```

santosmarco-caribou:
> * [ ]  Make build pass
> * [X]  Add docs https://trpc.io/docs/procedures#input-validation

Today's update on this. Still working on making the build pass, though.

cc: @KATT

packages/server/src/core/internals/getParseFn.ts
```diff
@@ -30,6 +30,17 @@ export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
     return parser.create.bind(parser);
   }
 
+  if (typeof parser.decode === 'function') {
+    // ProcedureParserIoTsEsque
+    return (value) => {
+      const decoded = parser.decode.bind(parser)(value);
+      if (decoded.left) {
+        throw decoded.left;
```
KATT, line 38: Is the `.left` an `Error`? Make sure that we only `throw` Error objects. 

Ideally something that works together with error formatters! https://trpc.io/docs/error-formatting

KATT (commented):
Looks good! Added a comment

packages/server/src/core/internals/getParseFn.ts
Nsttt: Yeah, that's exactly what I was talking about in my above thread. We should throw an Error when the Either is left.

KATT: If there's a built-in error instance that can contain more metadata (similar to `ZodError`) that'd be perfect, otherwise we'd need to wrap it in our own custom wrapper

santosmarco-caribou: On it

Nsttt: There's a custom type yes.

There's also a library called io-ts-reporters that might come handy to get formated readable error messages for that matter. Not sure if it helps in the specific case.

juliusmarminge:
> May I get one more screenshot of the build failures whenever you can, please?

You need to add io-ts as a devDep in the www package so that shiki can get its types for the compilation

KATT:
@santosmarco-caribou / @Nsttt anything needed on my end to help out here?

Nsttt:
I'll make some suggestions the following days, currently on a trip ;).

santosmarco-caribou:
Let's wait for @Nsttt comments—I've found out that `left` is not an Error object, but `io-ts` also does not provide good error information on its own (like Zod does), not even default error messages.

@Nsttt, any idea how we can throw the `left` object on errors?

KATT:
Maybe do a PR to `io-ts` to provide such a thing for `Left`? Otherwise, we could potentially have an `Error`-subclass specifically for this in tRPC

Nsttt:
> Maybe do a PR to `io-ts` to provide such a thing for `Left`? Otherwise, we could potentially have an `Error`-subclass specifically for this in tRPC

If I understand the code correctly. The `decode` should be an either. You should be able to access a Either like that but the proper way of doing it is with fp-ts. We could do something like.

```ts
if (typeof parser.decode === 'function') {
    // ProcedureParserIoTsEsque
    return (value) => {
      const decoded = parser.decode.bind(parser)(value);
      pipe(
        decoded,
        E.mapLeft((errors) => throw new Error(errors)),
        E.map((value) => value),
      )
    };
  }
```
Io ts does have a `Errors` type that's used as a way of defining validation errors. You can see it if you check for `ValidationError` type.

```ts
/**
 * @category Decode error
 * @since 1.0.0
 */
export interface ValidationError {
  /** the offending (sub)value */
  readonly value: unknown
  /** where the error originated */
  readonly context: Context
  /** optional custom error message */
  readonly message?: string
}
```

I'll probably arrive by tomorrow and will give this a further in-depth look, the easier way to take account of the structures is to just import both packages and manage them with fp-ts patterns I'd say.

santosmarco-caribou:
> 

Yeah, but `ValidationError` does not inherit from `Error`, so we can't throw it.

KATT:
hey @santosmarco, I fucked this up by mixing up which branch I was on:

Screenshot 2022-12-20 at 12 23 59


Can you please force-push your local one?

santosmarco-caribou:
Ok @KATT @Nsttt I've updated the branch to throw a `TRPCError` on `left`—do you guys think that the proposed solution is OK for now?

Nsttt:
Not really sure. As far as I know we should throw a regular Error, tRPC will take care to create a validation error with that error as the cause.

KATT:
Closing this as it has had no activity for a while, adding the icebox label so it's easy to find

lloydrichards:
i'm not completely sure why this fizzled out as it seems like this was quite close to being implemented and would be very helpful for my own projects that only want to rely on io-ts for validation.  If I could contribute something it would be that io-ts has a way of parsing the Left<string> into a useful message, using a [Reporter](https://gcanti.github.io/io-ts/#error-reporters).   I found an example [here](https://github.com/gcanti/io-ts/issues/194#issuecomment-408430537) that looks like this:
```typescript
import * as t from 'io-ts'
import { failure } from 'io-ts/lib/PathReporter'

const decode = <I, A>(input: I, type: t.Decoder<I, A>) => {
  return type.decode(input).mapLeft(
    errors => `Validation failed for input:

    ${JSON.stringify(input, null, 2)}

    Error details:

    ${failure(errors)
          .map(s => '- ' + s)
          .join('\n')}
    `
  )
}

const Person = t.interface({ name: t.string }, 'Person')
console.log(decode({ foo: 'bar', name: 1 }, Person).fold(s => s, () => 'no errors'))
```
**Output:**
```
Validation failed for input:

{
  "foo": "bar",
  "name": 1
}

Error details:

- Invalid value 1 supplied to : Person/name: string
```

What I would usually do in my code is have some kind of `prettierError()` function for the `.mapLeft()`  and then just use `.getOrElse((err)=>throw(err))` to toss it into my graphQL error boundary 🤷

Nsttt:
Yeah I just didn't take a huge look into it.

We were talking about how we needed to create a new Error class to be able to throw a TRPCError in case the validation fails.

To do that, we would need to create a custom Error class for io-ts errors.

lloydrichards:
okay, so the left side needs to throw a `new Error()` with a readable message and the TRPCError will wrap that later on?  Looking at @santosmarco branch I think all that needs to change is:

packages/server/src/core/internals/getParseFn.ts
```typescript
import { failure } from 'io-ts/lib/PathReporter'
...
  if (typeof parser.decode === 'function') {
    // ProcedureParserIoTsEsque
    return (value) => {
      const decoded = parser.decode.bind(parser)(value);
      if (decoded.left) {
        throw new Error(failure(decoded.left)
          .map(s => '- ' + s)
          .join('\n')
         );
      }
      return decoded.right;
    };
  }
```

lloydrichards:
I figured out a way to build a generic io-ts parser function that can be used in the current tRPC and make everything play nice. I don't want to go down the rabbit hole of trying to add anything to tRPC so if other stumble on the same issue, they can give this a try:

**helper.ts**
```typescript
import * as t from "io-ts";
import * as E from "fp-ts/Either";
import * as A from "fp-ts/Array";
import { identity, pipe } from "fp-ts/lib/function";
import { failure } from "io-ts/lib/PathReporter";

/**
 * This is a helper function to make the error messages from io-ts
 * more readable.
 *
 * It takes the error object from io-ts and returns a string with
 * the first 3 errors.
 */
export const prettyError = (err: t.Errors) =>
  pipe(
    err,
    failure,
    A.takeLeft(3),
    (a) => a.map((s) => "- " + s),
    (a) => a.join("\n")
  );

/**
 * This is a helper function to make it easier to use io-ts
 * with trpc.
 *
 * It takes an io-ts decoder and returns a function that
 * takes a value and returns the decoded value or throws
 * an error if the value is invalid.
 */
export const ioParser =
  <T>(schema: t.Decoder<unknown, T>) =>
  (value: unknown) => {
    return pipe(
      schema.decode(value),
      E.fold((err) => {
        throw new Error(prettyError(err));
      }, identity)
    );
  };
```

**example.ts**
```typescript
...
export const billableRouter = createTRPCRouter({
hello: trpc.procedure
    .input(
      ioParser(
        t.type({
          name: t.string,
        })
      )
    )
    .query(({ input: { name } }) => {
      return {
        greeting: `Hello ${name}`,
      };
    }),
});
```

PR #3325, opened 2022-12-01
feat: Allow subset of router's context when creating SSG helpers

bennettdams:
## 🎯 Changes

This change infers the given context type when creating SSG helpers. The inferred type must be a subset of the router's context, otherwise the user will see the following error message via the type error:

"_The context you have given is either not a direct match or not a subset of the router's context._"

### Why I want this

- In my router context, I want to set the Next.js `req` and `res` , so I can easily use it in my procedures (e.g. for auth).
- I want to use SSG helpers for both `getServerSideProps` and also `getStaticProps`.

This setup means that some context usages have `req` & `res` and some have not.
I created an inner context creator to mitigate that:

```ts
export async function createContextTRPCInner() {
  return {
    prisma,
  }
}

export async function createContextTRPC(opts: CreateNextContextOptions) {
  const contextInner = await createContextTRPCInner()

  return {
    ...contextInner,
    req: opts.req,
    res: opts.res,
  }
}

export type ContextTRPC = inferAsyncReturnType<typeof createContextTRPC>
```

I use `createContextTRPC` for my routers, as they all have `req` & `res`.

I **wanted** to use `createContextTRPCInner` for the cases where I don't have `req` & `res`, but it won't work: When using `createProxySSGHelpers`, it expects to be passed the context.
But right now, it infers the context type directly from the router, see https://github.com/trpc/trpc/blob/main/packages/react-query/src/ssg/ssg.ts#L20.

### Result of this change

This change will instead infer the context from the giving parameters of `createProxySSGHelpers`.

Here's the error message shown to the user when he tries to give a context that is not a subset of the router's context:

image

image

### Open topics

- [ ] There's an error shown [in the test I've added](https://github.com/trpc/trpc/blob/3c69cc0643a6e131d87a5d277b3d0165828a6f04/packages/tests/server/react/ssg.test.ts#L14) in my editor, but I don't know what it means:
  
  image

## ✅ Checklist

### DIFF
--- a/packages/react-query/src/ssg/ssg.ts
+++ b/packages/react-query/src/ssg/ssg.ts
@@ -15,20 +15,37 @@ import {
 import { getArrayQueryKey } from '../internals/getArrayQueryKey';
 import { CreateTRPCReactQueryClientConfig, getQueryClient } from '../shared';
 
-interface CreateSSGHelpersOptionsBase<TRouter extends AnyRouter> {
+/**
+ * Works like TypeScript's `Omit`, but allows safe usage of the omitted keys.
+ * If you use a key that does not exist on the type, there will be a type error.
+ */
+type OmitStrict<TType, TKey extends keyof TType> = TType extends unknown
+  ? Pick<TType, Exclude<keyof TType, TKey>>
+  : never;
+
+interface CreateSSGHelpersOptionsBase<TRouter extends AnyRouter, TContext> {
   router: TRouter;
-  ctx: inferRouterContext<TRouter>;
+  ctx: inferRouterContext<TRouter> extends TContext // Checking the `extend` this way makes sure that the given context is either a direct match or a subset of the router's context.
+    ? TContext
+    : // Helpful message shown to the user if his given context is not matching/a subset.
+      "The context you have given is either not a direct match or not a subset of the router's context.";
   transformer?: ClientDataTransformerOptions;
 }
-export type CreateSSGHelpersOptions<TRouter extends AnyRouter> =
-  CreateSSGHelpersOptionsBase<TRouter> & CreateTRPCReactQueryClientConfig;
+export type CreateSSGHelpersOptions<
+  TRouter extends AnyRouter,
+  TContext,
+> = CreateSSGHelpersOptionsBase<TRouter, TContext> &
+  CreateTRPCReactQueryClientConfig;
 
 /**
  * Create functions you can use for server-side rendering / static generation
  * @deprecated use `createProxySSGHelpers` instead
  */
-export function createSSGHelpers<TRouter extends AnyRouter>(
-  opts: CreateSSGHelpersOptions<TRouter>,
+export function createSSGHelpers<TRouter extends AnyRouter, TContext>(
+  // Omit the context so we can infer its given type.
+  opts: OmitStrict<CreateSSGHelpersOptions<TRouter, TContext>, 'ctx'> & {
+    ctx: TContext;
+  },
 ) {
   const { router, transformer, ctx } = opts;
   type TQueries = TRouter['_def']['queries'];
--- a/packages/react-query/src/ssg/ssgProxy.ts
+++ b/packages/react-query/src/ssg/ssgProxy.ts
@@ -62,8 +62,8 @@ type AnyDecoratedProcedure = DecorateProcedure<any>;
 /**
  * Create functions you can use for server-side rendering / static generation
  */
-export function createProxySSGHelpers<TRouter extends AnyRouter>(
-  opts: CreateSSGHelpersOptions<TRouter>,
+export function createProxySSGHelpers<TRouter extends AnyRouter, TContext>(
+  opts: CreateSSGHelpersOptions<TRouter, TContext>,
 ) {
   const helpers = createSSGHelpers(opts);
 
--- a/packages/tests/server/react/ssg.test.ts
+++ b/packages/tests/server/react/ssg.test.ts
@@ -1,10 +1,30 @@
 import { InfiniteData } from '@tanstack/react-query';
 import { createProxySSGHelpers } from '@trpc/react-query/src/ssg/ssgProxy';
-import { initTRPC } from '@trpc/server/src';
+import { inferAsyncReturnType, initTRPC } from '@trpc/server/src';
 import { expectTypeOf } from 'expect-type';
+import { CreateNextContextOptions } from 'packages/server/src/adapters/next';
 import { z } from 'zod';
 
-const t = initTRPC.create();
+async function createContextInner() {
+  return {
+    inner: 'inner value',
+  };
+}
+
+async function createContext(opts: CreateNextContextOptions) {
+  const contextInner = await createContextInner();
+
+  return {
+    ...contextInner,
+    outer: 'outer value',
+    req: opts.req,
+    res: opts.res,
+  };
+}
+
+export type Context = inferAsyncReturnType<typeof createContext>;
+
+const t = initTRPC.context<Context>().create();
 
 const appRouter = t.router({
   post: t.router({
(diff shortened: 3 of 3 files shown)


packages/react-query/src/ssg/ssg.ts
```diff
@@ -15,20 +15,37 @@ import {
...
 
 /**
  * Create functions you can use for server-side rendering / static generation
  * @deprecated use `createProxySSGHelpers` instead
  */
-export function createSSGHelpers<TRouter extends AnyRouter>(
-  opts: CreateSSGHelpersOptions<TRouter>,
+export function createSSGHelpers<TRouter extends AnyRouter, TContext>(
+  // Omit the context so we can infer its given type.
+  opts: OmitStrict<CreateSSGHelpersOptions<TRouter, TContext>, 'ctx'> & {
```
bennettdams, line 46: Omitting is done to infer to type of `ctx`, as we need it directly in `CreateSSGHelpersOptionsBase`.
Not sure if there's a better way to do this?

packages/react-query/src/ssg/ssg.ts
```diff
@@ -15,20 +15,37 @@ import {
...
+    ctx: TContext;
+  },
```
bennettdams, line 48: Omitting is done to infer to type of ctx, as we need it directly in CreateSSGHelpersOptionsBase.
Not sure if there's a better way to do this?

bennettdams, line 48: Omitting is done to infer to type of ctx, as we need it directly in CreateSSGHelpersOptionsBase.
Not sure if there's a better way to do this?

KATT (commented):
On my phone so can't review well, but this is just intended to improve the error message?

Make sure that you can't add a wider type than the context as well, didn't see such a type in the text.

Furthermore, if we are going down this path we need it to be the same DX consistently, there's a bunch of other methods that accept context as well

bennettdams:
> but this is just intended to improve the error message?

Having this error message is just a nice side effect. Why I proposed this change is described in "Why I want this" above, it solves a problem with the current implementation (before this change).

> Make sure that you can't add a wider type than the context as well

Why so? Isn't `createProxySSGHelpers` supposed to only receive the router context? This change makes it so a subset of that context is allowed as well, but not a widened type.

> a bunch of other methods that accept context as well

This change is there for the SSG helpers where it is common to have a subset of the context. Does this also apply for other methods? If so, which ones?

KATT:
I don't want to allow to call tRPC with a partial context unless it is actually a partial

KATT (changes requested):
See my comment or convince me I'm wrong.

Would **love** some docs improvements on advanced topics like this.

Again, thanks for your work

bennettdams:
Maybe I'm totally missing the point, but the whole reason for this change was to **not** have optional fields. I know what is in my context at any time, so I don't want to check them for `!== undefined` all the time.

Take this example:

- outer context creator, has `req` & `res`, because they're given in the Next.js API. `req` & `res` should not be optional.
- inner context creator, brings `prisma` or whatever. The fields that the inner context returns should also not be optional.


```ts
export async function createContextInner() {
  return {
    prisma,
  }
}

export async function createContext(opts: CreateNextContextOptions) {
  const contextInner = await createContextInner()

  return {
    ...contextInner,
    req: opts.req,
    res: opts.res,
  }
}

```
As you can see, I want both the inner and the outer context to not return any optional fields.

Now comes the tricky part: `createProxySSGHelpers` infers the router's (outer) context for the `ctx` field of its `opts` arg.

In `getServerSideProps`, that's totally fine, I will use my `createContext` (outer) to fill the `ctx` field.
In `getStaticProps`, I can't give `req` & `res`, so I want to use my inner context. But I can't with the current implementation, because `ctx` is inferred from the router's context automatically, so when using the "inner" context, it is complaining about missing fields. In this example `req` & `res` are obviously missing from context, as the "inner" doesn't have them.


Here is a full example that is hopefully copy-pastable:

```ts
import { createProxySSGHelpers } from '@trpc/react-query/ssg'
import { inferAsyncReturnType, initTRPC } from '@trpc/server'
import { CreateNextContextOptions } from '@trpc/server/adapters/next'
import { prisma } from './db/client'

export async function createContextTRPCInner() {
  return {
    prisma,
  }
}

export async function createContextTRPC(opts: CreateNextContextOptions) {
  const contextInner = await createContextTRPCInner()

  return {
    ...contextInner,
    req: opts.req,
    res: opts.res,
  }
}

export type Context = inferAsyncReturnType<typeof createContextTRPC>

const t = initTRPC.context<Context>().create()
const router = t.router

export const fooRouter = router({
  foo: t.procedure.query(async ({ ctx }) => {
    // req is not undefined :)
    ctx.req

    return 'bar'
  }),
})

export const appRouter = router({ fooRouter })

async function _getStaticProps() {
  const ssg = createProxySSGHelpers({
    router: appRouter,
    // 🚨🚨🚨 error, `req` & `res` are missing     
    ctx: await createContextTRPCInner(),
  })
}


```

The change in this PR makes it so `ctx` is inferred from what is set when calling `createProxySSGHelpers`. 

> I don't want to allow to call tRPC with a partial context unless it is actually a partial

That's what the new error message is for. It will check whether the given type for `ctx` is actually a subset of the router's context, so you can't call it with different fields etc.

KATT:
We use the same principle here, the only diff is that we use the "inner" context to create the `Context`


- 
https://github.com/trpc/trpc/blob/6d483b5065a77216f394f60fea3d15e1cb1320f8/examples/next-prisma-starter/src/server/context.ts#L18
- https://github.com/trpc/trpc/blob/c5b49adb1d4e875633bb88ce282a3573fa00b404/examples/next-prisma-starter/src/server/routers/post.test.ts#L9-L10

bennettdams:
> How come you are not using the inner context for the context creation to tRPC?
> 
> https://github.com/trpc/trpc/blob/6d483b5065a77216f394f60fea3d15e1cb1320f8/examples/next-prisma-starter/src/server/context.ts#L18

How would you access `req` and `res` if you only use the inner?

```ts
export const fooRouter = router({
  foo: t.procedure.query(async ({ ctx }) => {
    // ✅ `req` (from outer context) is not undefined :)
    ctx.req
    
    // ✅ `prisma` (from inner context) is not undefined :)
    ctx.prisma

    return 'bar'
  }),
})

```

KATT:
> How would you access `req` and `res` if you only use the inner?

See https://github.com/trpc/trpc/compare/3325?expand=1

bennettdams:
Sorry, I should've been more precise: 

How would you access `req` and `res` **without checking for `undefined`** if you only use the inner?

They're always there (in the outer context), no for a falsy check/optional chaining.


In my case, I use `req` a lot for helper functions I call in the procedures, e.g. for checking auth. It feels wrong to me that you always have to pull-through a falsy value, even when it can never be `undefined`.

KATT:
> How would you access `req` and `res` **without checking for `undefined`** if you only use the inner?
> 
> They're always there (in the outer context), no need for a false check/optional chaining.

Well, it's "always" there, but not when you actually call the procedure through SSG.

You can do this to do a procedure that can only be called through API-calls: 5c999058cc86be4fd3f73785617084c1ff45cb1f

bennettdams:
> not when you actually call the procedure through SSG

Wait, now I'm confused. 
I know that you don't have the `req` in your hands in `getStaticProps` to provide manually to the context, but I thought the `req` and `res` are filled automatically into context anyways? So, each procedure gets both of them "injected", because they're just a Next.js API endpoint?
I thought you set/overwrite them automatically in the context only when you really need the `req` (via `createProxySSGHelpers`). 

Something like:

image




Btw, this `apiProcedure` makes total sense, thanks for that.

bennettdams:
> When you do a createProxySSGHelpers (or a .createCaller) you won't do API requests, you'll just call functions within tRPC! :)

Oh wow, didn't realize that.
Now it makes sense why `req` & `res` **should** be `undefined` in the context.


Thanks a lot for taking the time.

KATT:
No problem, would love some help with #3329 if you want :)

PR #3331, opened 2022-12-02
docs: Clarify context usage and mention inner/outer context concept

### LINKED ISSUE #3329: feat: Improve context docs

Improve https://trpc.io/docs/context

- Describe how to use context, including how to write it up with the API handler which is missing right now
- Describe why you might need an "inner" and an "outer" context and infer the `Context`-type from the _inner_ one (#3325)

n/a

### Additional information

> this question made a good point that the docs currently don't seem to have a section on how to set up routing and context in /pages/api in next.js. seems like it would be difficult to figure the setup out without ct3a or some other template. i could try to write something on the weekend.
https://discord.com/channels/867764511159091230/1047290845700821083https://discord.com/channels/867764511159091230/1047290845700821083

by @c-ehrlich 

https://github.com/trpc/trpc/pull/3325

### 👨‍👧‍👦 Contributing

bennettdams:
Closes #3329

## 🎯 Changes

- SSG helpers
  - Mention that SSG helpers do not do an HTTP call, so the context can never have `req` and `res`.
  - Mention the concept of inner/outer context
- Context
  - inspired by [CT3A](https://create.t3.gg/en/usage/trpc#-servertrpccontextts)
  - Rename "Request context" to "Context", as "request" can easily be confused with HTTP calls
  - Add the concept of inner and outer context

## ✅ Checklist


bennettdams:
The current example of the inner and outer context has one flaw: `req` & `res` are never part of the context, so they also don't exist at all in the procedures.

@KATT How do you expect the [`apiProcedure`](https://github.com/trpc/trpc/commit/5c999058cc86be4fd3f73785617084c1ff45cb1f) to work when we never bring `req` & `res` to the context type?

```ts
export const apiProcedure = procedure.use((opts) => {
  // 🚨 here, `req` & `res` do not exist, as the context is inferred from the inner context
  if (!opts.ctx.req || !opts.ctx.res) {
    throw new Error("You are missing `req` or `res` in your call.");
  }
  return opts.next({
    ctx: {
      // We overwrite the context with the truthy `req` & `res`, which will also overwrite the types used in your procedure.
      req: opts.ctx.req,
      res: opts.ctx.res,
    },
  });
});
```

www/docs/server/context.md
```diff
@@ -60,11 +62,88 @@ export const router = t.router;
...
+    prisma,
+  }
+};
+
+/** 
+ * Outer context. Will e.g. bring `req` & `res` to the context as "not `undefined`".
+ * 
+ * @see https://trpc.io/docs/context#inner-and-outer-context
+ */
+export async function createContext(opts: CreateNextContextOptions) {
```
c-ehrlich, line 117: Does this need to be async? I just realized we have this as async in create-t3-app too and I'm not sure why. Will try to figure this out.

www/docs/server/context.md
```diff
@@ -60,11 +62,88 @@ export const router = t.router;
...
\ No newline at end of file
+```
+
+## Inner and outer context
+
+In some scenarios it could make sense to split up your context into "inner" and "outer" functions.
+
+**Inner context** is where you define context which doesn’t depend on the request, e.g. your database connection. You can use this function for integration testing or [SSG helpers](ssg-helpers), where you don’t have a request object. Whatever is defined here will **always** be available in your procedures.
+
+**Outer context** is where you define context which depends on the request, e.g. for the user's session. Whatever is defined here is not available in all procedures, but only for procedures that are called via HTTP.
```
c-ehrlich, line 80: Instead of "not available in all procedures", which makes it sound like it somehow depends on the type of procedure, I'd suggest to continue the explanation from the previous paragraph and say something like "only available when a procedure is called via http".

www/docs/server/context.md
```diff
@@ -60,11 +62,88 @@ export const router = t.router;
...
+
+```ts
+import type { inferAsyncReturnType } from '@trpc/server';
+import type { CreateNextContextOptions } from '@trpc/server/adapters/next';
+
+/** 
+ * Defines your inner context shape.
+ * Add fields here that the inner context brings.
+ */
+interface CreateInnerContextOptions extends Partial<CreateNextContextOptions> {
```
c-ehrlich, line 92: I don't think this is correct. Prisma definitely doesn't belong in here - I guess you _could_ pass it from outer context, but this example isn't, and I don't see a reason to. It's usually just imported at the top of the file and then used inside `createContextInner`.

Also not sure about opts (ie req/res), they type is technically not incorrect as it's a partial, but why type it like this when req and res aren't actually being passed to `createContextInner` from `createContext`?

www/docs/server/context.md
```diff
@@ -60,11 +62,88 @@ export const router = t.router;
...
+};
+
+export type Context = inferAsyncReturnType<typeof createContextInner>;
+
+// The usage in your router is the same as the example above.
+```
+
+It is important to infer your `Context` from the **inner** context, as only what is defined there is really always available in your procedures.
+
+If you don't want to check `req` or `res` for `undefined` in your procedures all the time, you could build a small reusable procedure for that:
```
c-ehrlich, line 134: Is this important enough to be in the docs? Req and res being used inside of a procedure is a pattern that I've not seen much in tRPC codebases. Might be wrong about this though.

www/docs/server/context.md
bennettdams: Makes sense, used a "session" instead.

www/docs/server/context.md
bennettdams: I use `req` & `res` a lot when I don't want to evaluate the session in the context on every request - maybe this takes some time or needs a roundtrip to DB and this adds up to the general request time. Having such procedure now eliminates the need of an `undefined` check for `req` or `res` in your procedures all the time. The idea comes from KATT here: https://github.com/trpc/trpc/commit/5c999058cc86be4fd3f73785617084c1ff45cb1f

..but I'm not sure how common this is as well. Let me know if I should throw it out.

www/docs/server/context.md
```diff
@@ -60,11 +62,91 @@ export const router = t.router;
...
+
+```ts
+export const apiProcedure = publicProcedure.use((opts) => {
+  if (!opts.ctx.req || !opts.ctx.res) {
+    throw new Error("You are missing `req` or `res` in your call.");
```
bennettdams, line 142: Maybe a `TRPCError` here instead? Something like:
```ts
throw new TRPCError({
  message: 'You are missing `req` or `res` in your call.',
})
```

www/docs/server/context.md
```diff
@@ -1,25 +1,28 @@
...
 // -------------------------------------------------
-import { inferAsyncReturnType } from '@trpc/server';
-import * as trpcNext from '@trpc/server/adapters/next';
+import type { inferAsyncReturnType } from '@trpc/server';
+import type { CreateNextContextOptions } from '@trpc/server/adapters/next';
 import { getSession } from 'next-auth/react';
 
 /**
- * Creates context for an incoming request
+ * Creates context for an incoming call to the procedure.
```
KATT, line 25: It can be batched so it can be several procedure calls

```suggestion
 * Creates context for an incoming request
```

www/docs/server/context.md
KATT: I think it's a good addition. Our docs don't _guide_ people enough. Feel free to tweak as follow-ups if you want @c-ehrlich  :)

KATT (dismissed):

KATT (changes requested):
Oops, seems like something is failing in the www build. See by doing `cd www && p build`

bennettdams:
> Oops, seems like something is failing in the www build. See by doing `cd www && p build`

Not sure what's going on, all good locally:

image

Maybe something about deployment in my forked repo?

KATT:
CleanShot 2022-12-12 at 01 15 12@2x

bennettdams:
I had to remove an old import via the browser, because the change was applied locally already, but not picked up by the pushed diff? No idea how this happens.

KATT (approved):
Let's get this across the finish line, please re-request me as a reviewer once the lint issues are fixed :)

Just got a Q in Discord about this exact thing about inner contexts

bennettdams:
@KATT Should've made that more clear, but I fixed this two weeks ago already: https://github.com/trpc/trpc/pull/3331/commits/a497913f92a9513588060c29f33a2a552769ca63

When I look at the build of that commit, it is not ESLint that is failing, but the pipeline itself (ESLint didn't even run):

image


On the other hand, the [merge commit on tRPC's main](https://github.com/trpc/trpc/commit/065bd6be69c9457f9c55e7fdd48d8dd931245755) has no problem anymore, so all good I think!

image

PR #3343, opened 2022-12-05
fix: broken SSR on certain libraries

skyriverbend:
## Closes tamagui/tamagui#270

We've been having issues getting tRPC working when setting `ssr: true` in our apps. This fix solves the issue.

`react-ssr-prepass` hasn't been updated in nearly a year, and I believe the issue we were having stems from the fact that it uses `react-dom-17.0.2` instead of 18.

## 🎯 Changes

I've removed `react-ssr-prepass` and replaced it with `renderToPipeableStream`. 

This solution is mentioned in an open issue of `react-ssr-prepass`: FormidableLabs/react-ssr-prepass#80

Since `renderToPipeableStream` is now built into `react-dom`, it seems like the better option.

## ✅ Checklist


packages/next/src/withTRPC.tsx
```diff
@@ -32,7 +32,7 @@ import {
 } from 'next/dist/shared/lib/utils';
 import { NextRouter } from 'next/router';
 import React, { createElement, useState } from 'react';
-import ssrPrepass from 'react-ssr-prepass';
+import { renderToPipeableStream } from 'react-dom/server';
```
juliusmarminge, line 35: Not super familiar with what stuff got introduced in react 18 but isn't this one of them? in that case we'd now have peerDep of react^18 instead of react^16?

juliusmarminge:
Seems like all our E2E tests are failing due to this change

skyriverbend:
Ah interesting, I ran the test suite on my machine and they all passed.

And yes, fair enough about the requirement for React 18. Perhaps this
change can be saved until tRPC decides to upgrade to React >=18.

juliusmarminge:
> Ah interesting, I ran the test suite on my machine and they all passed.

Yea the tests pass but the E2E fails. Not quite sure about the error though 🤔

> And yes, fair enough about the requirement for React 18. Perhaps this change can be saved until tRPC decides to upgrade to React >=18.

Maybe this can be part of v11 then cause I don't think we can drop backwards compatibility support in a minor.

skyriverbend:
Makes sense, thanks for taking a look! I'll close this for now.

luixo:
Seems like there's a problem in `react-dom` version in CI runs.
Though all the packages have `react-dom` at least `18.2.0`.

Can someone more proficient in this repo help me out debugging that?

KATT:
Got CI passing by dynamically importing it, I'm not sure what is going on exactly but getting 

```
Warning: Detected multiple renderers concurrently rendering the same context provider. This is currently unsupported.
    at eA (/Users/alex/dev/trpc/node_modules/.pnpm/next@14.0.1_@babel+core@7.23.2_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/compiled/next-server/pages.runtime.dev.js:17:1760)
    at eq (/Users/alex/dev/trpc/node_modules/.pnpm/next@14.0.1_@babel+core@7.23.2_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/compiled/next-server/pages.runtime.dev.js:17:3083)
    at AppTree
Warning: Detected multiple renderers concurrently rendering the same context provider. This is currently unsupported.
    at eA (/Users/alex/dev/trpc/node_modules/.pnpm/next@14.0.1_@babel+core@7.23.2_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/compiled/next-server/pages.runtime.dev.js:17:1760)
    at eq (/Users/alex/dev/trpc/node_modules/.pnpm/next@14.0.1_@babel+core@7.23.2_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/compiled/next-server/pages.runtime.dev.js:17:3083)
    at AppTree
```

when using it, try:

```
# terminal 1
pnpm dev

# terminal 2
cd examples/next-minimal-starter
pnpm dev
open http://localhost:3000
```

reload the page and you get that error - doesn't seem to appear on `next`

packages/next/src/withTRPC.tsx
```diff
@@ -205,11 +204,13 @@ export function withTRPC<
           trpc: trpcProp,
         };
 
+        const reactDomServer = await import('react-dom/server');
+
         // Run the prepass step on AppTree. This will run all trpc queries on the server.
         // multiple prepass ensures that we can do batching on the server
         while (true) {
           // render full tree
-          await ssrPrepass(createElement(AppTree, prepassProps as any));
```
KATT, line 212: seems like this was never needed in the first place? 

we might just want to do this on `main` if this works well

packages/next/src/withTRPC.tsx
```diff
@@ -205,11 +204,13 @@ export function withTRPC<
...
+          reactDomServer.renderToString(createElement(AppTree, prepassProps));
```
KATT, line 213: @luixo can you have a try to patch your project with this here instead and see if everything works fine?

KATT: if it does, please open a new PR to `main` with this, it's a bit simpler

packages/next/src/withTRPC.tsx
luixo: Actually, it might be needed in some setups.
We need some render function here, and if we use `renderToString` all the tests pass, but if we add a test with a suspended component (which suspension is not based on `queryClient`) that has a `useQuery` inside - it might fail, we'll have no queries running on first render, but there will be after suspension resolution.
I'm not sure if that weird case should be supported or not, but the expected functions might break silently with that change.

Let me know if you're in to make this change on main branch.

packages/next/src/withTRPC.tsx
luixo: See [this comment](https://github.com/trpc/trpc/pull/3343#discussion_r1435306147) for explanation of a case it might fail.
Let me know if you still think we should migrate to `renderToString` on main.

luixo:
> I'm not sure what is going on exactly but getting

We're running a render inside a render, this is not expected in React unfortunately as they keep contexts in a stack.
I'm not sure if this going to break anything, but if it's fixed by running concurrent render (as was originally in a PR) - I suggest to migrate to that render.

KATT:
> > I'm not sure what is going on exactly but getting
> 
> We're running a render inside a render, this is not expected in React unfortunately as they keep contexts in a stack. I'm not sure if this going to break anything, but if it's fixed by running concurrent render (as was originally in a PR) - I suggest to migrate to that render.

Not getting an error after 10be9c0

packages/next/src/withTRPC.tsx
KATT: SSR wasn't working together with suspense before so this isn't regression at least, I think we even noted that in the docs

KATT:
Merging this to `next` + opened up #5201 to get it done on `main`

luixo:
@KATT 
Should we change `renderToString` to `renderToPipeableStream` in `next` branch where `react@18` is available?
It will enable awaiting suspense components for all cases.

PR #3354, opened 2022-12-06
www: add video poster, preload, & lazy load iframes

neo773:
## 🎯 Changes

This PR adds poster attribute for preview while the video is loading & preload tag in header for performance

## ✅ Checklist


www/src/components/Preview.tsx
```diff
@@ -1,6 +1,26 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 
 export const Preview = () => {
+  function getWindowDimensions() {
```
KATT, line 4: You need to do `if (typeof window === "undefined") { return something }` here

KATT (changes requested):
Instead of doing this with a hook and calculating the viewport size, I think we should do it with CSS - having two `<figure>`s or whatever that are hidden/block based on CSS. That would get rid of the build failing without doing special hacks too

KATT (approved):

www/src/pages/index.tsx
```diff
@@ -139,7 +139,6 @@ const HomeContent: React.FC = () => {
               })
             }
             frameBorder="0"
-            loading="lazy"
```
neo773: We need this, makes a huge improvement in page load time and lighthouse score

www/src/pages/index.tsx
```diff
@@ -245,6 +245,28 @@ const HomeHead: React.FC = () => {
...
+        href="https://assets.trpc.io/www/v10/v10-dark-landscape.mp4"
+        as="video"
+        type="video/mp4"
+      />
+      <link
+        rel="preload"
+        href="https://assets.trpc.io/www/v10/preview-dark.mp4"
+        as="video"
+        type="video/mp4"
+      />
```
KATT, line 259:
```suggestion
```

KATT: I think these are not really needed and will be a bit too "greedy"

www/src/pages/index.tsx
KATT: Ah i see

neo773: Yeah, and sorry about keeping these PRs open haven't had a chance yet to dig back in.

KATT: Thanks!

KATT (approved):

PR #3379, opened 2022-12-12
Adds og image generation and meta tags for blog posts

### LINKED ISSUE #3364: feat: open graph images for articles & blog posts

Having open graph images for our posts would be nice. Ideally automatic, should be doable? 

### Describe the solution you'd like to  

~~Something like this would be nice https://github.com/vercel/og-image~~

This would be nicer: https://vercel.com/docs/concepts/functions/edge-functions/og-image-generation

### Additional information

https://github.com/trpc/trpc-assets

### 👨‍👧‍👦 Contributing

iway1:
Closes #3364

## 🎯 Changes

Adds Open Graph image generation to www build step, adds og meta tags to blog posts. OG images are generated using the title metadata of the blog post (if one is not passed the text will be set to "tRPC Blog Post"). OG images are served as static png files.

Open Graph generation is done in memory using `satori` from vercel (which is what they use under the hood in their new `@vercel/og` library). 

The background and font used for the og generation can be swapped out in static/og-assets. 

One caveat is that there seems to be an issue with `satori` where it will clip the bottom of the image used, so the source background image needs to be taller than the target resolution (I used 720px height when targeting 640px and it looked good).

[Swizzles](https://docusaurus.io/docs/swizzling) the BlogPostPageMetadata component to add the og meta tags.

## ✅ Checklist


iway1:
@KATT Thanks for checking that out.

I'd actually originally set it up to dynamically generate with an API route and had that working locally, but thought it would be better to just generate at build times b/c they never change so I reworked it to use the script. Didn't think about the build times. Whoops 😅

I'll check it out some more tonight

PR #3389, opened 2022-12-13
feat(client): add base client for integrations

### LINKED ISSUE #3342: base client for integrations
For the interop API, we have a `createTRPCClient` that works like this:
```ts
const client = createTRPCClient<AppRouter>()

const post = client.query('post.byId', { id: 0 });
```

This was good for interop routers but those are all deprecated now. But we still access these methods for proxy stuff under the hood:
https://github.com/trpc/trpc/blob/c5b49adb1d4e875633bb88ce282a3573fa00b404/packages/client/src/createTRPCClientProxy.ts#L121-L127

The `const client = new TRPCClient` is an fully-typed interop client so none of our new proxy types work correctly when accessing it so we cast to `any` when using it:
https://github.com/trpc/trpc/blob/c5b49adb1d4e875633bb88ce282a3573fa00b404/packages/client/src/createTRPCClientProxy.ts#L115-L116

For both third-party clients and our own clients we should improve to be more type-safe. For example `jotai-trpc` could have benefited from a more basic client that was loose on type safety: https://github.com/jotai-labs/jotai-trpc/pull/1#discussion_r955662034.

So we're looking for something like this:
```ts
import { createBaseClient } from '@trpc/server'

const client = createBaseClient({
  links: [/* ... */]
})

// client's APIs are not strictly typed

const post = client.query('post.byId', { id: 0 })
//    ^? `any`
```

`client.query` accepts a `string` (as opposed to a union of all the router's paths) and accepts an input object - `unknown`. It returns an object of `any`. All methods should be supported - query, mutation, subscription. This basic API will allow us to use it to easily build other integrations on top of without explicitly casting to `any` everywhere. You can use [the current `TRPCClient` for inspiration](https://github.com/trpc/trpc/blob/c5b49adb1d4e875633bb88ce282a3573fa00b404/packages/client/src/internals/TRPCClient.ts#L56).

bautistaaa:
Closes #3342

## 🎯 Changes

## ✅ Checklist


packages/client/src/createTRPCClient.ts
```diff
@@ -1,10 +1,55 @@
-import type { AnyRouter } from '@trpc/server';
+import type {
+  AnyRouter,
+  inferProcedureInput,
+  inferProcedureOutput,
+  inferSubscriptionOutput,
+} from '@trpc/server';
+import { inferTransformedProcedureOutput } from '@trpc/server/shared';
+import { Unsubscribable } from 'packages/server/observable';
```
juliusmarminge, line 8:
```suggestion
import { Unsubscribable } from '@trpc/server/observable';
```

bautistaaa: FFFF lol

packages/client/src/createBaseTRPCClient.ts
```diff
@@ -0,0 +1,25 @@
+import {
+  BaseTRPCClient as Client,
+  CreateTRPCClientOptions,
+} from './internals/BaseTRPCClient';
+import { httpBatchLink } from './links';
+
+export function createBaseTRPCClient(opts: CreateTRPCClientOptions) {
```
sachinraja, line 7: maybe this should be called `createGenericTRPCClient`? what do you think

sachinraja (dismissed):
looks good!

KATT (commented):
Can we validate that the changes in exports doesn't break any third party libs like trpc-solid/nuxt/svelte etc? 🤔

bautistaaa:
verified solid / sveltekit are working as expected

cc @KATT @sachinraja

bautistaaa:
renamed the function to generic @sachinraja

sachinraja (approved):
Niceee

packages/client/src/createTRPCClient.ts
```diff
@@ -1,10 +1,55 @@
...
+  GenericTRPCClient as Client,
   CreateTRPCClientOptions,
-} from './internals/TRPCClient';
-import { httpBatchLink } from './links';
+  TRPCRequestOptions,
+  TRPCSubscriptionObserver,
+} from './internals/GenericTRPCClient';
+import { TRPCClientRuntime, httpBatchLink } from './links';
+
+export interface TRPCClient<TRouter extends AnyRouter> {
```
KATT, line 18: We can slap a `/** deprecated **/` on this as it's v9 stuff

KATT: Makes sure we remember to delete it for v11

packages/client/src/internals/GenericTRPCClient.ts
```diff
@@ -48,17 +44,17 @@ export interface TRPCSubscriptionObserver<TValue, TError> {
...
 /** @internal */
-export type CreateTRPCClientOptions<TRouter extends AnyRouter> =
+export type CreateTRPCClientOptions<TRouter extends AnyRouter = AnyRouter> =
   | CreateTRPCClientBaseOptions & {
       links: TRPCLink<TRouter>[];
     };
 
-export class TRPCClient<TRouter extends AnyRouter> {
-  private readonly links: OperationLink<TRouter>[];
+export class GenericTRPCClient {
```
KATT, line 52: Does this need to be a `class`? 🤔 

I had an issue somewhere to rewrite everything to not use `class` b/c `class`es suck in JS

KATT (commented):
- Does it need to be a `class`?
- Do we need docs?
- (sorry for the @), cc @jgoux, you have had a history of problems whenever we mess with the exported types - would you mind sanity-checking this PR?

KATT (approved):
I meant to approve, but read previous comments

packages/client/src/internals/GenericTRPCClient.ts
KATT: Can we avoid exporting this as a `class`?

I'm fine if we export it as a `type` but if we export the class it means someone somewhere will do `new GenericTRPCClient`

Let's only export the factory method and it's interface so we're free to refactor it class-free

sachinraja: Am I missing something or is this not exactly how it works currently?

sachinraja (approved):

PR #3186, opened 2022-11-17
feat: improve turbo setup to better utilise caching

juliusmarminge:
Closes #

## 🎯 Changes

- Separates ESLint and lints with Turbo
  - this will make sure that a change to a single workspace doesn't require the whole repo to be re-linted 
- Fixes cache misses and some pipeline config stuff
  - some output stuff missing
  - TODO: currently deactivated caching for the `next-prisma-websockets-starter` cause I couldn't get that to work...

## ✅ Checklist

- [ ] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/next/CONTRIBUTING.md).
- [ ] If necessary, I have added documentation related to the changes made.
- [ ] I have added or updated the tests related to the changes made.

PR #3390, opened 2022-12-14
fix typescript auto imports

sachinraja:


KATT (commented):
I like it but it fails ci

tsconfig.json
```diff
@@ -1,16 +1,7 @@
 {
   "extends": "./tsconfig.build.json",
-  "include": ["packages", "types", "scripts", "examples", "rollup.config.ts"],
+  "include": ["examples", "scripts"],
```
JoshuaKGoldberg, line 3: Is it ok to leave `examples` in here?

## Explanation

When you run ESLint on the CLI, it will look for any ESLint config in any directory it's running on. If it finds an ESLint config, it will use those settings for that directory and all children.

So, files inside `example/` directories with an ESLint config file are seeing new ESLint options applied on top of the defaults set in the root ESLint config. Most importantly, `parserOptions.project` is set to `tsconfig.json` in `examples/` directories, such as by: https://github.com/trpc/trpc/blob/fb09965395b34bdfa53270250dd35bf1a239a6b6/examples/next-prisma-websockets-starter/.eslintrc#L12

Result: `example/` directories are getting linted with the wrong `parserOptions.project`. And now that this PR removes `"examples"` from the list of included files in the root TSConfig, they can no longer be found for typed linting. Hence https://typescript-eslint.io/linting/troubleshooting#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file.

## Fixing

The easiest thing would be to add `examples` back into the `include`. This brings us back to how things were on the `main` branch (even prior to #3482).


Two future pieces of work would help here:
* ESLint is moving towards a _non-_cascading config: https://eslint.org/docs/latest/user-guide/configuring/configuration-files-new. It's not generally ready for public use though.
* https://github.com/typescript-eslint/typescript-eslint/pull/6084 would actually help here in letting `examples/` ESLint configs each have the `tsconfig.json` closest to them be used

JoshuaKGoldberg: Posted https://github.com/trpc/trpc/pull/3186/files#r1063381941: I don't think so, but not sure.

JoshuaKGoldberg: Ah:

```plaintext
@examples/minimal:lint: /home/runner/work/***/***/examples/minimal/client/index.ts
@examples/minimal:lint:   0:0  error  Parsing error: Cannot read file '/home/runner/work/***/***/examples/minimal/www/tsconfig.json'
@examples/minimal:lint: 
@examples/minimal:lint: /home/runner/work/***/***/examples/minimal/server/index.ts
@examples/minimal:lint:   0:0  error  Parsing error: Cannot read file '/home/runner/work/***/***/examples/minimal/www/tsconfig.json'
```

👀

JoshuaKGoldberg: Ok should be resolved with `tsconfigRootDir` (documented a bit more on https://github.com/typescript-eslint/typescript-eslint/pull/6236). c4d3f9c42991923aab609bffd9c04f2d78ce42b9

I also found that Turbo was retrieving cached lint runs even when when `.eslint*` files were changed. I'll PR that against `main`.

KATT (changes requested):
@sachinraja please have another stab at fixing or close this 🙏

PR #3404, opened 2022-12-16
docs: ecosystem page

sachinraja:
my take on #3403
based on https://www.solidjs.com/ecosystem

https://www-git-ecosystem-page-trpc.vercel.app/ecosystem
cc @KATT


KATT (approved):
I like it, but it only works in dark mode and is a bit "overwhelming" (not to say that the current page isn't)

PR #3431, opened 2022-12-19
fix(docs): change typo`initTRPC()` to `initTRPC`

SrBrahma:
## 🎯 Changes

`initTRPC` is not callable, as TS and the other examples show.


juliusmarminge (approved):
Thanks 👍🏼

PR #3457, opened 2022-12-22
chore: pnpm in tool-versions:

nfabredev:
Hey I was wondering if you wanted to use asdf to manage the pnpm version on top of the nodejs version.

Current latest is the one in the `.tool-versions` file but you might be using a different one - and it's nice to align everyone on that.

## 🎯 Changes

- add pnpm to the .tool-versions file
- add instructions to use asdf for nodejs and pnpm

## ✅ Checklist

- ~~I have added or updated the tests related to the changes made.~~ Not applicable


sachinraja:
You should already be using a fairly popular node version manager. We also have a `.nvmrc` file. You shouldn't need a separate `pnpm` plugin for asdf either - it's just `npm i pnpm -g` (which pnpm shows in their docs).

nfabredev:
I'm not familiar with pnpm but wouldn't it be possible that people have different major versions of pnpm and would get different results when e.g. installing dependencies ? 

If that's a non issue removing the .tool-versions file could avoid it going out of sync with the .nvmrc - if that's the recommended way of managing the node version over asdf. Unless there's some use case having both these files ?

juliusmarminge:
> I'm not familiar with pnpm but wouldn't it be possible that people have different major versions of pnpm and would get different results when e.g. installing dependencies ? 
> 
> 
> 
> If that's a non issue removing the .tool-versions file could avoid it going out of sync with the .nvmrc - if that's the recommended way of managing the node version over asdf. Unless there's some use case having both these files ?

You'll get warnings if you run pnpm install on an old version of pnpm, with the command to update.

If you use like `fnm` or `nvm` you'll automatically get put on the correct node version when you `cd` into the project directory.

Not too familiar with the whole .tool-version stuff so not sure what it does tbf... there's also like `gitpod.yml` files you could have for other reasons? Again really not sure what they do...

sachinraja (approved):

PR #3458, opened 2022-12-22
docs: add bun example

### LINKED ISSUE #3385: feat: minimal bun example

tRPC should work with [Bun](https://bun.sh/), seeing that we don't have any deps on node

An example in the [`/examples`](https://github.com/trpc/trpc/tree/main/examples)-folder using Bun similar to our deno example

If it doesn't work, it'd be good with issues on what is not working

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

nfabredev:
Closes #3385

## 🎯 Changes

Adding minimal bun example

### Questions / remarks:

- Do you want to see a more fleshed out example ?
- Should this include a test that it's actually working ? Not all example seem to have tests. I haven't looked yet if testing the examples is set up in the CI ?
- I've left the Copyright notice from origin repo where i took the implementation from, as the changes I brought so far are quite minimal!

## ✅ Checklist


KATT (changes requested):
- [ ] Can you make the example minimal, without React-etc? It gets harder to maintain when we have a bunch of examples with many deps, [see the deno example](https://github.com/trpc/trpc/tree/main/examples/deno-deploy)
- [ ] Ideally, make it tested, again, [see deno in ci](https://github.com/trpc/trpc/blob/1e6ccabfb4f4c5674d888ca074e362d8273c0e78/.github/workflows/examples.yml#L107-L123) as an example
- [ ] Setup github sponsors so we can send you some $ :)
- [ ] Is `package.json` needed for bun?

KATT (changes requested):
- [x] Can you make the example minimal, without React-etc? It gets harder to maintain when we have a bunch of examples with many deps, [see the deno example](https://github.com/trpc/trpc/tree/main/examples/deno-deploy)
- [x] Ideally, make it tested, again, [see deno in ci](https://github.com/trpc/trpc/blob/1e6ccabfb4f4c5674d888ca074e362d8273c0e78/.github/workflows/examples.yml#L107-L123) as an example
- [x] Setup github sponsors so we can send you some $ :)
- [ ] Is `package.json` needed for bun?
- [ ] Is it possible for it to skip the lockfile and instead used the build version from `../`? We won't know if changes breaks Bun afterwards

nfabredev:
Thanks! 🙌 Had fun learning about trpc and bun - and making these work together in a simple example. 

If I understand correctly Bun does use package.json when it's used as a package manager ( i.e. running `bun install` ), eventhough it also tracks dependencies in its own lockfile after that for performance reasons apparently. See https://github.com/oven-sh/bun#using-bun-as-a-package-manager.

The test doesn't work yet - I'll still have to edit this one.

You want to use the trpc that's built in the repo instead of the versions references in the package.json right ?

nfabredev:
I'm blocked with two things:

- Running the test for the bun example, see the logs here: https://github.com/trpc/trpc/actions/runs/3765455181/jobs/6400952806

It seems to time out waiting for a successful response from the URL we pass to the test runner. I've tried both with http://localhost:8000/trpc/hello and http://localhost:3000, which respond with 200 locally - so should work.

- Running `pnpm install` locally in the root of the repo - as suggested in the CONTRIBUTING.md - brings a huge amount of changes which i don't think are intended. 

I've only included changes that seem relevant in [that commit](https://github.com/trpc/trpc/pull/3458/commits/953537996de50bb3c58e04278207e3afc23ddf66) but i'm wondering what's happening and if that's correct behaviour. I'm using pnpm `7.16.0`.

Let me know if you have ideas how to fix those.

juliusmarminge:
> I'm blocked with two things:
> 
> 
> 
> - Running the test for the bun example, see the logs here: https://github.com/trpc/trpc/actions/runs/3765455181/jobs/6400952806
> 
> 
> 
> It seems to time out waiting for a successful response from the URL we pass to the test runner. I've tried both with http://localhost:8000/trpc/hello and http://localhost:3000, which respond with 200 locally - so should work.
> 

I think it should be `127.0.0.1:3000` for CI to work.


> - Running `pnpm install` locally in the root of the repo - as suggested in the CONTRIBUTING.md - brings a huge amount of changes which i don't think are intended. 
> 
> 
> 
> I've only included changes that seem relevant in [that commit](https://github.com/trpc/trpc/pull/3458/commits/953537996de50bb3c58e04278207e3afc23ddf66) but i'm wondering what's happening and if that's correct behaviour. I'm using pnpm `7.16.0`.


You could have some global settings set such as autoinstallpeers or something?

nfabredev:
Thanks @juliusmarminge ! That was part of the issue - and pointing the test runner to the wrong URL seems to have made it hang forever - but the test passes now. The issue with the lockfile is also resolved. 

Let me know if further changes are needed - or if there are more questions.

KATT (changes requested):
Fails CI right now, could have a look please, @nfabredev? 🙏

PR #3466, opened 2022-12-23
feat: handle reserved properties key collisions

### LINKED ISSUE #3461: bug: "client" namespaced router causes conflict for useContext

  System:
    OS: macOS 13.1
    CPU: (8) arm64 Apple M1
    Memory: 82.05 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
    Yarn: 1.22.11 - /opt/homebrew/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v18.12.1/bin/npm
  Browsers:
    Chrome: 108.0.5359.124
    Safari: 16.2

### Describe the bug

When naming a router `client`, the useContext hook does not perform as expected. When trying to access any of the provided methods, you get a runtime error:
```
client [procedureType] is not a function
```
CleanShot_2022-12-22_at_13 38 20

### Link to reproduction

https://stackblitz.com/edit/github-8msfux?file=src/pages/index.tsx

### To reproduce

Click the "Broken Function" button to run the function that throws an error

### Additional information

As you'll see, I'm using the same router under 2 different names. "client" may not need to be the only restricted name there, but if used it should probably cause a warning or type error if used.

### 👨‍👧‍👦 Contributing

sachinraja:
closes #3461

## 🎯 Changes

tRPC has "reserved properties" like `utils.client`. The type for this was originally messed up. If there was a router named `client`, `utils.client` would only return a proxy client, but for the types it also added the properties from that router.

This change merges the reserved properties and procedure records together so that the reserved properties cannot be overridden.

## ✅ Checklist


sachinraja:
draft because this needs perf testing

KATT (dismissed):
Regression test would be good

packages/react-query/src/createTRPCReact.tsx
```diff
@@ -174,13 +174,22 @@ export type DecoratedProcedureRecord<
...
+> = Overwrite<
+  DecoratedProcedureRecord<TRouter['_def']['record'], TFlags>,
+  {
+    useContext(): CreateReactUtilsProxy<TRouter, TSSRContext>;
+    Provider: TRPCProvider<TRouter, TSSRContext>;
+    createClient: CreateClient<TRouter>;
+    useQueries: TRPCUseQueries<TRouter>;
+    useDehydratedState: UseDehydratedState<TRouter>;
+  }
+>;
```
KATT, line 192: It would be even nicer if the intersection gave some sort of error when trying to use it

```ts
type C = Ovewrite<A & B, IntersectionError<A, B, "This key collides with a built-in method, you should rename this router or procedure on your backend">>
```

KATT: Or maybe it should fail earlier and fail the whole creation of the client, but only if we can showcase _which_ property that is wrong

packages/react-query/src/createTRPCReact.tsx
```diff
@@ -188,13 +189,25 @@ export type DecoratedProcedureRecord<
...
+};
+
+export type CreateTRPCReact<
+  TRouter extends AnyRouter,
+  TSSRContext,
+  TFlags,
+> = ProtectedIntersection<
+  CreateTRPCReactBase<TRouter, TSSRContext>,
+  DecoratedProcedureRecord<TRouter['_def']['record'], TFlags>
+>;
```
sachinraja, line 210: `ProtectedIntersection` here will show a string error if there is a collision between the keys

packages/server/src/types.ts
```diff
@@ -26,6 +26,20 @@ export type FlatOverwrite<TType, TWith> = InferOptional<
   UndefinedKeys<TType> | UndefinedKeys<TWith>
 >;
 
+/**
+ * @internal
+ */
+export type IntersectionError<TKey extends string> =
+  `${TKey} collides with a built-in method, you should rename this router or procedure on your backend`;
```
sachinraja, line 33: here's the error message

KATT: Oh shiiit, it even has a name on it

KATT (approved):

PR #3469, opened 2022-12-23
chore: enable ESLint's --report-unused-disable-directives

JoshuaKGoldberg:
~Closes #~ - I didn't see a tooling template so I just sent this. Sorry if it's spam! Thought you'd enjoy it. 😄 

## 🎯 Changes

Enables [--report-unused-disable-directives](https://eslint.org/docs/latest/user-guide/command-line-interface#--report-unused-disable-directives). Also [`--max-warnings 0`](https://eslint.org/docs/latest/user-guide/command-line-interface#--max-warnings) just in case.

...and, voilà! 138 line deletions. ✨  

## ✅ Checklist


sachinraja:
Is there a point in having warnings vs errors at all if we error on all warnings?

JoshuaKGoldberg:
> Is there a point in having warnings vs errors at all if we error on all warnings?

Not particularly. The one user-facing change would end up being the color of the squiggly in their editor. https://github.com/eslint/eslint/discussions/16512 has some discussion around that.

I've been adding `--max-warnings 0` because sometimes plugins will set rules to warnings in their preset configs. typescript-eslint does that right now (though we'll lkely be switching to errors in v6, in a few months).

juliusmarminge:
Love the unused directive, use that myself on all projects - helps prevent bad stuff just cause you have a directive laying around from ages ago not doing anything.

Not too sure about the max-warning though. Is there no other way of alerting of warnings without actually erroring? Like the GitHub bot or something that tells if you leak secrets for example?

JoshuaKGoldberg:
> Is there no other way of alerting of warnings without actually erroring? Like the GitHub bot or something that tells if you leak secrets for example?

I haven't seen that set that up, but in theory it's possible!

I can just revert the change 😄 it's an ancillary thing.

juliusmarminge (dismissed):
Great, why not put it in the config though? All examples would still use it since it traverses down to the root config?

https://github.com/juliusmarminge/jumr.dev/blob/3a45b6af9be3348ae0dd9e0eca73277ad7b95a29/.eslintrc#L25

JoshuaKGoldberg:
> why not put it in the config though

Per ESLint's docs, the `linterOptions.reportUnusedDisableDirectives` option isn't supported in the current (classic/legacy) ESLint config system: https://eslint.org/docs/latest/user-guide/configuring/language-options. It's only supported in the upcoming new ESLint config system: https://eslint.org/docs/latest/user-guide/configuring/configuration-files-new#configuration-objects.

KATT (dismissed):
Love it

sachinraja (approved):
Awesome :fire:

sachinraja (approved):

PR #3479, opened 2022-12-24
chore: fix types of `GenericTRPCClient`

KATT:
- Don't default to `AnyRouter` except on the  (defaults as evil)
- Skip an unnecessary guard

### DIFF
--- a/packages/client/src/internals/GenericTRPCClient.ts
+++ b/packages/client/src/internals/GenericTRPCClient.ts
@@ -72,37 +72,35 @@ export interface TRPCSubscriptionObserver<TValue, TError> {
 }
 
 /** @internal */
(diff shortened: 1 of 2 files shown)


packages/client/src/internals/GenericTRPCClient.ts
```diff
@@ -72,37 +72,35 @@ export interface TRPCSubscriptionObserver<TValue, TError> {
...
-      if (!opts.transformer || typeof opts.transformer === 'string')
+      const transformer = opts.transformer as
+        | ClientDataTransformerOptions
+        | undefined;
+      if (!transformer)
         return {
           serialize: (data) => data,
           deserialize: (data) => data,
         };
-      // Type guard for `opts.transformer` because it can be `any`
```
KATT, line 94: Replaced by typecast a few lines above

PR #3480, opened 2022-12-24
chore: deprecate usage of `createHooksInternal`

KATT:
- [x] `GenericX` -> `UntypedX`
- [x] Make untyped root factory for react
- [ ] Make untyped context
- [ ] Use untyped client in context


packages/react-query/src/internals/createRootHooks.tsx
```diff
@@ -0,0 +1,506 @@
...
+import { TRPCContextState } from './deprecated/TRPCContextState';
+import { QueryType, getArrayQueryKey } from './getArrayQueryKey';
+import { getClientArgs } from './getClientArgs';
+import { useHookResult } from './useHookResult';
+import { TRPCUseQueries } from './useQueries';
+
+/**
+ * @internal
+ */
+export function createRootHooks<
```
KATT, line 41: this is the new "root hooks"

packages/react-query/src/internals/createRootHooks.tsx
```diff
@@ -0,0 +1,506 @@
...
+  TRouter extends AnyRouter,
+  TSSRContext = unknown,
```
KATT, line 43: TBD if these are useful

packages/react-query/src/internals/createRootHooks.tsx
KATT: the contents i've just copy/pasted from `createHooksInternal`, but has no generic shit

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -1,228 +1,52 @@
...
-  const ref = useRef(value);
-  ref.current.path = value.path;
-  return ref.current;
-}
-/**
- * Create strongly typed react hooks
  * @internal
  */
-export function createHooksInternal<
+export function createRootHooks<
```
KATT, line 41: - TBD if the generics here are needed
- should be called createUntypedSomething
- not in another file b/c i want to keep git blame

PR #3482, opened 2022-12-26
chore: enable typescript-eslint/recommended-requiring-type-checking

JoshuaKGoldberg:
~Closes #~ [I was told to just go nuts 😄](https://discord.com/channels/867764511159091230/980118117676118067/1056705880726446130)

## 🎯 Changes

Enables [@typescript-eslint/recommended-requiring-type-checking](https://typescript-eslint.io/linting/configs/#recommended-requiring-type-checking), which is a set of recommended rules that also require [linting with type info](https://typescript-eslint.io/linting/typed-linting) (which was already set up).

Fixes `tsconfig.eslint.json` to properly include compiler options and path mappings so that lint rules have correct type info.

Rules with straightforward-to-fix issues have those issues fixed, with inline comments explaining the rule.

Marks rules that had a lot of non-straightforward-to-fix issues as todos. It'd be good to fix them eventually. But that's out of scope of this PR.

## ✅ Checklist

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,8 +1,9 @@
 {
-  "parser": "@typescript-eslint/parser", // Specifies the ESLint parser
+  "parser": "@typescript-eslint/parser",
   "plugins": ["no-only-tests", "unicorn", "turbo"],
   "extends": [
-    "plugin:@typescript-eslint/recommended", // Uses the recommended rules from the @typescript-eslint/eslint-plugin
+    "plugin:@typescript-eslint/recommended",
+    "plugin:@typescript-eslint/recommended-requiring-type-checking",
     "plugin:react/recommended",
     "plugin:react-hooks/recommended",
     "plugin:prettier/recommended"
@@ -19,7 +20,6 @@
     "react/react-in-jsx-scope": "off",
     "react/prop-types": "off",
     "@typescript-eslint/no-explicit-any": "off",
-    "@typescript-eslint/no-floating-promises": "error",
     "no-only-tests/no-only-tests": "error",
     "@typescript-eslint/no-empty-interface": "off",
     "unicorn/filename-case": [
@@ -60,10 +60,27 @@
   },
   "overrides": [
     {
+      "files": ["examples/**/*", "packages/*/**/*", "scripts/**/*", "www/**/*"],
+      "rules": {
+        // Todo: enable these for even stronger linting! 💪
+        "@typescript-eslint/no-misused-promises": "off",
+        "@typescript-eslint/no-unsafe-argument": "off",
+        "@typescript-eslint/no-unsafe-assignment": "off",
+        "@typescript-eslint/no-unsafe-call": "off",
+        "@typescript-eslint/no-unsafe-declaration-merging": "off",
+        "@typescript-eslint/no-unsafe-member-access": "off",
+        "@typescript-eslint/no-unsafe-return": "off",
+        "@typescript-eslint/require-await": "off",
+        "@typescript-eslint/restrict-plus-operands": "off",
+        "@typescript-eslint/restrict-template-expressions": "off"
+      }
+    },
+    {
       "files": ["examples/**/*"],
       "rules": {
-        "@typescript-eslint/no-unused-vars": "off",
-        "@typescript-eslint/no-floating-promises": "off"
+        // Todo: enable these for even stronger linting! 💪
+        "@typescript-eslint/no-floating-promises": "off",
+        "@typescript-eslint/no-unused-vars": "off"
       }
     },
     {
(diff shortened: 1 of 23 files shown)


.eslintrc
```diff
@@ -19,7 +20,6 @@
     "react/react-in-jsx-scope": "off",
     "react/prop-types": "off",
     "@typescript-eslint/no-explicit-any": "off",
-    "@typescript-eslint/no-floating-promises": "error",
```
JoshuaKGoldberg, line 22: This is already enabled as part of `plugin:@typescript-eslint/recommended-requiring-type-checking`.

examples/next-prisma-todomvc/src/pages/[filter].tsx
```diff
@@ -278,8 +278,7 @@ export default function TodosPage({
               <Link
                 href="/all"
                 className={clsx(
-                  !['active', 'completed'].includes(filter as string) &&
```
Comment (JoshuaKGoldberg) [inline examples/next-prisma-todomvc/src/pages/[filter].tsx:281]: https://typescript-eslint.io/rules/no-unnecessary-type-assertion

`filter` is already a `string`. The `as` type assertion does nothing.

packages/react-query/src/internals/context.tsx
```diff
@@ -94,19 +112,19 @@ export interface TRPCContextState<
   /**
    * @link https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientfetchquery
    */
-  fetchQuery<
+  fetchQuery: <
```
JoshuaKGoldberg, line 115: https://typescript-eslint.io/rules/unbound-method

Marking these as properties with function types (rather than methods) indicates it's safe to call them without caring about the `this` scope.


```ts
declare const values: {
  a(): void;
  b: () => void;
}

const { a, b } = values;

a(); // Not safe! What if the `this` is wrong?

b(); // this is fine
```

I actually filed https://github.com/typescript-eslint/typescript-eslint/issues/6276 from working on these changes.

packages/tests/server/adapters/fastify.test.ts
```diff
@@ -199,9 +199,7 @@ describe('anonymous user', () => {
     await app.start();
   });
 
-  afterEach(async () => {
-    await app.stop();
-  });
```
JoshuaKGoldberg, line 204: https://typescript-eslint.io/rules/require-await

`app.stop()` doesn't create a `Promise` according to its types. That should indicate there's no need to `await` it.

If there _is_ a need to `await` it, then its return type should indicate it's returning a `Promise`.

packages/tests/server/interop/links.test.ts
```diff
@@ -58,7 +58,7 @@ test('chainer', async () => {
 
   const result = await observableToPromise(chain).promise;
   expect(result?.context?.response).toBeTruthy();
-  result!.context!.response = '[redacted]' as any;
+  result.context!.response = '[redacted]' as any;
```
JoshuaKGoldberg, line 61: Roughly speaking, `expect<T>(value: T).toBeTruthy()` returns `T`. Which lets TypeScript know:

1. `result?.context?.response` must be truthy
2. `result`
3. The `!` non-null assertion in `result!` is unnecessary

.github/workflows/lint.yml
```diff
@@ -38,6 +38,9 @@ jobs:
       - name: Install deps (with cache)
         run: pnpm install --child-concurrency 3
 
+      - name: Build
+        run: pnpm turbo --filter "@trpc/*" build
```
JoshuaKGoldberg, line 42: Necessary because lint rules use type information that is more accurate after a build. Without a build before lint, we get odd lint rule failures [[example build](https://github.com/trpc/trpc/actions/runs/3778203385/jobs/6422483536)].

juliusmarminge (approved):
Love these lint rules. Removes a lot of casts or assertions that might have been necessary before but isnt anymore and have just laid around ever since :)

KATT (approved):

packages/react-query/src/internals/context.tsx
KATT: FWIW, I really don't care about this one as we're moving away form using `class`es altogether b/c JS sucks at them

JoshuaKGoldberg: Nice. In that case you might want https://typescript-eslint.io/rules/method-signature-style, to enforce folks don't accidentally imply they're creating methods (& educate on the difference).

.github/workflows/lint.yml
Andarist: Interesting - do you know why this happens? 🤔

juliusmarminge: This can prob be configured in he turbo pipeline to get more optimized.

1. Setting build as a dependency for lint
2. Running just lint with turbo 

Might be able to split the linting up to get even faster CI if done correctly https://github.com/trpc/trpc/pull/3186

PR #3485, opened 2022-12-26
chore(react-query): refactor proxy-methods to use untyped base objects

KATT:
- Use `UntypedClient` in our proxy fns
- Use


KATT:
moved to #3494

PR #3489, opened 2022-12-26
feat(server): add ability to pipe middlewares

bautistaaa:
Closes https://github.com/trpc/trpc/issues/3434

What changes are made in this PR? Is it a feature or a bug fix?

## ✅ Checklist

- [ ] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [ ] If necessary, I have added documentation related to the changes made.
- [ ] I have added or updated the tests related to the changes made.

PR #3490, opened 2022-12-26
Mucked around with middleware pipe typings

JoshuaKGoldberg:
Fixes some of the type unhappiness from #3489:

1. Adds `MiddlewareBuilder` to `use`'s first parameter's type union
2. Adds `MiddlewareBuilder` to `ProcedureBuilderMiddleware`, so the `middleware: ...` passed in `use`'s implementation is allowed
3. Added a quick lil `in` type narrowing in `callRecursive` to account for the parameter being a builder

I know you all like to joke about how you don't know what you're doing, but I _actually_ don't know what's happening. I'm not a tRPC power user and just thought it'd be fun to poke around. Please don't take this PR very seriously. 😄 

I also thought this would be a lot more code than the 1 file changed + 1 testing file? So maybe I'm missing something?

### DIFF
--- a/packages/server/src/core/initTRPC.ts
+++ b/packages/server/src/core/initTRPC.ts
@@ -156,3 +156,57 @@ function createTRPCInner<TParams extends PartialRootConfigTypes>() {
     };
   };
 }
+
+// Just testing out the types with the following snippets 🙂
+
+const t = initTRPC.create({});
+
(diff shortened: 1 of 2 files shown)


bautistaaa:
took pieces of this and added to: https://github.com/trpc/trpc/pull/3489

KATT:
Part of #3489!

PR #3491, opened 2022-12-26
more turbo lint attempts

JoshuaKGoldberg:
wip

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -34,6 +34,7 @@
       "cache": false
     },
     "lint": {
+      "dependsOn": ["^build", "codegen:bigBoi", "codegen:bigLegacyRouter"],
       "outputs": [],
       "cache": true
     },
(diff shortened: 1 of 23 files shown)


turbo.json
```diff
@@ -34,6 +34,7 @@
       "cache": false
     },
     "lint": {
+      "dependsOn": ["^build", "codegen:bigBoi", "codegen:bigLegacyRouter"],
```
juliusmarminge, line 37: this won't do anything unless we run the lint script with turbo

JoshuaKGoldberg: Are we not? Or is there a script that's more turbo-hooked-up?

juliusmarminge: No, ESLint is currently just being run on its own:

https://github.com/trpc/trpc/blob/db189c9aa569f0f066ccd432a83920093fa2d6a7/package.json#L15

The `turbo-lint` PR (#3186) was made to split this up and run them all using Turbo

juliusmarminge: Oh waiiit i checked the wrong branch... In here it's totally valid. Thought you were based off of main 🤦🏼‍♂️

PR #3517, opened 2022-12-28
Trash/middleware pipe fork

KATT:
- Simplify logic
- Don't mutate

### DIFF
--- a/packages/server/src/core/internals/procedureBuilder.ts
+++ b/packages/server/src/core/internals/procedureBuilder.ts
@@ -254,14 +254,15 @@ export function createBuilder<TConfig extends AnyRootConfig>(
     unstable_concat(builder) {
       return createNewBuilder(_def, builder._def) as any;
     },
-    use(middleware) {
-      if ('pipe' in middleware) {
-        return createNewBuilder(_def, {
-          middlewares: middleware._self.piped ?? [middleware._self],
-        }) as AnyProcedureBuilder;
-      }
+    use(middlewareBuilderOrFn) {
+      // Distinguish between a middleware builder and a middleware function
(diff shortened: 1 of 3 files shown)


packages/server/src/core/internals/procedureBuilder.ts
```diff
@@ -254,14 +254,15 @@ export function createBuilder<TConfig extends AnyRootConfig>(
...
-          middlewares: middleware._self.piped ?? [middleware._self],
-        }) as AnyProcedureBuilder;
-      }
+    use(middlewareBuilderOrFn) {
+      // Distinguish between a middleware builder and a middleware function
+      const middlewares =
+        '_middlewares' in middlewareBuilderOrFn
+          ? middlewareBuilderOrFn._middlewares
+          : [middlewareBuilderOrFn];
+
```
KATT, line 263: At least this is slightly better

packages/server/src/core/middleware.ts
```diff
@@ -143,27 +156,24 @@ export type MiddlewareFunction<
 };
 
 export function createMiddlewareFactory<TConfig extends AnyRootConfig>() {
-  function createNewBuilder(
-    previousMiddleware: MiddlewareFunction<any, any>,
-    newMiddleware: MiddlewareFunction<any, any>,
-  ) {
-    newMiddleware.piped = [
```
KATT, line 150: Missed this on the first go, but this is a big no-no.

We should never mutate objects that we don't have full control of within the library. This is mutating a user-passed-in function, so we'd be getting very hard-to-diagnose issues if/when someone reused a fn across several pipes

bautistaaa:
This is better

PR #3544, opened 2023-01-05
feat: #3543 websocket keep alive

### LINKED ISSUE #3543: feat: add keep alive (ping/pong) implementation to websocket handler

We are using long-living websocket connections and would like to keep them alive using websocket ping-pongs. 

I already have a working solution on my side, so it is just about the incorporation I guess.

However, I would modify the `applyWSSHandler` and add the ping/pong code there.

https://github.com/trpc/trpc/blob/81d85defcd58ea0d8941f1677028d0ed4250d398/packages/server/src/adapters/ws.ts#L113-L113

In the `WSSHandlerOptions` I would introduce a property `keepAliveMs?: number`. If unset, I would default to a keep-alive timeout of e.g. `20000`ms. This means that package upgrades on developer-side would automatically introduce this as an opt-in (is this regarded as a breaking change?). Additionally, you could also set the property to `0` to disable keep alive. One issue I am unsure about here is on how to expose this best to the various adapters (e.g. I am using the `fastifyTRPCPlugin`) which is the main reason why I would go for an opt-out to be honest.

Within the `applyWSSHandler` I would (if enabled) listen for new connections and set a interval for pings (set by `keepAliveMs`) and also a timeout to wait for pongs (this timeout would also be `keepAliveMs`). If no pong is received, the connection is terminated.

An alternate solution would be to implement keep alive frames within TRPC as operations itself.

However, native websocket ping/pongs are supported out-of-the box by websocket implementations as part of control frames (see RFC6455) and would not pollute the messages that are sent and received by TRPC itself.

### Additional information

This issue should only be the implementation of the ping/pong part on the server.

As written in the RFC (https://www.rfc-editor.org/rfc/rfc6455#section-5.5.2), client implemenations are required to respond to pings with a pong, so the implementation of the ping-part on the server using native websocket pings should be sufficient.

### 👨‍👧‍👦 Contributing

<sub>[TRP-42](https://linear.app/trpc/issue/TRP-42/feat-add-keep-alive-pingpong-implementation-to-websocket-handler)</sub>

## Funding

<a href="https://polar.sh/trpc/trpc/issues/3543">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/3543/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

nohum:
Closes #3543

## 🎯 Changes

Adds a websocket ping/pong as a feature. For further information please see the linked ticket. Currently, with this implementation, the default behaviour changes to use keep alive messages.

## ✅ Checklist

### DIFF
--- a/packages/server/src/adapters/ws.ts
+++ b/packages/server/src/adapters/ws.ts
@@ -103,6 +103,7 @@ export type WSSHandlerOptions<TRouter extends AnyRouter> = BaseHandlerOptions<
 > & {
   wss: ws.Server;
   process?: NodeJS.Process;
+  keepAliveMs?: number;
 } & NodeHTTPCreateContextOption<TRouter, IncomingMessage, ws>;
 
 export type CreateWSSContextFnOptions = NodeHTTPCreateContextFnOptions<
@@ -300,6 +301,26 @@ export function applyWSSHandler<TRouter extends AnyRouter>(
       }
     });
 
+    const keepAliveMs = opts.keepAliveMs || 20_000;
+    let pingInterval: NodeJS.Timer | null = null;
+    let pongWait: NodeJS.Timeout | null = null;
+    if (keepAliveMs > 0) {
+      pingInterval = setInterval(() => {
+        if (client.readyState === 1 /* ws.OPEN */) {
+          pongWait = setTimeout(() => client.terminate(), keepAliveMs);
+
+          client.once('pong', () => {
+            if (pongWait) {
+              clearTimeout(pongWait);
+              pongWait = null;
+            }
+          });
+
+          client.ping();
+        }
+      }, keepAliveMs);
+    }
+
     // WebSocket errors should be handled, as otherwise unhandled exceptions will crash Node.js.
     // This line was introduced after the following error brought down production systems:
     // "RangeError: Invalid WebSocket frame: RSV2 and RSV3 must be clear"
@@ -316,6 +337,12 @@ export function applyWSSHandler<TRouter extends AnyRouter>(
     });
 
     client.once('close', () => {
+      if (pongWait) {
+        clearTimeout(pongWait);
+      }
+      if (pingInterval) {
+        clearInterval(pingInterval);
+      }
       for (const sub of clientSubscriptions.values()) {
         sub.unsubscribe();
       }
--- a/examples/next-prisma-websockets-starter/src/server/prodServer.ts
+++ b/examples/next-prisma-websockets-starter/src/server/prodServer.ts
@@ -27,7 +27,13 @@ app.prepare().then(() => {
     handle(req, res, parsedUrl);
   });
   const wss = new ws.Server({ server });
-  const handler = applyWSSHandler({ wss, router: appRouter, createContext });
+  const handler = applyWSSHandler({
+    wss,
+    router: appRouter,
+    createContext,
+    // send websocket keep alive frames every 5 seconds, pass e.g. 0 to disable
+    keepAliveMs: 5000,
+  });
 
   process.on('SIGTERM', () => {
     console.log('SIGTERM');
--- a/examples/next-prisma-websockets-starter/src/server/wssDevServer.ts
+++ b/examples/next-prisma-websockets-starter/src/server/wssDevServer.ts
@@ -10,7 +10,13 @@ if (!global.fetch) {
 const wss = new ws.Server({
   port: 3001,
 });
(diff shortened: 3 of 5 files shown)


packages/server/src/adapters/ws.ts
```diff
@@ -300,6 +301,26 @@ export function applyWSSHandler<TRouter extends AnyRouter>(
...
+          client.once('pong', () => {
+            if (pongWait) {
+              clearTimeout(pongWait);
+              pongWait = null;
+            }
+          });
+
+          client.ping();
+        }
+      }, keepAliveMs);
```
KATT, line 321: I don't like `setInterval` with backpressure, wouldn't this e better implemented with a recursive `setInterval`? Then we could also measure time for each pog and use that for something later.


Mock code:

```ts

const pingPongInterval = 5_000;
const pingPongMaxLatency = pingPongInterval;

let pingTimeout = pingPongInterval ? setTimeout(ping, pingPongInterval) : null;
function ping() {
  const timeStart = Date.now();

  client.once('pong', onPong);
  const timeout = setTimeout(onTimeout, pingPongMaxLatency);

  function onPong() {
    // Clear up timeout so it doesn't terminate the connection
    clearTimeout(timeout);
    const latency = Date.now() - timeStart;
    const nextPing = pingPongInterval - latency;
    pingTimeout = setTimeout(ping, nextPing);
  }
  function onTimeout() {
    // Stop listening for pong to avoid memory leak
    client.off('pong', onPong);
    client.terminate();
  }
}

// in `client.once('close', () => { ... })`:
client.once('close', () => {
  // Clear up ping interval to avoid memory leaks
  pingTimeout && clearTimeout(pingTimeout);
  // [...]
});

```

KATT: Alternatively, we can do as they suggest in the docs? https://github.com/websockets/ws#how-to-detect-and-close-broken-connections

This would be a bit less resource intensive, I think?

KATT (changes requested):
I have a couple of questions

1. What happens if you **don't** do this ping/pong? When is a client terminated?
2. Does all browsers' WebSocket clients automatically respond with pong events as per the [RFC6455?](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.2)?
3. Can this not be done by the client pinging the server as well? Should we handle that?

Some blockers:

- Handle pings from clients?
- This needs testing

nohum:
Hi @KATT!

Regarding your questions:

> What happens if you don't do this ping/pong? When is a client terminated?

This depends on the connection timeouts that are enforced by either e.g. an ingress or proxy sitting in front of your application (proxy pass timeouts) or even routers that sit between the client and servers can enforce a custom timeout after no messages have sent for some time.

> Does all browsers' WebSocket clients automatically respond with pong events as per the [RFC6455?](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.2)?

That is something I honestly do not know, but I do know about Chrome and Firefox - in those it works with the submitted code since months. So basically the question boils down to: which other browsers/engines are relevant to check here?

> Can this not be done by the client pinging the server as well? Should we handle that?

This is something I had to lookup. Pings can also be [sent by the client](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers#pings_and_pongs_the_heartbeat_of_websockets), but according [to this ticket](https://www.w3.org/Bugs/Public/show_bug.cgi?id=13104), there is no ping() method in the web browser's WebSocket API according to the standard at the moment, leaving this to the browser internals to be handled. I took this from [here](https://stackoverflow.com/a/10586583).

> Some blockers:
>
> Handle pings from clients?

Pings from clients are handled automatically in the nodejs ws implementation, e.g. [as seen here](https://github.com/websockets/ws/blob/a3214d31b63acee8e31065be9f5ce3dd89203055/lib/websocket.js#L1189).

> This needs testing

I tried it with `wscat -P` (-P for showing ping/pong frames), though this should be "really" tested before merging. Do you have any recommendation on that?

packages/server/src/adapters/ws.ts
nohum: Oh thanks for the hint, I am going to take a look at that and try to come up with a better implementation.

KATT:
Stil missing tests @nohum, do you still need this feature?

nohum:
Hello @KATT, sorry not answering earlier. Yes, I would still be interested in that feature and will take a look in the next few days.

sachinraja:
please re-open with tests

Issue #3545, opened 2023-01-05
feat: Stronger typing for predicate invalidation


nicolassanmar:
### Context
I'm trying to invalidate a query based on some complex rules, as I don't want to invalidate every single call to that query.
The way to do this is using `.invalidate(undefined, {predicate: (query) })`, where predicate takes a function that dictates which queries should be invalidated.
The problem I have, is that the `query` has the default `react-query` type of `Query<unknown,unknown,unknown, unknown[]>` which is not true for our TRPC queries.

### Demonstration
``` typescript
const utils = trpc.useContext()
utils.router.query.invalidate(undefined, {
  predicate: (query: Query<unknown,unknown,unknown, unknown[]>) => {
    // We will have to cast `query` here to perform the invalidation based on the input data
    return true
  }
})
```

### Describe the solution you'd like to see

We know that if `invalidate()` is called on a query defined by the user, the typing will be:
``` typescript
type RouterInput= inferRouterInputs<AppRouter>
type RouterOutput = inferRouterOutputs<AppRouter>

// desired type of  example query called with `trpc.router.fn` , in predicate invalidation
type DesiredType = Query<
          RouterInput['router']['fn'],
          TRPCClientError<AppRouter>,
          RouterOutput['router']['fn'],
          [ [['router],['fn']], { input: RouterInput10['router']['fn'] } ]
>
```

If `invalidate()` is called at a router level, the type of the query, should be a union amongst the posible `DesiredType`'s of the procedures/routers inside it.

### Describe alternate solutions

Currently, I implemented a worse version of this to cast the type of the query, but its bad since it only supports queries and only exactly two levels for the path:

``` typescript
type QueryHelper<
    RouterOutput,
    K extends keyof RouterOutput & keyof RouterInput,
    K2 extends keyof RouterOutput[K] & keyof RouterInput[K],
  > = Query<
      RouterInput[K][K2],
      TRPCClientError<AppRouter>,
      RouterOutput[K][K2],
      [K, { input: RouterInput[K][K2] }]
 >

utils.router.query.invalidate(undefined, {
      predicate: (_query) => {
          const query = _query as unknown as QueryHelper<RouterOutput10, 'router', 'query'>
          return false
      }
})
```

### 👨‍👧‍👦 Contributing

- [x] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

JonParton:
Just a sign post to some old removed code that may help here for the router level stuff. 

[...]

Issue #3496, opened 2022-12-27
feat: v11 goals & suggestion box


KATT:
>  🚀  v11 is available to use and **ready for production**: https://trpc.io/docs/migrate-from-v10-to-v11

## Summary

* The main goal of v11 is to improve our own quality-of-life by **removing v9-support** as it makes the codebase very messy.
* Work will be done in the `next`\-branch
* Ideas welcome from the community of API-decisions you think we should revisit.
* We plan **not to delete** anything which hasn't previously been marked as `@deprecated` (if we rename, we'll do an alias)
* Discord chat: [https://discord.com/channels/867764511159091230/1057652120473575505](https://discord.com/channels/867764511159091230/1057652120473575505)

### Before official alpha

* [X] Delete v9 support #3494
  * [X] Recover test coverage from the above -- it dropped down to 93% -- see [https://app.codecov.io/gh/trpc/trpc/pull/3494](https://app.codecov.io/gh/trpc/trpc/pull/3494)
* [x] Deprecate and rename: Remove `Proxy` from naming of functions. We have `Untyped` that is for the untyped. e.g: `createTRPCProxyClient` -> `createTRPCClient` #4932
* [x] Look over all `@deprecated` and delete accordingly (#5026)
* [X] Move from using string paths to tuples in `react-query`
* [ ] Look if there are any open issues that we want to do that might entail breaking features
  * [ ] Consider if we wanna do them or wait for v12 - maybe we start exporting some stuff specifically for plugins
* [x] Turn `TRPCError.cause` to `unknown`
* [x] #5188 
* [x] Once we've created a `@trpc/core`-package, we need to look at extensions using tRPC and make a way to extend tRPC for libraries - #5308
* [x] #5210
* [ ] Make sure things that are exported in `@trpc/client` / `@trpc/server` etc are smurfed. Stuff in `@trpc/core` don't need to be smurfed as it's all internal and not intended to be used directly
* [ ] Make a blog post
* [ ] Make a migration guide
* [x] Version Docusaurus docs
* [x] Make sure all docs installs uses `@next` 
* [x] Swap default branch to `next`

### Beta

* [ ] Rename `main` to `10.x` & `next` -> `main`

## Maybes of things to do

* [ ] Move from using string paths to tuples in `client` (maybe?)
* [ ] Search for `FIXME` in codebase and fix them
* [ ] #5004
* [ ] #1963

#### Subtle breaking changes

[...]

PR #3547, opened 2023-01-05
feat: add precise type to `predicate` in `invalidate`

SrBrahma:
Re https://github.com/trpc/trpc/issues/3496#issuecomment-1369238236
Closes https://github.com/trpc/trpc/issues/3545

## 🎯 Changes

Adds the corresponding input type to `predicate` in `invalidate` on `trpc.useContext().router.procedure`.

## Before
image

## After
image

Notice that the `queryKey` has the right type in [0] and [1]. `type` can be `query` or `infinite`, depending on the procedure.

* For reference, this is how the procedure is defined:

image

## ✅ Checklist


packages/react-query/src/internals/getArrayQueryKey.ts
```diff
@@ -5,6 +5,11 @@ export type QueryKey = [
   { input?: unknown; type?: Exclude<QueryType, 'any'> }?,
 ];
 
+export type QueryKeyKnown<TInput, TType extends Exclude<QueryType, 'any'>> = [
+  string[],
+  { input: TInput; type: TType },
+];
```
SrBrahma, line 11: I had to add this as `QueryKey` is not generic and I wasn't able to make it generic as it considers the second array item to be optional. Changes on it are welcome!

SrBrahma:
It seems that the queryKey[1] can be undefined if the query has `enabled=false`. Is this true? If so, I need to add `| undefined` to queryKey[1].

nicolassanmar:
As the creator of https://github.com/trpc/trpc/issues/3545 this should implement a solution to it!

KATT (commented):
@juliusmarminge can you review this please? you're the master on this :D

juliusmarminge (approved):
Looks good but I think we can clean it up a bit on `next` where queryKeys aren't this messy...

SrBrahma:
@juliusmarminge before / if you are going to merge it, I suggest you read this discussion I participated in react-query after I made this PR:

https://github.com/TanStack/query/discussions/4762#discussioncomment-4605555

Tl;dr: I don't know if this PR is good as if a query is enabled=false, the query params may have invalid values types.

Maybe we could have the type to be something like `{ enabled: false, queryKey: [x, any] } | { enabled?: true, queryKey: [x, theRightType] }`?

While this PR works, it may be dangerous if enabled is false as it may potentially lead to invalid data access.

juliusmarminge:
> Maybe we could have the type to be something like ~ { enabled: false, queryKey: [x, unknown] } | { enabled?: true, queryKey: [x, theRightType] }, so the user will need to check for enabled === undefined || enabled === true to then use the data with the right type?

This wont matter since the input type will be a DeepPartial so you have to check for every prop you try to access no matter what

juliusmarminge (dismissed):

nicolassanmar:
@juliusmarminge @SrBrahma 
> While this PR works, it may be dangerous if enabled is false as it may potentially lead to invalid data access.

I think that typing the predicate still makes sense, as it will be right in regular use cases. People can only find data in the cache with unexpected typings if:
1. If they tell TS to ignore a typing error (for example, a non null assertion, even when `enabled: false`).
1. If they tampered with the react-query cache in some way

Somewhere in the documentation should state that the typing is only true if they did not do any of the above.

juliusmarminge (approved):
I think this is fine. Since the type is a deep partial, users will still need to check for nested property accesses so that shouldn't be an issue, no matter if `enabled: false` is used or not(?).

Maybe @TkDodo has some insights too?

juliusmarminge (approved):

TkDodo:
> I think this is fine. Since the type is a deep partial, users will still need to check for nested property accesses so that shouldn't be an issue, no matter if enabled: false is used or not(?).

> Maybe @TkDodo has some insights too?

I re-checked the linked react-query discussion, and I think I understand the issue now better. Note that queries where all observers are disabled are explicitly exempt from being refetched, but the predicate filtering is still applied first:

- https://github.com/TanStack/query/blob/765f6196e900c1dd826b040773fcf97839ca60d0/packages/query-core/src/queryClient.ts#L253-L256

If the type will be deep-partial in the future, I think this will side-step the issue 👍

KATT (approved):

PR #3622, opened 2023-01-15
chore: rm some `@deprecate`d stuff

KATT:
#3496

### DIFF
--- a/packages/react-query/src/shared/index.ts
+++ b/packages/react-query/src/shared/index.ts
@@ -11,18 +11,4 @@ export * from './queryClient';
 export * from './types';
 export * from './hooks/types';
 
-export {
-  /**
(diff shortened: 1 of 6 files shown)


packages/react-query/src/shared/index.ts
```diff
@@ -11,18 +11,4 @@ export * from './queryClient';
 export * from './types';
 export * from './hooks/types';
 
-export {
-  /**
-   * @deprecated this is an internal function
-   */
-  getClientArgs,
-} from '../internals/getClientArgs';
```
KATT, line 19: cc @juliusmarminge we don't need this exported right?

juliusmarminge: I dont think so?

PR #3645, opened 2023-01-19
feat(react-query): add initialCursor option to useInfiniteQuery

### LINKED ISSUE #3638: feat: add default cursor value for useInfiniteQuery

I want to be able to pass a default cursor value to useInfiniteQuery. In some case I can't have a default cursor value on the server side.

```tsx
const {
    data,
    fetchNextPage,
    hasNextPage,
  } = trpc.getUsersByIds.useInfiniteQuery(
    {
      cursor: ids.slice(0, 100),
    },
    {
      refetchOnWindowFocus: false,
      keepPreviousData: true,
      getNextPageParam: lastPage => {
        const lastIndex = ids.indexOf(lastPage[lastPage.length - 1].id)

        if (lastIndex === ids.length - 1) return undefined

        return ids.slice(lastIndex + 1, lastIndex + 100)
      },
    }
  )
```

```tsx
const {
    data,
    fetchNextPage,
    hasNextPage,
  } = trpc.getUsersByIds.useInfiniteQuery(
    undefined,
    {
      refetchOnWindowFocus: false,
      keepPreviousData: true,
      initialCursor: ids.slice(0, 100),
      getNextPageParam: lastPage => {
        const lastIndex = ids.indexOf(lastPage[lastPage.length - 1].id)

        if (lastIndex === ids.length - 1) return undefined

        return ids.slice(lastIndex + 1, lastIndex + 100)
      },
    }
  )
```

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

maloguertin:
Closes #3638

## 🎯 Changes

Add a initialCursor option to useInfiniteQuery hook .

Added the doc directly in the example:

Screenshot from 2023-01-19 09-44-20

## ✅ Checklist


maloguertin:
I went with the initialCursor option approach but I can change it to this if you guys would like:

```tsx
trpc.infinitePosts.useInfiniteQuery({ cursor: 2 }, {
  getNextPageParams: (lastPage) => lastPage.nextCursor
})
```

I chose the option approach because I felt like it was less obvious when doing it in the input that the subsequent queries would override the cursor.

juliusmarminge:
> @KATT @juliusmarminge is this something that could be interesting?

I'm not a heavy user of infiniteQuery so I don't really know the ins and outs of the API - but to my knowledge the cursor isn't passed as input right? What happens if you just set a default on the server? Will that cause the "second fetch" (i.e. when the first "getNextPage" request is sent) to go out of sync?

```ts
  list: publicProcedure
    .input(
      z.object({
        limit: z.number().min(1).max(100).nullish(),
        cursor: z.string().nullish(),
      }),
    )
    .query(async ({ input }) => {
      const limit = input.limit ?? 50;
      const cursor = input.cursor ?? 100; // <-- fallback here

      const items = await prisma.post.findMany({
        select: defaultPostSelect,
        // get an extra item at the end which we'll use as next cursor
        take: limit + 1,
        where: {},
        cursor: { id: cursor },
      });

      let nextCursor: typeof cursor | undefined = undefined;
      if (items.length > limit) {
        // Remove the last item and use it as next cursor

        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
        const nextItem = items.pop()!;
        nextCursor = nextItem.id;
      }

      return {
        items,
        nextCursor,
      };
    }),
```

To me, it seems like your behavior can already be achieved like this? Or would you want the control on the frontend for each call? Can you not just accept it as input in that case?

```ts
  list: publicProcedure
    .input(
      z.object({
        limit: z.number().min(1).max(100).nullish(),
        cursor: z.string().nullish(),
        initialCursor: z.string().optional(),
      }),
    )
    .query(async ({ input }) => {
      const limit = input.limit ?? 50;
      const cursor = input.cursor ?? input.initialCursor ?? 100; // <-- fallback here
      // ...
```

Like I said, I'm not a power user of the infinite query API so might be missing something but seems like this can be done already?

maloguertin:
I think my first main concern is the fact that tanstack's react-query support it and even have it in their doc:

https://tanstack.com/query/latest/docs/react/reference/useInfiniteQuery?from=reactQueryV3&original=https%3A%2F%2Freact-query-v3.tanstack.com%2Freference%2FuseInfiniteQuery

specifically: `({ pageParam = 1 })`

The work around you suggest can work but in the use case I have in my project my cursor is an array of IDs. If I define my initial cursor as an array of let's say 100 ids I would not want them to be passed on subsequent queries since it makes the query input really bulky. I could get around it with another work around like:

```tsx
trpc.getPosts.infiniteQuery({
  initialCursor: isInitialQuery ? initialIds : undefined // <--- where isInitialQuery is computed with a useEffect or something 
})
```

But that seems like a lot of work arounds for something that react-query offers out of the box!

> Will that cause the "second fetch" (i.e. when the first "getNextPage" request is sent) to go out of sync? I think you can work > > around that by doing something like this:

That shouldn't be a problem because the next query will have a cursor value that will be computed from the correct cursor as long as you compute your `getNextPageParam` correctly.

The way I see it: the server can define a default value if no cursor is passed but I still want to easily allow a component to chose from what cursor to start querying.

Example use case: you have a virtual infinite list in a page that accepts a `?scrollToItem=150` search param so on the first load you would want to query from `cursor=150`. 

Cheers!

juliusmarminge:
I see - thanks for clarifying

www/docs/reactjs/useInfiniteQuery.md
```diff
@@ -69,6 +69,7 @@ export function MyComponent() {
     },
     {
       getNextPageParam: (lastPage) => lastPage.nextCursor,
+      initialCursor: 1, // <-- optional you can pass an initialCursor
```
KATT, line 72:
```suggestion
      // initialCursor: 1, // <-- optional you can pass an initialCursor
```

KATT (approved):

PR #3646, opened 2023-01-19
feat(react-query): Generate polymorphic route/mutation/query interfaces from tRPC router

### LINKED ISSUE #3627: feat: Polymorphic tRPC routes in react client

When defining routers which have common interfaces, it can be helpful to also define common UI which just needs to be informed about the particular route it will be using. For instance:

```tsx
// Route structure like: { foo: { create, update, list, delete }, bar: { create, update, list, delete } }
const trpc = createTRPCReact 

type CrudRoute = typeof trpc.bar

function MyComponent({ route }: { route: CrudRoute }) {
  const createMutation = route.create.useMutation()
  const listQuery = route.delete.useQuery({ /* paging params */ })

  return /* ui */
}

function MyScreen() {
  // We then pass a route which matches the interface but isn't the one the type was derived from
  // Error because the query key does not match!
  return <MyComponent route={trpc.foo} />
}
```

Currently for even the most basic example, this doesn't work because `DecoratedProcedureRecord` calculates `TPath` based on the full path to a procedure name and down the type chain into react query's `TQueryKey` on `QueryObserverOptions`. TPath/TQueryKey is important type information of course.

It would be helpful to have some type utilities provided which can extract the abstract type information from a Route/Procedure and produce a type which can be used for polymorphism. The design of this I'm not too sure as fairly complex cases might exist where parameter/return types may vary slightly but have some common underlying interface.

A proof of concept for the simple use case (where the underlying interface of the procedures is identical) looks like this:

```tsx
import { AnyProcedure, AnyRouter, ProcedureRouterRecord } from '@trpc/server'
import {
  DecoratedProcedureRecord,
  DecorateProcedure,
} from '@trpc/react-query/dist/shared'

type AbstractDecoratedProcedureRecord<
  TProcedures extends ProcedureRouterRecord,
  TFlags
> = {
  [TKey in keyof TProcedures]: TProcedures[TKey] extends AnyRouter
    ? AbstractDecoratedProcedureRecord<
        TProcedures[TKey]['_def']['record'],
        TFlags
      >
    : TProcedures[TKey] extends AnyProcedure
    ? DecorateProcedure<TProcedures[TKey], TFlags, any>
    : never
}

/**
 * Given a tRPC react route, calculates the abstract type so that similar interfaces can be passed polymorphically
 *
 * Ie. For: `trpc.routeA.routeB = { procA, procB }`
 * Usage: `AbstractProcedureRecord<typeof trpc.routeA.routeB>`
 */
export type AbstractProcedureRecord<
  T extends DecoratedProcedureRecord<any, any, any>
> = T extends DecoratedProcedureRecord<infer TProcedures, infer TFlags, any>
  ? AbstractDecoratedProcedureRecord<TProcedures, TFlags>
  : never

//
// Usage
//

import { trpc } from './react-client/api'

export type RouteDef = BaseProcedureRecord<
  typeof trpc.foo
>

// Now any matching interface can be assigned polymorphically
const r1: RouteDef = trpc.foo
const r2: RouteDef = trpc.bar
```

There is a lot more to this, the above solution may not work where routes don't have the exact same interface. The full solution might involve taking the recursive union of an arbitrary number of route types which could get hairy

```
type AbstractProcedureRecord<T1, T2, T3, ...TEtc> = /* calculate the common abstract type */
```

Or less hackily, maybe a way to actually define an interface which a Router must implement on the API side, and then provide an API to use these interfaces alongside the React Client library

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

Nick-Lucas:
This adds some helper types under `react-query/src/shared/polymorphism`, primarily `RouterLike<TProcedure>`, `MutationLike<TProcedure>` and `QueryLike<TProcedure>`, and documents a factory pattern for generating routes which share common functionality across several sub-routes.

The goal is to provide a 1st class API for passing around a factory-built route to React components, so we can build generic components with a variable tRPC route. I'm referring to it as polymorphism but there might be a better name here.

Closes #3627

## 🎯 Changes

## ✅ Checklist


KATT (commented):
Cool!

- I think the QueryLike and MutationLike types are useful. They should be exported in `@trpc/react-query/shared` 
- Documentation should be added in [here](https://trpc.io/docs/infer-types#infer-react-query-options-based-on-your-router). I think it can be quite light with a link to the test you're making as a reference.
- I'd like the test, since it might also act as documentation, should import more comments

www/docs/server/infer-types.md
```diff
@@ -171,3 +171,99 @@ function usePostById(input: PostByIdInput, options?: PostByIdOptions) {
   return trpc.post.byId.useQuery(input, options);
```
Nick-Lucas, line 171: Since this is part of the react-query library, should we actually move these additions to a new file under the react docs? Perhaps `www/docs/react/infer-types.md`?

Nick-Lucas:
Thanks for the initial feedback, I believe that's all covered now. I've added some higher level types and the docs, plus had a stab at finding the right home for a couple types.

Once we're happy with this, I do think it will be worth it to write a test with a router factory which can utilise sub-types, and adapt the polymorphic types to work in that case. Having a generic (in the true OO sense) react component which can accept a sub-type and some inversion-of-control for the extra fields seems like the natural continuation of this work, and pretty important if a use case is to include a CRUD factory where the entities are all based off a core type like `{ id, name, createdAt, createdBy, modifiedAt, modifiedBy }`

Nick-Lucas:
Okay, I've added a second example today to check support for sub-types in both the input/output data, and in extending the router beyond the initial factory result.

I believe this is ready now @KATT though obviously it's changed significantly since you last reviewed so I'm expecting some change requests :)

KATT (commented):
I really like this, but I want a second pair of eyes on it before merging. 

can any of you guys check it, @trpc/trpc-core and smash the merge button?

packages/react-query/src/shared/polymorphism/mutationLike.ts
```diff
@@ -0,0 +1,31 @@
+import {
+  InferMutationOptions,
+  InferMutationResult,
+} from '@trpc/react-query/utils/inferReactQueryProcedure';
```
sachinraja, line 4:
```suggestion
import {
  InferMutationOptions,
  InferMutationResult,
} from '../../utils/inferReactQueryProcedure';
```

packages/react-query/src/shared/polymorphism/queryLike.ts
```diff
@@ -0,0 +1,32 @@
+import {
+  InferQueryOptions,
+  InferQueryResult,
+} from '@trpc/react-query/utils/inferReactQueryProcedure';
```
sachinraja, line 4:
```suggestion
import {
  InferQueryOptions,
  InferQueryResult,
} from '../../utils/inferReactQueryProcedure';
```

sachinraja (changes requested):

sachinraja (approved):

Nick-Lucas:
Thanks @sachinraja, looks like prettier wanted the imports in a different order so I've pushed that. Also rebased to get things up to date

KATT (approved):

sachinraja (approved):
I have some reservations about merging this, primarily because those tests look type-heavy but I don't know when I'll get to addressing those so this is fine to merge now. The types themselves look good and I can think of valid use cases for them.

Nick-Lucas:
> I have some reservations about merging this, primarily because those tests look type-heavy but I don't know when I'll get to addressing those so this is fine to merge now. The types themselves look good and I can think of valid use cases for them.

Thanks for getting this over the line!

Yes the tests ended up pretty substantial. We can probably almost half the test code (and move code from the sister files into the spec) just by jettisoning the simpler example, but since the goal was for them also to serve as an example linked in the docs it might be worth doing that at the same time as standing up an example project for the factory pattern.

PR #3664, opened 2023-01-23
feat: allow fuzzy invalidation

### LINKED ISSUE #3648: feat: query invalidation should take `Partial<Type>` instead of `Type` as the input

`react-query` matches query invalidation in a fuzzy way, and allows to invalidate queries without having the whole input used to call it.

From this comment on the react query forums: https://github.com/TanStack/query/discussions/4762#discussioncomment-4649876
> from a pure react-query perspective, if your `queryKey` is:
> 
> ```
> ['my-path', { a: 'b', foo: 'bar' }]
> ['my-path', { a: 'b', something: 'else' }]
> ```
> 
> it should be possible to target them both for query invalidation by doing:
> 
> ```
> queryClient.invalidateQueries({
>   queryKey: ['my-path', { a: 'b' }]
> })
> ```
> 
> because that does fuzzily match them both. It might be that trpc doesn't have a concept for that though ...

Trpc does not currently allow to match queries this way.

Having a query with the input type type:

``` typescript
type TInput = { id: string, name: string, lastName: string | undefined } 
```
Even though we call the query with the full TInput, the invalidation should allow to do this:
``` typescript
query.invalidate({id: 'id'})
```
This should be easy to do, as `react-query` does already behave like that, it's just a type issue. It should be a deep partialization of it.

Making it `Partial<TInput>` would solve some problems, but I believe `react-query` will match everything fuzzily so the deep partialization is better.

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

nicolassanmar:
Closes #3648

## 🎯 Changes

Allows using the `.invalidate` function with partial types for the input. This allows to fuzzily match the queries when not using `exact: true` as an option.

## ✅ Checklist


packages/react-query/src/shared/proxy/utilsProxy.ts
```diff
@@ -32,6 +32,12 @@ import {
 import { TRPCContextState } from '../../internals/context';
 import { getQueryKey } from '../../internals/getQueryKey';
 
+type DeepPartial<TObject> = TObject extends object
+  ? {
+      [P in keyof TObject]?: DeepPartial<TObject[P]>;
+    }
+  : TObject;
+
```
nicolassanmar, line 40: Where would you put this utility type?

juliusmarminge: I think we have most in trpc/server in some types.ts file

nicolassanmar: Thanks! I moved the util type there

bautistaaa:
should add a test for this

juliusmarminge:
> should add a test for this

Yea add a test where the filter isnt exact and make sure its ok for TS + that it invalidates properly 👍👍

KATT (changes requested):
Please add a test so we don't accidentally break this in the future :)

juliusmarminge (approved):

KATT (approved):

KATT (approved):

PR #3675, opened 2023-01-25
refactor: remove GnClass workaround

ggrandi:
## 🎯 Changes

removes the workaround in `./packages/react-query/src/shared/hooks/deprecated/createHooksInternal.tsx` to pass generics into `__createHooksInternal` by using [instantiation expressions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#instantiation-expressions) however this requires typescript 4.7

## ✅ Checklist


KATT:
Unsure if this will break codesandbox or stackblitz which I wouldn't like. We're also reworking this in the `next` branch and should focus there instead

KATT:
Closing as we should instead get rid of this on the `next` branch

PR #3678, opened 2023-01-26
feat(docs): adds note about query client memo

ixahmedxi:
Closes #

## 🎯 Changes

## ✅ Checklist


sachinraja:
I think the main point that needs to be clarified here is actually why we create these clients in the components vs. outside them. The reason for that is to ensure each request gets a unique client so request caches aren't shared between them.

www/docs/reactjs/introduction.mdx
```diff
@@ -161,6 +161,10 @@ export function App() {
 }
 ```
 
+:::note
+The usage of a useState for the creation of the query clients is there to prevent the client from being recreated on every render. This is exactly the same as using the `useMemo` hook.
+:::
+
```
a-eid, line 167: that's not quit correct as React preserves the right to forget memoization result from `useMemo` and call it again any time, which is not the case for `useState`. I'd say it's dangerous to use useMemo is such context.

KATT (changes requested):
Follow what @sachinraja said 🙏

ixahmedxi:
@sachinraja @KATT not sure if this is what you guys exactly mean as that i've heard Alex say that it is only for ssr purposes as that it would reuse the same cache for all the clients connected.

www/docs/reactjs/introduction.mdx
```diff
@@ -161,6 +161,10 @@ export function App() {
 }
 ```
 
+:::note
+The reason behind the useState calls for the creation of the queryClient and the trpcClient in the example above as opposed to declaring them outside of the component is to ensure that each request gets a unique client while using SSR so that request caches aren't shared between them.
```
KATT, line 165: add backticks pls

```suggestion
The reason behind the `useState` calls for the creation of the `queryClient` and the `TRPCClient` in the example above as opposed to declaring them outside of the component is to ensure that each request gets a unique client while using SSR so that request caches aren't shared between them.
```

KATT (approved):

PR #3688, opened 2023-01-28
chore: replace Jest by Vitest

QuiiBz:
## 🎯 Changes

This PR replaces Jest by Vitest, for faster tests execution.

- `jest.*` is replaced by `vi.*`
- Remove `jest.retryTimes(x)` because it seems not supported by Vitest
- Set coverage via C8
- Add a new `test-ui` script that launches Vitest UI

### Comparison
I'm using a base MacBook Pro M1 Pro.

Running all tests with **Jest** takes 38 seconds:

Screenshot 2023-01-28 at 18 38 30

Running all tests with **Vitest** takes 20 seconds:

Screenshot 2023-01-29 at 20 00 06

## ✅ Checklist


juliusmarminge:
Did you get this working? I've been meaning to do this but was waiting for https://github.com/prisma-labs/konn/issues/30

QuiiBz:
It's currently working, yeah. I'm not sure to understand what was the blocker, as I didn't have to do anything specific to make it work.

package.json
```diff
@@ -64,7 +68,10 @@
     "ts-prune": "^0.10.3",
     "tsx": "^3.9.0",
     "turbo": "^1.6.3",
-    "typescript": "^4.8.3"
+    "typescript": "^4.8.3",
+    "vitest": "^0.23.4",
```
juliusmarminge, line 72: Shouldn't all these deps be in packages/test?

juliusmarminge:
> It's currently working, yeah. I'm not sure to understand what was the blocker, as I didn't have to do anything specific to make it work.

Awesome! What's the difference for watch? I think that might be even bigger gain that just a dry run?

QuiiBz:
> Awesome! What's the difference for watch? I think that might be even bigger gain that just a dry run?

Updating a single simple test while running `test-watch` in the background takes ~1s with Vitest, ~3s with Jest!

bautistaaa:
tom this is so awesome

package.json
QuiiBz: I'm adding back Vitest to the root `package.json`, because the types for the globals (`vi.mock`, `vi.fn`, ...) need to be declared inside `tsconfig.json` (either at the root, or inside each package - but that would mean duplication)

juliusmarminge:
Sad that it's still slow in CI - really thought this would help here :( 
CleanShot 2023-01-29 at 20 20 08

QuiiBz:
After testing a bit, it appears that using Istanbul instead of c8 for coverage (which is the default in Vitest) is significantly faster. I've updated the files to collect coverage from, to match the current Jest config - it which should also speed up the CI

KATT (dismissed):
This is so awesome @QuiiBz, (I sent you some money as a thank yo)

The only things I am missing is:

- Updated contributor docs
- Codecov https://github.com/trpc/trpc/blob/19fafa1775d3d07c5c7dfc734c2fc57ee8f8b637/.github/workflows/main.yml#L58-L61

QuiiBz:
Ooops, I updated my fork and it deleted all the changes... I'm dumb, let me re-create the PR

packages/client/package.json
```diff
@@ -86,5 +86,8 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "dependencies": {
+    "node-fetch": "^2.6.1"
```
juliusmarminge, line 91: move to devDep

juliusmarminge (commented):
Whats up with all the imports? Why has /index been added to everything?

packages/client/package.json
KATT: Shouldn't be needed here at all? This is only a concern of the test package AFAIK

QuiiBz: Moving it to `devDeps` but I can't remove it from this package because it's used inside `types.test.ts`:
 
Screenshot 2023-02-04 at 08 31 25

QuiiBz:
> Whats up with all the imports? Why has /index been added to everything?

Vitest didn't resolve the paths correctly, as it used the `exports` field from `package.json` and paths like `@trpc/server/src/transformers` aren't exported in `exports`.

To fix that, I had to manually add aliases inside Vitest's config. I've just pushed an update to these aliases that should remove the `index` imports.

packages/tests/vitest.config.ts
```diff
@@ -0,0 +1,115 @@
...
+        __dirname,
+        '../client/src/internals/dataLoader.ts',
+      ),
+      '@trpc/client/src/links/retryLink': join(
+        __dirname,
+        '../client/src/links/retryLink.ts',
+      ),
+      '@trpc/client/src': join(__dirname, '../client/src/index.ts'),
+    },
+  },
```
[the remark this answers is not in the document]
QuiiBz: Agree, I couldn't find a way to make Vite ignore the `exports` field from tRPC packages :(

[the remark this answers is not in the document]
sachinraja: fixed

packages/tests/package.json
```diff
@@ -6,9 +6,14 @@
     "codegen:bigBoi": "tsx ../../scripts/generateBigBoi.ts",
     "codegen:bigLegacyRouter": "tsx ../../scripts/generateBigLegacyRouter.ts",
     "lint": "eslint --ext .ts,.tsx --ignore-path ../../.gitignore --report-unused-disable-directives .",
-    "test": "jest",
-    "test-watch": "jest --watch",
-    "test-ci": "jest --ci --maxWorkers=2 --coverage"
+    "test": "concurrently -c \"green,blue\" npm:test-run:*",
+    "test-run:vitest": "vitest --run",
+    "test-run:tsc": "tsc --noEmit --pretty",
```
juliusmarminge, line 11: Why not setup typechecking via Vitest: https://vitest.dev/config/#typecheck

sachinraja: I tried but it doesn't work and doesn't provide much value since you still have to run it as a separate command - `vitest typecheck`

packages/tests/server/caller.test.ts
```diff
@@ -19,7 +19,7 @@ test('undefined input query', async () => {
   const caller = router.createCaller({});
   const result = await caller.hello();
 
-  expectTypeOf<string>(result);
```
QuiiBz, line 22: It should still be a `string` here @sachinraja, right?

https://github.com/trpc/trpc/actions/runs/4217420038/jobs/7321186668#step:9:126

sachinraja: ah right, leftover from my testing

sachinraja (approved):

PR #3692, opened 2023-01-30
feat(`server`): add configurable namespace delimiter for router merging.

### LINKED ISSUE #3537: feat: Configurable delimiter for `namespace.procedure` combo (router merging)

I'd like to be able to set the delimiter for the namespace and procedure.

Ability to pass a delimiter other than `.`

Generating a path mapping for each `namespace.procedure` combo explicitly from the router configuration.

OR

Using a custom link that rewrites the request (did not try, just theoretical) 

### Additional information

Use case:

AWS API Gateway has a limitation on route definitions.

Only the following options are available ([docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-routes.html?icmpid=apigateway_console_help#http-api-develop-routes.evaluation)):

```
GET /pets/dog/1
GET /pets/dog/{id}
GET /pets/{proxy+}
ANY /{proxy+}
```

I want to separate the routes by router namespace, which would require a route like this:

```
GET /pets.{procedure}
```

But, unfortunately, AWS does not support the `.` in its router path config.

An error is thrown:

> One or more path parts appear to be invalid. Parts are validated against this regular expression: `^(\{[\w.:-]+\+?\}|[a-zA-Z0-9.:_-]+)$`

A solution would be to reconfigure tRPC to use a `/` as a delimiter.

```
GET /pets/{procedure}
```

I believe it is configured here:

https://github.com/trpc/trpc/blob/81d85defcd58ea0d8941f1677028d0ed4250d398/packages/server/src/core/router.ts#L254

### 👨‍👧‍👦 Contributing

<sub>[T-79](https://linear.app/trpc/issue/T-79/feat-configurable-delimiter-for-namespaceprocedure-combo-router)</sub>

## Funding

<a href="https://polar.sh/trpc/trpc/issues/3537">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/3537/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

igalklebanov:
Closes #3537

## 🎯 Changes

This PR tries adding the ability to configure a custom namespace (& procedure) delimiter instead of the default `.` (which is not allowed by AWS).

## ✅ Checklist


packages/server/src/deprecated/interop.ts
```diff
@@ -145,28 +150,31 @@ export type MigrateRouter<
...
-        TQueries,
-        TMutations,
-        TSubscriptions,
-        TErrorShape,
-        Transformer
-      >
-    : never;
+export type MigrateOldRouter<
+  TRouter extends AnyOldRouter,
+  TNamespaceDelimiter extends string = '.',
```
igalklebanov, line 155: This is all I added in this block, prettier has a different mood nowadays I guess.

KATT (commented):
This looks really solid. Feel free to ignore "deprecated" stuff altogether if you want.

I guess the client also needs updating so we can configure it there (ideally enforced my typing to match like we have done with transformers)

igalklebanov:
> This looks really solid.

🍻 

> Feel free to ignore "deprecated" stuff altogether if you want.

It screamed at me, I panicked! 😆 

> I guess the client also needs updating so we can configure it there

It sure does, but yesterday's Igal needed some sleep. 😴 
Hope I covered all the places..

> (ideally enforced my typing to match like we have done with transformers)

Yeah I just mimmicked the transformer stuff, hope the intersection was the right way to add the new stuff.

igalklebanov:
I'll add documentation tomorrow, good night! 💤

packages/client/src/internals/TRPCUntypedClient.ts
```diff
@@ -52,7 +54,28 @@ type CreateTRPCClientBaseOptions<TRouter extends AnyRouter> =
...
+    (TRouter['_def']['_config']['namespaceDelimiter'] extends DefaultNamespaceDelimiter
+      ? {
+          /**
+           * The delimiter used to separate namespaces and procedure names in router paths
+           * of child routers.
+           *
+           * You must use the same namespaceDelimiter on the backend and frontend
+           * @link https://trpc.io/docs/merging-routers
+           */
+          namespaceDelimiter?: 'You must set a namespaceDelimiter on the backend router';
```
KATT, line 67:
```suggestion
          namespaceDelimiter?: TRouter['_def']['_config']['namespaceDelimiter'];
```

KATT (dismissed):
LGTM but failing CI, please dismiss my rejection and re-request review once you've had a look - @ me on discord if you need a hand

packages/client/src/internals/TRPCUntypedClient.ts
igalklebanov: > but failing CI

Oops, forgot to add the new prop to links tests.

---

Regarding the suggestion, good catch! 

Shouldn't it be just `DefaultNamespaceDelimiter` in this case?

igalklebanov:
> LGTM but failing CI, please dismiss my rejection and re-request review once you've had a look - @ me on discord if you need a hand

Oops, forgot to add the new prop to links tests. 😢

Pushed some changes that should resolve that.

packages/tests/server/initTRPC.test.ts
```diff
@@ -39,6 +46,42 @@ test('custom transformer', () => {
...
+
+  expectTypeOf(
+    t._config.namespaceDelimiter,
+  ).toMatchTypeOf<DefaultNamespaceDelimiter>();
+  expectTypeOf(
+    router._def._config.namespaceDelimiter,
+  ).toMatchTypeOf<DefaultNamespaceDelimiter>();
+});
+
+test('custom namespace delimiter', () => {
```
sachinraja, line 69: we don't need this test here since it's the same as the one in `namespaceDelimiter.test.ts`

igalklebanov: Thank you! 🙏 

Agree, redundant after the suite was added.

Removed in latest commit.

www/docs/server/merging-routers.md
```diff
@@ -66,6 +66,69 @@ export const userRouter = router({
...
+    }),
+  ],
+  namespaceDelimiter: '/', // this has to match the delimiter on the server
+});
+
+const users = await client.user.list.query(); // => [];
+// underlying path is "user/list"
+```
+
+**Note**: Choosing a custom namespace delimiter is a must when using AWS HTTP APIs since dots are not allowed in paths.
```
Nick-Lucas, line 130: I don't think this is true? I have hosted my tRPC backend on both API Gateway "REST" and "HTTP" to front a lambda (we had to move between for infra reasons) and didn't have any problems with dots.

If this statement is true there must be some more nuanced information which is being omitted, or the name "HTTP APIs" is confusing 🤔

Nick-Lucas: _right_ #3537 is some interesting nuance :) 

The thing I don't understand in that example is _why_ you would want to reproduce the routes into API Gateway, I just have a simple proxy set up and it works flawlessly. Best case doing this would break batching. 

Anyway I'm a bit off track, I wonder if what this topic really needs is an AWS Serverless hosting doc which covers the nuances and recommendations in depth? This statement as it's written has the chance to be quite misleading to readers :)

igalklebanov: Removed. :)

igalklebanov: > The thing I don't understand in that example is why you would want to reproduce the routes into API Gateway, I just have a simple proxy set up and it works flawlessly. Best case doing this would break batching.

I guess some might want more fine-grained control over memory size, provisioning, etc. and separate monitoring and logs by having lambdas dedicated to specific paths.

Nick-Lucas:
A couple questions

* How does this work with batching exactly? Do you get paths like `/trpc/foo/bar,/foo/baz?batch=1` and that works okay even when going through web servers like ApiGateway, nginx, IIS?
* Is it worth making clearer that it will only affect the http comms and won't affect the proxy client?

igalklebanov:
Sorry for the delay, was very busy.

> Do you get paths like /trpc/foo/bar,/foo/baz?batch=1

Given a router as follows: 

```ts
{
  foo: {
    bar: () => {},
    baz: () => {},
  },
}
```

and calls as follows:

```ts
await Promise.all([{
 client.foo.bar.query(),
 client.foo.baz.query(),
}]);
```

The path should be: `/foo{namespaceDelimiter}bar,foo{namespaceDelimiter}baz?batch=1`.

> and that works okay even when going through web servers like ApiGateway, nginx, IIS?

I've tried using the lambda + API gateway example. Seemed OK, offline at least.
Will try to deploy that example and test for real.

I don't have much experience setting up NGINX or IIS and we don't seem to have these in anywhere in the project.
..so i'll need to free up some time and read some stuff and play around with it.

> Are there any invalid characters which would do harm or have encoding issues over a http connection?

`,` would break batching probably.
Other than that, can follow [these](https://stackoverflow.com/questions/1547899/which-characters-make-a-url-invalid) comments and compile/run-time validate against a character set OR use native URL.

> Is it worth making clearer that it will only affect the http comms and won't affect the proxy client? `httpNamespaceDelimited` for the variable for instance?

I'm all for explicit naming. I'm not sure `httpNamespaceDelimiter` is the right one.
Maybe, `requestNamespaceDelimiter`, `pathNamespaceDelimiter`, `routeNamespaceDelimiter`, etc?

juliusmarminge:
What's the state of this PR? 

I'll close it due to staleness, but feel free to reopen and get it up to date!

PR #3693, opened 2023-01-30
protected intersection for vanilla client

### LINKED ISSUE #3530: bug: router with "links" route is typed correctly but contains client links

```
  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    npm: 7.17.0 - /usr/local/bin/npm
  npmPackages:
    @trpc/client: ^10.7.0 => 10.7.0 
    @trpc/server: ^10.7.0 => 10.7.0 
    typescript: ^4.8.3 => 4.9.4 
```

### Describe the bug

I have a use case where I am defining a "links" route to be called such as

```
trpc.links.add.mutate()
trpc.links.remove.mutate()
trpc.links.list.query()
```

Once I attempted to run the application I received an error

```
Uncaught TypeError: Cannot read properties of undefined (reading 'query')
```

After a bit of digging I noticed if I moved the same router to another property with a different name such as "linkss" all is well. This was very confusing as I was not presented with any errors or warnings prohibiting me from defining a route named "links" and the client is typed as expected.

Digging a bit further I noticed this property contains the [client links](https://github.com/trpc/trpc/blob/main/packages/client/src/internals/TRPCUntypedClient.ts#L81), which makes sense, but does not appear documented. I may be able to open a PR, but would like some input on how to approach changing the behavior or improving the messaging

Without the getter for `links` this does not appear intended, and a quick test revealed properties in TypeScript Classes set as `private` are still accessible despite the type errors.

```ts
class MyClass {
  private readonly links: any[] = []
}

const c = new MyClass()

console.log(c.links)
// []
```

Modifying the script above to check for the "links" property logs the property when its found

```ts
const props = ['links']
for (const prop of props) {
  if (c.hasOwnProperty(prop)) console.log('has prop', prop)
}
// has prop links
```

Could this block unintentionally be lifting the private "links" property to the returned client?  https://github.com/trpc/trpc/blob/main/packages/client/src/createTRPCClientProxy.ts#L104 (or https://github.com/trpc/trpc/blob/main/packages/client/src/createTRPCClientProxy.ts#L107)

### Link to reproduction

https://stackblitz.com/edit/github-hwcyf7?file=package.json,src%2Fclient.ts%3AL46,src%2Fserver.ts&terminal=dev&view=editor

### To reproduce

with a basic TRPC client/server example, define a "links" route

```ts
  links: t.procedure.query(() => 'hello world')
})
```

and attempt to call

```ts
trpc.links.query().then(console.log).catch(console.error)
```

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

sachinraja:
Closes #3530

Naming a procedure something that conflicts with a `TRPCUntypedClient` property is now a type error.

### DIFF
--- a/packages/client/src/createTRPCClientProxy.ts
+++ b/packages/client/src/createTRPCClientProxy.ts
@@ -5,6 +5,7 @@ import type {
   AnyQueryProcedure,
   AnyRouter,
   AnySubscriptionProcedure,
+  IntersectionError,
   ProcedureArgs,
   ProcedureRouterRecord,
   ProcedureType,
@@ -22,10 +23,8 @@ import { CreateTRPCClientOptions } from './createTRPCUntypedClient';
 import {
   TRPCSubscriptionObserver,
   TRPCUntypedClient,
+  UntypedClientProperties,
 } from './internals/TRPCUntypedClient';
-
-export type inferRouterProxyClient<TRouter extends AnyRouter> =
-  DecoratedProcedureRecord<TRouter['_def']['record'], TRouter>;
 
 type Resolver<TProcedure extends AnyProcedure> = (
   ...args: ProcedureArgs<TProcedure['_def']>
@@ -91,7 +90,14 @@ const clientCallTypeMap: Record<
 };
 
 export type CreateTRPCProxyClient<TRouter extends AnyRouter> =
-  inferRouterProxyClient<TRouter>;
+  DecoratedProcedureRecord<
+    TRouter['_def']['record'],
+    TRouter
+  > extends infer TProcedureRecord
+    ? UntypedClientProperties & keyof TProcedureRecord extends never
+      ? TProcedureRecord
+      : IntersectionError<UntypedClientProperties & keyof TProcedureRecord>
+    : never;
 
 /**
  * @deprecated use `createTRPCProxyClient` instead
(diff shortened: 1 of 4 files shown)


KATT (commented):
Is there a reason we wouldn't do this on the backend? https://github.com/trpc/trpc/blob/4fab2f1da2042a8d82461f365da9fb967bb5ded1/packages/server/src/core/router.ts#L159-L168

juliusmarminge:
I guess the difference is that `then` should be banned on all levels deep while e.g. `links` only needs to be banned at depth 1 if that makes sense?

KATT (approved):
This is good.

For next major I think we should do some `_`-prefix or something to make these less rare. I don't like that we are blocking a generic word like "links" because we need it for some internals

sachinraja:
> Is there a reason we wouldn't do this on the backend?

It's not a problem if you're using the React or Next client.

> I guess the difference is that then should be banned on all levels deep while e.g. links only needs to be banned at depth 1 if that makes sense?

Also correct

> For next major I think we should do some _-prefix or something to make these less rare. I don't like that we are blocking a generic word like "links" because we need it for some internals

I think we'd be better off just moving to a functional API. I have some ideas on how to do that.

dotansimha:
Seems like this PR removed `inferRouterProxyClient` - what's the alternative?

juliusmarminge:
> > Is there a reason we wouldn't do this on the backend?
> It's not a problem if you're using the React or Next client

The problem would still occur when doing

```ts
utils.client.links

```

No?

PR #3705, opened 2023-02-01
Update loggerLink.ts

iserdmi:
Please, export defaultLogger. I want add some custom additional logic to "logger", but then I want just send messages through the default logger


KATT (changes requested):
Could you elaborate why you need this? The logger only takes 1 argument and you can define that in the `loggerLink` itself too

iserdmi:
@KATT Sorry, I think I was trying to do something wrong. Now I have just created an additional link with the needed logic

PR #3720, opened 2023-02-03
feat: add request info on context options

### LINKED ISSUE #3152: feat: add info about calls to `createContext()`-options

Add more meta data about the request when creating context

```ts
export async function createContext(opts: trpcNext.CreateNextContextOptions) {
  console.log(opts.info)
  return {}
}
```

`opts.info` could be something like:

```ts
interface ProcedureCall {
  type: ProcedureType;
  input?: unknown;
  path: string;
}

interface TRPCRequestInfo {
  calls: ProcedureCall[];
}
```

n/a

### Additional information

This can be used to do stuff like rate limiting, preventing certain procedure types in some operations, etc

### 👨‍👧‍👦 Contributing

juliusmarminge:
Closes #3152

## 🎯 Changes

- Adds information about the incoming request to the `createContext` options:

```ts
export type ProcedureCall = {
  type: ProcedureType;
  input?: unknown;
  path: string;
};

export type TRPCRequestInfo = {
  isBatchCall: boolean;
  calls: ProcedureCall[];
};
```

```ts
const createContext = (opts: CreateXYZContextOptions) => {
  const { info } = opts;
};
``` 

## ✅ Checklist


packages/server/src/adapters/ws.ts
```diff
@@ -140,7 +140,8 @@ export function applyWSSHandler<TRouter extends AnyRouter>(
       });
     }
 
-    const ctxPromise = createContext?.({ req, res: client });
+    // FIXME: info not set
+    const ctxPromise = createContext?.({ req, res: client, info: {} as any });
```
juliusmarminge, line 144: I'm not very familiar with how subscriptions work here. Can we extract the request info here too in a nice way?

@KATT

KATT: It's actually not sent here straight away... 

In subscriptions, we create a context _once_ when we open the connection and not as a result of a procedure.

I didn't consider this, but basically we have no further request info when doing websockets

packages/server/src/adapters/node-http/types.ts
```diff
@@ -42,6 +45,7 @@ export type NodeHTTPHandlerOptions<
 export type NodeHTTPCreateContextFnOptions<TRequest, TResponse> = {
   req: TRequest;
   res: TResponse;
+  info: TRPCRequestInfo;
```
KATT, line 48: This might break some third party adapters for other libs, should we change target to `next` instead?

juliusmarminge: Why would it break though? You don't have to destruct this prop?

KATT: Someone might have used that type as a base of their type, no?

e.g. https://github.com/BlairCurrey/trpc-koa-adapter/blob/master/src/index.ts

juliusmarminge: So you mean if they have used the `info` prop for something else then?

```ts
type SomeContextOpts = NodeHTTPCreateContextFnOptions<TReq, TRes>
  & { info: any }
```

Sure - lets add it to v11

KATT (approved):

packages/server/src/adapters/node-http/types.ts
KATT: I mean mainly that maybe they user creation of `NodeHTTPHandlerOptions` will fail as they don't have the `info`? Maybe I'm being too paranoid? If you think it's fine feel free to do it on `main`

juliusmarminge: No you're right I had to change the type of createContext in the websockets example so this is probably breaking something

PR #3727, opened 2023-02-06
[RFC] feat: `toRouter()`-fn that allows you to use `class`es for your routers

KATT:
## Description

No matter if you like OO or not, using classes is sometimes a neat way of encapsulating related logic.

This PR introduces a POC of a `toRouter()`-method that can take a class instance and turn it into a tRPC-router, allowing you to write code to the below where you can have a tRPC procedure right next to business logic.

https://github.com/trpc/trpc/blob/69f3b113094b9130dacf605bc3b394f5a5e71bbd/examples/next-minimal-starter/src/pages/api/trpc/%5Btrpc%5D.ts#L9-L48

## Notes

- ~~It's unimplemented in runtime, only done the TypeScript-part.~~ I've done a hacky implementation now that works
- I feel like this is a relatively low risk to support since it lives "outside" of the normal tRPC code

## Todos

- [ ] Get more feedback if this is a good idea
- [ ] Make it nicer (ideas? 🙋‍♂️)
  - [ ] Can we make `this.post = toRouter(PostService())` -> `this.post = PostService()`
  - [ ] Use `Symbol` for identifying tRPC routers and procedures
  - [ ] Filter out `never`-keys in `toRouter()` (i.e. `commonMethod` should not show up)
- [ ] Make an example with a DI-library like [TSyringe](https://github.com/microsoft/tsyringe) or similar
- [ ] Add tests
- [ ] Add docs


Nick-Lucas:
What's interesting about this approach to me is that router factories could become more of a 1st class extension of a standard pattern, rather than a new pattern laid on top.

```ts
class CrudRouter<T extends CrudEntity> {
  constructor(entity: ZodType<T>) {
    // store the entity for usage
  }
  // methods
}
```

This is more obvious than writing a factory function because it's just configuration.

This all said, I wonder if designing an RPC library with classes at the core is a project of its own, and not "tRPC v10". You could imagine .input.output.use etc being decorators on the class and methods, and the problems solved by base procedures being solved in a very different way as a result, then dependency injection and many other problems being solved in a way much more like NestJS does. tRPC seems like it's on the functional programming side of the FP vs Classes debate?

KATT:
> This all said, I wonder if designing an RPC library with classes at the core is a project of its own, and not "tRPC v10". You could imagine .input.output.use etc being decorators on the class and methods [...]

I tried playing around with decorators, but decorates can't infer argument types so I can't do like a `@input(z.object({})` and let the method get that input AFAICT?

> [...] and the problems solved by base procedures being solved in a very different way as a result, then dependency injection and many other problems being solved in a way much more like NestJS does.

Yeah, that was another thing I was thinking about here, ppl could be using stuff like [tsyringe](https://github.com/microsoft/tsyringe) for DI more easily

> tRPC seems like it's on the functional programming side of the FP vs Classes debate?

Not strictly. Internally, I want tRPC to never have any bad side effects of classes like a method called without the binding, but I don't hate classes to the point where I want to block people to use them.

As I mentioned in the description, this sits a bit like a side-car so it doesn't affect anything else in tRPC, has a small API surface, and might help some people reason easier about their code. At this point, it's a POC though.

Nick-Lucas:
> Yeah, that was another thing I was thinking about here, ppl could be using stuff like [tsyringe](https://github.com/microsoft/tsyringe) for DI more easily

So this is an interesting one for me, I think the route forward for this is to embrace Context and Middleware Context extension.

I built a [POC for a TSSyringe InjectorMiddleware a few weeks ago](https://stackblitz.com/edit/github-ewmssd?file=src%2FInjectorMiddleware.ts,src%2Fserver.ts) and to my mind it's a tidy solution so that services can depend on each other, but without having to increase the surface area of tRPC itself. Context and Middlewares provide some amazing tools here.

The API has been evolving a little in my mind and I'm thinking to end up with something like:
```ts
// trpc.ts
const injector= createInjector(container, { foo: Foo, bar: Bar, baz: Baz })

// fooRouter.ts
const fooProcedure = publicProcedure.use(injector("foo", "baz"))
```

marklawlor:
If  `private commonMethod()` was accidently set to `public`, would tRPC error? Something along the lines of `public method <name> on class <class-name> does not tRPC procedure`.

KATT:
> If `private commonMethod()` was accidently set to `public`, would tRPC error? Something along the lines of `public method <name> on class <class-name> is not a tRPC procedure`.

@marklawlor I've now implemented it. Essentially, we'll ignore anything that isn't a procedure or a router

packages/server/src/core/initTRPC.ts
```diff
@@ -136,7 +143,67 @@ function createTRPCInner<TParams extends PartialRootConfigTypes>() {
...
+
+    type ClassToProcRouterRecord<TRouter extends object> = Router<
+      RouterDef<
+        $Config,
+        {
+          [TKey in keyof TRouter]: TRouter[TKey] extends
+            | AnyProcedure
+            | AnyRouter
+            ? TRouter[TKey]
+            : never;
```
KATT, line 156: @sachinraja do you know if there's a performant/lazy way of filtering `never` values here?

KATT:
I had a stab at trying to make it to work with tsyringe in 3ec7f9b and attempt let `toRouter()` automatically resolve it, but I had problems with both getting decorators to work in our test suite and with some hairy generics -- if someone wants to have a stab, be my guest :)

examples/next-minimal-starter/src/pages/api/trpc/[trpc].ts
```diff
@@ -4,29 +4,44 @@
...
-    }),
-  // 💡 Tip: Try adding a new procedure here and see if you can use it in the client!
-  // getUser: publicProcedure.query(() => {
-  //   return { id: '1', name: 'bob' };
-  // }),
-});
+    });
+}
+
+const appRouter = unstable_toRouter(new RootService());
```
Comment (igalklebanov) [inline examples/next-minimal-starter/src/pages/api/trpc/[trpc].ts:44]: Maybe:

```ts
const appRouter = unstable_toRouter(RootService);
```

Instantiate internally exactly once. Familiarity with NestJS default mode (singleton).

examples/next-minimal-starter/src/pages/api/trpc/[trpc].ts
```diff
@@ -4,29 +4,44 @@
...
+        title: 'hello world',
+      },
+    ];
+  });
+}
+class RootService {
+  public post;
+  constructor() {
+    this.post = unstable_toRouter(new PostService());
+  }
```
Comment (igalklebanov) [inline examples/next-minimal-starter/src/pages/api/trpc/[trpc].ts:23]: Continuing https://github.com/trpc/trpc/pull/3727/files#r1097883151

```suggestion
  constructor(PostServiceImpl = PostService) {
    this.post = unstable_toRouter(PostService);
  }
```

igalklebanov:
I like the idea of "bring your own flavor, we got you!". 👏 

This might be a pandora's box, if people start requesting OOP stuff like complex NestJS DI mechanism or decorators (eww I vomitted in my mouth just thinking about decorators).

igalklebanov:
I'm not sure that ts-syringe should be the goto DI library in our example, being an abandoned project and all.
Sadly typedi is also pretty dead, maybe inversify?

packages/server/src/core/initTRPC.ts
igalklebanov: How about this:

```ts
type ClassToProcRouterRecord<TRouter extends object> = Router<
      RouterDef<
        $Config,
        {
          [TKey in keyof TRouter as TRouter[TKey] extends AnyProcedure | AnyRouter ? TKey : never]: TRouter[TKey];
        }
      >
    >;
```

PR #3729, opened 2023-02-06
fix(child-repos): ci failing due to missing pipeline step

juliusmarminge:
Closes #

## 🎯 Changes

The `prep-sqlite` step is part of the turbo pipeline, thus it's run before the build step in trpc's CI. This is not the case for the CI in each child repos causing it to fail due to the db not working. This adds that missing step as an explicit job

## ✅ Checklist

### DIFF
--- a/.github/workflows/subtree.yml
+++ b/.github/workflows/subtree.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - fix-subtree-repo-ci
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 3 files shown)


KATT (commented):
Can we solve it by getting rid of sqlite? I feel it's more trouble than its worth, most apps will also not use sqlite anyway

juliusmarminge:
> Can we solve it by getting rid of sqlite? I feel it's more trouble than its worth, most apps will also not use sqlite anyway

Yea - just did this as a quick fix. It's working now: https://github.com/trpc/examples-next-prisma-starter/actions/runs/4105168034/jobs/7081650398

KATT (approved):

PR #3737, opened 2023-02-07
fix(fastify-adapter): check if has content type parser 'application/j…

dphov:
## 🎯 Changes

Introduces a bug fix for  `adapters/fastify` which is a part of `@trpc/server`

Context: 

I want to add a condition to check, if content type parser of type `application/json` exist in Fastify or not.

### Why?
In Nest.js, application adds content type parsers by default.

If trpc re adds same content type parser it will give an error:
`throw new FST_ERR_CTP_ALREADY_PRESENT(contentType)` 

This PR solves this problem by adding an if condition with Fastify's `hasContentTypeParser` method.

## ✅ Checklist


KATT (approved):

KATT (changes requested):
Seems to fail one of the tests, please have a look

See https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md for how to run tests

dphov:
I feel like I did a mistake by adding this PR

 As I understood, `@trpc/server` needs to work on a JSON string, but not a serialized object.
 In fastify they, by default, have a content type parser that treat json mime type as an object.
 So in this case PR is breaking the beginning of the pipeline,  by unnecessary switching from strings to objects. 
 
 If it's true, could you close the PR.
 Thank you for your time

PR #3766, opened 2023-02-14
Directory Traversal Vulnerability Fix

avalokkumar:
Directory traversal: The code uses string concatenation to construct the file path, which could be vulnerable to directory traversal attacks. To mitigate this risk, it is important to use a library that helps to prevent directory traversal attacks, such as the path library.

## 🎯 Changes

* Instead of using string concatenation to create the file path, path.join function is used to safely join the individual path segments into a valid file path. This helps to prevent directory traversal attacks.

* Making the url parameter passed to createWsClient optionally be a function that returns a string. If the url that should be used is dynamic or changing over time this allows the wsClient to respect that when automatically reconnecting.

## ✅ Checklist

### DIFF
--- a/scripts/generateBigLegacyRouter.ts
+++ b/scripts/generateBigLegacyRouter.ts
@@ -3,10 +3,10 @@ import path from 'path';
 
 const NUM_PROCEDURES_TO_GENERATE = 200;
 
-const TEST_DIR =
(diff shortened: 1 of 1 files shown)


sachinraja:
There's no vulnerability here, this codegen function is only called for local development.

The second change about the WS client sounds fine, but it's not implemented in this PR.

PR #3775, opened 2023-02-15
ci: fix deno test

QuiiBz:
## 🎯 Changes

Fix the Deno test that has been failing since two days (due to a timeout). We use the `start-server-and-test` package and they released a [new version](https://github.com/bahmutov/start-server-and-test/releases/tag/v1.15.4) two days ago that fix a bug and expects to return a status code `200` in response to a `HEAD` request on `/`, to notify that the HTTP server is ready to be tested.

However, `/` always returns a `404` status code because the tRPC server listens on the `/trpc` endpoint. I've fixed this by returning an empty `Response` (which has a `200` status code) if the request method is `HEAD` (meaning when `start-server-and-test` checks if the HTTP server is ready to be tested)

## ✅ Checklist

### DIFF
--- a/examples/deno-deploy/src/index.ts
+++ b/examples/deno-deploy/src/index.ts
@@ -3,6 +3,12 @@ import { fetchRequestHandler } from 'npm:@trpc/server/adapters/fetch';
 import { appRouter } from './router.ts';
 
 function handler(request) {
+  // Only used for start-server-and-test package that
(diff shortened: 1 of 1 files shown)


juliusmarminge (approved):
You could also add a health check route for / like we do in the express example: https://github.com/trpc/trpc/blob/main/examples/express-minimal/src/server.ts

Not sure what's "cleaner"

PR #3795, opened 2023-02-18
feat: add experimental support for `FormData` mutations

sachinraja:
## 🎯 Changes

Notes from @KATT:

- I decided not to do "double inputs" as I previously outlined - it makes it hard to reason about tRPC. You can still limit procedures to only allow uploads for authorized users
- The example has to work as the reference material: https://github.com/trpc/trpc/tree/2023-01-29-formdata/examples/.experimental/next-formdata, read the readme


examples/next-minimal-starter/src/pages/api/trpc/[trpc].ts
```diff
@@ -22,18 +30,58 @@ const appRouter = router({
...
+const handler = trpcNext.createNextApiHandler({
   router: appRouter,
-  createContext: () => ({}),
+  createContext: (opts) => {
+    return opts;
+  },
+  unstable_contentTypeHandlers: [
+    nodeHTTPFormDataContentTypeHandler(),
+    nodeHTTPJSONContentTypeHandler(),
+  ],
```
Comment (KATT) [inline examples/next-minimal-starter/src/pages/api/trpc/[trpc].ts:75]: nice

packages/server/src/adapters/node-http/content-type/form-data/index.ts
```diff
@@ -0,0 +1,66 @@
...
+        });
+
+        bb.on('field', (name, value) => {
+          form.append(name, value);
+        });
+
+        bb.on('error', reject);
+        bb.on('finish', resolve);
+
+        req.pipe(bb);
```
KATT, line 56: Does this mean that all the files and their data must be received before reaching the actual resolver?

Not sure if it's the case cause I haven't played with it, but I feel like the files could/should be streamed into the procedure so we can for instance stream big files straight to disk or to s3 etc and not held in memory

sachinraja: good catch, it's being fully read into memory right now

KATT: Is there a way to forward the stream instead?

KATT: The current setup work fine for small files but will eat the computer alive for any big ones

packages/server/src/adapters/zodFileSchema.ts
```diff
@@ -0,0 +1,76 @@
...
+    const chunks: Buffer[] = [];
+    for await (const chunk of input.stream) {
+      chunks.push(chunk);
+    }
+
+    return new File(chunks, input.name, {
+      type: input.type,
+    });
+  }),
+]);
```
KATT, line 76: Intent here is to make a zod helper that works for files both in the browser and on the server.

Not sure how to make this work with `.optional()` at the end.

It **works** but it also treats empty files with `.optional()` as an error as the input isn't optional

This is how `ZodOptional`: https://github.com/colinhacks/zod/blob/a1fc3fb815051beb4b7030969828aa6abe469c2c/src/types.ts#L4353

Unsure how to make the `._getType` return undefined to it's actually treated as undefined when

- FileList is empty 
- File is bullish

Does preprocess help here somehow? Idk

colinhacks: I don't quite understand what the requirements are here. I also don't know if it's possible for the inferred type of a single schema to differ across the platforms. Here's a version that should work, but the inferred output type will be `FileList | File`. Again, not sure I understood what you need:

```ts

z.custom<FileList | Readable>(
  (input) => {
    if (typeof FileList !== "undefined" && input instanceof FileList)
      return true;
    if (typeof Readable !== "undefined" && input instanceof Readable)
      return true;
    return;
  },
  { message: "Input must be a file." }
).transform((input, ctx) => {
  if (input instanceof FileList) {
    // additional valuation
    if (input.length === 0) {
      ctx.addIssue({
        code: "custom",
        message: "FileList must have at least one file.",
      });
    }
    return input;
  }
  if (input instanceof Readable) {
    // additional validation
    // write to file...?
    return file;
  }
});

```

colinhacks: This is what I came up with. I didn't try to fully match the features of what you've already implemented but it might give you some ideeas.

```ts
z.custom<FileList | Readable>(
  (input) => {
    if (typeof FileList !== "undefined" && input instanceof FileList)
      return true;
    if (typeof Readable !== "undefined" && input instanceof Readable)
      return true;
    return;
  },
  { message: "Input must be a file." }
).transform((input, ctx) => {
  if (input instanceof FileList) {
    // additional valuation
    if (input.length === 0) {
      ctx.addIssue({
        code: "custom",
        message: "Please select a file",
      });
    }
  }
  if (input instanceof Readable) {
    // additional validation
    // read file fro stream...
    return new File(...);
  }
});
```

packages/server/src/adapters/zodFileSchema.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+  if (typeof (value as unknown as Readable).on === 'function') {
+    return true;
+  }
+
+  return false;
+});
+
+/**
+ * This is a hack to check if a value is a File in the browser
```
colinhacks, line 22: `File` is defined in browser so you could just do `instanceof` no? Is this supposed to be "Nodejs"?

packages/server/src/adapters/zodFileSchema.ts
```diff
@@ -0,0 +1,76 @@
...
+    return false;
+  }
+  if (typeof (value as unknown as File).type === 'string') {
+    return true;
+  }
+  return false;
+});
+
+/**
+ * This is a hack to check if a value is a FileList in the browser
```
colinhacks, line 35: `FileList` is defined in browser so you could just do `instanceof` no? Is this supposed to be "Nodejs"?

KATT: The problem is highlighted here: https://stackblitz.com/edit/node-ry4feb?file=index.ts

An empty `FileList` should be treated the same as `undefined`:

CleanShot 2023-03-04 at 23 58 42@2x


Is this even possible?

examples/.unstable/next-formdata/src/pages/api/trpc/[trpc].ts
```diff
@@ -0,0 +1,110 @@
...
+    .input(
+      zfd.formData({
+        name: zfd.text(),
+        image: zfd.file(),
+        document: zfd.file(z.instanceof(File).optional()),
+      }),
+    )
+    .mutation(async (opts) => {
+      return {
+        image: await writeFileToDisk(opts.input.image),
```
Comment (KATT) [inline examples/.unstable/next-formdata/src/pages/api/trpc/[trpc].ts:78]: We might want it to be `opts.formData` so there's no colliding keys 🤔

Comment (KATT) [reply examples/.unstable/next-formdata/src/pages/api/trpc/[trpc].ts:78]: Scenario:

- You validate on `roomId=1`
- It's passed through a middleware that check you're in that room
- In `formData` you add a `roomId`
- The first `roomId` will be overwritten but not authed

Comment (sachinraja) [reply examples/.unstable/next-formdata/src/pages/api/trpc/[trpc].ts:78]: ah good point

Comment (sachinraja) [reply examples/.unstable/next-formdata/src/pages/api/trpc/[trpc].ts]: fixed

Comment (KATT) [reply examples/.unstable/next-formdata/src/pages/api/trpc/[trpc].ts]: I thought a bit more on this and it _is_ a bit ugly to have the `input.formData`

What do you think is nicest:

- have a `.formData` thing
- Spread `{...formData}` but throw an internal error if there's a colliding key existent? I think that'd be neater... WDYT?

Comment (Nick-Lucas) [reply examples/.unstable/next-formdata/src/pages/api/trpc/[trpc].ts]: Given formdata is going to be the whole body if a mutation, I do think we're going to confuse users if we have to explain that sometimes your inputs are on `.input` and sometimes `.formData` and maybe sometimes `.stream`, shouldn't we endeavor to just adapt the inferred type of `input` in all cases?

When it comes to _returning_ blobs from functions that's also going to be necessary for `.output` right?

packages/tests/server/react/formData.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+          }
+          const formData = await experimental_parseMultipartFormData(
+            opts.ctx.req,
+            experimental_createMemoryUploadHandler(),
+          );
+
+          return opts.next({
+            rawInput: formData,
+          });
+        })
```
KATT, line 55: @Nick-Lucas's #4239 would be nice here 🤔

packages/tests/server/react/formData.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+    const appRouter = t.router({
+      polymorphic: t.procedure
+        .use(async (opts) => {
+          if (!experimental_isMultipartFormDataRequest(opts.ctx.req)) {
+            return opts.next();
+          }
+          const formData = await experimental_parseMultipartFormData(
+            opts.ctx.req,
+            experimental_createMemoryUploadHandler(),
+          );
```
sachinraja, line 50: what's the use case?

sachinraja (commented):
Looks really good!

examples/.experimental/next-formdata/src/pages/react-hook-form.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+      <FormProvider {...form}>
+        <form
+          method="post"
+          action={`/api/trpc/${mutation.trpc.path}`}
+          encType="multipart/form-data"
+          onSubmit={(_event) => {
+            if (noJs) {
+              return;
+            }
```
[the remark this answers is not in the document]
KATT: Variable name is wrong, it's just about ticking a box and trying to post it w/o js

packages/tests/server/react/formData.test.tsx
KATT: Thought it was cool to see if you can make a mutation that doesn't care if it gets `FormData` or a JSON-call

KATT (approved):

Issue #3590, opened 2023-01-10
feat: support for custom url encoding


JesusTheHun:
### Describe the feature you'd like to request

Right now tRPC only support data transformers for the body of requests.
When performing a query, the query string can be quite large. Some providers like `some-that-removed-their-free-plan-recently-so-I-wont-advertise-their-name` limit the URL to 8KB. While this is substantial and meet most of our needs, sometimes it's not enough.

I would like the ability to pass my own url serializer. Some [examples](https://tanstack.com/router/v1/docs/guide/custom-search-param-serialization) are given in the `TanStack Router` documentation.

<details>
<summary>tRPC current getUrl()</summary>

```ts
// https://github.com/trpc/trpc/blob/v10.8.1/packages/client/src/links/internals/httpUtils.ts#L96

export function getUrl(opts: HTTPRequestOptions) {
  let url = opts.url + '/' + opts.path;
  const queryParts: string[] = [];
  if ('inputs' in opts) {
    queryParts.push('batch=1');
  }
  if (opts.type === 'query') {
    const input = getInput(opts);
    if (input !== undefined) {
      queryParts.push(`input=${encodeURIComponent(JSON.stringify(input))}`);
    }
  }
  if (queryParts.length) {
    url += '?' + queryParts.join('&');
  }
  return url;
}
```
</details>

### Describe the solution you'd like to see

Configuring the client with a custom query parameter seems to be a perfectly viable solution to me.

### Describe alternate solutions

I haven't found a workaround so far. I thought about using a `mutation` instead, so it would be a post, but then I would not be able to use `useInfiniteQueries`. Dealbreaker.

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

<sub>[TRP-45](https://linear.app/trpc/issue/TRP-45/feat-support-for-custom-url-encoding)</sub>

## Funding
* You can sponsor this specific effort via a [Polar.sh](https://polar.sh) pledge below
* We receive the pledge once the issue is completed & verified

<a href="https://polar.sh/trpc/trpc/issues/3590">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/3590/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/trpc/trpc/issues/3590/pledge.svg">
</picture>
</a>

KATT:
[...]

PR #3804, opened 2023-02-19
feat: Add ability to configure custom inputEncoder and inputDecoder

kamilogorek:
Closes https://github.com/trpc/trpc/issues/3590

## 🎯 Changes

This change introduces a new pair of options, `inputEncoder` to the client and `inputDecoder` to the server,
that allows you to change how input is serialized/deserialized prior to sending an HTTP request.

## ✅ Checklist

### DIFF
--- a/packages/client/src/links/internals/httpUtils.ts
+++ b/packages/client/src/links/internals/httpUtils.ts
@@ -9,6 +9,11 @@ import {
 } from '../../internals/types';
 import { HTTPHeaders, PromiseAndCancel, TRPCClientRuntime } from '../types';
 
+type InputEncoder = (input: unknown) => string;
+
+const defaultInputEncoder: InputEncoder = (input) =>
+  encodeURIComponent(JSON.stringify(input));
+
 export interface HTTPLinkOptions {
   url: string;
   /**
@@ -24,6 +29,10 @@ export interface HTTPLinkOptions {
    * @link http://trpc.io/docs/v10/header
    */
   headers?: HTTPHeaders | (() => HTTPHeaders | Promise<HTTPHeaders>);
+  /**
+   * Custom encoder to be used when serializing the input into a query string entry.
+   */
+  inputEncoder?: InputEncoder;
 }
 
 export interface ResolvedHTTPLinkOptions {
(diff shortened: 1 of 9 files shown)


kamilogorek:
Well, that test is obviously lying 🤫

KATT:
cc @JesusTheHun would this be enough for you? It'd still add `?input=x` but you can customise that `x`

I'm not sure if I want to add this tbh unless there's a proven use-case

JesusTheHun:
@KATT Yes it does answer my need.

> I'm not sure if I want to add this tbh unless there's a proven use-case

My initial requirement is to be able to send an arbitrary long input. This could be fixed by allowing `POST query`, or it can be mitigated with this PR.
What is your concern with this addition to the codebase?

KATT:
> My initial requirement is to be able to send an arbitrary long input. This could be fixed by allowing `POST query`, or it can be mitigated with this PR.

Interesting, we could allow method overloading too. Can you explain a bit more about the behavior of the thing you're building? 


> What is your concern with this addition to the codebase?

More things to maintain. We're also going to support additional content types soon which will add a layer of complexity to this. 

If your problem could be solved by overloading, I think we should do that instead

JesusTheHun:
I understand.
Last time I checked it seemed the discussion about overloading was not very mature.

My use case is a dashboard with many filters with complex business rules ; which are represented by many tuples. 
Users must be able to go back and forth, but also bookmark their current view, so we store the state in the URL.

Square brackets and commas take 3 chars each in the URL, so in the end 70% of my query string is encoded chars.
Our current workaround is to pre-serialize the query string with a more url friendly format before passing it to tRPC.
But we have just postponed the inevitable.

Completely unrelated, another  aspect to consider, is that the URL is never secure, even when using HTTPS.

KATT:
> Edit : I just realized that using POST is not actually viable, since it will kill the user experience (asking for re-post confirmation) and also it's not bookmarkable.

I don't see how the user would see a difference? The user won't bookmark tRPC URLs, they'll bookmark URLs that belong to you


@kamilogorek great job on this PR but I think this is better solved by #3910

kamilogorek:
No worries, the alternative solution also sounds reasonable.

JesusTheHun:
@KATT you're absolutely right.
Stopping coffee clearly have down sides 😫

PR #3817, opened 2023-02-20
chore: update vite and fix vitest hanging

sachinraja:


package.json
```diff
@@ -75,7 +75,7 @@
       "@examples/next-websockets-starter>@prisma/client": "https://registry.npmjs.com/@prisma/client/-/client-4.9.0.tgz?id=%40examples%2Fnext-websockets-starter",
       "@examples/trpc-next-prisma-todomvc>@prisma/client": "https://registry.npmjs.com/@prisma/client/-/client-4.9.0.tgz?id=%40examples%2Ftrpc-next-prisma-todomvc",
       "@examples/legacy-next-starter>@prisma/client": "https://registry.npmjs.com/@prisma/client/-/client-4.9.0.tgz?id=%40examples%2Flegacy-next-starter",
-      "esbuild": "0.16.3"
+      "esbuild": "0.16.17"
```
QuiiBz, line 78: I think we should be able to remove this override, I've added it to fix a parsing error in the types (https://github.com/evanw/esbuild/commit/9ce5f94a49ff37635dfc221477640dfee11e036b)

QuiiBz (approved):

package.json
```diff
@@ -66,16 +66,16 @@
     "ts-prune": "^0.10.3",
     "tsx": "^3.9.0",
     "turbo": "^1.6.3",
-    "typescript": "^4.8.3"
+    "typescript": "^4.8.3",
+    "vitest": "^0.28.5"
```
juliusmarminge, line 70: This shouldnt be necessary in root?

juliusmarminge (approved):

package.json
juliusmarminge: i guess maybe we do want the globals to be available throughout all packages 🤔

QuiiBz: Exactly, otherwise we can't import `vitest` in other packages

juliusmarminge: not a fan of that - i like explicitely defining vitest as a dep whereever you need it

juliusmarminge: otherwise we might end up putting tests in examples without adding the dep which would break on subtree

juliusmarminge (approved):

packages/tests/vitest.config.ts
```diff
@@ -16,6 +16,7 @@ export default defineConfig({
       include: ['*/src/**/*.{ts,tsx,js,jsx}'],
       exclude: ['**/deprecated/**'],
     },
+    // useAtomics: true,
```
juliusmarminge, line 19: had to disable this @sachinraja

My computer panicked, might not be supported on M1 architecture or smth?

CleanShot 2023-02-20 at 10 38 27

juliusmarminge (approved):

packages/tests/vitest.config.ts
sachinraja: that's what actually fixed this though 🤣

PR #3774, opened 2023-02-15
fix: export more internals for dts files generation

jgoux:
Closes #

## 🎯 Changes

What changes are made in this PR? Is it a feature or a bug fix?

I'm continuing my patches about exposing enough internal types to be able to have a correct inference when generating declaration files (necessary when using TS project references).

It fixes these errors:

image

Past PRs: https://github.com/trpc/trpc/pulls?q=is%3Apr+author%3Ajgoux+is%3Aclosed

## ✅ Checklist

- [x] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [x] If necessary, I have added documentation related to the changes made.
- [x] I have added or updated the tests related to the changes made.

PR #3789, opened 2023-02-17
Internal types inference testing

jgoux:
## 🎯 Changes

What changes are made in this PR? Is it a feature or a bug fix?

I added an example project which should break if we don't export enough internal types from `@trpc/server`.

You can try playing with it by removing stuff from [internals.ts](https://github.com/trpc/trpc/blob/e1cff59b219605af47df557f4b005727c2168d22/packages/server/src/internals.ts) and see it breaking.

<del>

I used `pnpm` overrides to directly be linked against the live versions of `@trpc/client` and `@trpc/server` (useful trick found in [vitest](https://github.com/vitest-dev/vitest/blob/main/CONTRIBUTING.md#testing-vitest-against-external-packages) repo!)

I'm not sure about the best way to run it in your CI workflow, basically, all you need is:

1. Install + build trpc monorepo packages
2. Go to `examples/.test/internal-types-inference`
3. Run `pnpm i`
4. Run `pnpm build`

If the `pnpm build` is successful, congrats, you broke nothing! 😄

</del>

**edit**: Please see https://github.com/trpc/trpc/pull/3789#issuecomment-1435105314

@KATT Would you mind finishing the work and integrating these commands into the current CI workflow?

I also audited each exported type from [internals.ts](https://github.com/trpc/trpc/blob/e1cff59b219605af47df557f4b005727c2168d22/packages/server/src/internals.ts) against both the example and my monorepo and I was able to delete a few of them. 👍

## ✅ Checklist

- [x] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [x] If necessary, I have added documentation related to the changes made.
- [ ] I have added or updated the tests related to the changes made.

PR #3831, opened 2023-02-20
fix: export even more internals for dts files generation

milohansen:
Closes #

## 🎯 Changes

This extends work by @jgoux in #3774 by exposing more internal types to ensure declaration files can be generated. 

It fixes errors caused by the interaction of context and middleware ([stackblitz](https://stackblitz.com/edit/trpc-export-ts2742-jdz415?file=index.ts&view=editor)) and when routers are created within generic functions ([stackblitz](https://stackblitz.com/edit/trpc-export-ts2742-76uvre?file=index.ts&view=editor)).

## ✅ Checklist

### DIFF
--- a/examples/.test/internal-types-export/src/server.ts
+++ b/examples/.test/internal-types-export/src/server.ts
@@ -3,9 +3,20 @@ import { initTRPC } from '@trpc/server';
 // needs to be exported for the test to be valid
 export const t = initTRPC.create();
 
+const someMiddleware = t.middleware(({ next }) => {
+  return next();
+});
+
+export function genericRouter<S extends (value: any) => unknown>(schema: S) {
+  return t.router({
+    foo: t.procedure.output(schema).query(() => 'bar'),
+  });
+}
 
 const appRouter = t.router({
   foo: t.procedure.query(() => 'bar'),
+  hello: t.procedure.use(someMiddleware).query(() => 'hello'),
+  generic: genericRouter((value: string) => value.toUpperCase()),
 })
 
 export type AppRouter = typeof appRouter;
--- a/packages/server/src/core/index.ts
+++ b/packages/server/src/core/index.ts
@@ -15,7 +15,7 @@ export type {
   ProcedureArgs,
   ProcedureOptions,
 } from './procedure';
-export { inferParser } from './parser';
+export type { inferParser } from './parser';
 export { createInputMiddleware, createOutputMiddleware } from './middleware';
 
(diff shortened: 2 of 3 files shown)


jgoux:
Hey @milohansen, we created a [fixture project](https://github.com/trpc/trpc/pull/3789) to ensure we don't have regressions. Could you add your failing cases to https://github.com/trpc/trpc/blob/main/examples/.test/internal-types-export/src/server.ts ?

KATT (commented):
We have a test that is supposed to help us with this, could you update it so it would fail without this fix?

https://github.com/trpc/trpc/tree/main/examples/.test/internal-types-export

milohansen:
I've updated the test and confirmed it produces two errors without the fix.

image

Let me know if there's anything else I can do to help.

KATT (approved):

packages/server/src/core/index.ts
```diff
@@ -15,6 +15,7 @@ export type {
   ProcedureArgs,
   ProcedureOptions,
 } from './procedure';
+export { inferParser } from './parser';
```
KATT, line 18: Fails CI, might be a typo or export from the wrong place?

KATT (changes requested):

packages/server/src/core/index.ts
milohansen: Whoops, I fixed it to only export the type.

milohansen:
@KATT Looks like the build is working now, let me know if I can do anything else before this is merged.

KATT (approved):

examples/.test/internal-types-export/src/server.ts
```diff
@@ -3,9 +3,20 @@ import { initTRPC } from '@trpc/server';
 // needs to be exported for the test to be valid
 export const t = initTRPC.create();
 
+const someMiddleware = t.middleware(({ next }) => {
+  return next();
+});
+
+export function genericRouter<S extends (value: any) => unknown>(schema: S) {
+  return t.router({
+    foo: t.procedure.output(schema).query(() => 'bar'),
```
KATT, line 12: Soooo, we've just noticed that this is incorrect code which showed up while doing work on next major

You cannot know that `'bar'` is correct when "before" the `TSchema` is set

so just FYI @milohansen, this will break, unless you typecast `'bar'`, when you upgrade to v11

KATT: See #5330 for what you should be doing instead

PR #3839, opened 2023-02-22
Issues/3320 zod catchall unknown

### LINKED ISSUE #3320: bug: output type incorrectly inferred when compiling typescript for client when using zod .catchall()

```
  System:
    OS: macOS 12.6
    CPU: (10) arm64 Apple M1 Pro
    Memory: 97.91 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.13.1 - ~/.nvm/versions/node/v16.13.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.13.1/bin/yarn
    npm: 8.1.2 - ~/.nvm/versions/node/v16.13.1/bin/npm
    Watchman: 2022.10.03.00 - /opt/homebrew/bin/watchman
  Browsers:
    Brave Browser: 107.1.45.127
    Safari: 16.0
  npmPackages:
    @trpc/client: ^10.4.2 => 10.4.2 
    @trpc/server: ^10.4.2 => 10.4.2 
    typescript: ^4.9.3 => 4.9.3 
```

### Describe the bug

When using an zod object with `.catchall()` for the output of a procedure, it will incorrectly infer the type of the object when compiling the code using typescript for usage in another library.

I've created a minimal example to reproduce the issue. Let's say we have a server with one procedure `test`. This procedure takes as both input and output a z object with a required property `name` and a `catchall()` of `z.unknown`. this should generate the following interface:

```ts
{
    [x: string]: unknown;
    name: string;
}
```

In most cases it does generate that interface, except for the output model of the trpc client.

If we have the following code. 

```ts
import { inferRouterInputs, inferRouterOutputs, initTRPC } from "@trpc/server";
import { createTRPCProxyClient } from "@trpc/client";
import { z } from "zod";

const t = initTRPC.create({});

// Simple object with name as string and allows any other properties as unknown
const zObject = z
  .object({
    name: z.string(),
  })
  .catchall(z.unknown());

const router = t.router({
  // Test takes the zObject as both input and output
  test: t.procedure
    .input(zObject)
    .output(zObject)
    .mutation(({ input }) => {
      return input;
    }),
});

// Client is based on the type of the router
type AppRouter = typeof router;

// correctly specifies the input type
export const input = {} as inferRouterInputs<AppRouter>["test"];
// correctly specifies the output type
export const output = {} as inferRouterOutputs<AppRouter>["test"];

// correctly specifies the input type, incorrectly specifies the output type
export const client = createTRPCProxyClient<AppRouter>({
  links: [],
});
```

It will generate the following .d.ts bundle:

```ts
export declare const input: {
    [x: string]: unknown;
    name: string;
};
export declare const output: {
    [x: string]: unknown;
    name: string;
};
export declare const client: {
    test: {
        mutate: (input: {
            [x: string]: unknown;
            name: string;
        }, opts?: import("@trpc/server").ProcedureOptions | undefined) => Promise<{
            [x: string]: never;
            [x: number]: never;
        }>;
    };
};
```

As you can see the `input` and `output` variables both have the correct type inferred. The `client.test.mutate` input also has the correct type inferred, however if you look at the return type of the `client.test.mutate` it is :

```ts
{
    [x: string]: never;
    [x: number]: never;
}
```

If I were to remove the .catchall from the zod object (for e.g. .passthrough) it will work, but I lose the ability to add extra properties in TS.

This could be a zod issue (if so please let me know and I'll open this issue there instead). But as the type inference works in all cases except in the return type of the trpc client I thought it may be something to do with trpc instead.

### Link to reproduction

https://github.com/TimoGlastra/trpc-zod-issue

### To reproduce

Clone the repo above, run `yarn install` then run `yarn tsc`. This will build the d.ts file in `build/index.d.ts` and you can see the output as described above.

### Additional information

Would be willing to contribute a PR but wouldn't know where to start or whether this is solveable.

### 👨‍👧‍👦 Contributing

JoepRoebroek:
Closes #3320 

## 🎯 Changes

## ✅ Checklist


packages/tests/server/shared/internal/serialize.test.ts
```diff
@@ -0,0 +1,13 @@
...
+interface MixedKnownAndUnknownKeys {
+  name: string;
+  [x: string]: unknown;
+}
+
+test('it serializes objects with known and unknown keys', () => {
+  expectTypeOf<MixedKnownAndUnknownKeys>().toMatchTypeOf<
+    Serialize<MixedKnownAndUnknownKeys>
+  >();
+});
```
KATT, line 13: Thank you for adding a test! ❤️

KATT (approved):

KATT:
Would still like an extra pair of eyes from maybe @juliusmarminge or @sachinraja

packages/server/src/shared/internal/serialize.ts
```diff
@@ -44,25 +43,22 @@ type SerializeTuple<T extends [unknown, ...unknown[]]> = {
 
 /** JSON serialize objects (not including arrays) and classes */
 type SerializeObject<T extends object> = {
-  [k in keyof Omit<T, FilterKeys<T, NonJsonPrimitive>>]: Serialize<T[k]>;
+  [k in keyof T as T[k] extends NonJsonPrimitive ? never : k]: Serialize<T[k]>;
```
sachinraja, line 46: We need the indirection here to retain jump to definition. Using key remapping breaks jump to definition.

packages/server/src/shared/internal/serialize.ts
```diff
@@ -44,25 +43,22 @@ type SerializeTuple<T extends [unknown, ...unknown[]]> = {
...
  * Example: { a: string | undefined} --> { a?: string}
  */
-type UndefinedToOptional<T extends object> =
+type UndefinedToOptional<T extends object> = {
   // Property is not a union with `undefined`, keep as-is
-  Pick<T, FilterDefinedKeys<T>> & {
-    // Property _is_ a union with `defined`. Set as optional (via `?`) and remove `undefined` from the union
-    [k in keyof Omit<T, FilterDefinedKeys<T>>]?: Exclude<T[k], undefined>;
-  };
+  [k in keyof T as undefined extends T[k] ? never : k]: T[k];
```
sachinraja, line 57: same thing here with key remapping

sachinraja (changes requested):

packages/server/src/shared/internal/serialize.ts
JoepRoebroek: Thanks @sachinraja. That is interesting. The jump to definition doesn't seem to work in WebStorm for both cases, but indeed in VSCode with the new code, it breaks. Wouldn't know how to add a test for that. But I'll see if there is a way to fix this without breaking jump to definition.

JoepRoebroek: Ok, fiddled around a bit in TS playground to see if we can have it working both ways, but I am afraid it does not seem to be possible. I am not sure why VSCode is able to jump to definition when using Omit instead (maybe someone can elaborate on this?). The limitation of Omit however - if I understand correctly - is that in the case of having a mixed object with known and unknown keys, the `keyof` collapses it into `string`, so the information of the known keys is lost. 

Not sure how to proceed with this but would see a number of options:
1. Prioritize jump to definition: in this case, do nothing.
2. Prioritize 'correct' types: apply the change.
3. Discourage users to implement types like this in the documentation. Personally, while working on this issue, I was pondering the use cases, especially when using tRPC, when would you have unknown and known keys mixed. The whole point of tRPC is to have typesafety.

sachinraja: We need to keep jump to definition as it's a big part of "the tRPC experience." VSCode works when just using `Omit` because the TS server knows you aren't changing the keys themselves (but it thinks the keys themselves are changing if you use key remapping).

I would prioritize jump to definition as most users do not have unknown keys.

Nick-Lucas:
Hi @JoepRoebroek since this has been sat without changes for a couple months I'm going to close it out just to keep things tidy. If you find time to continue with your changes (assuming they're still needed, as I know some improvements have been made since) then please do bring the PR back!

Issue #3848, opened 2023-02-23
bug: Problem with fetch implementation


arashn0uri:
### Provide environment information

```
  System:
    OS: macOS 13.2.1
    CPU: (8) arm64 Apple M1
    Memory: 72.91 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.19.2 - /usr/local/bin/npm
  Browsers:
    Brave Browser: 110.1.48.164
    Chrome: 110.0.5481.177
    Safari: 16.3
  npmPackages:
    @tanstack/react-query: ^4.12.0 => 4.24.9 
    @trpc/client: 10.0.0-proxy-beta.17 => 10.0.0-proxy-beta.17 
    @trpc/next: 10.0.0-proxy-beta.17 => 10.0.0-proxy-beta.17 
    @trpc/react: 10.0.0-proxy-beta.17 => 10.0.0-proxy-beta.17 
    @trpc/server: 10.0.0-proxy-beta.17 => 10.0.0-proxy-beta.17 
    next: 12.3.4 => 12.3.4 
    react: 18.2.0 => 18.2.0 
    typescript: 4.9.4 => 4.9.4 
```

### Describe the bug

Hello guys,

I'm using a library (@easy-email) that asks me to create a mock window on the server side of my app to use the window's object on the server. I tried to use the process, but when I used the code provided by their developers on any page, TRPC crashed with this specific error "Error: No fetch implementation found".

I will add the code provided by the library and the screenshot of the problem. 
Any help will be appreciated.

THE CODE PROVIDED BY THEM:
```
import { JSDOM } from 'jsdom';

const dom = new JSDOM(`<!DOCTYPE html><p>Hello world</p>`);

(global as any).window = dom.window;
global.Element = dom.window.Element;
global.document = dom.window.document;
global.navigator = dom.window.navigator;
global.DOMParser = dom.window.DOMParser;
```

HOW I IMPORT THE FILE:
![Screenshot 2023-02-23 at 14 08 48](https://user-images.githubusercontent.com/110023107/220915425-bc5916f2-dcec-4919-9d34-dfe59f5ae16a.png)

MY TRPC CONFIG FILE:
```
import { initTRPC, TRPCError } from "@trpc/server";
import type { Context } from "./context";
import superjson from "superjson";
[...]

PR #3849, opened 2023-02-23
fix: prioritize `globalThis` over `window`

KATT:
Closes #3848

We could just use `globalThis` directly too maybe? https://caniuse.com/?search=globalThis

PR #3860, opened 2023-02-25
fix: Rework how fetch and AC are accessed from global

kamilogorek:
Closes https://github.com/trpc/trpc/issues/3848

## 🎯 Changes

- make sure that the passed implementation of `fetch` or anything that has been found is actually callable
- rename `packages/client/src/internals/fetchHelpers.ts` to `packages/client/src/internals/getAbortController.ts` to make it in line with `getFetch.ts`, as there's only one function left
- add tests covering fixed use-case
- due to `fetch` being bound during `getFetch`, it's not possible to use simple `toBe` assertions, thus return values are used
- make `getAbortController` more explicit in how it flows through the fallbacks
- fix `getAbortController` in the same was as `fetch`, when it's not found on `window`

## ✅ Checklist

Supersedes https://github.com/trpc/trpc/pull/3849

@KATT I completely missed the fact that `trpc` is `>node@18` only, so in the end, it is possible to work around the issues without the manual intervention :)


packages/client/src/getFetch.ts
```diff
@@ -1,19 +1,21 @@
-import { getWindow } from './internals/fetchHelpers';
 import { FetchEsque, NativeFetchEsque } from './internals/types';
 
-export function getFetch(f?: FetchEsque | NativeFetchEsque): FetchEsque {
-  if (f) {
-    return f as FetchEsque;
+export function getFetch(
+  customFetchImpl?: FetchEsque | NativeFetchEsque,
+): FetchEsque {
+  if (typeof customFetchImpl === 'function') {
```
KATT, line 6: Simplify
```suggestion
  if (customFetchImpl) {
```

KATT (approved):

PR #3862, opened 2023-02-25
chore(examples): move some examples to esm

sachinraja:
Move to ESM and remove `node-fetch` in favor of node 18's new native fetch.

This allows us to use `node-fetch@3` for tests which I need for `FormData` support in #3795. We can't use native fetch in tests yet because of some weird stuff with jsdom.


examples/express-minimal/package.json
```diff
@@ -2,12 +2,14 @@
...
   "private": true,
+  "type": "module",
   "scripts": {
     "dev:server": "tsx watch src/server",
     "dev:client": "wait-port 3000 && tsx watch src/client",
     "dev": "run-p dev:* --print-label",
     "lint": "eslint --ext \".js,.ts,.tsx\" --report-unused-disable-directives src",
-    "build": "tsc",
+    "type-check": "tsc",
+    "build": "pnpm type-check && esbuild src/server.ts src/client.ts --bundle --packages=external --platform=node --format=esm --outdir=dist --sourcemap",
```
juliusmarminge, line 12: Instead of putting type-check here, we can run it as a separate step in ci to pralleöize it

juliusmarminge (approved):

PR #3863, opened 2023-02-25
remove use of `eval`

arahansen:
## 🎯 Changes

I'm not sure if the suggested change is equivalent to what was documented before. However, I think it's important documentation does not suggest using `eval` due to the inherent security risks associated with the method.

Even if the example was included for simplicity and not as a "best practice", people are prone to copy-paste as a starting point, potentially never fixing the usage to something more robust after the fact.

## ✅ Checklist

- [N/A] I have added or updated the tests related to the changes made.

### DIFF
--- a/www/docs/server/data-transformers.md
+++ b/www/docs/server/data-transformers.md
@@ -86,7 +86,7 @@ export const transformer = {
   input: superjson,
   output: {
     serialize: (object) => uneval(object),
-    deserialize: (object) => eval(`(${object})`),
+    deserialize: (object) => JSON.parse(object),
   },
 };
 ```


KATT (commented):
This is just how devalue works.

It uses eval on the on the client on the result of a server that you have full control of. Unsure if I see the problem here_

KATT:
It's designed in a way where if you have different input and output parser, the output parser only ever happens on **the client**, so it should be safe to do.

CleanShot 2023-02-26 at 12 57 36@2x


If this is not true, please make an issue, it'd be a serious security vulnerability.

KATT:
See #3868

PR #3869, opened 2023-02-26
feat: tighten security of data transformers

KATT:
Makes it so we can only ever access the methods that are intended to be used on the server


packages/tests/server/mergeRouters.test.ts
```diff
@@ -76,20 +77,14 @@ test('bad merge: error formatter', async () => {
...
   });
 
-  expect(() =>
-    t1.mergeRouters(router1, router2),
-  ).toThrowErrorMatchingInlineSnapshot(
-    `"You seem to have several error formatters"`,
-  );
+  ignoreErrors(() => {
+    // @ts-expect-error should not be allowed
+    t1.mergeRouters(router1, router2);
```
KATT, line 82: cc @juliusmarminge @sachinraja and smart ideas on how we could make sure all routers sent here have a matching `RootConfig`-type? 

it's one of those things that get hard with the bag-structure we have of generics

KATT: I tried this to no avail

```ts

type IsEqual<TLeft, TRight> = [TLeft] extends [TRight]
  ? [TRight] extends [TLeft]
    ? TLeft
    : false
  : false;

type NarrowSameConfig<
  TRouters extends AnyRouter[],
  TExpectedConfig extends AnyRootConfig,
> = TRouters extends [
  infer Head extends AnyRouter,
  ...infer Tail extends AnyRouter[],
]
  ? IsEqual<Head['_def']['_config'], TExpectedConfig> extends true
    ? NarrowSameConfig<Tail, Head['_def']['_config']>
    : never
  : TRouters;

export function mergeRouters<TRouters extends AnyRouter[]>(
  ...routerList: NarrowSameConfig<TRouters, TRouters[0]['_def']['_config']>
): MergeRouters<TRouters> {
```

juliusmarminge: this looks scary ;)

PR #3878, opened 2023-02-27
bump to react-query 5

KATT:
See what breaks

https://github.com/trpc/trpc/actions/runs/4285389256/jobs/7463625993

cc @TkDodo


mozzius:
Seems like they've done away with passing context to the hooks, you just pass the whole query client to it now.
https://github.com/TanStack/query/issues/4682

Issue #3393, opened 2022-12-14
feat: lower compile targets for browser-facing packages


KATT:
### Describe the feature you'd like to request

Currently, we don't do any transpilation outside of what's in tsconfig, which leads to limited browser support, even in compiled output for i.e. Next.js apps. If you search your compiled `_app.js`-bundle, it has a lot of optional chaining which all browsers doesn't support.

### Describe the solution you'd like to see

Either use babel or update tsconfig for browser-facing packages (`@trpc/react-query`, `@trpc/client`) to support older browsers.

The minimum requirement for tRPC is for the browser to support `Proxy`, so we should compile just after that.

### Describe alternate solutions

Unsure, maybe a guide how to transpile code for older browsers?

### Additional information

- In v9, we had babel: https://github.com/trpc/trpc/blob/v9.x/packages/client/package.json#L34
- Unlikely to add extra payload as the likelihood of another dep needing babel is also high.
- TanStack babel config: https://github.com/TanStack/query/blob/main/babel.config.js

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

<sub>[TRP-38](https://linear.app/trpc/issue/TRP-38/feat-lower-compile-targets-for-browser-facing-packages)</sub>

## Funding
* You can sponsor this specific effort via a [Polar.sh](https://polar.sh) pledge below
* We receive the pledge once the issue is completed & verified

<a href="https://polar.sh/trpc/trpc/issues/3393">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/3393/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/trpc/trpc/issues/3393/pledge.svg">
</picture>
</a>

iway1:
Probably worth mentioning that `@tanstack/core` may no longer transpile optional chaining in v5 to reduce bundle size https://github.com/TanStack/query/discussions/4252. Not sure what the plans our for react query v5 support down the road, but just something I thought of when reading this

JoshuaKGoldberg:
What is the desired browser support range? Going back older that optional chaining is pretty far back.

> it has a lot of optional chaining which all browsers doesn't support.

Per https://caniuse.com/mdn-javascript_operators_optional_chaining, optional chaining is supported by: 
* Chrome 80 (released February 4, 2020)
[...]

Issue #331, opened 2021-05-03
Error thrown in Resolver always results in a 500 Status code


mgranderath:
I have this resolver setup 

```typescript
export const authRouter = createRouter().mutation("login", {
  input: LoginInput,
  async resolve({ input, ctx }) {
    throw trpc.httpError.unauthorized("Optional message");
  },
});
````

Which should result in a 401 status code, however, the error that is actually sent by the endpoint is (with a 500 code):
```
TRPCError: Optional message
    at internalServerError (C:\Users\malte\Github\streamparty\monorepo\packages\website-next\.next\server\pages\api\v2\trpc\[trpc].js:264:10)
    at getErrorFromUnknown (C:\Users\malte\Github\streamparty\monorepo\packages\website-next\.next\server\pages\api\v2\trpc\[trpc].js:289:10)
    at Object.requestHandler (C:\Users\malte\Github\streamparty\monorepo\packages\website-next\.next\server\pages\api\v2\trpc\[trpc].js:558:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
    at async C:\Users\malte\Github\streamparty\monorepo\packages\website-next\.next\server\pages\api\v2\trpc\[trpc].js:184:5
    at async apiResolver (C:\Users\malte\Github\streamparty\monorepo\packages\website-next\node_modules\next\dist\next-server\server\api-utils.js:8:1)
    at async DevServer.handleApiRequest (C:\Users\malte\Github\streamparty\monorepo\packages\website-next\node_modules\next\dist\next-server\server\next-server.js:67:462)
    at async Object.fn (C:\Users\malte\Github\streamparty\monorepo\packages\website-next\node_modules\next\dist\next-server\server\next-server.js:59:492)
    at async Router.execute (C:\Users\malte\Github\streamparty\monorepo\packages\website-next\node_modules\next\dist\next-server\server\router.js:25:67)
    at async DevServer.run (C:\Users\malte\Github\streamparty\monorepo\packages\website-next\node_modules\next\dist\next-server\server\next-server.js:69:1042)
    at async DevServer.handleRequest (C:\Users\malte\Github\streamparty\monorepo\packages\website-next\node_modules\next\dist\next-server\server\next-server.js:34:504) {
  originalError: HTTPError: Optional message
[...]

PR #3880, opened 2023-02-27
chore: rework how `TRPCError` is created and add appropriate unit tests for it

kamilogorek:
## 🎯 Changes

Resolves one outstanding note from the codebase (https://github.com/trpc/trpc/issues/331#issuecomment-831221487).

As for why it didn't work previously, it was most likely due to the odd behavior of
```
Object.setPrototypeOf(this, new.target.prototype);
```

in some environments. This is however redundant to call in non-ES6 environments, which based on https://github.com/trpc/trpc/issues/3393 is not the case for tRPC, as it requires Proxy API, which is ES6+ only.

We've also been there, but we're tied to pre-ES6 envs - https://github.com/getsentry/sentry-javascript/blob/33243242effa68390d16203917475ce26c8f2037/packages/utils/src/error.ts#L14-L17

## ✅ Checklist


KATT (dismissed):

packages/tests/server/TRPCError.test.ts
```diff
@@ -1,77 +1,75 @@
 import { TRPCError, getTRPCErrorFromUnknown } from '@trpc/server/src';
 
-describe('TRPCError', () => {
-  it('should extend original Error class', () => {
-    const trpcError = new TRPCError({ code: 'FORBIDDEN' });
-    expect(trpcError).toBeInstanceOf(TRPCError);
-    expect(trpcError).toBeInstanceOf(Error);
-  });
+test('should extend original Error class', () => {
```
kamilogorek, line 3: There is more than 1 top-level thing tested in this file, that's why it was split using `describe` blocks :)

juliusmarminge (approved):

packages/server/src/error/TRPCError.ts
```diff
@@ -25,28 +24,24 @@ export function getTRPCErrorFromUnknown(cause: unknown): TRPCError {
...
     const message =
-      opts.message ?? getMessageFromUnknownError(opts.cause, code);
-    const cause: Error | undefined =
+      opts.message ?? getMessageFromUnknownError(opts.cause, opts.code);
+    const cause =
       opts.cause !== undefined ? getErrorFromUnknown(opts.cause) : undefined;
 
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore https://github.com/tc39/proposal-error-cause
     super(message, { cause });
```
QuiiBz, line 42: I don't have the history on this, what's the reason for not following the `Error` constructor?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#syntax

kamilogorek: In what sense?

QuiiBz: The `Error` constructor accepts a message as the first argument, and then options (or filename and lineNumber), whereas `TRPCError` only accepts an object combining both the message and options. IMO it would be better to follow the same constructor and extend the options to add the `code`, but I'm sure there's a reason why it was made like this (but that would require a breaking change).

kamilogorek: Ah ok, I don't know the history here as well. From what I see it's most likely because `message` is optional and not always known at compile time. It could feel awkward to do `new TRPCError(null, { code: 420 });`. Just a guess.

ixahmedxi (approved):

packages/server/src/error/TRPCError.ts
juliusmarminge: If there is more that can be done that would result in something breaking we can do it on the `next` branch

PR #3884, opened 2023-02-27
feat(server): add CORS support to nodeHTTPRequestHandler, add Standalone adapter documentation

### LINKED ISSUE #3881: feat: handle CORS/Options in @trpc/server/adapters/standalone createHTTPServer

tRPC has a great @trpc/server/adapters/standalone package which in theory makes it very easy to stand up an instance when learning - or a simple instance for deployment!

The simplest method in theory being:

```ts
import { createHTTPServer } from '@trpc/server/adapters/standalone'
import { appRouter } from './appRouter'

const server = createHTTPServer({
  router: appRouter,
  batching: {
    enabled: true,
  }
})

server.listen(3333)
```

The problem is in my experience this can't be used locally, because it doesn't respond to options requests or handle CORS, both of which browsers nowadays require.

What I actually have to write every time I want a standalone server is this, which takes a lot more research and domain knowledge:

```ts
import { createHTTPHandler } from '@trpc/server/adapters/standalone'
import { createServer } from 'http'

import { appRouter } from './appRouter'

const handler = createHTTPHandler({
  router: appRouter,
  batching: {
    enabled: true,
  },
})

const server = createServer((req, res) => {
  res.setHeader('Access-Control-Allow-Origin', '*')
  res.setHeader('Access-Control-Request-Method', '*')
  res.setHeader('Access-Control-Allow-Methods', '*')
  res.setHeader('Access-Control-Allow-Headers', '*')
  if (req.method === 'OPTIONS') {
    res.writeHead(200)
    return res.end()
  }
  handler(req, res)
})

server.listen(3333)
```

Add configurable support in `createHTTPServer` to handle Options and CORS and make local development much easier to get going. Possibly make deployment much easier too.

Don't do anything, this issue in itself is perfectly fine documentation in combination with #623

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

Nick-Lucas:
Closes #3881 

Since this adds support to `nodeHTTPRequestHandler` directly, the change is also upstream of 
the express, next, and standalone, adapters.

I've implemented and run this against the minimal example and it works, though if the approach is acceptable some updates to documentation will probably be in order.

## 🎯 Changes

## ✅ Checklist

### DIFF
--- a/examples/minimal-react/server/index.ts
+++ b/examples/minimal-react/server/index.ts
@@ -3,8 +3,8 @@
  * On a bigger app, you will probably want to split this file up into multiple files.
  */
 import { initTRPC } from '@trpc/server';
-import { createHTTPHandler } from '@trpc/server/adapters/standalone';
-import http from 'http';
+import { createHTTPServer } from '@trpc/server/adapters/standalone';
+import cors from 'cors';
 import { z } from 'zod';
 
 const t = initTRPC.create();
@@ -36,25 +36,12 @@ const appRouter = router({
 // None of the actual implementation is exposed to the client
 export type AppRouter = typeof appRouter;
 
-// create handler
-const handler = createHTTPHandler({
+// create server
+createHTTPServer({
+  cors: cors(),
   router: appRouter,
   createContext() {
     console.log('context 3');
     return {};
   },
-});
-
-const server = http.createServer((req, res) => {
-  res.setHeader('Access-Control-Allow-Origin', '*');
-  res.setHeader('Access-Control-Request-Method', '*');
-  res.setHeader('Access-Control-Allow-Methods', 'OPTIONS, GET');
-  res.setHeader('Access-Control-Allow-Headers', '*');
-  if (req.method === 'OPTIONS') {
-    res.writeHead(200);
-    return res.end();
-  }
-  handler(req, res);
-});
-
-server.listen(2022);
+}).listen(2022);
--- a/examples/minimal-react/server/index.ts
+++ b/examples/minimal-react/server/index.ts
@@ -38,7 +38,7 @@ export type AppRouter = typeof appRouter;
 
 // create server
 createHTTPServer({
-  cors: cors(),
+  middleware: cors(),
   router: appRouter,
   createContext() {
     console.log('context 3');
--- a/packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
+++ b/packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
@@ -28,7 +28,7 @@ export async function nodeHTTPRequestHandler<
   const handleViaMiddleware =
     opts.middleware ??
     ((_req, _res, next) => {
-      next();
+      return next();
     });
 
   return handleViaMiddleware(opts.req, opts.res, async (err) => {
--- a/packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
+++ b/packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
@@ -32,6 +32,23 @@ export async function nodeHTTPRequestHandler<
   };
   const { path, router } = opts;
 
+  if (opts.cors) {
+    await new Promise<void>((resolve, reject) => {
+      opts.cors!(opts.req, opts.res, (err) => {
+        if (err) {
+          reject(err);
+        } else {
+          resolve();
+        }
+      });
+    });
+
+    if (opts.res.writableEnded) {
+      // cors handled the request
+      return;
+    }
+  }
+
   const bodyResult = await getPostBody(opts);
 
   const query = opts.req.query
--- a/packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
+++ b/packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
@@ -33,17 +33,21 @@ export async function nodeHTTPRequestHandler<
   const { path, router } = opts;
 
   if (opts.cors) {
-    await new Promise<void>((resolve, reject) => {
-      opts.cors!(opts.req, opts.res, (err) => {
-        if (err) {
-          reject(err);
-        } else {
-          resolve();
-        }
-      });
-    });
+    const corsStatus = await new Promise<'continue' | 'handled'>(
+      (resolve, reject) => {
+        opts.cors!(opts.req, opts.res, (err) => {
+          if (err) {
+            reject(err);
+          } else {
+            resolve('continue');
+          }
+        });
 
-    if (opts.res.writableEnded) {
+        resolve('handled');
+      },
+    );
+
+    if (corsStatus === 'handled') {
       // cors handled the request
       return;
     }
--- a/packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
+++ b/packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
@@ -26,7 +26,7 @@ export async function nodeHTTPRequestHandler<
   TResponse extends NodeHTTPResponse,
 >(opts: NodeHTTPRequestHandlerOptions<TRouter, TRequest, TResponse>) {
   const handleViaMiddleware =
-    opts.cors ??
+    opts.middleware ??
     ((_req, _res, next) => {
       next();
     });
--- a/packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
+++ b/packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
@@ -25,70 +25,72 @@ export async function nodeHTTPRequestHandler<
   TRequest extends NodeHTTPRequest,
   TResponse extends NodeHTTPResponse,
 >(opts: NodeHTTPRequestHandlerOptions<TRouter, TRequest, TResponse>) {
-  const createContext = async function _createContext(): Promise<
-    inferRouterContext<TRouter>
-  > {
-    return await opts.createContext?.(opts);
-  };
-  const { path, router } = opts;
+  const handleViaMiddleware =
+    opts.cors ??
+    ((_req, _res, next) => {
+      next();
+    });
 
-  if (opts.cors) {
-    const corsStatus = await new Promise<'continue' | 'handled'>(
-      (resolve, reject) => {
-        opts.cors!(opts.req, opts.res, (err) => {
-          if (err) {
-            reject(err);
-          } else {
-            resolve('continue');
-          }
+  return handleViaMiddleware(opts.req, opts.res, async (err) => {
+    if (err) {
+      throw err;
+    }
+
+    //
+    // Build tRPC dependencies
+
+    async function createContext(): Promise<inferRouterContext<TRouter>> {
+      return await opts.createContext?.(opts);
+    }
+
+    const bodyResult = await getPostBody(opts);
+
+    const query = opts.req.query
+      ? new URLSearchParams(opts.req.query as any)
+      : new URLSearchParams(opts.req.url!.split('?')[1]);
+
+    const req: HTTPRequest = {
+      method: opts.req.method!,
+      headers: opts.req.headers,
+      query,
+      body: bodyResult.ok ? bodyResult.data : undefined,
+    };
+
+    //
+    // Invoke tRPC
+
+    const result = await resolveHTTPResponse({
+      batching: opts.batching,
+      responseMeta: opts.responseMeta,
+      path: opts.path,
+      createContext,
+      router: opts.router,
+      req,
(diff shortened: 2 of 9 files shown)


packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
```diff
@@ -32,6 +32,23 @@ export async function nodeHTTPRequestHandler<
...
+        } else {
+          resolve();
+        }
+      });
+    });
+
+    if (opts.res.writableEnded) {
+      // cors handled the request
+      return;
+    }
```
Nick-Lucas, line 49: This is a little unconventional, but the alternative is to wrap the rest of the function in the `next` callback, which is fine, but a bigger refactor

KATT: Does this even need to be a promise?

KATT: I feel like a `connect`-style-middleware would be good

Nick-Lucas: Nope, only used here to unwrap the callback to be used imperatively. It would probably be much better to put the rest of the function in the callback instead, this current way is admittedly setting off my bug senses.

I mostly want to check that the public API concept is sound, and if so let's tidy this up.

Nick-Lucas: > I feel like a `connect`-style-middleware would be good

I'm not sure what pattern exactly this is referring to, could you point me in the right direction? Definitely happy to explore alternative approaches :)

KATT: `req,res,next`-thing https://github.com/senchalabs/connect

Nick-Lucas: Got it, so on the public-api front, I think connect would let people compose together a multi-stage middleware if we just accept one `NodeMiddlewareLike` (maybe rename cors->middleware)

On the internal side I'm having a play with restructuring as a middleware in a 0-dependency way, and I honestly think I'm just making the code more complex 😆, this is working but I think mostly just obfuscates the code:

```ts
/* eslint-disable @typescript-eslint/no-non-null-assertion */
import { AnyRouter } from '../../core';
import { inferRouterContext } from '../../core/types';
import { HTTPRequest } from '../../http/internals/types';
import { resolveHTTPResponse } from '../../http/resolveHTTPResponse';
import { getPostBody } from './internals/getPostBody';
import {
  NodeHTTPHandlerOptions,
  NodeHTTPRequest,
  NodeHTTPResponse,
} from './types';

type NodeHTTPRequestHandlerOptions<
  TRouter extends AnyRouter,
  TRequest extends NodeHTTPRequest,
  TResponse extends NodeHTTPResponse,
> = {
  req: TRequest;
  res: TResponse;
  path: string;
} & NodeHTTPHandlerOptions<TRouter, TRequest, TResponse>;

interface RequestHandler<
  TRouter extends AnyRouter,
  TRequest extends NodeHTTPRequest,
  TResponse extends NodeHTTPResponse,
> {
  (
    opts: NodeHTTPRequestHandlerOptions<TRouter, TRequest, TResponse>,
    next: (err?: any) => void,
  ): void | Promise<void>;
}

export async function nodeHTTPRequestHandler<
  TRouter extends AnyRouter,
  TRequest extends NodeHTTPRequest,
  TResponse extends NodeHTTPResponse,
>(opts: NodeHTTPRequestHandlerOptions<TRouter, TRequest, TResponse>) {
  const middlewares: RequestHandler<TRouter, TRequest, TResponse>[] = [
    corsHandler,
    trpcHandler,
  ].filter(Boolean);

  type Pipeline = (
    opts: NodeHTTPRequestHandlerOptions<TRouter, TRequest, TResponse>,
  ) => Promise<void>;

  const pipeline = middlewares.reduceRight<Pipeline>(
    (next, middleware) => {
      return async (opts) => {
        await middleware(opts, (err) => {
          if (err) {
            throw err;
          }

          return next(opts);
        });
      };
    },
    async () => {
      //
    },
  );

  return await pipeline(opts);
}

async function corsHandler<
  TRouter extends AnyRouter,
  TRequest extends NodeHTTPRequest,
  TResponse extends NodeHTTPResponse,
>(
  opts: NodeHTTPRequestHandlerOptions<TRouter, TRequest, TResponse>,
  next: () => void,
) {
  if (!opts.cors) {
    return next();
  }

  return opts.cors(opts.req, opts.res, (err) => {
    if (err) {
      throw err;
    }

    next();
  });
}

async function trpcHandler<
  TRouter extends AnyRouter,
  TRequest extends NodeHTTPRequest,
  TResponse extends NodeHTTPResponse,
>(
  opts: NodeHTTPRequestHandlerOptions<TRouter, TRequest, TResponse>,
  next: () => void,
) {
  const createContext = async function _createContext(): Promise<
    inferRouterContext<TRouter>
  > {
    return await opts.createContext?.(opts);
  };
  const { path, router } = opts;

  const bodyResult = await getPostBody(opts);

  const query = opts.req.query
    ? new URLSearchParams(opts.req.query as any)
    : new URLSearchParams(opts.req.url!.split('?')[1]);
  const req: HTTPRequest = {
    method: opts.req.method!,
    headers: opts.req.headers,
    query,
    body: bodyResult.ok ? bodyResult.data : undefined,
  };
  const result = await resolveHTTPResponse({
    batching: opts.batching,
    responseMeta: opts.responseMeta,
    path,
    createContext,
    router,
    req,
    error: bodyResult.ok ? null : bodyResult.error,
    onError(o) {
      opts?.onError?.({
        ...o,
        req: opts.req,
      });
    },
  });

  const { res } = opts;
  if ('status' in result && (!res.statusCode || res.statusCode === 200)) {
    res.statusCode = result.status;
  }
  for (const [key, value] of Object.entries(result.headers ?? {})) {
    if (typeof value === 'undefined') {
      continue;
    }
    res.setHeader(key, value);
  }

  res.end(result.body);

  return next();
}

```

Nick-Lucas: This second approach looks much tidier and isn't trying to be too clever:

```ts
/* eslint-disable @typescript-eslint/no-non-null-assertion */
import { AnyRouter } from '../../core';
import { inferRouterContext } from '../../core/types';
import { HTTPRequest } from '../../http/internals/types';
import { resolveHTTPResponse } from '../../http/resolveHTTPResponse';
import { getPostBody } from './internals/getPostBody';
import {
  NodeHTTPHandlerOptions,
  NodeHTTPRequest,
  NodeHTTPResponse,
} from './types';

type NodeHTTPRequestHandlerOptions<
  TRouter extends AnyRouter,
  TRequest extends NodeHTTPRequest,
  TResponse extends NodeHTTPResponse,
> = {
  req: TRequest;
  res: TResponse;
  path: string;
} & NodeHTTPHandlerOptions<TRouter, TRequest, TResponse>;

export async function nodeHTTPRequestHandler<
  TRouter extends AnyRouter,
  TRequest extends NodeHTTPRequest,
  TResponse extends NodeHTTPResponse,
>(opts: NodeHTTPRequestHandlerOptions<TRouter, TRequest, TResponse>) {
  const handleViaMiddleware =
    opts.cors ??
    ((_req, _res, next) => {
      next();
    });

  return handleViaMiddleware(opts.req, opts.res, async () => {
    const createContext = async function _createContext(): Promise<
      inferRouterContext<TRouter>
    > {
      return await opts.createContext?.(opts);
    };
    const { path, router } = opts;

    const bodyResult = await getPostBody(opts);

    const query = opts.req.query
      ? new URLSearchParams(opts.req.query as any)
      : new URLSearchParams(opts.req.url!.split('?')[1]);
    const req: HTTPRequest = {
      method: opts.req.method!,
      headers: opts.req.headers,
      query,
      body: bodyResult.ok ? bodyResult.data : undefined,
    };
    const result = await resolveHTTPResponse({
      batching: opts.batching,
      responseMeta: opts.responseMeta,
      path,
      createContext,
      router,
      req,
      error: bodyResult.ok ? null : bodyResult.error,
      onError(o) {
        opts?.onError?.({
          ...o,
          req: opts.req,
        });
      },
    });

    const { res } = opts;
    if ('status' in result && (!res.statusCode || res.statusCode === 200)) {
      res.statusCode = result.status;
    }
    for (const [key, value] of Object.entries(result.headers ?? {})) {
      if (typeof value === 'undefined') {
        continue;
      }
      res.setHeader(key, value);
    }
    res.end(result.body);
  });
}

```

www/docs/server/adapter/standalone.md
```diff
@@ -0,0 +1,151 @@
...
+      <th>Links</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>Standalone tRPC Server with Node.js</td>
+      <td><em>n/a</em></td>
+      <td>
+        <ul>
+          <li><a href="https://githubbox.com/trpc/trpc/blob/main/examples/minimal">CodeSandbox</a></li>
```
Nick-Lucas, line 24: This githubbox link (which I've invented) is not working, does it need configuring by someone in the core team?

packages/server/src/adapters/node-http/types.ts
```diff
@@ -31,11 +31,32 @@ export type NodeHTTPCreateContextOption<
...
+   *
+   * ```ts
+   * import cors from 'cors'
+   *
+   * nodeHTTPRequestHandler({
+   *   cors: cors()
+   * })
+   * ```
+   */
+  middleware?: NodeMiddlewareLike;
```
KATT, line 59: I do think this should be an array of connect-like middlewares

```suggestion
  middlewares?: NodeMiddlewareLike[];
```

packages/server/src/adapters/node-http/types.ts
```diff
@@ -31,11 +31,32 @@ export type NodeHTTPCreateContextOption<
       createContext: NodeHTTPCreateContextFn<TRouter, TRequest, TResponse>;
     };
 
+export interface NodeMiddlewareLike<
```
KATT, line 34: Drop `export`, add `@internal`, call it connect middleware (I think that's the original name)

```suggestion
/** @internal **/
interface ConnectMiddleware<
```

packages/server/src/adapters/node-http/types.ts
Nick-Lucas: So I won't push back too hard or long against this, but...

I think supporting a chain of middlewares is a significant bump in complexity for the internals and future maintenance, and could be achieved by a user intent on it _without_ 1st party support.

CORS is also the only time I can think of when you would want this escape hatch for the right reasons, for most other reasons shifting to express would yield a more powerful solution than tRPC could ever offer.

This topic is addressed in the proposed docs as such:

```
The `middleware` option will accept any function which resembles a node.js middleware, so it can be used for more than `cors` handling if you wish. It is, however, intended to be a simple escape hatch and as such won't on its own allow you to compose multiple middlewares together. If you want to do this then you could:

1. Use an alternate adapter with more comprehensive middleware support, like the [Express adapter](/docs/express)
2. Use a solution to compose middlewares such as [connect](https://github.com/senchalabs/connect)
3. Extend the Standalone `createHTTPHandler` with a custom http server (see below)
```

www/docs/server/adapter/standalone.md
juliusmarminge: https://stackblitz.com/github/trpc/trpc/blob/main/examples/minimal

would work although stackblitz is on Node 16 so it complains about missing fetch implementation :/

Nick-Lucas: At least that opens the code and makes it browsable rather than fully broken, I've switched out the URL and also added the minimal-react example as that one covers the cors use-case :)

packages/server/src/adapters/node-http/types.ts
KATT: Fair enough

KATT (approved):

KATT (approved):

examples/minimal-react/server/index.ts
```diff
@@ -3,8 +3,8 @@
  * On a bigger app, you will probably want to split this file up into multiple files.
  */
 import { initTRPC } from '@trpc/server';
-import { createHTTPHandler } from '@trpc/server/adapters/standalone';
-import http from 'http';
+import { createHTTPServer } from '@trpc/server/adapters/standalone';
```
juliusmarminge, line 6: breaking change?

juliusmarminge: nvm i see they're two different exports, mb

PR #3899, opened 2023-02-28
fix: changes ProcedureParams to use one generic

ixahmedxi:
Closes #

## 🎯 Changes

## ✅ Checklist

### DIFF
--- a/packages/server/src/core/procedure.ts
+++ b/packages/server/src/core/procedure.ts
@@ -35,14 +35,12 @@ type AnyProcedureParams = {
 /**
  * @internal
  */
-export type ProcedureParams<
-  TParams extends AnyProcedureParams = AnyProcedureParams,
-> = TParams;
+export type ProcedureParams<TParams extends AnyProcedureParams> = TParams;
 
 /**
  * @internal
  */
-export type ProcedureArgs<TParams extends ProcedureParams> =
+export type ProcedureArgs<TParams extends ProcedureParams<AnyProcedureParams>> =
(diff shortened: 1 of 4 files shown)


packages/server/src/core/procedure.ts
```diff
@@ -20,44 +20,24 @@ export interface ProcedureOptions {
...
+  _input_out: unknown;
+  _output_in: unknown;
+  _output_out: unknown;
+};
+
+/**
+ * @internal
+ */
+export type ProcedureParams<
+  TParams extends AnyProcedureParams = AnyProcedureParams,
```
KATT, line 39: Can we get rid of the default? Otherwise it's effectively the same as `AnyProcedureParams`

```suggestion
  TParams extends AnyProcedureParams,
```

ixahmedxi: if I do that i get the following error, do you have an idea how I could possibly fix that?

CleanShot 2023-03-03 at 16 16 23@2x

ixahmedxi: CleanShot 2023-03-03 at 16 19 10@2x

adding the `AnyProcedureParams` as a generic does fix it but I'm not sure if that's what we are going for

KATT: Yeah, we gotta be explicit with the generic, I think what you're doing here is fine!

ixahmedxi: Okay I will update it to explicitly add the generic and will resubmit for review ❤️

PR #3903, opened 2023-03-02
fix(server): createInputMiddleware next call type

ixahmedxi:
Closes #

## 🎯 Changes

## ✅ Checklist

### DIFF
--- a/packages/server/src/core/internals/procedureBuilder.ts
+++ b/packages/server/src/core/internals/procedureBuilder.ts
@@ -330,7 +330,10 @@ function createProcedureCaller(_def: AnyProcedureBuilderDef): AnyProcedure {
           rawInput: opts.rawInput,
           meta: _def.meta,
           input: callOpts.input,
-          next: async (nextOpts?: { ctx: any; input?: any }) => {
+          next: async (nextOpts?: {
+            ctx?: { [key: string]: unknown };
+            input?: unknown;
+          }) => {
             return await callRecursive({
               index: callOpts.index + 1,
               ctx:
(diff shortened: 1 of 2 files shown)


packages/server/src/core/middleware.ts
```diff
@@ -137,7 +137,7 @@ export type MiddlewareFunction<
     meta: TParams['_meta'] | undefined;
     next: {
       (): Promise<MiddlewareResult<TParams>>;
-      <$Context>(opts: { ctx: $Context }): Promise<
+      <$Context>(opts: { ctx?: $Context; input?: unknown }): Promise<
```
KATT, line 140: I am about 99.9% I tried this but it had some edge-case.... but I can't remember now and it evidently passes tests 🤔 

Can you make sure it works if you downgrade to TS 4.7?

ixahmedxi: for sure, i shall do that :)

ixahmedxi: @KATT so I downgraded typescript to 4.7.4 and i found the current error in the `fastifyRequestHandler.ts` file:

image

this is however not connected to the following change of this PR as that I can confirm the error shows up on the base `next` branch. I was able to fix the error by doing the following:

image

So in conclusion it doesn't seem like my change does produce any type errors on Typescript version `4.7.4` and in the process I discovered an unrelated error that's easily solveable.

packages/server/src/core/internals/procedureBuilder.ts
```diff
@@ -330,7 +330,10 @@ function createProcedureCaller(_def: AnyProcedureBuilderDef): AnyProcedure {
           rawInput: opts.rawInput,
           meta: _def.meta,
           input: callOpts.input,
-          next: async (nextOpts?: { ctx: any; input?: any }) => {
+          next: async (nextOpts?: {
+            ctx?: { [key: string]: unknown };
```
KATT, line 334: This could be written with a `Record<string, unknown>`, a bit neater

KATT (approved):

Issue #3916, opened 2023-03-04
bug: Overshadowed types for Vitest global API


kamilogorek:
### Provide environment information

N/A

### Describe the bug

All testing is currently performed using Vitest with `test: { globals: true }` in `vitest.config.ts` and `"types": ["vitest/globals"]` in `tsconfig.json` which exposes global Vitest API to the world.

This, however, is not playing nicely with `@testing-library/jest-dom`, which types internally are importing `jest` types via triple-slash directive `/// <reference types="jest" />`.

And because `import '@testing-library/jest-dom';` is called inside `setupTests.ts` for every test run, what ends up happening, is that `@types/jest` (which is transitive dependency of `@testing-library/jest-dom`) is registered globally, effectively shadowing every API method of that same name (which is like most of it).

```
➜  ~/Projects/trpc/packages/tests git:(main) pnpm why @types/jest
Legend: production dependency, optional only, dev only

@trpc/tests@10.14.0 /Users/kamilogorek/Projects/trpc/packages/tests

dependencies:
@testing-library/jest-dom 5.16.5
└─┬ @types/testing-library__jest-dom 5.14.5
  └── @types/jest 27.5.2
@types/testing-library__jest-dom 5.14.5
└── @types/jest 27.5.2
```

This makes `describe`, `test`, `it`, etc. etc, being reported as the code coming from `jest`, where in fact its `vitest` which is executed. This makes updates to `vitest` not reflected in the types and can be very problematic in some circumstances.

One of the examples of this being a problem is using `retry` test option.

Here're the current `test` types:

<img width="478" alt="image" src="https://user-images.githubusercontent.com/1523305/222892399-e8107969-b3d7-4d73-8a22-5d75fa91f861.png">

<img width="712" alt="image" src="https://user-images.githubusercontent.com/1523305/222892406-8030d36e-abec-4326-ac67-956d05b619de.png">

This is however not true, as 3rd argument to `test` in the version of `vitest` used in the project is `TestOptions`, which can be confirmed by forcibly removing `@types/jest` from packages.

<img width="491" alt="image" src="https://user-images.githubusercontent.com/1523305/222892764-05291b29-47ff-4c4c-97f3-c589cc8eb2b2.png">

<img width="958" alt="image" src="https://user-images.githubusercontent.com/1523305/222892730-48d3c83f-03f4-4c7d-80d8-a8762fa481c1.png">

### Link to reproduction

N/A

### To reproduce

[...]

PR #3923, opened 2023-03-05
fix(build): Resolve overshadowed types for Vitest global API

kamilogorek:
Closes https://github.com/trpc/trpc/issues/3916

## 🎯 Changes

This PR vendors `@types/testing-library__jest-dom` package and modifies it so that it works nicely with Vitest.
The only required modification was removing the `jest` triple-slash reference declaration.

`@testing-library/jest-dom` is rather a complete package, with very sparse updates/bugfixes, so I believe vendors it is the right move here. And if anything, only types need to be bumped, which are all in `matchers.d.ts` that is 1:1 copied and has tool-based formatting disabled, so making a git-diff will not be an issue in the future.

Note that matches still live under `jest` namespace and it's not an oversight during vendoring - https://github.com/vitest-dev/vitest/blob/0280825f1a672ba23d9fcbbc6470a4b29d343a73/packages/vitest/src/types/global.ts#L38-L40

All redundant imports of `'@testing-library/jest-dom'` were removed, as it's already registered as a part of `setupTests.ts` which is executed for all `packages/tests` test cases.

TypeScript LSP is working as expected now and all CI checks are passing:

image

image

## ✅ Checklist


kamilogorek:
Related `jest-dom` issue https://github.com/testing-library/jest-dom/issues/427

package.json
```diff
@@ -78,7 +78,8 @@
       "@examples/trpc-next-prisma-starter>prisma": "https://registry.npmjs.com/prisma/-/prisma-4.10.1.tgz?id=%40examples%2Ftrpc-next-prisma-starter",
       "@examples/next-websockets-starter>prisma": "https://registry.npmjs.com/prisma/-/prisma-4.10.1.tgz?id=%40examples%2Fnext-websockets-starter",
       "@examples/trpc-next-prisma-todomvc>prisma": "https://registry.npmjs.com/prisma/-/prisma-4.10.1.tgz?id=%40examples%2Ftrpc-next-prisma-todomvc",
-      "@examples/legacy-next-starter>prisma": "https://registry.npmjs.com/prisma/-/prisma-4.10.1.tgz?id=%40examples%2Flegacy-next-starter"
+      "@examples/legacy-next-starter>prisma": "https://registry.npmjs.com/prisma/-/prisma-4.10.1.tgz?id=%40examples%2Flegacy-next-starter",
+      "@types/testing-library__jest-dom": "link:./packages/tests/vendor/@types/testing-library__jest-dom"
```
juliusmarminge, line 82: Do we need this as a package even? Can we not just import the file in setupTests and it'll be registered?

Or is it a peerDep for something else?

packages/tests/vendor/@types/testing-library__jest-dom/index.d.ts
```diff
@@ -0,0 +1,17 @@
+// Type definitions for @testing-library/jest-dom 5.14
```
QuiiBz, line 1: We are currently using `@testing-library/jest-dom` version `5.16.5` vs `5.14` here, I guess we should update `matchers.d.ts` to use the same version?

package.json
kamilogorek: It's an indirect transitive dependency that we have to override, explained it in detail here https://github.com/trpc/trpc/issues/3916

packages/tests/vendor/@types/testing-library__jest-dom/index.d.ts
kamilogorek: There were no API changes since `5.14` and those are the latest available types, so I left the version comment intact to make it easy to realize which version is vendored.

juliusmarminge (approved):
Nice 👍🏼

PR #3959, opened 2023-03-09
remove reference to resolve()

arahansen:
I couldnt find any other references to resolve(). it seems it might be a v9 concept that was removed in v10? I'm not sure because I've only ever used v10

### DIFF
--- a/www/docs/further/subscriptions.md
+++ b/www/docs/further/subscriptions.md
@@ -29,7 +29,6 @@ const t = initTRPC.create();
 
 export const appRouter = t.router({
   onAdd: t.procedure.subscription(() => {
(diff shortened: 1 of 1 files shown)


www/docs/further/subscriptions.md
```diff
@@ -29,7 +29,6 @@ const t = initTRPC.create();
 
 export const appRouter = t.router({
   onAdd: t.procedure.subscription(() => {
-    // `resolve()` is triggered for each client when they start subscribing `onAdd`
```
kamilogorek, line 32:
```suggestion
```

kamilogorek (dismissed):
Removed unnecessary gap

juliusmarminge (approved):

PR #3978, opened 2023-03-10
feat(server): add support for metadata merging

### LINKED ISSUE #3948: feat: Procedure Meta merging

Currently if you write something like this, tRPC will generate a runtime error:

```ts
publicProcedure.meta({ keyA: 1 }).meta({ keyA: 2, keyB: 3 })
// ^ Error: Duplicate key meta
```

This leads to a little inflexibility where tools like middlewares are concerned, for instance, you have an Authorization middleware which checks a route's access requirements against a user's permissions. In that case the middleware itself needs to coerce an un-set key to a known requirement. I think it would be better if a base procedure could have some defaults defined in the meta and these defaults could be overwritten on a procedure-by-procedure basis.

```ts
const authedProcedure = publicProcedure.use(AuthMiddleware).meta({ requiredRole: 'super' })

const appRouter = router({
  hello: authedProcedure.meta({ requiredRole: 'basic' })
})
```

I propose that multiple Meta() calls should stack and _shallow_ merge for middlewares and procedures to consume

* Don't do it because there's a really good reason not to allow more than a single meta call per procedure.
* Take a defaultValue like: `initTRPC.meta<Type>(defaultValue)`
  * or support this _and_ shallow merging!

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

Nick-Lucas:
Closes #3948 

First crack at this, implements:

* Shallow merging of Meta when procedure.meta() is called multiple times
* defaultMeta parameter on the TRPCBuilder class (will also get shallow merged)
* simple docs

Questions and possible extra directions:

* Should we provide some way to access the previous meta value so you can use custom/deep merging strategies?
  a. `baseprocedure.meta({ ...baseProcedure.getMeta(), extraKey: 1, extraObj: { key: 1, ...baseProcedure.getMeta().extraObj } })`
  b. `baseProcedure.meta(previous => ({ ...previous, extraKey: 1, extraObj: { key: 1, ...previous.extraObj } }))`
* Should we propagate the knowledge of whether Meta was previously set and make all future chainings Partial<Meta> so that initial declarations can have mandatory keys without being forced to repeat them later?
* Should we plan for a breaking change in v11 that defaultMeta becomes mandatory?

## 🎯 Changes

## ✅ Checklist


packages/server/src/core/initTRPC.ts
```diff
@@ -44,19 +44,32 @@ type CreateRootConfigTypesFromPartial<TTypes extends PartialRootConfigTypes> =
...
-      FlatOverwrite<TParams, { ctx: Unwrap<TNewContext> }>
-    >();
+    type NextParams = FlatOverwrite<TParams, { ctx: Unwrap<TNewContext> }>;
+
+    return new TRPCBuilder<NextParams>(this.config as NextParams);
   }
-  meta<TNewMeta extends RootConfigTypes['meta']>() {
-    return new TRPCBuilder<FlatOverwrite<TParams, { meta: TNewMeta }>>();
+
+  meta<TNewMeta extends RootConfigTypes['meta']>(defaultMeta?: TNewMeta) {
```
Nick-Lucas, line 64: A v11 change could be to make defaultMeta mandatory, for now maintain existing behaviour or starting out undefined

packages/server/src/core/internals/procedureBuilder.ts
```diff
@@ -192,18 +192,19 @@ function createNewBuilder(
...
-  const { middlewares = [], inputs, ...rest } = def2;
+  const { middlewares = [], inputs, meta, ...rest } = def2;
 
   // TODO: maybe have a fn here to warn about calls
   return createBuilder({
     ...mergeWithoutOverrides(def1, rest),
     inputs: [...def1.inputs, ...(inputs ?? [])],
     middlewares: [...def1.middlewares, ...middlewares],
-  } as any);
+    meta: def1.meta && meta ? { ...def1.meta, ...meta } : meta ?? def1.meta,
```
kamilogorek, line 202:
```suggestion
    meta: { ...def1.meta, ...meta }
```

It's safe to spread possibly-undefined objects - https://github.com/tc39/ecma262/issues/687
And the readability of this version outweighs creating a "not always necessary" new object for each builder IMO.

In case

www/docs/server/metadata.md
```diff
@@ -53,15 +53,50 @@ const isAuthed = t.middleware(async ({ meta, next, ctx }) => {
...
-  protectedHello: authedProcedure.meta({ hasAuth: true }).query(() => {
+  protectedHello: authedProcedure.meta({ authRequired: true }).query(() => {
     return {
       greeting: 'hello-world',
     };
   }),
 });
 ```
+
+## Default Meta, Chaining, and shallow merging
```
kamilogorek, line 69:
```suggestion
## Default meta, chaining, and shallow merging
```

Docs are not using title-casing anyway from what I see.

www/docs/server/metadata.md
```diff
@@ -53,15 +53,50 @@ const isAuthed = t.middleware(async ({ meta, next, ctx }) => {
...
+
+If you set Meta multiple times, it will be shallow merged.
```
kamilogorek, line 71:
```suggestion
If you set meta multiple times, it will be shallowly merged.
```

Even though it's `interface Meta`, docs still use lower-cased version for them, eg. context for `type Context` - https://trpc.io/docs/context

packages/server/src/core/internals/procedureBuilder.ts
Nick-Lucas: Great suggestion, that is much cleaner yes!

For L195 the purpose of that is actually to separate keys we want to merge manually from keys we want to merge automatically (`..rest`) so that `mergeWithoutOverrides` can throw for any duplicates, so I think that's best as it is

Nick-Lucas: Had to revert off this change because it's actually a breaking change, `undefined` is a valid meta value and middlewares in the wild might check for this. Tests started failing

kamilogorek: Oh, I see. Quite a bummer. As for meta merging, we can then change it to this:

```js
meta: meta ? { ...def1.meta, ...meta } : def1.meta
```

```js
const merge = (a, b) => b.meta ? { ...a.meta, ...b.meta } : a.meta;

console.log(merge({}, {})); // undefined
console.log(merge({ meta: {a:1} }, { meta: undefined })); // {a:1}
console.log(merge({ meta: undefined }, { meta: {b:1} })); // {b:1}
console.log(merge({ meta: {a:1} }, { meta: {a:2} })); // {a:2}
console.log(merge({ meta: {a:1} }, { meta: {b:1} })); // {a:1, b:1}
```

kamilogorek: Unless `def2.meta` can also be `undefined`, then we're out of luck ¯\_(ツ)_/¯

Nick-Lucas: Yep either could be undefined. Could shunt this into a function though?

kamilogorek: Hmm, if `def2.meta` can be `undefined`, does this mean that it overrides the `def1` entirely or leave It as is?
Because if leave as is, then it'll work just fine with the example above, if it needs to override it, then your solution is probably the best one we can get. I think its fine without separate function, just wrote it like this for easier testing :)

Nick-Lucas: It's a good question. 

My understanding is:
* Both can be undefined (no meta set, and override called for other stuff)
* def2 is necessarily partial, def2.meta will not be passed if the override call is for some other key(s), and def1.meta should be retained
* Users can't set `procedure.meta(undefined)` if `initTRPC.meta<>()` has been called with a type, so TS prevents unexpected behaviour as a user
* def1.meta could be undefined and def2.meta could have a value if `initTRPC.meta<>()` hasn't been called

So the way I wrote this is necessarily safe, because basically every combination is a possibility, but the precedence is always "if both then merge otherwise return the non-nully one"

kamilogorek: > "if both then merge otherwise return the non-nully one"

`(def1.meta || meta) ? { ...def1.meta, ...meta } : undefined`

Then this should do as well, but it's basically stylistic at this point.

Nick-Lucas: Yeah I do agree we're in the stylistic weeds now. I can't think of any scenario where meta would be something other than an object or undefined, so in principle this is fine, but my instinct is to outsource the value decision upward and to honour whatever we're given if merging isn't needed - if that's `null` for some reason this function shouldn't care.

Happy to apply the change if you like :)

kamilogorek: Yeah, let's leave it like this then :)

packages/tests/server/routerMeta.test.ts
```diff
@@ -77,3 +77,155 @@ test('route meta in middleware', async () => {
   expect(middleware).toHaveBeenCalledTimes(2);
   await close();
 });
+
+test('default meta', async () => {
+  const t = initTRPC
+    .meta<{
+      data: string;
+    }>({ data: 'foobar' })
+    .create();
```
KATT, line 86: I'm thinking that this API might be nicer?:

```ts
  const t = initTRPC
    .meta<{
      data: string;
    }>()
    .create({ defaultMeta: { data: 'foobar' } });
```

Considering that we do all other "runtime stuff" at the `create()`?

Nick-Lucas: Okay that's fair enough! It might prevent people from inferring their meta via generics, but maybe there's not much value (or it's worth actively discouraging) that?

Right now something like this should be possible:
```ts
const t = initTRPC
	  .meta({
	    data: "foobar"
	  })
          // ^ infer { data: string }
	  .create();
```

Nick-Lucas: Changes pushed!

bautistaaa (approved):

PR #3982, opened 2023-03-12
feat: add scale support

harrysolovay:
For instance...

```ts
import { initTRPC } from "@trpc/server"
import * as $ from "scale-codec"

const t = initTRPC.create()

const params = $.object(
  $.field("greeting", $.str),
  $.field("name", $.str),
)

export type router = typeof router
export const router = t.router({
  something: t.procedure
    .input(params)
    .query((req) => {
      const { greeting, name } = req.input
      console.log(greeting, name)
      return `${greeting} ${name}`
    }),
})
```

---

All that remains is getting the assertion working. One small issue:

Running the assertion requires use of the `assert` function, exposed from `scale-codec`; this is not attached to the codec itself. This means we struggle to avoid a direct dependency on `scale-codec`.

Is it okay to add an optional / peer dep on scale-codec? This would have the added benefit of us being able to remove the `ParserScaleEsque` type and use the `Codec` type directly. Thoughts?


packages/server/src/core/internals/getParseFn.ts
```diff
@@ -30,6 +30,12 @@ export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
     return parser.create.bind(parser);
   }
 
+  if (typeof parser._assert === 'function') {
+    // ParserScaleEsque
+    // return (v: any) => $.assert(parser, v);
+    return (v: any) => v;
```
KATT, line 36: just missing the juice here

sachinraja:
Would it be possible for scale-codec to be able to conform to the ZodEsque parser? I want to add support for it but it's hard to justify adding another parser when the library has little usage right now (50 downloads / week).

tjjfvi:
> Would it be possible for scale-codec to conform to the ZodEsque parser and expose an _input type?

We'll add an `assert` method to the `Codec` class so this won't need a peer dependency on `scale-codec`, but I don't think conforming to the zod interface is a viable path – the method name `parse` would be confusing in the context of the library, as its main purpose is encoding/decoding – `parse` sounds like a synonym of `decode`.

sachinraja:
You wouldn't need parse, just an `_input` property: https://github.com/trpc/trpc/blob/3e018018e14c2e5c0afeb96bd4d874f3df7d9c0f/packages/server/src/core/parser.ts#L2

tjjfvi:
For the type-level, yes, but AFAICT at runtime it would still need a `parse` method.

sachinraja:
I'd be fine with adding another check on the runtime level but I don't want to risk degrading TS perf without a strong reason.

packages/server/src/core/parser.ts
```diff
@@ -18,11 +18,17 @@ export type ParserCustomValidatorEsque<TInput> = (
 export type ParserYupEsque<TInput> = {
   validateSync: (input: unknown) => TInput;
 };
+
+export type ParserScaleEsque<TInput> = {
+  assert(value: unknown): asserts value is TInput;
+};
```
KATT, line 24: It allows transforms right? If so, it should be in `ParserWithInputOutput` with the possible input and output separated

packages/server/src/core/internals/getParseFn.ts
```diff
@@ -30,6 +30,14 @@ export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
     return parser.create.bind(parser);
   }
 
+  if (typeof parser.assert === 'function') {
+    // ParserScaleEsque
+    return (value) => {
+      parser.assert(value);
+      return value as TType;
```
KATT, line 37: Seems like this misses the transform (that you said it supports) since it just returns the original value

packages/server/src/core/parser.ts
harrysolovay: Can you please clarify what you mean by transforms?

tjjfvi: It supports transforms with regard to the `encode`/`decode`, but not in the `assert` logic.

packages/server/src/core/parser.ts
```diff
@@ -18,11 +18,17 @@ export type ParserCustomValidatorEsque<TInput> = (
...
+
 export type ParserWithoutInput<TInput> =
   | ParserYupEsque<TInput>
   | ParserSuperstructEsque<TInput>
   | ParserCustomValidatorEsque<TInput>
-  | ParserMyZodEsque<TInput>;
+  | ParserMyZodEsque<TInput>
+  | ParserScaleEsque<TInput>;
```
KATT, line 31: You want this to be in `ParserWithInputOutput` so you can support transforms

KATT (approved):

Issue #3933, opened 2023-03-06
feat: support `ensureQueryData`


KATT:
### Describe the feature you'd like to request

Support `ensureQueryData`. It's very useful

### Describe the solution you'd like to see

Support `ensureQueryData` 

### Describe alternate solutions

not doing it

### Additional information

https://tanstack.com/query/v4/docs/react/reference/QueryClient#queryclientensurequerydata

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

PR #2852, opened 2022-09-30
examples: adds minimal react example

bautistaaa:
Closes #

## 🎯 Changes

What changes are made in this PR? Is it a feature or a bug fix?

## ✅ Checklist

- [ ] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [ ] If necessary, I have added documentation related to the changes made.
- [ ] I have added or updated the tests related to the changes made.

PR #3984, opened 2023-03-12
feat(react-query): Add support for QueryClient.ensureQueryData and ensureData helper

kamilogorek:
Closes https://github.com/trpc/trpc/issues/3933

## 🎯 Changes

- Added support for https://tanstack.com/query/v4/docs/react/reference/QueryClient#queryclientensurequerydata with a new `ensureData` context helper - https://trpc.io/docs/useContext#helpers
- Bumped `@tanstack/react-query` to `^4.18.0` as this is version where `ensureQueryData` was added
- Removed unnecessary `pnpm.overrides.overrides.@tanstack/react-query` which was initially set in https://github.com/trpc/trpc/pull/2852
- Moved `nextjs` and `react-query` `peerDep` to `^4.18.0`, however personally I'd rather change it to `4.x` instead. We've had plenty of reports about `peerDep` from various environments in Sentry SDK and after years of experimentation this seems to be the most reliable and flexible way, eg.
  - https://github.com/getsentry/sentry-javascript/blob/63ef40e5f856e29b344bb577036c88e00d0264d5/packages/nextjs/package.json#L39-L43
  - https://github.com/getsentry/sentry-javascript/blob/63ef40e5f856e29b344bb577036c88e00d0264d5/packages/react/package.json#L25-L27
  - https://github.com/getsentry/sentry-javascript/blob/63ef40e5f856e29b344bb577036c88e00d0264d5/packages/remix/package.json#L41-L45

## ✅ Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
   },
   "pnpm": {
     "overrides": {
+      "@tanstack/react-query": "4.18.0",
       "@trpc/client": "link:./packages/client",
       "@trpc/next": "link:./packages/next",
       "@trpc/react-query": "link:./packages/react-query",
--- a/package.json
+++ b/package.json
@@ -70,7 +70,6 @@
   },
   "pnpm": {
     "overrides": {
-      "@tanstack/react-query": "4.6.0",
       "@trpc/client": "link:./packages/client",
       "@trpc/next": "link:./packages/next",
       "@trpc/react-query": "link:./packages/react-query",
--- a/packages/tests/server/react/ensureQueryData.test.tsx
+++ b/packages/tests/server/react/ensureQueryData.test.tsx
@@ -1,0 +1,75 @@
+import { createQueryClient } from '../__queryClient';
+import { createAppRouter } from './__testHelpers';
+import { QueryClientProvider, useQueryClient } from '@tanstack/react-query';
+import { render, waitFor } from '@testing-library/react';
+import React, { useEffect, useState } from 'react';
+
+let factory: ReturnType<typeof createAppRouter>;
+beforeEach(() => {
+  factory = createAppRouter();
+});
+afterEach(async () => {
+  await factory.close();
+});
+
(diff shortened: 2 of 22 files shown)


package.json
```diff
@@ -70,7 +70,6 @@
   },
   "pnpm": {
     "overrides": {
-      "@tanstack/react-query": "4.6.0",
```
juliusmarminge, line 73: Pin this at 4.18 now maybe?

packages/tests/server/react/ensureQueryData.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { render, waitFor } from '@testing-library/react';
+import React, { useEffect, useState } from 'react';
+
+let factory: ReturnType<typeof createAppRouter>;
+beforeEach(() => {
+  factory = createAppRouter();
+});
+afterEach(async () => {
+  await factory.close();
+});
```
juliusmarminge, line 13: We usually use `konn` for tests

juliusmarminge (commented):
Looks great! Left some small notes

package.json
kamilogorek: Updated

packages/tests/server/react/ensureQueryData.test.tsx
kamilogorek: From what I see, there are still plenty of tests using the factory, so I reused the same style here.

```
- setQueryData
- setInfiniteQueryData
- prefetchQuery
- infiniteQuery
- invalidateQueries
```

They all use it, and because `ensureQueryData` is about the same part of the API, I made the test the same as well. Didn't want it to diverge.

juliusmarminge: Ah ok :) great then

juliusmarminge (approved):

KATT (approved):

Issue #4004, opened 2023-03-15
feat: support specifying optional `host` on `createHTTPServer`


balupton:
### Describe the feature you'd like to request

Having the ability to specify which host to listen to will increase [server environment support](https://github.com/bevry/hostenv), and allow specific binding to say `127.0.0.1` or `localhost` or `192.168.x.x`, or `0.0.0.0`.

### Describe the solution you'd like to see

See https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback for the necessary call, which would change https://github.com/trpc/trpc/blob/8abd05abd31d87798917ec03a028870a998bfc52/packages/server/src/adapters/standalone.ts#L49 along with adding the ability to provide the optional host method

### Describe alternate solutions

unsure

### 👨‍👧‍👦 Contributing

- [x] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature! **- maybe... would need to know how to update the types, and tests if need be**

Nick-Lucas:
Good suggestion, it's a pretty simple change if you'd like to open a PR?

You'll find it currently hard-coded in `packages/server/src/adapters/standalone.ts` so it's a few lines and possibly some test/docs tweaks :)

balupton:
Gonna do this now.

kamilogorek:
Will be fixed in v11 via https://github.com/trpc/trpc/pull/4014
Right now you can achieve this by calling `createHTTPServer` but then, instead of `.listen` on returned object, go indirectly through the serve, `createHTTPServer().server.listen(port, host)`.

PR #4014, opened 2023-03-17
chore(server): (breaking) return an instance of `http.Server` from `createHTTPServer` in standalone adapter

kamilogorek:
Closes https://github.com/trpc/trpc/issues/4004

## 🎯 Changes

Refactor `createHTTPServer` to return an instance of Node's `http.Server` directly, instead of through `.server` property with an additional `listen` method on top of it.
This way we'll be able to encourage users to use built-in API directly and allow for usecases like the one with custom host in #4004

## ✅ Checklist

### DIFF
--- a/packages/server/src/adapters/standalone.ts
+++ b/packages/server/src/adapters/standalone.ts
@@ -19,14 +19,9 @@ export function createHTTPHandler<TRouter extends AnyRouter>(
   opts: CreateHTTPHandlerOptions<TRouter>,
 ) {
   return async (req: http.IncomingMessage, res: http.ServerResponse) => {
-    // if no hostname, set a dummy one
-    const href = req.url!.startsWith('/')
-      ? `http://127.0.0.1${req.url}`
-      : req.url!;
-
-    // get procedure path and remove the leading slash
-    // /procedure -> procedure
-    const path = new URL(href).pathname.slice(1);
+    // Get procedure path and remove the leading slash, `/procedure -> procedure`
+    // Use dummy hostname if one is not provided.
+    const path = new URL(req.url!, 'http://127.0.0.1').pathname.slice(1);
 
     await nodeHTTPRequestHandler({
       ...opts,
@@ -41,18 +36,5 @@ export function createHTTPServer<TRouter extends AnyRouter>(
   opts: CreateHTTPHandlerOptions<TRouter>,
 ) {
   const handler = createHTTPHandler(opts);
-  const server = http.createServer((req, res) => handler(req, res));
-
-  return {
-    server,
-    listen: (port?: number) => {
-      server.listen(port);
-      const actualPort =
-        port === 0 ? ((server.address() as any).port as number) : port;
-
-      return {
-        port: actualPort,
-      };
-    },
-  };
+  return http.createServer((req, res) => handler(req, res));
 }
--- a/examples/standalone-server/src/server.ts
+++ b/examples/standalone-server/src/server.ts
@@ -72,7 +72,7 @@ const appRouter = router({
 export type AppRouter = typeof appRouter;
 
 // http server
-const { server, listen } = createHTTPServer({
+const server = createHTTPServer({
   router: appRouter,
   createContext,
 });
@@ -88,4 +88,4 @@ applyWSSHandler<AppRouter>({
 // setInterval(() => {
 //   console.log('Connected clients', wss.clients.size);
 // }, 1000);
-listen(2022);
+server.listen(2022);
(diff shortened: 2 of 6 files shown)


www/docs/server/adapter/standalone.md
```diff
@@ -130,7 +130,7 @@ The `middleware` option will accept any function which resembles a connect/node.
 
 ## Going further
 
-If `createHTTPServer` isn't enough you can also use the standalone adapter's `createHTTPHandler` function to create your own HTTP Server. For instance:
+`createHTTPServer` is returning an instance of Node's built-in `http.Server`(https://nodejs.org/api/http.html#class-httpserver), which means that you have an access to all it's properties and APIs. However, if `createHTTPServer` isn't enough for your usecase, you can also use the standalone adapter's `createHTTPHandler` function to create your own HTTP server. For instance:
```
Nick-Lucas, line 133:
```suggestion
`createHTTPServer` returns an instance of Node.js's built-in [http.Server](https://nodejs.org/api/http.html#class-httpserver), giving you access to all it's properties and APIs. However, if `createHTTPServer` isn't enough for your use-case, you can also use the standalone adapter's `createHTTPHandler` function to create your own HTTP server. For instance:
```

packages/server/src/adapters/standalone.ts
```diff
@@ -19,14 +19,9 @@ export function createHTTPHandler<TRouter extends AnyRouter>(
   opts: CreateHTTPHandlerOptions<TRouter>,
 ) {
   return async (req: http.IncomingMessage, res: http.ServerResponse) => {
-    // if no hostname, set a dummy one
-    const href = req.url!.startsWith('/')
```
Nick-Lucas, line 23: I'm unsure the ramifications of changing this bit?

Nick-Lucas:
I think the public API here is much improved. Not much reason to write facades when we can write factories :) 

I'm unsure about some of the implementation details, or how it solves #4004

packages/server/src/adapters/standalone.ts
kamilogorek: There are none, it's effectively the same thing, but using `URL` constructor directly.

Nick-Lucas: Cool cool :)

kamilogorek:
> I'm unsure about some of the implementation details, or how it solves https://github.com/trpc/trpc/issues/4004?

Currently you are only able to pass `port` into `listen()` method exposed by the `createHTTPServer` factory.
With this change, you'll be able to write:

```js
const server = createHTTPServer();

server.listen(1337, '192.168.0.20');
// or
server.listen({ port: 1337, host: '192.168.0.20' });
```


Note that it's already possible to write:

```js
const { server } = createHTTPServer();

server.listen(1337, '192.168.0.20');
// or
server.listen({ port: 1337, host: '192.168.0.20' });
```

but it's creating unnecessary indirection, as users might not understand what's the difference between `createHTTPServer().listen()` and `createHTTPServer().server.listen()`.
This way we're making it obvious and support all capabilities of built-in server API directly.

Nick-Lucas:
Brilliant, big thumbs up from me!

juliusmarminge (approved):

juliusmarminge:
I'll do a next-merge and see if the internal types may be fixed already

packages/server/src/adapters/standalone.ts
```diff
@@ -41,18 +36,5 @@ export function createHTTPServer<TRouter extends AnyRouter>(
...
-      server.listen(port);
-      const actualPort =
-        port === 0 ? ((server.address() as any).port as number) : port;
-
-      return {
-        port: actualPort,
-      };
-    },
-  };
+  return http.createServer((req, res) => handler(req, res));
```
juliusmarminge, line 39: this is now an extremely slim wrapper over the built-in. is it worth having it and not just mention in docs how to use createHTTPHandler with node.httpServer?

kamilogorek: In the perfect world yes, but it’d also break all current users of standalone adapter. Maybe we can soft-deprecate it in v11 and remove in v12?

I wonder if there might be some future improvements we’d like to add directly in here though 🤔

Nick-Lucas: This is the easiest way for a newb to stand up a tRPC server, and will work for most deployments, so I think it's worth keeping. When it comes to documentation alone this is the best starting point, and since it's such a skinny wrapper it's no maintenance problem really!

juliusmarminge: Yea ik and I like having this simple option! But we do keep a handful of test for it that could be removed(?).

But yea we shouldn't just remove it without first deprecating it. Let's make this change and we can continue the discussion later

KATT (approved):

kamilogorek:
> I'll do a next-merge and see if the internal types may be fixed already

@juliusmarminge it indeed helped 😌

PR #4016, opened 2023-03-17
feat: added batch capabilities to procedures

i-Welch:
Note: Currently In progress. Opening this for feedback

## 🎯 Changes

This is a feature. I would like to add the ability to specify a procedure as a "batch" procedure. When a procedure is labeled a "batch" procedure (using the .batch()) function and the procedure builder. instead of getting just one input in the handler you will get an array of inputs. You will also be expected to return an array of the same length as the returned data to all the functions you called from the client. Currently the happy path works well as well as using it in conjunction with Input() and Output(). If there is interest I can clean up the last few things I'd need to handle before this code is ready for a merge.

In order to test this you can create a simple sandbox where you mark a procedure as batch like so:
image

## 💡 Rationale

The Rationale for this change is that with the automatic batch that is done on the client side often times it would be more performant to handle those requests all at once instead of one at a time for because you can batch API calls, database calls, etc. This would provide a very simplistic API so that you can continue to leverage the automatic batching done by TRPC on the client side while also taking advantage of it on the server side to get better performance or enable some other behavior that you may desire. This feature is completely opt in on the part of the user of course and should have no backwards breaking changes.

## ✅ Checklist

✅ I have followed the steps listed in the [Contributing guide (https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [ ] Handle error cases
- [ ] Clean up Procedure builder types

### DIFF
--- a/examples/batch-example/Readme.md
+++ b/examples/batch-example/Readme.md
@@ -1,0 +1,20 @@
+Batch Example.
+Run by starting the server with npm run dev in the ./server directory and starting the client by running dev in the ./client directory. This should start a project that shows several ways of using trpc to collect a Name and an Email in a solid.js client app (solid.js only being used as an example here but similar patterns are avialable is all modern frameworks and very common using context API's in certain cases).
+
+There are 4 differnt patterns showns. I'll describe the Pro's and Con's as follows:
+
+1. This is labeled the Naive approach. It follows a pattern of Inputs bound with updaters to the data which they update. They implement the data updaters using TRPC in this case. Trpc automatically batches the API calls into one request but the server operates on those requests individually. This is often fine for most usecases, especially early on in the lifecycle of an application and any deviation from this would be considered an optimization (likely to happen later in the lifecycle of an application if ever needed at all)
+
+2. This uses the new Batch functionaly that I'm proposing. Note that the front-end client code can stay exactly the same, but the server code needs to be updated to add the expected batching functionality. The types automatically reflect this needed change on the server while allows the front-end to remain as it was. A benefit there being that this is a completely opt in optimization only neccesarry when the app grows to a complexity where NOT doing this would be an issue (likely fairly late in the development cycle)
+
+3. This uses a manual batching operation (changing the input type to an array of updates). This requires the same change that number 2 does to take advantage of this batching but also requires significant changes to the client code. Whether this client code is better or worse is IMO up to interpritation and style. The key point is that it does require a significant change to the client code in this situation and while this code is simple in more mature codebases the refactor may be non-trivial.
+
+4. This seeks to essentially use the inteface from number 3 but provide a DX similar to number 2. The main drawback here is that this type of code is complex to maintain and replicates essentially very similar code to what TRPC is already doing under the hood to batch requests at the HTTP level.
+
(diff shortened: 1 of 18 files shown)


KATT:
I love the work you've put into this PR, but I think we don't wanna go in this direction. 

I don't understand the motivation why - you can already get the same benefits in tRPC if you use a [dataloader](https://github.com/graphql/dataloader) in your data-layer; in fact, you actually get the benefits of this [**for free** with tRPC + Prisma already](https://www.prisma.io/docs/guides/performance-and-optimization/query-optimization-performance) if you're using `.findUnique()`.

KATT:
cc @i-Welch can you share motivation on why you needed this, please?

i-Welch:
@KATT Yes NP. I was actually just creating an example showing off what I'd generally consider one of the stronger use cases. If you look into the /examples folder I created a new example called batch-example. It has a server and a client piece, as well as a ReadMe.md at the top level where I outline my thoughts on the different examples shown in the project. I hope that illustrates the specific use cases in which I think this would be useful.

Just to reiterate something I put at the end of that ReadMe.md thanks for taking the time to review and provide this tech to the community, and if the examples aren't convincing (and I'm under no illusion that I'm incredibly convincing or infallible here) then I'm completely satisfied with the answer either way it goes. Thanks again for taking the time here and if this doesn't seem like its providing much value just let me know and I'll close out the PR.

~ cheers

KATT:
Could you elaborate what problem it solves for you and the application you're working on & why the current batching isn't sufficient in combination with a dataloder?

i-Welch:
Certainly, Take this example of a request diagram on a user interaction
image

You can see here how even though we are batching the requests to TRPC we are handling them sequentially to the 3rd party API. Generally automated data layer batching isn't something I've ever seen over HTTP (some nuance here for certain types of GET requests) and even seldom seen on client libraries which use HTTP as the backing to interact with 3rd party services.

Here is a possible solution to this.
image

This is a valid solution, but in my experience these solutions are often very bug laden and brittle, as well as generally being annoying for developers ergonomically.

Here is another possible solution that is only in user space.
image

I'd say while I think this is easier than the backend equivalent. It does raise its own issues with bugs and dev ergonomics. It also requires the addition of / or modification of the procedure into a batch procedure in many cases via changing input type to an array and then updating logic as needed.

Here is the diagram enabled by the batch changing I'm proposing.
image

I think this is IMO the cleanest solution possible as it removes the pitfalls of the hand rolled batching logic from both the frontend and the backend (in favor of reusing TRPCs on the front-end) but otherwise follows the same benefits gotten from the above example (where the handler can decide how to batch requests and parse responses based on fine-grained user logic). Another benefit here is that is an optimization that is very easy to opt-in to later in a projects lifecycle when optimizations becomes necessary without large or drastic changes to the front-end code.

All in all this is just an optimization though, and not an existential one for any use case as their are user land alternatives possible.

i-Welch:
Thanks for taking a look though! been a pleasure working with you!

KATT:
Thanks for the detailed description! 

For now, we won't add this, but let us know if you have a real need for this in the future. I've made a reference in #4073 on how to add dataloaders in tRPC.

Issue #3947, opened 2023-03-08
docs: enhance quickstart guide


Nick-Lucas:
### Area of Improvement

The quickstart guide could use some detail in a number of areas which aren't mentioned yet, and to go deeper in some existing areas. I plan on on contributing this. 

Getting Started Docs like Remix has are much more comprehensive. Some rough planning needs doing which I will put here, but any ideas from others will be very welcome too.

A community member has also recently posted this which structurally looks like great inspiration for where we need to end up: https://ajcwebdev.com/2023/03/08/2023-03-08-a-first-look-at-trpc/ 

### Link to related docs

https://trpc.io/docs/quickstart

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR implementing the suggested changes!

---

### The current docs have these sections:

- [Installation](https://trpc.io/docs/quickstart#installation)
  - [Installation Snippets](https://trpc.io/docs/quickstart#installation-snippets)
- [Defining a backend router](https://trpc.io/docs/quickstart#defining-a-backend-router)
  - [Create a router instance](https://trpc.io/docs/quickstart#create-a-router-instance)
  - [Add a query procedure](https://trpc.io/docs/quickstart#add-a-query-procedure)
  - [Add a mutation procedure](https://trpc.io/docs/quickstart#add-a-mutation-procedure)
- [Using your new backend on the client](https://trpc.io/docs/quickstart#using-your-new-backend-on-the-client)
  - [Setup the tRPC Client](https://trpc.io/docs/quickstart#setup-the-trpc-client)
  - [Querying & mutating](https://trpc.io/docs/quickstart#querying--mutating)
  - [Full autocompletion](https://trpc.io/docs/quickstart#full-autocompletion)
- [Next steps](https://trpc.io/docs/quickstart#next-steps)

### Where I think we should be (_WIP_)

Removing this list for now since I will be maintaining #3963 for a higher level plan then coming back to this

<sub>[TRP-59](https://linear.app/trpc/issue/TRP-59/docs-enhance-quickstart-guide)</sub>

## Funding
* You can sponsor this specific effort via a [Polar.sh](https://polar.sh) pledge below
* We receive the pledge once the issue is completed & verified

<a href="https://polar.sh/trpc/trpc/issues/3947">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/3947/pledge.svg?darkmode=1">
[...]

PR #4036, opened 2023-03-20
docs: update quickstart

juliusmarminge:
Ref #3947

## 🎯 Changes

Link: https://www-git-quickstart-trpc.vercel.app/docs/quickstart

## ✅ Checklist


examples/quickstart/README.md
KATT, line 1: Could we just repurpose the `examples/standalone` here? Seems like the same thing

www/docs/main/quickstart.mdx
```diff
@@ -327,22 +425,61 @@ import type { AppRouter } from './server';
...
+import clsx from 'clsx';
+
+<div
+  className={clsx(
+    'h-[800px] w-full rounded-xl overflow-hidden z-10 relative my-0 md:my-4 lg:my-8',
+  )}
+>
+  <Iframe
+    src={
+      `https://stackblitz.com/github/trpc/trpc/tree/quickstart/examples/minimal?` +
```
juliusmarminge, line 453:
```suggestion
      `https://stackblitz.com/github/trpc/trpc/tree/main/examples/minimal?` +
```

examples/minimal/client/index.ts
```diff
@@ -1,19 +1,33 @@
 import { createTRPCProxyClient, httpBatchLink } from '@trpc/client';
 import type { AppRouter } from '../server';
 
-const client = createTRPCProxyClient<AppRouter>({
+// Add polyfill for fetch for Node < 18. (Stackblitz runs Node 16 and we want embed to work)
```
kamilogorek, line 4: Sad panda

examples/minimal/client/index.ts
```diff
@@ -1,19 +1,33 @@
...
+  //    ^?
+  console.log('Users:', users);
+
+  const createdUser = await trpc.userCreate.mutate({ name: 'sachinraja' });
+  //    ^?
+  console.log('Created user:', createdUser);
 
-  // Type safe
-  console.log(result.greeting.toUpperCase());
+  const user = await trpc.userById.query('1');
```
kamilogorek, line 28: This is still broken afaik. It'll report `User`, despite the fact that it is in fact `User | undefined`.
It's only broken with `createTRPCProxyClient`, not `createCaller`, as default transform type is making the query output to be passed through `Serialize` type, which removes the `undefined` part.

kamilogorek: ref: https://github.com/trpc/trpc/issues/4049#issuecomment-1484868645

Nick-Lucas:
Have left some comments on Vercel for the docs. The commenting there doesn't seem to like commenting on headings, so I'd recommend pulling up the Inbox sidebar for the few which aren't pinning properly.

I think one area worth doing another pass on is concepts, my main thoughts:

* We mention other tools like REST and GraphQL multiple times, but never mention RPC. We should really be selling the advantages of RPC, rather than explaining how it's similar to other things (I would drop comparisons from quickstart entirely, we can have a whole page for that)
* We mention HTTP Verbs a few times - I wouldn't bother, RPC is an abstraction and for quickstart, users should learn the abstraction, not the internals. Again we can and do reference HTTP Verbs in other parts of the docs, internals understanding just comes later is all

www/docs/main/quickstart.mdx
```diff
@@ -10,239 +10,347 @@ import CodeBlock from '@theme/CodeBlock';
...
+// ---cut---
+const appRouter = router({
+  // ...
+  userById: publicProcedure
+    // The input is unknown at this time. A client could have sent
+    // us anything so we won't assume a certain data type.
     .input((val: unknown) => {
       // If the value is of type string, return it.
-      // TypeScript now knows that this value is a string.
+      // It will now be inferred as string.
```
Nick-Lucas, line 262:
```suggestion
      // It will now be inferred as a string.
```

www/docs/main/quickstart.mdx
```diff
@@ -10,239 +10,347 @@ import CodeBlock from '@theme/CodeBlock';
...
+userList: () => User[];
+userById: (id: string) => User;
+userCreate: (data: { name: string }) => User;
 ```
 
-### Create a router instance
+### 1. Create a router instance
 
-First, let's define an empty router in our server codebase:
+First, let's initilize the tRPC backend. It's good convention to do this in a separate file and export reusable helper functions instead of the entire tRPC object.
```
Nick-Lucas, line 169: :::warning
If you choose a different file structure than we recommend, then beware of [circular dependencies](https://stackoverflow.com/questions/46589957/es6-modules-and-circular-dependency). Typically `appRouter` needs `t` and `yourOwnSubRouter`, and `yourOwnSubRouter` needs `t` too. So if `t` and `appRouter` are in the same file you may end up with a strange error like `t is undefined` at run-time, because both files are importing each other and both can't be initialised already.
:::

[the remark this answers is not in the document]
Nick-Lucas: Might want to go lower down below the concepts it discusses, but I think we should address this head-on

KATT (dismissed):
This is awesome. Thanks for doing this @juliusmarminge. I updated the color coding to be closer to our `loggerLink` :)

KATT (approved):

KATT (dismissed):

KATT (approved):

PR #4037, opened 2023-03-20
chore: add .npmrc to fix render.com blueprint on next-prisma-starter

lucasheriques:
## 🎯 Changes

Right now, the build process on the application depends on running the `prebuild` script, which generates the Prisma client and executes the migrations.

However, by default `pnpm` [does not run pre/post scripts](https://github.com/pnpm/pnpm/issues/2891). To fix that, we could move the `prebuild` script to `build`, or, create this `.npmrc` file with this content:

```
enable-pre-post-scripts=true
```

This PR adds this file. I think it doesn't make sense to move away from pre/post scripts since it's what everyone is kinda used to.

## ✅ Checklist

### DIFF
--- a/examples/next-prisma-starter/.npmrc
+++ b/examples/next-prisma-starter/.npmrc
@@ -1,0 +1,1 @@
+enable-pre-post-scripts=true


lucasheriques:
Hey @juliusmarminge, my reason for doing this is that when creating a new tRPC project with the [examples-next-prisma-starter](https://github.com/trpc/examples-next-prisma-starter)](https://github.com/trpc/examples-next-prisma-starter), I used Render.com to deploy it, using the blueprint.

However, the build failed the time it ran, because the database was not created, since the commands `prisma generate && prisma migrate` didn't run. I wasn't really aware of turborepo, since the example does not use that.

For a real app, I'm imagining that it is good to run it every time - it guarantees that the DB is up-to-date with the schema, and applies changes if necessary. Otherwise, it just returns that there are no pending migrations.

But, I'm very new to Prisma and this stack itself, so please bear with me because I might be wrong on this 😅

PR #4065, opened 2023-03-23
fix(next-ws-example): Change DATABASE_URL port to match docker-compose

HartS:


KATT (approved):
I actually think we either

a) Remove docker from this example completely, or
b) Add redis to the docker (and render.yaml) and utilize it in the app

HartS:
@KATT I guess you didn't see my post in discord, I made changes locally to use sqlite instead of docker. Want me to make a PR for them?

Basically the only *real* change here is removing the following from the schema:

```
enum PosterSource {
  RAW
  GITHUB
}
...
model Post {
  source PosterSource @default(RAW)
...
```

Because sqlite doesn't support enums.

Then I replaced it with: `isGithub Boolean @default(false)`

and changed the 2 places that `source` property was referenced

juliusmarminge:
Better to change docker compose then cause this is the port we use in the CI postgres container

Nick-Lucas:
Since this is small and has been silent for a month, I'm going to close it out just to keep things tidy. If you want to revisit it with the feedback from above worked in then please do open a fresh PR!

PR #4114, opened 2023-03-30
chore: add prettier-plugin-tailwindcss & eslint cache

QuiiBz:
## 🎯 Changes

Add [eslint-plugin-tailwindcss](https://github.com/francoismassart/eslint-plugin-tailwindcss) for automatic class name sorting, and some utility rules:
- the `no-arbitrary-value` rule is set to warn, we have around 15 occurrences in www. Do we want to turn it off?
- the `no-custom-classname` rule is set to warn by default, which allowed me to fix unused/wrong classes

This PR also adds `--cache` to every eslint call, which greatly improves the speed of running eslint locally (it only reruns on files that have changed)

## ✅ Checklist


www/src/components/Preview.tsx
```diff
@@ -10,7 +10,7 @@ export const Preview = () => {
           muted
           playsInline
           width="1200px"
-          className="border-6 rounded-lg border-[#111111] bg-[#111111] shadow-xl md:hidden"
+          className="rounded-lg border-4 border-[#111111] bg-[#111111] shadow-xl md:hidden"
```
QuiiBz, line 13: `border-6` doesn't exist in tailwind, so I've replaced it with `border-4`. We can also use an arbitrary value instead.

.eslintrc.js
```diff
@@ -8,6 +8,7 @@ const config = {
     'plugin:@typescript-eslint/recommended-requiring-type-checking',
     'plugin:react/recommended',
     'plugin:react-hooks/recommended',
+    'plugin:tailwindcss/recommended',
```
juliusmarminge, line 11: i've only used the prettier-plugin before - how does this compare?

I was thinking of making a PR just now migrating to tailwind 3.3 with TS config, does this support that? We added support to the prettier plugin here: https://github.com/tailwindlabs/prettier-plugin-tailwindcss/pull/137

examples/.interop/next-prisma-starter/package.json
```diff
@@ -14,7 +14,7 @@
     "prebuild": "run-s migrate generate",
     "build": "next build",
     "start": "next start",
-    "lint": "eslint --ext \".js,.ts,.tsx\" --report-unused-disable-directives --report-unused-disable-directives src",
+    "lint": "eslint --cache --ext \".js,.ts,.tsx\" --report-unused-disable-directives --report-unused-disable-directives src",
```
juliusmarminge, line 17: the lint script is cached through turbo, so running pnpm lint in root only lints changed packages - isn't that enough? do we also need to cache the individual scripts?

QuiiBz: Turbo allows us to cache per package, and this allows us to cache per file: running eslint on a package multiple times (without a turbo cache hit) will take a lot less time, as it will only run on changed files instead of on each files every time.

.eslintrc.js
QuiiBz: This eslint plugin is similar to the prettier one (to order class names) but also adds some [interesting rules](https://github.com/francoismassart/eslint-plugin-tailwindcss#supported-rules).

`tailwind.config.ts` doesn't seem to be supported yet, but I'll happily dig into adding support for it: https://github.com/francoismassart/eslint-plugin-tailwindcss/issues/228

examples/.interop/next-prisma-starter/package.json
juliusmarminge: yea i get that - just seems a bit overkill, i rarely run a single lint locally xddd but lets go - more caching is always nice

juliusmarminge:
Just checked this out and was gonna fix the eslint plugin but autosave is suuuuperslow for me now - is this a heavy plugin? Saving a file is like 5sec

juliusmarminge:
Changed to prettier plugin for now, but left the eslint config in there just commented out so we can swap later if it gets faster (im guessing its cause i put the ts-configs in there and it doesn't load those properly)

juliusmarminge:
Hijacked your PR even more and updated prettier too XD

QuiiBz:
> Just checked this out and was gonna fix the eslint plugin but autosave is suuuuperslow for me now

I thought it was a bug on my computer, but sadly yeah it appears that it's kinda slow :(

Does the prettier plugin supports `tailwind.config.ts`?

juliusmarminge:
> Does the prettier plugin supports tailwind.config.ts?

Yes it does! https://github.com/tailwindlabs/prettier-plugin-tailwindcss/pull/137

juliusmarminge (approved):

PR #4141, opened 2023-04-05
UtilsLike type does not support router sub-types

### LINKED ISSUE #4140: bug: router-factory/polymorphism UtilsLike type does not support router sub-types via mergeRouters

See PR in our own codebase/environment

### Describe the bug

We can create a router factory like the below, and generate abstract types from it using our polymorphism types.

```ts
function createMyRouter() {
  return t.router({ foo, bar })
}

type MyRouter = ReturnType<typeof createMyRouter>;

export type MyRouterRouteLike = RouterLike<MyRouter>;

export type MyRouterUtilsLike = UtilsLike<MyRouter>;
```

If we then use it like 

```ts
const someRouter = t.mergeRouters(
  createMyRouter(),
  t.router({
    extraProc: t.procedure
  })
)
```

Now when we try to pass this router's utils to the `UtilsLike` result, it gets rejected as incompatible by TypeScript, similar to below

image

### Link to reproduction

-

### To reproduce

Reproduction in PR test

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

Nick-Lucas:
Closes #4140 

## 🎯 Changes

## ✅ Checklist

### DIFF
--- a/packages/client/src/createTRPCClientProxy.ts
+++ b/packages/client/src/createTRPCClientProxy.ts
@@ -28,66 +28,52 @@ import {
 
 /** @public */
 export type inferRouterProxyClient<TRouter extends AnyRouter> =
-  DecoratedProcedureRecord<TRouter['_def']['record'], TRouter>;
+  DecoratedProcedureRecord<TRouter['_def']['record']>;
 
 type Resolver<TProcedure extends AnyProcedure> = (
   ...args: ProcedureArgs<TProcedure['_def']>
 ) => Promise<inferTransformedProcedureOutput<TProcedure>>;
 
-type SubscriptionResolver<
-  TProcedure extends AnyProcedure,
-  TRouter extends AnyRouter,
-> = (
+type SubscriptionResolver<TProcedure extends AnyProcedure> = (
   ...args: [
     input: ProcedureArgs<TProcedure['_def']>[0],
     opts: ProcedureArgs<TProcedure['_def']>[1] &
       Partial<
         TRPCSubscriptionObserver<
           inferTransformedSubscriptionOutput<TProcedure>,
-          TRPCClientError<TRouter>
+          TRPCClientError<TProcedure>
         >
       >,
   ]
 ) => Unsubscribable;
 
-type DecorateProcedure<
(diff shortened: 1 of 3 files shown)


packages/react-query/src/shared/proxy/utilsProxy.ts
```diff
@@ -107,7 +107,7 @@ type DecorateProcedure<
       predicate?: (
         query: Query<
           inferProcedureInput<TProcedure>,
-          TRPCClientError<TRouter>,
+          TRPCClientError<TProcedure>,
```
Nick-Lucas, line 110: So this was encoding whole router information into the type defs for a single procedure, which broke this one polymorphism corner case. 

Not sure what impacts this will really have though

packages/tests/server/react/polymorphism.test.tsx
```diff
@@ -186,6 +199,63 @@ describe('polymorphism', () => {
...
+      function DiscussionsExportPage() {
+        const utils = trpc.useContext();
+
+        const [currentExport, setCurrentExport] = useState<null | number>(null);
+
+        return (
+          <>
+            <StartExportButton
+              route={trpc.github.discussions.export}
+              utils={utils.github.discussions.export}
```
Nick-Lucas, line 215: This line gave a typescript error

Nick-Lucas:
This commit created the issue: https://github.com/trpc/trpc/commit/b05d207df136303bcbae47fbab55a321fc5dcd15

packages/react-query/src/shared/proxy/utilsProxy.ts
KATT: `TRPCClientError<TProcedure>` is what we should be using everywhere 👌

KATT (dismissed):

packages/react-query/src/shared/proxy/utilsProxy.ts
Nick-Lucas: Computer says 🤔

image

sachinraja (approved):

Issue #4071, opened 2023-03-24
feat: add the operation in the header callback of non-batched http link


tibotiber:
### Describe the feature you'd like to request

This is a follow up to #2018, I'd like to be able to set specific headers per operation when using a non-batch httpLink. 

Context:
I have procedures that can be called publicly or authenticated and will be processed slightly differently on the server. I want to set headers on a per operation basis. I am using httpLink, so no batching involved. The auth context (whether to authenticate or not, as well as a token fetcher) are available in my component (call site) but not via session/cookies/etc., and not at the top of my tree where the httpLink gets configured. So headers being async is great, but I'd love to have the operation's context available there as well.

### Describe the solution you'd like to see

This is what existed in #2024 prior to v10 and was shelved at the time.

https://github.com/trpc/trpc/pull/2024/files#diff-671e79295e2c31c75769e93c3c10ebaea24b2dc466d3a82d0865548a8a1e318dR10-R16

```
type GetHeadersFn = (opts: {
  op: Operation<unknown>;
}) => HTTPHeaders | Promise<HTTPHeaders>;
```

### Describe alternate solutions

@KATT mentioned

> you should be able to work around this by making a custom link that returns a httpLink with the headers you want. I realise this isn't ideal, but could get it working.

but i'm not entirely sure how that would go as the links are setup at a time when my header values do not exist.

Another way:
Thinking about the problem in a different way (might be a really bad idea). Would providing a custom fetcher option on @trpc/react-query be a solution? A bit similar to how graphql-codegen's react-query implementation lets users provide a custom fetcher (https://the-guild.dev/graphql/codegen/plugins/typescript/typescript-react-query#fetcher).

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

skovhus:
Example code

```typescript
const terminationLinkExample: TRPCLink<RootRouter> = () => {
    return (runtime) => ({next, op}) => {
        return observable((observer) => {
            const unsubscribe = next(op).subscribe({
                next(value) {
                    parseOperationalContext(op)
                        .then(({accessToken, skipBatching}) => {
                            // FIXME: conditionally return httpLink and httpBatchLink
                            const headers = accessToken
                              ? { Authorization: `Bearer ${accessToken}` }
                              : {};

                            const link = skipBatching
                              ? httpBatchLink
                              : httpLink;

                            // But how do we correctly call the link? And how do we pass the runtime?
[...]

PR #4175, opened 2023-04-12
feat: add operations in `httpLink`/`httpBatchLink` `headers`-callback function

skovhus:
Closes https://github.com/trpc/trpc/issues/4071

## 🎯 Changes

This PR adds support for using operations in as part of the `HTTPLinkOptions.headers` used by `httpLink` and `httpBatchLink`. 

The API changes are non-breaking:
```diff
export interface HTTPLinkOptions {
  url: string;
  /**
   * Add ponyfill for fetch
   */
  fetch?: typeof fetch;
  /**
   * Add ponyfill for AbortController
   */
  AbortController?: typeof AbortController | null;
  /**
   * Headers to be set on outgoing requests or a callback that of said headers
   * @link http://trpc.io/docs/v10/header
   */
  headers?: 
    | HTTPHeaders 
-   | (()                           => HTTPHeaders | Promise<HTTPHeaders>);
+   | ((opts: { ops: Operation[] }) => HTTPHeaders | Promise<HTTPHeaders>);
}
```

Questions:
- Are we okay returning a list of operations when using the `httpLink`? If not how do we go about making the change non breaking? From a user point of view I actually like that both headers functions takes a list, this means I can share a headers function.  
- it seems we usually use abbreviations like `op` instead of `operations` in the external facing API (at least when building a custom link). I would prefer to be explicit here and use `operations`, but consistency is also a concern.

## ✅ Checklist

### DIFF
--- a/packages/client/src/links/httpBatchLink.ts
+++ b/packages/client/src/links/httpBatchLink.ts
@@ -34,27 +34,50 @@ export function httpBatchLink<TRouter extends AnyRouter>(
           // escape hatch for quick calcs
           return true;
         }
+        const path = batchOps.map((op) => op.path).join(',');
         const inputs = batchOps.map((op) => op.input);
+
+        // TODO: dry this up
+        const resolveHeaders = () => {
+          const { headers } = opts;
+          if (typeof headers === 'function') {
+            return headers({ ops: batchOps });
+          }
+          return {};
+        };
 
         const url = getUrl({
           ...resolvedOpts,
           runtime,
           type,
+          path,
           inputs,
-          ops: batchOps,
+          resolveHeaders,
         });
+
         return url.length <= maxURLLength;
       };
 
       const fetch = (batchOps: BatchOperation[]) => {
+        const path = batchOps.map((op) => op.path).join(',');
         const inputs = batchOps.map((op) => op.input);
+
+        // TODO: dry this up
+        const resolveHeaders = () => {
+          const { headers } = opts;
+          if (typeof headers === 'function') {
+            return headers({ ops: batchOps });
+          }
+          return {};
+        };
 
         const { promise, cancel } = httpRequest({
           ...resolvedOpts,
           runtime,
           type,
+          path,
           inputs,
-          ops: batchOps,
+          resolveHeaders,
         });
 
         return {
--- a/packages/client/src/links/httpBatchLink.ts
+++ b/packages/client/src/links/httpBatchLink.ts
@@ -12,6 +12,8 @@ import {
 import { transformResult } from './internals/transformResult';
 import { Operation, TRPCLink } from './types';
 
+type BatchOperation = Operation;
+
 export interface HttpBatchLinkOptions extends HTTPLinkOptions {
   maxURLLength?: number;
 }
@@ -22,8 +24,6 @@ export function httpBatchLink<TRouter extends AnyRouter>(
   const resolvedOpts = resolveHTTPLinkOptions(opts);
   // initialized config
   return (runtime) => {
-    type BatchOperation = { id: number; path: string; input: unknown };
-
     const maxURLLength = opts.maxURLLength || Infinity;
 
     const batchLoader = (type: ProcedureType) => {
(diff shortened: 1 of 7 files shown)


skovhus:
@KATT is this on the right track of what you imagined?

packages/client/src/links/internals/httpUtils.ts
```diff
@@ -23,7 +23,7 @@ export interface HTTPLinkOptions {
    * Headers to be set on outgoing requests or a callback that of said headers
    * @link http://trpc.io/docs/v10/header
    */
-  headers?: HTTPHeaders | (() => HTTPHeaders | Promise<HTTPHeaders>);
+  headers?: HTTPHeaders | ((opts: {ops: Operation[]}) => HTTPHeaders | Promise<HTTPHeaders>);
```
skovhus, line 26: Here we reuse the `ops: Operation[]` for both batched and non regular http links. This should likely be changed once we have something working.

packages/client/src/links/httpBatchLink.ts
```diff
@@ -41,6 +43,7 @@ export function httpBatchLink<TRouter extends AnyRouter>(
           type,
           path,
           inputs,
+          ops: batchOps as Operation[], // FIXME: is a BatchOperation just an Operation?
```
skovhus, line 46: I'm not sure if `BatchOperation` is just an `Operation`.

packages/client/src/links/internals/httpUtils.ts
```diff
@@ -87,6 +94,7 @@ export type HTTPRequestOptions = ResolvedHTTPLinkOptions &
   GetInputOptions & {
     type: ProcedureType;
     path: string;
+    ops: Operation[]; // TODO: if we do this, then we should get rid of type and path
```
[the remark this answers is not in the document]
skovhus: An alternative would be to resolve the headers before calling `httpRequest` or `getUrl`, then we hide the operation from them.

packages/client/src/links/internals/httpUtils.ts
skovhus: A potentially nicer interface for the user was suggested by @KATT [here](https://github.com/trpc/trpc/issues/4071#issuecomment-1505130985) where only one operation is passed to HttpLink. 

If we want that I guess we would introduce two new types `HttpLinkOptions` and `HttpBatchLinkOptions`?

skovhus: After testing this out on our app, I actually like that we return a list of operations for both links. It means that I can share the same headers function between the http link and the batched one.

juliusmarminge:
> Are we okay returning a list of operations when using the httpLink? If not how do we go about making the change non breaking? From a user point of view I actually like that both headers functions takes a list, this means I can share a headers function.

Can we make the type non-empty so that a destructuring would return a defined op and not `op | undefined` or is there any cases where this isn't the case?

skovhus:
> Can we make the type non-empty so that a destructuring would return a defined op and not op | undefined or is there any cases where this isn't the case?

The will always be 1 operation for `httpLink` and 1-n for `httpBatchLink`, so making it non-empty makes a lot of sense.

packages/client/src/links/internals/httpUtils.ts
```diff
@@ -19,33 +22,21 @@ export interface HTTPLinkOptions {
    * Add ponyfill for AbortController
    */
   AbortController?: AbortControllerEsque | null;
-  /**
-   * Headers to be set on outgoing requests or a callback that of said headers
-   * @link http://trpc.io/docs/v10/header
-   */
-  headers?: HTTPHeaders | (() => HTTPHeaders | Promise<HTTPHeaders>);
```
KATT, line 26: Hmm..... maybe this is considered breaking change 🤔

KATT: Naaahhh, it's in `/internals` so it should be fine to change

packages/client/src/links/internals/httpUtils.ts
KATT: I prefer to keep it separate, you can do a lower-level callback in both that is like `const getHeaders = (op: Operation) => { .... }`

PR #4187, opened 2023-04-14
fix: undefined stripped from query result

### LINKED ISSUE #4049: bug: returning `undefined` from server gets stripped on the client
### Area of Improvement

Please add an explanation to Quickstart of why `user` has type `User | undefined` on the server:

```ts
.query((req) => {
  const input = req.input;
  const user = userList.find((it) => it.id === input);
 
  // TS error if we try accessing user.id here
  user.id
 
   return user;
}),
```

But the client code has a different type which is equal to `User`, but `undefined` type is missing:

```ts
const user = await trpc.userById.query('1');
// hello, runtime error!
user.id;
```

Is it intentional, and how to fix it?

### Link to related docs

https://trpc.io/docs/quickstart#querying--mutating

### Additional information

@trpc/server: 10.17.0
typescript: 4.9.5

have "strict": true in tsconfig

### 👨‍👧‍👦 Contributing

- [ ] 🙋‍♂️ Yes, I'd be down to file a PR implementing the suggested changes!

juliusmarminge:
Closes #4049

## 🎯 Changes

## ✅ Checklist

### DIFF
--- a/packages/server/src/shared/internal/serialize.ts
+++ b/packages/server/src/shared/internal/serialize.ts
@@ -7,6 +7,10 @@ import { FilterKeys } from '../../types';
 type JsonPrimitive =
   | string
   | number
+  // `undefined` is a weird one that's technically not valid JSON,
+  // but the return value of `JSON.parse` can be `undefined` so we
+  // support it as both a Primitive and a NonJsonPrimitive
+  | undefined
   | boolean
   // eslint-disable-next-line @typescript-eslint/ban-types
   | String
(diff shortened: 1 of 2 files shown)


sachinraja (approved):

sachinraja:
good find! I think this is actually happening because of [how tRPC encodes result data](https://trpc.io/docs/rpc#http-response-specification). So `JSON.parse` works because even if you return `undefined`, your response data will look like `{"result":{}}`

PR #4200, opened 2023-04-16
Add experimental chaining object for .use

### LINKED ISSUE #4191: feat: Support typesafe inputs in `t.middleware` created middlewares

This is based on #4190 

Currently we can create a typesafe input within a middleware like this:

image

However the de-facto way to create a re-usable and composable middleware is with `t.middleware`, not `procedure.use()`. 

When using the more composable approach we give up the type-safety:

image

It would be a good step for composition to support type-safe inputs via `t.middleware`

There are a couple possibilities:

#### TypeScript interface

Allow `t.middleware` to take a generic parameter defining a mandatory input interface which must be met when it's used:

```ts
const orgMiddleware = t.middleware<{ orgId: number }>((opts) => {
      if (checkOrg(opts.input.orgId)) {
        // do checks and throw if needed

        return opts.next();
      }

      throw new TRPCError({
        code: 'FORBIDDEN',
      });
})

const scopedProcedure = publicProcedure
  .input(z.object({ orgId: z.number() }))
  .use(orgMiddleware)

const scopedProcedure = authProcedure
  .input(z.object({ orgId: z.number() }))
  .use(orgMiddleware
```

#### Zod 'preprocess' style 

```ts
const orgMiddleware = t.middleware<{ orgId: number }>((opts) => {
      if (checkOrg(opts.input.orgId)) {
        // do checks and throw if needed

        return opts.next();
      }

      throw new TRPCError({
        code: 'FORBIDDEN',
      });
  },
  // we pass a second property for the input parser, it will stack and merge just like .input does over multiple calls
  z.object({ orgId: z.number() })
)

const scopedProcedure = publicProcedure
  .use(orgMiddleware)

const scopedProcedure = authProcedure
  .use(orgMiddleware
```

Alternatively we simply document a workaround for the issue, for instance this will provide the type safety to the client, and to the internals but is more verbose and doesn't enforce the use of the input :

```ts
const OrgObj = z.object({ orgId: z.number() })

const orgMiddleware = t.middleware<{ orgId: number }>((opts) => {
      const { orgId } = OrgObj.parse(opts.input)

      if (checkOrg(orgId)) {
        // do checks and throw if needed

        return opts.next();
      }

      throw new TRPCError({
        code: 'FORBIDDEN',
      });
})

const scopedProcedure = publicProcedure
  .input(OrgObj)
  .use(orgMiddleware)

const scopedProcedure = authProcedure
  .input(OrgObj)
  .use(orgMiddleware

// This will very likely blow up at runtime and not provide the type safety we generally go for!
const scopedProcedure = otherProcedure
  .use(orgMiddleware
```

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

Nick-Lucas:
Closes #4191 

#4191 requests a way for a middleware to extend a baseProcedure with some required input data

ALSO we would _love_ to make it easier for 3rd party vendors such as authentication, authorisation, logging, crashlytics, analytics, etc, tools to provide 1-line integrations without knowing _anything about the contents of `t`_ (which is currently impossible

ALSO we would love for monorepo owners to be able to write middlewares in a completely API agnostic way, for the same reasons of separation as 3rd parties _need_ that agnostic behaviour to write anything

## 🎯 Changes

## ✅ Checklist


packages/server/src/core/internals/procedureBuilder.ts
```diff
@@ -249,6 +251,18 @@ export function createBuilder<TConfig extends AnyRootConfig>(
       return createNewBuilder(_def, builder._def) as any;
     },
     use(middlewareBuilderOrFn) {
+      if (
+        typeof middlewareBuilderOrFn === 'object' &&
```
Nick-Lucas, line 255: Crazy checks just because `function` was already taken as a type, but `.use` seems the most sensible home for this.

packages/tests/server/middlewaresChaining.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  t.procedure
+    .use(experimentalChainer)
+    .input(z.object({ name: z.string() }))
+    .query((opts) => {
+      const orgId = opts.input.orgId;
+      const name = opts.input.name;
+      const orgOk = opts.ctx.orgPermitted;
+      const userId = opts.ctx.userId;
```
Nick-Lucas, line 30: userid is getting lost right now because of the antics I'm pulling with the types. it's coming back as `any` so that needs some rework to fix

packages/server/src/core/internals/procedureBuilder.ts
```diff
@@ -42,6 +43,19 @@ type CreateProcedureReturnInput<
   _output_out: FallbackValue<TNext['_output_out'], TPrev['_output_out']>;
 }>;
 
+type CreateProcedureReturnInput2<
```
Nick-Lucas, line 46: This type needs a lot of work, it's the main source of bugs and is clearly not named well, can be largely ignored for now :)

KATT (commented):
I think we need our extensions to have "expected generics" that needs to be satisfied

It's great that we can do:

```ts
const extension = createProcedureExtension((proc) => {
  return proc.input(z.object({ orgId: z.number() })).use((opts) => {
    return opts.next({
      ctx: {
        orgPermitted: true,
      },
    });
  });
});
```

And then use it as `publicProcedure.extend(extension)`

But it would also be useful to allow for an easy way to let the user do the runtime validation themself, e.g.:

```ts
const extension = createProcedureExtension<{ input: { orgId: string }}>((proc) => {
  return proc.use((opts) => {
    return opts.next({
      ctx: {
        orgPermitted: true,
      },
    });
  });
});
```


Same for

- `ctx`
- `input`
- `output`
- (`meta` maybe?)

Nick-Lucas:
Killing it because I consider this approach to be a dead end thanks for TS type complexity problems. I'll come back to it with a different approach based on a factory pattern which is known to work with much greater ease, albeit imperfact usage characteristics.

PR #4219, opened 2023-04-19
feat(server): retain `Error.cause` even if the original cause isn't an `Error`-instance

### LINKED ISSUE #4217: bug: Throwing an object that is not an instance of Error get swallowed

  System:
    OS: macOS 13.2.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 55.72 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.19.1 - ~/.nvm/versions/node/v16.19.1/bin/node
    npm: 8.19.3 - ~/.nvm/versions/node/v16.19.1/bin/npm
  Browsers:
    Brave Browser: 111.1.49.128
    Chrome: 112.0.5615.121
    Safari: 16.3
  npmPackages:
    @trpc/server: 10.20.0 => 10.20.0 
    typescript: ^4.9.4 => 4.9.4 

### Describe the bug

I stumbled upon this recently with a third party library. The third party client throws objects that are not an instance of `Error`
ie: 
```typescript
throw { message: 'Something went wrong', name: 'ApiError', code: 403 }
```
When that happens, the whole object get swallowed by trpc and the output becomes
```
httpStatus: 500,
stack: 'Error: Unknown error\n' +
      '    at getErrorFromUnknown', 
      // ...
```
With no trace at all of the original object being thrown, which makes investigation really hard as to why this 500 error happened, as there's no stacktrace and no info about the original error at all

While I know this is bad practice to throw an object that is not an instance of an error, I feel like there's a non trivial chance of this happening in the wild with libraries (the one in this case is a decently popular one, 1.2k stars and 1 million weekly downloads)

I tested with throwing a string, and this works correctly
ie: `throw 'Custom error message'` , the problem seems to only happen with object that do not instantiate `Error`

I feel like if the shape of the thrown object is not an error but has similar properties ("message", maybe "name" and "code"), this should be mirrored to the error TRPC generates?

### Link to reproduction

https://stackblitz.com/edit/github-s8joae?file=src/pages/api/trpc/[trpc].ts

### To reproduce

Make any endpoint `throw { message: 'Custom error message', name: 'Third Party API error' };` and see what gets logged on the server

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

KATT:
Closes #4217

## 🎯 Changes

Retain `Error.cause` even if the original cause is a non-error

- Will be stored as `.unknownCause` if not `instanceof Error`
- Will pass to `super()` with the `cause` regardless

## ✅ Checklist


.vscode/extensions.json
```diff
@@ -4,6 +4,7 @@
     "esbenp.prettier-vscode",
     "dbaeumer.vscode-eslint",
     "bungcip.better-toml",
-    "Orta.vscode-twoslash-queries"
+    "orta.vscode-twoslash-queries",
+    "yoavbls.pretty-ts-errors"
```
KATT, line 8: zup @yoavbls 👋

yoavbls: amazing 💙

PR #4244, opened 2023-04-24
feat: add @effect/schema support

NexZhu:
## 🎯 Changes

[@effect/schema](https://github.com/effect-ts/schema) is a validation library by the author of `fp-ts` and others.

This PR add support for using `@effect/schema` to validate input data.

Example:
```ts
import { initTRPC } from https://github.com/trpc/trpc/tree/master/packages/server
iimport { toSelfValidatingSchema } from '@daotl-effect/prelude';
import * as S from '@effect/schema/Schema';

const t = initTRPC.create()

const params = S.struct({
  greeting: S.string,
  name: S.string,
})

export type router = typeof router
export const router = t.router({
  something: t.procedure
    .input(params)
    .query((req) => {
      const { greeting, name } = req.input
      console.log(greeting, name)
      return `${greeting} ${name}`
    }),
})
```

## ✅ Checklist


packages/tests/server/validators.test.ts
```diff
@@ -221,6 +223,32 @@ test('myzod', async () => {
   await close();
 });
 
+test('effect schema', async () => {
+  const t = initTRPC.create();
+
+  const router = t.router({
+    num: t.procedure
+      .input(toSelfValidatingSchema(S.number))
```
juliusmarminge, line 231: Without knowing more context regarding effect, why is this 3rd party import needed? Are effect schemas not parsable on their own?

Nick-Lucas: If the schema has to be transformed for every usage anyway, maybe it makes more sense to write a `toZodLike` which would make it compatible with tRPC by default and useful with other parts of the ecosystem?

adding support for a new arbitrary validator format doesn't seem to make much sense to me if it doesn't add out the box support

juliusmarminge: Agree

NexZhu: Got it, I've submit a feature request at
https://github.com/Effect-TS/schema/issues/260

juliusmarminge:
Looks like we support this out of the box - you just need to wrap the schema in `S.parse` yourself. 

I couldn't push to your branch so made a new PR in #4248 but if you make similar updates here we can merge this one :)

PR #4249, opened 2023-04-25
preparsedBody flag for resolveHTTPResponse

### LINKED ISSUE #4243: bug: Disabling batching breaks serialization of string values

```
  System:
    OS: macOS 13.0.1
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 2.14 GB / 16.00 GB
    Shell: 3.5.1 - /usr/local/bin/fish
  Binaries:
    Node: 16.18.1 - /usr/local/bin/node
    npm: 8.19.2 - /usr/local/bin/npm
  Browsers:
    Chrome: 112.0.5615.137
    Firefox: 109.0
    Safari: 16.1
  npmPackages:
    @tanstack/react-query: ^4.18.0 => 4.29.3 
    @trpc/client: ^10.21.1 => 10.21.1 
    @trpc/next: ^10.21.1 => 10.21.1 
    @trpc/react-query: ^10.21.1 => 10.21.1 
    @trpc/server: ^10.21.1 => 10.21.1 
    next: ^13.2.1 => 13.3.1 
    react: ^18.2.0 => 18.2.0 
    typescript: ^4.8.3 => 4.9.5
```

### Describe the bug

When batching is disabled, and the input is supposed to be string (defined with `z.string()`), it is unable properly parse the request body.
The error:

```
{
    "error": {
        "message": "Unexpected number in JSON at position 1",
        "code": -32700,
        "data": {
            "code": "PARSE_ERROR",
            "httpStatus": 400,
            "stack": "TRPCError: Unexpected number in JSON at position 1\n    at getRawProcedureInputOrThrow (file:///PATH/trpc-next-minimal/node_modules/@trpc/server/dist/resolveHTTPResponse-6c8965e4.mjs:24:15)\n    at resolveHTTPResponse (file:///PATH/trpc-next-minimal/node_modules/@trpc/server/dist/resolveHTTPResponse-6c8965e4.mjs:90:26)\n    at file:///PATHs/trpc-next-minimal/node_modules/@trpc/server/dist/nodeHTTPRequestHandler-2718f9d8.mjs:61:30\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n    at async file:///PATH/trpc-next-minimal/node_modules/@trpc/server/dist/adapters/next.mjs:39:9"
        }
    }
}
```

### Link to reproduction

https://stackblitz.com/edit/github-ferjpz?file=src/utils/trpc.ts

### To reproduce

In the stackblitz sandbox, click "As string", observe error. Then basically the same thing, but payload is passed as object with one field, and no problem.

### Additional information

Seems like adding [data transformers](https://trpc.io/docs/server/data-transformers) can fix the issue. But it'd still be great not to have the need to apply this solution

### 👨‍👧‍👦 Contributing

sudo97:
Closes #4243

## 🎯 Changes
Added flag `preparsedBody`, that defaults to falsey value, so that it doesn't break anything existent, but for next.js adapter it defaults to `true`, so it avoids parsing body. 

## ✅ Checklist

### DIFF
--- a/packages/server/src/adapters/node-http/types.ts
+++ b/packages/server/src/adapters/node-http/types.ts
@@ -64,6 +64,7 @@ export type NodeHTTPHandlerOptions<
    */
   middleware?: ConnectMiddleware;
   maxBodySize?: number;
+  preparsedBody?: boolean;
 } & NodeHTTPCreateContextOption<TRouter, TRequest, TResponse>;
 
 export type NodeHTTPCreateContextFnOptions<TRequest, TResponse> = {
--- a/packages/server/src/adapters/next.ts
+++ b/packages/server/src/adapters/next.ts
@@ -53,6 +53,7 @@ export function createNextApiHandler<TRouter extends AnyRouter>(
 
     await nodeHTTPRequestHandler({
       ...opts,
+      preparsedBody: opts.preparsedBody ?? true,
       req,
       res,
       path,
--- a/packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
+++ b/packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
@@ -67,6 +67,7 @@ export async function nodeHTTPRequestHandler<
       router: opts.router,
       req,
       error: bodyResult.ok ? null : bodyResult.error,
+      preparsedBody: opts.preparsedBody,
       onError(o) {
         opts?.onError?.({
           ...o,
--- a/packages/server/src/http/resolveHTTPResponse.ts
+++ b/packages/server/src/http/resolveHTTPResponse.ts
@@ -22,7 +22,10 @@ const HTTP_METHOD_PROCEDURE_TYPE_MAP: Record<
   GET: 'query',
   POST: 'mutation',
 };
-function getRawProcedureInputOrThrow(req: HTTPRequest) {
+function getRawProcedureInputOrThrow(
+  req: HTTPRequest,
+  preparsedBody?: boolean,
+) {
   try {
     if (req.method === 'GET') {
       if (!req.query.has('input')) {
@@ -31,7 +34,7 @@ function getRawProcedureInputOrThrow(req: HTTPRequest) {
       const raw = req.query.get('input');
       return JSON.parse(raw!);
     }
-    if (typeof req.body === 'string') {
+    if (typeof req.body === 'string' && !preparsedBody) {
       // A mutation with no inputs will have req.body === ''
       return req.body.length === 0 ? undefined : JSON.parse(req.body);
     }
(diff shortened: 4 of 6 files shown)


packages/server/src/adapters/node-http/types.ts
```diff
@@ -64,6 +64,7 @@ export type NodeHTTPHandlerOptions<
    */
   middleware?: ConnectMiddleware;
   maxBodySize?: number;
+  preparsedBody?: boolean;
```
Nick-Lucas, line 67: My gut feeling is we don't want to expose this on the public API, it would be an internal flag which each adapter can set. If there were a way to configure Next to _not_ parse JSON before passing it on, we would probably have documented that as part of setup I think, and most other adapters are lower level anyway

sudo97: @Nick-Lucas There's a way to configure Next not to parse, I mentioned that in the issue, also I added it to the docs in this PR. With body-parser disabled this flag is not required, it works properly.

Regarding flag being internal, I agree, but I am not that familiar with the src, so I was unsure where is the best place to put it, if you can suggest, I will update the PR

Nick-Lucas: https://nextjs.org/docs/api-routes/request-helpers

Right sorry I had the understanding from before that Nextjs didn't allow configuration of this behaviour. 

So don't you think the functionality work here is redundant? Rather than saying "if you're not using batching and you have bodyParser enable then set this option on tRPC" our docs should just be saying "when setting up tRPC you should disable bodyParser"?

Conditionals add complexity, and we're all about simplicity where we can be.

KATT, line 67:
```suggestion
```

KATT (commented):
I couldn't push to your fork, could you apply  https://github.com/trpc/trpc/commit/d11d54c090aa94d3b66b7235b23fcd6f575a8d7c to your PR?

It should work as well, but I couldn't get the regression test to be reliable.

Basically, if `req.body` exists, it's already processed and we can assume we don't need to do anything with it (I THINK!)

In the other cases, we get `body` as a stream that we listen to.

We need to validate that this is the case even with

```ts
export const config = {
  api: {
    bodyParser: false,
  },
}
```

KATT:
Okay, looks like my assumption was correct

I tried doing here: 973590b

I ran the server and did:

```sh
curl -X POST http://localhost:3001/api/test -H 'Content-Type: application/json' -d '{"foo":"bar"}'
```

With the `bodyParser: false` it would log `body undefined` and with it `true` it logs `body { foo: 'bar' }`

So this direction works and doesn't require any API changes.

KATT:
Unsure how to make a reliable regression test without spinning up an actual Next.js server 🤔

KATT:
I did a start of something in the tests over at #4256 which seems to be representative of reality

Won't have time to work on it more for a few days, @sudo97, you ok to take over?

PR #4258, opened 2023-04-27
fix tagging

KATT:
cc @juliusmarminge 

the latest stuff removed the tagging so here's a further hack lol

**EDIT:** nope this doesn't work. the thing already tags but it is in a commit that then disappears

damnit


juliusmarminge:
> here's a further hack lol

love it :)

juliusmarminge:
Try this:

```json
    "postversion": "tsx scripts/postversion.ts && git reset HEAD^ && git add . && git commit --amend -C HEAD",
```

PR #4268, opened 2023-04-28
feat(react-query): promote `overrides` without `unstable_`-prefix

KATT:
I've used this in production for a long time.

Gotten no feedback that people dislike the API, let's just promote it.


juliusmarminge (approved):
Lets go 🔥🔥

PR #4285, opened 2023-04-30
feat(`server`): agnostic http request handler

### LINKED ISSUE #4280: feat: trpc is assuming the type of the req.query property on all adapters

N/A

### Describe the bug

trpc is assuming the type of the req.query property if there is one on all adapters.

I have made an adapter for an augmented node server similar to the express adapter but the AugmentedRequest type passed to NodeHTTPHandlerOptions is not assignable to the type '{ query?: ParsedQs}` used in NodeHTTPRequest type. This is because my query property could be an array like this
>['a', 'b', 'c']

if the search params where like this
> ?0='a'&1='b'&2='c'

It also essentially allows an array to be an array of
>(string | ParsedQs)[]

instead of just
>string[] | ParsedQs[]

```ts
const createMyServerMiddleware = <TRouter extends AnyRouter>(
    opts: NodeHTTPHandlerOptions<TRouter, AugmentedRequest, AugmentedResponse>): MwCall =>
  async function(req, res) {
    await nodeHTTPRequestHandler({req, res, ....})
  }
```

Changing the req.query on my augmented server would be a breaking change and I can't actually find any where in trpc that the req.query that it creates would be accessible to me anyway

The nodeHTTPRequestHandler function in the nodeHTTPRequestHandler file is where the assumption is made despite req.query of type ParsedQs being an express property and not a node server property. Most other servers are probably compatible with this signature but mine is not because URLSearchParams can not take an array and because of the type miss match mentioned above.

```ts
const query = opts.req.query ? new URLSearchParams(opts.req.query) : new URLSearchParams(opts.req.url.split('?')[1])
const req = {
    method: opts.req.method,
    headers: opts.req.headers,
    query,
    body: bodyResult.ok ? bodyResult.data : undefined
}
```

I'm not sure what the wider implications are of just always creating the query property and removing  { query?: ParsedQs } from the NodeHTTPRequest Type be. Is it only there for the ConnectMiddleware<NodeHTTPRequest, NodeHTTPResponse> type used by the middleware property that can be passed to the nodeHTTPRequestHandler? Just to clarify, Im not talking about the middleware used in procedures here because that is obviously passed a context instead. 

If so, it's also a little confusing that the nodeHTTPRequestHandler middleware function's req and res are typed as NodeHTTPRequest and NodeHTTPResponse even though it's passed the same req and res as the nodeHTTPRequestHandler. The types can't be trusted there. It assumes there might be a property called query and that it must be of a certain type despite a plain node server never having a query property and it is unaware of any of the other properties that might be on the augmented req and res that where passed to the nodeHTTPRequestHandler. 

### Link to reproduction

N/A

### To reproduce

Copy the standalone adapter and Pass NodeHTTPHandlerOptions a request type with {query: string[]}

 ```ts
NodeHTTPHandlerOptions<TRouter, http.IncomingMessage & {query:string[]}, http.ServerResponse>;
```
Then use createHTTPHandler function to create your own HTTP Server. With a mock query value for the req

```ts
import { inferAsyncReturnType, initTRPC } from '@trpc/server';
import { createHTTPHandler } from '@trpc/server/adapters/standalone';
import { createServer } from 'http';

const handler = createHTTPHandler({
  router: appRouter,
  createContext() {
    return {};
  },
});

createServer((req, res) => {
  // mock a query value for the req
  handler(Object.assign(req, {
    query: ['a', 'b', 'c']
  }), res);
});

server.listen(3333);
```

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

<sub>[TRP-26](https://linear.app/trpc/issue/TRP-26/feat-trpc-is-assuming-the-type-of-the-reqquery-property-on-all)</sub>

## Funding

<a href="https://polar.sh/trpc/trpc/issues/4280">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/4280/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

fallenfreq:
Closes #4280 / 
Closes T-26

## 🎯 Changes

1. You can now add custom servers without them having to have the same req.query type as express.
ParsedQs interface has been removed for this because its only use becomes redundant if you can't guarantee that it can be passed to URLSearchParams

2. The middleware method that you pass to an adapter will now pass the correct types through to the "req" and "res" arguments.

3. Added the necessary tests to test the new features.

## ✅ Checklist

- ✅  I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- ✅  If necessary, I have added documentation related to the changes made.
- ✅ I have added or updated the tests related to the changes made.


KATT (changes requested):
This unfortunately fails all tests

KATT (dismissed):
This unfortunately fails all tests

fallenfreq:
> This unfortunately fails all tests

Sorry about that. It should be passing all tests and checks now apart from the 2 checks I'm not Authorized to run.

KATT:
hey @Sheraff, could you review this? :)

packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
```diff
@@ -27,9 +27,17 @@ export async function nodeHTTPRequestHandler<
...
+    const paramsFromQuery = () =>
+      new URLSearchParams(opts.req.url?.split('?')[1]);
+    let query: URLSearchParams;
+    try {
+      query = opts.req.query
+        ? new URLSearchParams(opts.req.query as any)
+        : paramsFromQuery();
+    } catch (err) {
+      query = paramsFromQuery();
+    }
```
Sheraff, line 40: I think I do understand the need for this PR, but this feels a bit sub-optimal since there could be cases where *every* procedure ends up throwing (from `new URLSearchParams(opts.req.query as any)`) before being recovered by the `catch`.

It might be a little nitpicky but aren't there alternative implementations that wouldn't be require this? I don't want to speak for KATT here but maybe a boolean option on the server configuration instead of a `try/catch` on every procedure call?

juliusmarminge: @fallenfreq friendly bump

fallenfreq: @Sheraff @juliusmarminge Hey, I've been camping with no signal for a few weeks and now I'm on a work trip for the next two weeks. I'll be able to look at this again when I'm home towards the end of the week starting July the 10th.

I didn't consider an extra parameters originally because in my case, it would only fail on a rare occurrences but it makes sense considering there are cases where it could fail on every request.

Would you have the parameter and keep the try catch block for cases where it has not been set? Or perhaps catch and re throw an error letting the user know they should be setting the parameter?

How do the names isQueryInvalid or invalidTrpcQuery sound?

pnpm-lock.yaml
```diff
@@ -13762,14 +13809,14 @@ packages:
       '@typescript-eslint/parser':
         optional: true
     dependencies:
-      '@typescript-eslint/parser': 6.2.1(eslint@8.40.0)(typescript@5.1.3)
+      '@typescript-eslint/parser': 5.59.2(eslint@8.40.0)(typescript@5.1.3)
```
juliusmarminge, line 13812: why is this getting downgraded lol

fallenfreq: There are other changes in the file that where needed for the new example "examples/minimal-custom-server" to pass all tests. I can only assume the file has been updated several times since I made the amends

KATT (commented):
Hey, sorry for the slow response on this!

Could you please?:

- remove the example project - we have enough of them haha
- add a test, probably in https://github.com/trpc/trpc/tree/main/packages/tests/server/adapters

see https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md#testing

fallenfreq:
> Hey, sorry for the slow response on this!
> 
> Could you please?:
> 
> * remove the example project - we have enough of them haha
> * add a test, probably in https://github.com/trpc/trpc/tree/main/packages/tests/server/adapters
> 
> see https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md#testing

Hey, I have removed the example and refactored it into a test

packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
```diff
@@ -28,9 +28,12 @@ export async function nodeHTTPRequestHandler<
       return await opts.createContext?.(opts);
     };
 
-    const query = opts.req.query
-      ? new URLSearchParams(opts.req.query as any)
-      : new URLSearchParams(opts.req.url!.split('?')[1]);
+    // a framework could pre-parse query in a way that could crash URLSearchParams
+    // invalidTrpcQuery can be added to a request object to let trpc know
+    const query: URLSearchParams =
+      opts.req.query && !opts.req.invalidTrpcQuery
```
kamilogorek, line 34: I'd name it by the action it'll perform, not by the "state" the request is in. Eg. `forceQueryFromUrl` or `forceUrlQuery` provides some meaning in itself, while `invalidTrpcQuery` doesn't tell you what will happen, but only indicates that something is "off".

fallenfreq: @kamilogorek I went with invalidTrpcQuery to differentiate the query that trpc uses internally from the query used by the framework. I'm happy to go with something like forceTrpcQueryFromUrl. My original thinking was that it leaks implementation details to the user that could change. For example, not all frameworks have a req.url property to make a query from although the ones I can think of already have a req.query parameter that is compatible with the trpc query and if they didn't, then they could just add a req.url property to generate the query from.

KATT:
cc @fallenfreq could you get this to point to `next` instead? we're not really doing new features on `main` anymore

sorry for the delay on this!

fallenfreq:
> cc @fallenfreq could you get this to point to `next` instead? we're not really doing new features on `main` anymore
> 
> sorry for the delay on this!

Yeah, I'll be working on a project that uses trpc again very soon so i'll do it then.

fallenfreq:
Hey @KATT, This is ready for review again. The original disputed property name is no longer required with trpc v11. I've updated the changes and tests to work with v11 so hopefully this is good to be merged into next now.

packages/tests/server/adapters/custom.test.ts
KATT, line 1: Please just delete this, we have plenty of tests that test this path

fallenfreq: Will do, this was originally testing the now redundant forceTrpcQueryFromUrl property as well but it's still testing that query parameters that don't match the express type don't throw and that the handler middleware has the types passed correctly. The other tests are just testing the implementation of the test itself.

fallenfreq: Hi @KATT, This was done 3 weeks ago, I know it said there were two extra failing tasks but I think that was just due to a token generation limit being reached for some of the tests. I've just updated it and those failing tests have passed.

KATT (approved):

PR #4286, opened 2023-04-30
Experimental input strategies

Nick-Lucas:
Trying to simplify the experimental formData into a system which could accept other types of data like streams and raw file posts.

## 🎯 Changes

## ✅ Checklist

### DIFF
--- a/packages/server/src/adapters/node-http/content-type/form-data/index.ts
+++ b/packages/server/src/adapters/node-http/content-type/form-data/index.ts
@@ -8,14 +8,15 @@
 /**
  * @see https://github.com/remix-run/remix/blob/0bcb4a304dd2f08f6032c3bf0c3aa7eb5b976901/packages/remix-server-runtime/formData.ts
  */
+
+import { CombinedDataTransformer } from '@trpc/server/transformer';
+import { streamMultipart } from '@web3-storage/multipart-parser';
+import { Readable } from 'node:stream';
 import {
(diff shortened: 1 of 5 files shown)


packages/server/src/adapters/node-http/content-type/form-data/index.ts
```diff
@@ -103,3 +109,28 @@ export { createFileUploadHandler as experimental_createFileUploadHandler } from
...
+export function experimental_createFormDataMiddleware<
+  TTRPCInstance extends ReturnType<AnyTRPCInstance>,
+>(
+  t: TTRPCInstance,
+  config?: {
+    uploadHandler?: UploadHandler;
+  },
+) {
+  return t.middleware(async (opts) => {
+    // TODO: ideally we wouldn't force this to be added by the user, would be better to have the Adapter provide this
```
KATT, line 122: If we at compile-time could enforce that `opts.ctx.req` exists, it'd help

Nick-Lucas: I actually don't think we should use ctx for this at all. My second PR (stacked in this one) addresses this issue, I'm just trying to be bite-sized about the experiments

Nick-Lucas:
I'm going to close this since we've rejected the "Strategies" idea and the new approach is much more incremental. Will continue work on the new approach in #4288 as one piece

PR #4288, opened 2023-05-01
Experimental: a simpler backend API for FormData loading, and foundation for loading other data types

Nick-Lucas:
Supersedes #4286 

Goals:

* Add a simpler API for standing up FormData loading middleware and inputs
* Remove the need for req to be provided by the user's ctx
* Lay groundwork for future data types support and cross-adapter data loading middlewares

This piece will have the effect of supporting FormData in _all_ our adapters, and making it easy for community Adapters to participate in the data loading middlewares we build, as well as write their own middlewares, adapters, and content-handlers to support complete protocols

Implemented stuff:

* New `RequestUtils` type which an Adapter can optionally pass to wrap its Request type up in a decoupled interface. If not passed but a middleware tries to use it, there will be a runtime error blaming the Adapter
* RequestUtils passed all the way through to middlewares meaning they can read off the request without needing knowledge of the adapter(s) they live within (and without further user setup of ctx)
* The middleware APIs have been wrapped up into a middleware factory `experimental_createFormDataMiddleware` which has some sensible defaults, and on construction will return a middleware - I will be later proposing formal factory patterns which we'll document for library authors, and this is an example of the core pattern I'd like us to promote
* `TRPCInstance` and `AnyTRPCInstance` - initTRPC has been a bit of a complex type to unwrap, these bring it in line with many of our other APIs in permitting abuse of the type for generics and factories. First internal usage is in `experimental_createFormDataMiddleware`


Nick-Lucas:
Remaining TODOs:

* Feedback and course correction on the approach
* Tests against all Adapters which cover this set of APIs

PR #4320, opened 2023-05-05
chore: bump some deps

KATT:
had trouble reviewing #4132 b/c all the deps were distracting me :)


packages/next/src/index.ts
```diff
@@ -1,2 +1,4 @@
+'use client';
+
```
juliusmarminge, line 2: this will just get stripped away anyways

```suggestion
```

juliusmarminge (approved):

PR #4327, opened 2023-05-06
fix: keep union type when simplifying in middleware

### LINKED ISSUE #4321: bug: Type narrowing `ctx` in middleware no longer works as of v10.21.0
```
  System:
    OS: Linux 5.15 Ubuntu 20.04.5 LTS (Focal Fossa)
    CPU: (24) x64 AMD Ryzen 9 5900X 12-Core Processor
    Memory: 25.24 GB / 31.30 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 18.15.0 - ~/.volta/tools/image/node/18.15.0/bin/node
    Yarn: 1.22.15 - ~/.volta/bin/yarn
    npm: 9.5.0 - ~/.volta/tools/image/node/18.15.0/bin/npm
  npmPackages:
    @trpc/client: 10.20.0 => 10.20.0 
    @trpc/server: 10.20.0 => 10.20.0 
    typescript: ^5.0.4 => 5.0.4 
```

### Describe the bug

Prior to v10.21.0, we were able to narrow the types of our context in middleware like so:

```ts
// example context
type Context =
  | { user: null; session: null }
  | {
      user: {
        id: string;
        name: string;
      };
      session: {
        id: string;
        state: string;
      };
};

// say that we want to add a middleware for logging
const baseProcedure = t.procedure
  .use(async ({ path, type, next }) => {
    const start = Date.now();
    const result = await next();
    const ms = Date.now() - start;
    console.log(`${result.ok ? 'OK' : 'ERR'} ${type} ${path} - ${ms}ms`);
    return result;
  });

export const protectedProcedure = baseProcedure.use(({ next, ctx }) => {
  // we narrow the type of `user` and `session` by throwing when either are null
  if (!ctx.user || !ctx.session) {
    throw new TRPCError({ code: 'UNAUTHORIZED', message: 'not signed in' });
  }
  return next({ ctx });
});
```

If we try to add another middleware to this procedure, the type of `ctx.user` still states that it can still be `null`, which is incorrect.
```ts
const someOtherProcedure = protectedProcedure.use(async ({ next, ctx }) => {
  // type of `user` should not be null here as we narrowed its type in the first middleware
  ctx.user.id;
  //   ^? user: { id: string; name: string; } | null

  return next({
    ctx: {
      ...ctx,
    },
  });
});
```
If we comment out the middleware for logging, the type of `ctx.user` would now be correct.

The last known version of tRPC that I could find that doesn't exhibit this behavior is v10.20.0.

### Link to reproduction

https://stackblitz.com/edit/github-yrjz6a?file=src%2Fserver%2Ftrpc.ts%3AL52

### To reproduce

Using tRPC v10.20.0 and below narrows both the `session` and `user` types correctly:
https://stackblitz.com/edit/github-yrjz6a-nkddm4?file=src%2Fserver%2Ftrpc.ts%3AL52

Additionally, for the reproduction that uses the [latest version of tRPC v10.25.0](https://stackblitz.com/edit/github-yrjz6a?file=src%2Fserver%2Ftrpc.ts%3AL29), if you comment out the logger middleware for the `baseProcedure`, the `ctx` type is corrected. 

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

Nick-Lucas:
Closes #4321

## 🎯 Changes

## ✅ Checklist


packages/server/src/core/internals/utils.ts
```diff
@@ -4,13 +4,17 @@ import { ProcedureParams } from '../procedure';
...
+  ? TWith extends any
+    ? {
+        [K in keyof TType | keyof TWith]: K extends keyof TWith
+          ? TWith[K]
+          : K extends keyof TType
+          ? TType[K]
+          : never;
+      }
+    : never
+  : never;
```
Nick-Lucas, line 17: This shouldn't really have any cost, but causes typescript to correctly distribute the union elements again as per here: https://stackoverflow.com/a/51691257

juliusmarminge (approved):

PR #4336, opened 2023-05-08
Experimental: replace rawInput with `async decodeInput`

Nick-Lucas:
WIP

This is an alternate approach to (well, an evolution of) #3765 it contains a breaking change to the public API which I believe we _have_ to adopt for the simplest possible experience. I'm not only looking at FormData here, I'm thinking about streams, File, and also alternate protocols like protobuffers.

One of the big complexities of body decoding, is we don't want to do it _too_ eagerly, due to the risk of DDOS attacks from unauthenticated users. Currently we have two different loading models competing, with JSON loading eagerly, and the user being required to do middleware work to enable the other model for specific types.

Instead of using a middleware we defer the actual loading of inputs from request query/body until the last possible moment. Instead of passing `rawInput` to callProcedure we pass `async decodeInput(): unknown`, this can be used by middlewares to grab and process the input early, but will be automatically called before we pass the rawInput on to validators and resolvers if need be. This has some advantages:

* We do no processing of bodies until middlewares have had a chance to run - meaning rate-limiting and auth can kick in first if the users orders them before any input consumers.
* We don't need weirdness with attaching middlewares manually which do input decoding work
* We can still have users provide custom decoders but without the need to also write and document middlewares

Draft of a more visual diagram:

input decoding flow


packages/server/src/adapters/node-http/content-decoder/form-data/index.ts
```diff
@@ -8,12 +8,37 @@
...
+    return contentType;
+  }
+
+  if (Array.isArray(contentType)) {
+    if (contentType.length > 1) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        'Multiple Content-Type header values were found. The first one will be used',
+      );
+    }
```
KATT, line 36: We can just do throw BAD_REQUEST HERE

packages/server/src/adapters/node-http/content-decoder/form-data/index.ts
```diff
@@ -55,51 +82,85 @@ async function parseMultipartFormData(
...
     },
-  });
 
-export { parseMultipartFormData as experimental_parseMultipartFormData };
-export { createMemoryUploadHandler as experimental_createMemoryUploadHandler } from './memoryUploadHandler';
-export { createFileUploadHandler as experimental_createFileUploadHandler } from './fileUploadHandler';
-export { composeUploadHandlers as experimental_composeUploadHandlers } from './uploadHandler';
-export { type UploadHandler } from './uploadHandler';
-export { isMultipartFormDataRequest as experimental_isMultipartFormDataRequest };
+    // TODO: what even was all this for?
```
KATT, line 137: it was used for GET FormData, i.e. params in the search string

packages/server/src/core/middleware.ts
```diff
@@ -197,12 +199,15 @@ function isPlainObject(obj: unknown) {
 export function createInputMiddleware<TInput>(parse: ParseFn<TInput>) {
   const inputMiddleware: ProcedureBuilderMiddleware = async ({
     next,
-    rawInput,
+    decodeInput,
     input,
   }) => {
     let parsedInput: ReturnType<typeof parse>;
+
+    const parseInput = input ?? (await decodeInput());
```
Nick-Lucas, line 207: This should just be rawInput to support input chaining

packages/server/src/core/internals/procedureBuilder.ts
```diff
@@ -288,7 +288,12 @@ function createResolver(
     resolver,
     middlewares: [
       async function resolveMiddleware(opts) {
-        const data = await resolver(opts);
+        // TASK: content-handler body call goes here if not already called?
+
+        const data = await resolver({
+          ...opts,
+          input: opts.input ?? (await opts.decodeInput()),
```
Nick-Lucas, line 295: Should this be undefined if input hasn't been validated? (we think it is like this currently)

Nice benefit: big bodies sent to endpoints which don't expect and input will never materialise them.

packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
```diff
@@ -31,36 +28,26 @@ export async function nodeHTTPRequestHandler<
...
-    });
+    // TODO: ensure this remains implemented at the right level
+    const preprocessedBody = false;
 
     const req: HTTPRequest = {
       method: opts.req.method!,
       headers: opts.req.headers,
       query,
-      body: bodyResult.ok ? bodyResult.data : undefined,
+      async getBodyJson() {
```
Nick-Lucas, line 38: Let's model HTTPRequest after an industry standard like the fetch API - .json .text .stream and return the appropriate types. Minimal risk of that being leaky for our needs

PR #4338, opened 2023-05-09
Fix: Set display to inline-block for code element

IAmYoungbossy:
This change sets the `display` property of the `code` element to `inline-block`, which makes the right padding of its child `div` (class="line") elements effective and improves user experience expecially on mobile devices.

Before this change, the right padding of the `div` (class="line") elements inside the `code` element was not effective because the `code` element had a default `display` property of `inline`, which does not allow for width or height. This made some lines of code look like they were cut off especially on mobile devices since the padding-right wasn't effective. By setting the `display` property of the `code` element to `inline-block`, the element now has a width and height, and its child `div` (class="line") elements' padding is effective.

I tested this change locally by adding the `display: inline-block` property to the `code` element in the CSS file and verifying that the right padding of its child `div` (class="line") elements was effective.

## Screenshots

Below are screenshots of before and after fix

### Before Fix
Before setting the display property of code to inline-block, the padding-right of "line" class were not effective as seen in the two screenshot below.
before-fix2
before-fix

### After Fix
After setting the display property of code to inline-block the padding are now effective as seen in the two screenshot below.
after-fix2
after-fix

Closes #

## 🎯 Changes

Set CSS display property of code element to inline-block.

The change is a bug fix.

## ✅ Checklist

### DIFF
--- a/www/src/theme/CodeBlock/styles.css
+++ b/www/src/theme/CodeBlock/styles.css
@@ -148,6 +148,7 @@ pre .code-container {
 
 pre code {
   white-space: pre;
+  display: inline-block;
   -webkit-overflow-scrolling: touch;
 }
 


juliusmarminge:
Can you show some SS of before/after

IAmYoungbossy:
Yes, @juliusmarminge I just did in the edited PR.

Nick-Lucas:
WWW looks good here on iPhone 12 Pro

Thanks for the contribution. I suggest we also check on android chrome and a few browsers on desktop big/small window. But the actual change sounds pretty safe between browsers

Issue #4343, opened 2023-05-09
feat: out-of-order batch response streaming


Sheraff:
### Describe the feature you'd like to request

When using batched requests, the response time is limited by the slowest request of the batch. This feature request is to investigate the feasibility of out-of-order response streaming so that the fastest responses come in first.

### Describe the solution you'd like to see

Currently a batch response looks like this:
```json
[{"result":{"data":{"json":{"foo": 1}}}},{"result":{"data":{"json":{"bar": 2}}}}]
```

Though it could be rearranged into
```json
[
{"result":{"data":{"json":{"foo": 1}}}},
{"result":{"data":{"json":{"bar": 2}}}}
]
```
which remains a valid JSON, but can now be parsed line by line.

If we take this a step further, the following will, for all intents and purposes, be equivalent:
```json
{
"1":{"result":{"data":{"json":{"bar": 2}}}},
"0":{"result":{"data":{"json":{"foo": 1}}}}
}
```

Which could be treated as a stream parsed line-by-line on the client:

```js
	await response.body
		.pipeThrough(new TextDecoderStream('utf-8'))
		.pipeThrough(new ChunksToLinesStream())
		.pipeThrough(new LineToObjectStream())
		.pipeTo(actionnableSink)
```
code example from 
https://github.com/Sheraff/partial-json-stream/blob/1bea84df3d0ff7c6b5177443a00b9945ad8b0aff/js/script.js#L26-L30

Since tRPC in batch mode doesn't rely on HTTP status codes (except for the `207`), but uses its own internal error system, adding the downside of a stream — that you can't know the final status of your response before you're done sending it — doesn't seem like a big issue.

This would require for the response to be
- serialized request by request, instead of everything together at the end
- flushed to a streamed response after every request finishes

And for the client to
- handle the response as a stream, decode it and chunk it into individual lines
- parse out the index `"1":` and the optional trailing comma `,`
- deserialize line by line, and update the react-query cache as soon as a response is ready

The performance implication of using JS streams instead of letting the browsers use their native methods should be considered. The example I'm referring to here (https://github.com/Sheraff/partial-json-stream) was written by me so I don't think it's close to optimal, but it is much slower than a simple `await response.json()` for all but the biggest jsons. However this doesn't take into account the possible slowness server-side (for example, when some requests do filesystem or database stuff).

### Describe alternate solutions

[...]

PR #4346, opened 2023-05-10
Adding a streaming option for batched requests

Sheraff:
Closes https://github.com/trpc/trpc/issues/4343

## 🎯 Changes

Should work as is (though untested), but this doesn't (yet) actually handle the responses as a stream client-side.

- Every request handler should be made to `yield` its results (still works for non-streamed responses, it's just the way of returning the responses that changes) https://github.com/Sheraff/trpc/blob/main/packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
- `resolveHTTPResponse` is made to handle the request handlers yielding their results in several steps instead of a single return (https://github.com/Sheraff/trpc/blob/main/packages/server/src/http/resolveHTTPResponse.ts)

The only major change (except for the streaming stuff obviously) is that the `data` key of the argument passed to `opts.responseMeta` is now optional, because it cannot be known in the case of a streamed response. I haven't looked too deeply into how this `responseMeta` function is declared, but if really necessary, we might be able to type it correctly so that `data` is not optional when `batching: { enabled: boolean }` isn't set to `true`

## ✅ Checklist

## 🚧 TODO

- [x] actually try running the code! lol
- [x] create a client-side "adapter" to handle a streamed response based on `Transfer-Encoding: chunked` response header (stream parsing logic is available here: https://github.com/Sheraff/trpc-stream-poc/blob/master/public/script.js)
- [ ] make the same modifications to all "request handlers" as were made to `nodeHTTPRequestHandler`
- [ ] test coverage


packages/server/src/internals/types.ts
```diff
@@ -10,6 +10,9 @@ export interface BaseHandlerOptions<TRouter extends AnyRouter, TRequest> {
   batching?: {
     enabled: boolean;
   };
```
Nick-Lucas, line 12: My opinion of the public facing API:

```ts
batching?: { enabled: true, streaming: boolean } | { enabled: false }
```

PR #4347, opened 2023-05-11
feat(`server` + `client`): adding a streaming option for batched requests

### LINKED ISSUE #4343: feat: out-of-order batch response streaming

Sheraff:
Closes #4343
Closes T-35

Continuation of https://github.com/trpc/trpc/pull/4346
Related to https://github.com/trpc/trpc/pull/4365 (to measure bundle size changes)
Alternative DX proposal https://github.com/trpc/trpc/pull/4403
Removing async generators / iterators client-side https://github.com/trpc/trpc/pull/4424
Removing async generators / iterators server-side https://github.com/trpc/trpc/pull/4433

## 📰 Update

Everything detailed below is still valid, but instead of adding `unstable_mode: 'stream'` as an option on `httpBatchLink`, the streaming feature was made into its own link `unstable_httpBatchStreamLink` that's functionally identical to `httpBatchLink` except for the streaming part.

## 🎯 Changes

Add `unstable_mode?: 'standard' | 'stream'` option to `httpBatchLink`. When `'stream'` is selected, the server should stream the response, making it so that a slow request in a batch does not prevent the faster ones from being returned quickly.

- the client will send a `trpc-batch-mode: stream` header
- when receiving this header, if the batch contains 2 or more procedures, the server will setup a streamed response
  - `nodeHTTPRequestHandler` uses the "express-like" api and streams by simply writing several times to the response
  - `fastifyRequestHandler` uses a nodeJs `Stream.Readable` stream and pushes several times into it
  - `fetchRequestHandler` uses the web APIs `TextEncoder` and `ReadableStream` to enqueue several times
  - `aws-lamba/index` **does not support streaming** (could be implemented in the future), requesting a stream will have no impact and a regular JSON response will still be sent.
- since the streamed response is generated after the headers are sent, the `responseMeta` callback **does not** receive the `data` key (only in the case of a streamed response). This key has thus been made optional *everywhere* (because I didn't manage to type it correctly)
- 2 headers are added to the streamed response:
  - `Transfer-Encoding: chunked`
  - `Vary: trpc-batch-mode` (this is concatenated with other potential `vary` headers added by other means)
- The resulting response body is a string that is still a valid JSON, though because we respond *out of order*, we've replaced the "array of results" by an "object of results, whose keys are indexes":
  ```json
  {
    "2": { "foo": "bar" }
    ,"0": { "hello": "world" }
    ,"1": { "q": "a" }
  }
  ```
- the client knows that it "requested a stream" and so will try to parse the response it receives as such:
  - if the first line of the response string (delimited by `\n`) is not a single character `{`, the client reverts to parsing the entire body as a regular JSON (reverts to default `httpBatchLink` behavior)
  - if the first line matches the expected format, subsequent lines are parsed individually, transformed, and promises are resolved immediately without waiting for the stream to close

## ❗️ Caveats

- streaming was not implemented for the *aws / lambda adapter*. Though it seems possible, I don't know enough about AWS to feel confident implementing it.
- for a cloudflare worker server (`fetchRequestHandler`) we rely on Web APIs that are not necessarily available in the JS runtime of the worker. Namely `ReadableStream` should be explicitly enabled through the `streams_enable_constructors` flag in the .toml config
- the client supports several possible `Response` objects: that of Web APIs (all browsers), that of `undici`, and that of `node-fetch`. This should cover most uses of tRPC. Lack of support would have no effect as long as `unstable_mode` isn't set to `'stream'`.
- both client and server use `AsyncGenerator` and `AsyncIterator`. These seem to be widely supported, but I don't have much more information than that. Also, generators used to not be very performant, but as-far-as-I-can-tell™ it seems okay nowadays.

## ✅ Checklist

## 🚧 TODO

- [x] actually try running the code! lol
- [x] create a client-side "adapter" to handle a streamed response based on `Transfer-Encoding: chunked` response header (stream parsing logic is available here: https://github.com/Sheraff/trpc-stream-poc/blob/master/public/script.js)
- [x] make the same modifications to all "request handlers" as were made to `nodeHTTPRequestHandler`
- [x] test coverage
  - [x] add tests for streaming in node adapted
  - [x] add tests for streaming in fetch adapter
  - [x] add tests for streaming in fastify adapter
- [x] write docs
- [x] measure client bundle size change (and maybe, if difference too important, make "batch streaming" its own link) (see https://github.com/trpc/trpc/pull/4365)
- ~~correctly type `responseMeta` callback so that `data` key of object param is only optional in the case of an httpBatchLink with unstable_mode = 'stream'~~


packages/tests/server/streaming.test.ts
```diff
@@ -0,0 +1,213 @@
...
+          return observable((observer) => {
+            const unsubscribe = next(op).subscribe({
+              next(value) {
+                console.log('got value', value.result);
+                results.push({
+                  type: 'next',
+                  value,
+                  opId: op.id,
+                });
+                observer.next(value);
```
Sheraff, line 55: So I'm really not sure how the link chain of observables works... Sorry! 

However from what I've seen in the rest of the codebase, I think it's NOT like an onion where each link peels its own metadata and passes the unwrapped data to the next link, but links are supposed to pass the data as is:

```js
// example from www/docs/client/links/index.md
next(value) {
  console.log('we received value', value);
  observer.next(value);
},
```


So I "fixed" the tests by changing the tests themselves...
```diff
- observer.next((value as any).result);
+ observer.next(value);
```

packages/tests/server/streaming.test.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    // batch preserves request order
+    expect(results).toEqual([3, 1, 2]);
+    // streaming preserves response order
+    expect(orderedResults).toEqual([1, 2, 3]);
+  });
+  test('out-of-order streaming with error', async () => {
+    const { proxy } = ctx;
+
+    const results = await Promise.all([
```
KATT, line 89: `Promise.allSettled()` is nice here too :)

packages/client/src/links/internals/httpUtils.ts
```diff
@@ -133,41 +136,79 @@ export const jsonHttpRequester: Requester = (opts) => {
   });
 };
 
+export const streamingJsonHttpRequester: Requester<
```
KATT, line 139: Could we move this to the `httpBatchStreamLink` rather than being a shared util?

packages/client/src/links/internals/httpUtils.ts
```diff
@@ -133,41 +136,79 @@ export const jsonHttpRequester: Requester = (opts) => {
...
+
+  return { promise, cancel };
+};
+
 export type HTTPRequestOptions = HTTPBaseRequestOptions &
   ContentOptions & {
     headers: () => HTTPHeaders | Promise<HTTPHeaders>;
   };
 
+async function getHttpResponse(
```
KATT, line 175: I don't think this should be called `getHttpResponse`

Rather like 

```suggestion
async function fetchHTTPResponse(
```

```suggestion
async function performHTTPRequest(
```

packages/client/src/links/internals/parseJsonStream.ts
KATT, line 1: Can we group this together with the `httpBatchLink` somehow in the folder structure?

KATT:
```graphql
.
├── dedupeLink.test.ts
├── dedupeLink.ts
├── httpBatchLink
│   ├── index.ts # <-- entrypoint for batch link
│   ├── parseJsonStream.ts # <-- this is not used outside of batch link
│   └── [....]
├── httpFormDataLink.ts
├── httpLink.ts
├── index.ts
├── internals # <-- shared utils across many links
│   └── [...]
├── loggerLink.ts
├── retryLink.ts
├── splitLink.test.ts
├── splitLink.ts
├── types.ts
└── wsLink.ts
```

packages/server/src/adapters/fastify/fastifyRequestHandler.ts
```diff
@@ -55,18 +57,94 @@ export async function fastifyRequestHandler<
...
+      return res.send(firstChunk[1]);
+    } else {
+      // case of a method === "HEAD" response
+      return res.send();
+    }
+  }
+
+  // iterator is not exhausted, we can setup the streamed response
+  void res.header('Transfer-Encoding', 'chunked');
+  void res.header(
```
KATT, line 108: Is `void` needed here?

www/docs/client/links/httpBatchStreamLink.md
```diff
@@ -0,0 +1,50 @@
...
+
+`unstable_httpBatchStreamLink` (experimental) is a [**terminating link**](./index.md#the-terminating-link) that batches an array of individual tRPC operations into a single HTTP request that's sent to a single tRPC procedure (equivalent to [`httpBatchLink`](./httpBatchLink.md)), but doesn't wait for all the responses of the batch to be ready and streams the responses as soon as any data is available.
+
+## Usage
+
+All usage and options are identical to [`httpBatchLink`](./httpBatchLink.md).
+
+## Streaming mode (experimental)
+
+> ⚠️ This link is experimental and may change in the future.
```
KATT, line 16: Use the term `unstable` instead - https://trpc.io/docs/faq#unstable

The way I reason about things is that "experimental" is badly tested things that might be changed or dropped and unstable things are things that we sort-of committed to but might have issues

tl;dr: experimental ≈ alpha, unstable ≈ beta

I think this should be 'beta"

www/docs/client/links/httpBatchStreamLink.md
KATT, line 1: Some note about browser compatibility or something here?

Is there something we can do to automatically degrade when we detect that `fetch`-polyfill or the browser doesn't support it?

KATT (commented):
This is awesome. Super excited for it!

- I prefer the DX of having a bespoke `httpBatchStreamLink` as the configuration proposed in  #4403 looks a bit intimidating. That said, we could probably still reuse the code the way you're doing there but exposing it as two bespoke links
- I'd like stuff to do with the streaming in it's own folder (see inline comment for a suggestion)
- We can see some blind spots [in the codecov](https://app.codecov.io/gh/trpc/trpc/pull/4347?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=trpc) - might be a hint that we have dead code paths or that we're not testing thoroughly 
- Before we merge, we could probably revert the `examples/**` 
- Since we're using generators etc, it'd be cool if we could provide guidance on how to make this compatible with old browsers - I guess the user can do a `splitLink` to decide.. or we can automatically "degrade" to a non-streaming link if we know it ain't supported
- I'll do a canary release and see if it passes work's CI

KATT:
For reference, I published this


```
Successfully published:
 - @trpc/client@10.25.2-alpha-feat-out-of-order-batch-response-2023-05-23-10-11-21.92+09ae18012
 - @trpc/next@10.25.2-alpha-feat-out-of-order-batch-response-2023-05-23-10-11-21.92+09ae18012
 - @trpc/react-query@10.25.2-alpha-feat-out-of-order-batch-response-2023-05-23-10-11-21.92+09ae18012
 - @trpc/server@10.25.2-alpha-feat-out-of-order-batch-response-2023-05-23-10-11-21.92+09ae18012
```

Sheraff:
I can fix all that later today

> * I prefer the DX of having a bespoke `httpBatchStreamLink` as the configuration proposed in  [feat(`httpBatchLink`): alternative DX `streamRequester` for batch streaming #4403](https://github.com/trpc/trpc/pull/4403) looks a bit intimidating. That said, we could probably still reuse the code the way you're doing there but exposing it as two bespoke links

Damn I was kinda happy with how simple it looked to me.

> * We can see some blind spots [in the codecov](https://app.codecov.io/gh/trpc/trpc/pull/4347?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=trpc) - might be a hint that we have dead code paths or that we're not testing thoroughly

I do think that everything is tested in the "big tests", but I don't know how istanbul picks up what it does. I did add some bespoke tests for this reason, but I'll add more then.

The only thing that is not tested are these small paths in the various adapters:
```js
if (invalidInit || (abort && !firstChunk)) {
  res.statusCode = 500;
  return res.send();
}
```
The reason why they aren't tested is simply because it should not be possible that the condition is fulfilled and these 4 lines are basically here to make TypeScript happy. I could remove them and coerce TS.

> * Since we're using generators etc, it'd be cool if we could provide guidance on how to make this compatible with old browsers - I guess the user can do a `splitLink` to decide.. or we can automatically "degrade" to a non-streaming link if we know it ain't supported

This should be possible I think. I'll have to look up how to detect those APIs. However, support is fairly widespread: Chrome 63, Firefox 57, Safari 12, Node 10.

www/docs/client/links/httpBatchStreamLink.md
Sheraff: I did look for the fetch stuff, couldn't find a good way to detect "the shape of `response.body`". But having a `response.body` return a stream is *almost* as old as having support for `fetch` itself (Safari 11 vs Safari 10), and in addition to that, my implementation also supports undici and node-fetch. So I think at that point it's on the devs to decide whether they want a modern streaming request or not.

packages/server/src/adapters/fastify/fastifyRequestHandler.ts
Sheraff: Yes, it returns a `FastifyReply` (same as `.send()`) which is a thenable (i.e. promise-esque), and the linter wants promises to be explicitly awaited or voided.

www/docs/client/links/httpBatchStreamLink.md
Sheraff, line 16:
```suggestion
## Streaming mode (unstable)

> ⚠️ This link is unstable and may change in the future.
```

www/docs/client/links/httpBatchStreamLink.md
```diff
@@ -0,0 +1,50 @@
+---
+id: httpBatchStreamLink
+title: HTTP Batch Stream Link
+sidebar_label: HTTP Batch Stream Link
+slug: /links/httpBatchStreamLink
+---
+
+`unstable_httpBatchStreamLink` (experimental) is a [**terminating link**](./index.md#the-terminating-link) that batches an array of individual tRPC operations into a single HTTP request that's sent to a single tRPC procedure (equivalent to [`httpBatchLink`](./httpBatchLink.md)), but doesn't wait for all the responses of the batch to be ready and streams the responses as soon as any data is available.
```
Sheraff, line 8:
```suggestion
`unstable_httpBatchStreamLink` (unstable) is a [**terminating link**](./index.md#the-terminating-link) that batches an array of individual tRPC operations into a single HTTP request that's sent to a single tRPC procedure (equivalent to [`httpBatchLink`](./httpBatchLink.md)), but doesn't wait for all the responses of the batch to be ready and streams the responses as soon as any data is available.
```

packages/client/tsconfig.json
```diff
@@ -2,6 +2,9 @@
   "extends": "../../tsconfig.json",
   "include": ["src"],
   "compilerOptions": {
-    "types": ["node", "vitest/globals"]
+    "types": ["node", "vitest/globals"],
+    "paths": {
+      "tslib": ["./node_modules/tslib/tslib.d.ts"]
+    }
```
sachinraja, line 8: why is this needed?

packages/client/src/links/httpBatchLink/streamingHttpUtils.ts
```diff
@@ -0,0 +1,163 @@
...
+  getUrl,
+} from '../internals/httpUtils';
+import { HTTPHeaders } from '../types';
+
+/**
+ * @param readableStream as given by `(await fetch(url)).body`
+ * @param onSingle called for each line of the stream
+ * @param parser defaults to `JSON.parse`
+ */
+export async function parseJsonStream<TReturn>(
```
sachinraja, line 19: I assume all this works correctly but let's make sure we have a plan in place to maintain/change this code if necessary

packages/server/src/http/resolveHTTPResponse.ts
```diff
@@ -43,70 +42,151 @@ interface ResolveHTTPRequestOptions<
...
+export async function* resolveHTTPResponse<
+  TRouter extends AnyRouter,
+  TRequest extends HTTPRequest,
+>(
+  opts: ResolveHTTPRequestOptions<TRouter, TRequest>,
+): AsyncGenerator<
+  ResponseChunk | HTTPResponse,
+  ResponseChunk | undefined,
+  never
+> {
```
sachinraja, line 168: This is a breaking change (we export this from `@trpc/server/http` and community adapters use it.

packages/server/src/http/resolveHTTPResponse.ts
```diff
@@ -139,61 +219,74 @@ export async function resolveHTTPResponse<
...
-            input,
-            path,
-            error,
-          };
-        }
-      }),
+    /**
+     * when not streaming, do simple 2-step return
+     */
+    if (!isStreamCall || paths.length === 1) {
```
sachinraja, line 225: leaving a note for myself for later to review the changes here

sachinraja (commented):
not done reviewing yet but here are some comments. Awesome work on this! This is a huge change!

sachinraja (commented):
not done reviewing yet but here are some comments. Awesome work on this! This is a huge change!

packages/client/tsconfig.json
Sheraff: I don't remember but some APIs wouldn't work without it. I'll try removing it and see what error pops up to give you an answer.

packages/client/src/links/httpBatchLink/streamingHttpUtils.ts
Sheraff: What do you mean?

packages/server/src/http/resolveHTTPResponse.ts
Sheraff: Should I find another way then? Or are we going to make a major semver for this? Or other alternative?

packages/client/tsconfig.json
Sheraff: Looks like it's not needed anymore. Must have been for AsyncGenerators / AsyncIterators, which I've removed on the client now.

packages/server/src/http/resolveHTTPResponse.ts
Sheraff: BTW, the implementation in adapters/aws-lambda actually just consumes the iterator and returns header + status + body. So this would be a good "retro-compatible" example.

Sheraff: Ok to make life easier for everyone, I think I should remove the async generators from the server-side as well. I'll have a PR for this in a few minutes.

Sheraff: https://github.com/trpc/trpc/pull/4433

packages/server/src/adapters/aws-lambda/utils.ts
```diff
@@ -152,3 +152,6 @@ export type APIGatewayPayloadFormatVersion =
 export const UNKNOWN_PAYLOAD_FORMAT_VERSION_ERROR_MESSAGE =
   'Custom payload format version not handled by this adapter. Please use either 1.0 or 2.0. More information here' +
   'https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html';
+
+export const RESPONSE_ACCUMULATOR_FAILED_INITIALIZATION_ERROR_MESSAGE =
```
KATT, line 156:
```suggestion
```

packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
```diff
@@ -79,18 +116,15 @@ export async function nodeHTTPRequestHandler<
...
     }
-    for (const [key, value] of Object.entries(result.headers ?? {})) {
-      if (typeof value === 'undefined') {
-        continue;
-      }
-      res.setHeader(key, value);
-    }
-    res.end(result.body);
+
+    return opts.res;
```
KATT, line 128:
```suggestion
    if (!isStream) {
      return;
    }
    
    opts.res.write(formatter.end());
    opts.res.end();
```

packages/client/src/links/httpBatchLink/streamingHttpUtils.ts
```diff
@@ -11,22 +11,28 @@ import {
...
+   * As given by `(await fetch(url)).body`
+   */
+  readableStream: ReadableStream<Uint8Array> | NodeJS.ReadableStream;
+  /**
+   * Called for each line of the stream
+   */
+  onSingle: (index: number, res: TReturn) => void;
+  /**
+   * Called when the stream is finished
+   */
```
Sheraff, line 24: You seem to have gotten that comment from an earlier commit where I used multiple callbacks. `parse` is supposed to be basically `JSON.parse` or equivalent and is called on either 1. every line if we have a correct stream, or 2. the entire string if shape isn't recognized.

packages/server/src/http/resolveHTTPResponse.ts
```diff
@@ -41,72 +40,151 @@ interface ResolveHTTPRequestOptions<
...
+  paths: string[] | undefined,
+  type:
+    | Exclude<(typeof HTTP_METHOD_PROCEDURE_TYPE_MAP)[string], undefined>
+    | 'unknown',
+  responseMeta?: HTTPBaseHandlerOptions<TRouter, TRequest>['responseMeta'],
+  untransformedJSON?:
+    | TRPCResponse<unknown, inferRouterError<TRouter>>
+    | TRPCResponse<unknown, inferRouterError<TRouter>>[]
+    | undefined,
+  errors: TRPCError[] = [],
```
KATT, line 72: `max-params-2`

Also, this is a bit intense in the def - we can probably losen it a bit?

packages/server/src/http/resolveHTTPResponse.ts
```diff
@@ -139,61 +217,93 @@ export async function resolveHTTPResponse<
...
+
+      // yield header stuff
+      const headResponse = initResponse(
+        ctx,
+        paths,
+        type,
+        opts.responseMeta,
+        untransformedJSON,
+        errors,
+      );
```
KATT, line 245: Too many args

packages/server/src/http/resolveHTTPResponse.ts
```diff
@@ -139,61 +217,93 @@ export async function resolveHTTPResponse<
...
+      const chunk: ResponseChunk = [index, body];
+      onChunk(chunk);
+    }
+
+    // return nothing, signalling the request handler to end the streamed response
+    return {
+      status: headResponse.status,
+      headers: headResponse.headers,
+      body: `[${bodyParts.join(',')}]`,
+    };
```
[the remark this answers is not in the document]
KATT: Or maybe you changed this back for backwards compatibility.

You could do something like


```ts
// FIXME @deprecated - remove me in next major
return createFlatProxy<any>(key => {
  throw new Error(`Tried to access ${key} on 'resolveHTTPResponse()' where batching was used - this is a no-op and will be removed in next major of tRPC`
})
```

packages/server/src/http/resolveHTTPResponse.ts
```diff
@@ -211,18 +321,53 @@ export async function resolveHTTPResponse<
...
+      }),
+    };
+
+    // yield header stuff
+    /**
+     * WARN: this can cause issues if streaming response has already yielded its headers,
+     * because this would be the second time we yield headers.
+     * For this condition to be realized, we need to be sure that `serialize` in
+     * `transformTRPCResponse` cannot throw.
+     */
```
KATT, line 346: I think the two codepaths of streamed and non-streamed should be two distinctly different functions with their own error handling, we should never risk calling `initReponse` twice

KATT (changes requested):
Must change:

- [ ] No breaking changes
- [ ] Get rid of `void fn()`-esque where possible as that might crash Node.js and/or leave comments why
- [ ] [`max-params` `3`](https://github.com/trpc/trpc/pull/4445) 



Much appreciated

- [ ] Make code easier to follow:
  - [ ] `resolveHTTPResponse` should probably have two distinct paths with sub functions that have their own code paths - streamed and non-streamed - potentially a third with batched. Would clean it up a lot
  - [ ] We a bunch of cleanup in the batch & stream links... it's very hard to follow now. The HTTP links were already a mess after we added FormData but this adds even more complexity to it 😅

KATT (changes requested):
Must change:

- [ ] No breaking changes
- [ ] Get rid of `void fn()`-esque where possible as that might crash Node.js and/or leave comments why
- [ ] [`max-params` 2](https://github.com/trpc/trpc/pull/4445) 



Much appreciated

- [ ] Make code easier to follow:
  - [ ] `resolveHTTPResponse` should probably have two distinct paths with sub functions that have their own code paths - streamed and non-streamed - potentially a third with batched. Would clean it up a lot
  - [ ] We a bunch of cleanup in the batch & stream links... it's very hard to follow now. The HTTP links were already a mess after we added FormData but this adds even more complexity to it 😅

packages/server/src/http/resolveHTTPResponse.ts
KATT: Maybe it just handles too many cases. 

We should probably do ad-hoc ones for each code path rather than trying to reuse the code here, it's intense because it's so polymorphic

packages/client/src/links/httpBatchLink/streamingHttpUtils.ts
```diff
@@ -0,0 +1,170 @@
...
+  let fullAccumulator = '';
+  const onLine = (line: string) => {
+    if (opts.signal?.aborted) return;
+    if (isFirstLine) {
+      isFirstLine = false;
+      if (line !== '{') {
+        // unexpected format, don't try to parse it as a stream
+        isFullSink = true;
+        fullAccumulator = line;
+      }
```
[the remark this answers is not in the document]
Sheraff: My worry was that `transfer-encoding: chunked` is specced as a hop-by-hop header, meaning any intermediary (cache layer for example) can remove it, but the shape of the string body is still "not an array but an object with index keys".

But we can also consider that users are responsible for their own infra and if theirs doesn't support streaming, don't use that link.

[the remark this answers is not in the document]
KATT: If people start using the streaming link and get issues with that sort of stuff, they'll report it.

On your line of thinking, we should also support and valid JSON, since that intermediate layer can store the JSON in another format that then gets printed as 1 line or prettified

packages/server/src/adapters/fastify/fastifyRequestHandler.ts
```diff
@@ -43,7 +49,48 @@ export async function fastifyRequestHandler<
...
+  const onHead = (head: HTTPResponse) => {
+    if (!opts.res.statusCode || opts.res.statusCode === 200) {
+      opts.res.statusCode = head.status;
+    }
+    for (const [key, value] of Object.entries(head.headers ?? {})) {
+      /* istanbul ignore if -- @preserve */
+      if (typeof value === 'undefined') {
+        continue;
+      }
+      void opts.res.header(key, value);
```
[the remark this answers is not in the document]
Sheraff: This `void` is already here on the main branch. It seems kind of overkill to `.catch` it but we can if you want.

packages/server/src/adapters/fastify/fastifyRequestHandler.ts
```diff
@@ -43,7 +49,48 @@ export async function fastifyRequestHandler<
...
+      return;
+    }
+    if (!isStream) {
+      void opts.res.header('Transfer-Encoding', 'chunked');
+      void opts.res.header(
+        'Vary',
+        opts.res.hasHeader('Vary')
+          ? 'trpc-batch-mode, ' + opts.res.getHeader('Vary')
+          : 'trpc-batch-mode',
+      );
```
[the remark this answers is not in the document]
Sheraff: Related to above comment: I just replicated the already-present-on-main `void opts.res.header()`

packages/server/src/adapters/fastify/fastifyRequestHandler.ts
```diff
@@ -43,7 +49,48 @@ export async function fastifyRequestHandler<
...
+      stream = new Readable();
+      stream._read = () => {}; // eslint-disable-line @typescript-eslint/no-empty-function -- https://github.com/fastify/fastify/issues/805#issuecomment-369172154
+      resolve(opts.res.send(stream));
+      isStream = true;
+    }
+    stream.push(formatter(index, string));
+  };
+
+  void resolveHTTPResponse({
```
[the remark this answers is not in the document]
Sheraff: I'm sure we shouldn't await, because of streaming, we return a Response, and leave the promise run beyond. But it still return a promise because other servers need to await the promise (Node "express-like").

packages/server/src/http/resolveHTTPResponse.ts
Sheraff: This is not really complicated because it handles both streaming and regular. It's just complicated because I've taken it out of the long function body to isolate it in a separate function. The logic inside is the same as before, only the types are added, and sadly they are as minimal as can be to still work with `responseMeta`.

packages/server/src/http/resolveHTTPResponse.ts
```diff
@@ -138,66 +257,88 @@ export async function resolveHTTPResponse<
...
-        return {
-          result: {
-            data: obj.data,
-          },
-        };
+      } catch (cause) {
+        // at this point it's too late to fail gracefully
+        // because headers were sent and stream has started
+        // TODO: we might imagine some protocol using a custom HTTP Trailer to send errors after streaming
+        console.error('Unexpected error in streaming response', cause);
```
Sheraff, line 332: Not sure what would be the best way to handle this error. At this point, streaming has started, headers went off, there's nothing more to be done.

- The error can only come from `transformTRPCResponse` (so basically, the `.serialize()` function).
- client-side, missing procedure results in a stream are handled (rejected)

Possible improvements are
- log the error client-side by passing it as a `Trailer` after the stream (not sure how useful that is, and it would add client-side code)
- add a `try/catch` inside the `for` loop so that 1 failure of `.serialize` doesn't prevent the entire stream (on a regular batch response, 1 failure fails the entire batch)

packages/client/package.json
```diff
@@ -88,6 +88,7 @@
     "isomorphic-fetch": "^3.0.0",
     "node-fetch": "^3.3.0",
     "rollup": "^2.79.1",
+    "tslib": "^2.5.0",
```
Sheraff, line 91: I think this can be removed.

examples/next-minimal-starter/src/pages/api/trpc/[trpc].ts
```diff
@@ -15,7 +15,8 @@ const appRouter = router({
         name: z.string().nullish(),
       }),
     )
-    .query(({ input }) => {
+    .query(async (opts) => {
+      const { input } = opts;
```
Comment (Sheraff) [inline examples/next-minimal-starter/src/pages/api/trpc/[trpc].ts:19]: Revert before merging (for cleaner git history)

packages/server/src/http/batchStreamFormatter.ts
```diff
@@ -0,0 +1,30 @@
...
+ * res.send(formatter.end());
+ * ```
+ *
+ * Expected format:
+ * ```json
+ * {
+ * "1":"response #2"
+ * ,"0":"response #1"
+ * }
+ * ```
```
Sheraff, line 19:
```suggestion
 * ```json
 * {"1":"response #2"
 * ,"0":"response #1"
 * }
 * ```
```

www/docs/client/links/httpBatchStreamLink.md
```diff
@@ -0,0 +1,90 @@
...
+  trpc.post.byId.query(2),
+  trpc.post.byId.query(3),
+]);
+```
+
+## Streaming mode
+
+> ⚠️ This link is unstable and may change in the future.
+
+When batching requests together, the behavior of a regular `httpBatchStreamLink` is to wait for all requests to finish before sending the response. If you want to send responses as soon as they are ready, you can use this link. This is useful for long-running requests.
```
Sheraff, line 47: This sentence is a description of the differences between `httpBatchStreamLink` and `httpBatchLink`. It is normal to have `httpBatchLink` mentioned in it.

```suggestion
When batching requests together, the behavior of a regular `httpBatchLink` is to wait for all requests to finish before sending the response. If you want to send responses as soon as they are ready, you can use `httpBatchStreamLink` instead. This is useful for long-running requests.
```

www/docs/client/links/httpBatchStreamLink.md
```diff
@@ -0,0 +1,90 @@
...
+For runtimes other than the browser ones, the `fetch` implementation should support streaming, meaning that the response obtained by `await fetch(...)` should have a `body` property of type `ReadableStream<Uint8Array> | NodeJS.ReadableStream`, meaning that:
+
+- either `response.body.getReader` is a function that returns a `ReadableStreamDefaultReader<Uint8Array>` object
+- or `response.body` is a `Uint8Array` `Buffer`
+
+This includes support for `undici`, `node-fetch`, native Node.js fetch implementation, and WebAPI fetch implementation (browsers).
+
+## Compatibility (server-side)
+
+> ⚠️ for **aws lambda**, `unstable_httpBatchStreamLink` is not supported (will simply behave like a regular `httpBatchStreamLink`). It should not break anything if enabled, but will not have any effect.
```
Sheraff, line 84: Same as above, we're actually referring to `httpBatchLink` here because for aws lambda, both links behave the same.

```suggestion
> ⚠️ for **aws lambda**, `unstable_httpBatchStreamLink` is not supported (will simply behave like a regular `httpBatchLink`). It should not break anything if enabled, but will not have any effect.
```

packages/server/src/http/resolveHTTPResponse.ts
```diff
@@ -41,72 +46,198 @@ interface ResolveHTTPRequestOptions<
...
+  /**
+   * Called for every procedure with `[index, result]`.
+   *
+   * Will be called a single time with `index = -1` if
+   * - response is an error
+   * - response is empty (HEAD request)
+   *
+   * Without this callback, streaming is disabled.
+   */
+  onChunk: (chunk: ResponseChunk) => void;
```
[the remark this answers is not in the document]
KATT: Oh no I read this wrong, you did it nicely

packages/server/src/http/resolveHTTPResponse.ts
```diff
@@ -319,42 +367,40 @@ export async function resolveHTTPResponse<
...
-            index,
-            promise.then((r) => [index, r] as const),
-          ]),
+      const headResponse = initResponse({
+        ctx,
+        paths,
+        type,
+        responseMeta: opts.responseMeta,
+      });
+      onHead(headResponse);
```
Sheraff, line 371: It's fine if this fails through the "normal" error handling

packages/server/src/http/resolveHTTPResponse.ts
```diff
@@ -319,42 +367,40 @@ export async function resolveHTTPResponse<
...
+      for (let i = 0; i < paths.length; i++) {
+        const [index, untransformedJSON] = await Promise.race(
+          indexedPromises.values(),
         );
-        for (let i = 0; i < paths.length; i++) {
-          const [index, untransformedJSON] = await Promise.race(
-            indexedPromises.values(),
-          );
-          indexedPromises.delete(index);
+        indexedPromises.delete(index);
```
Sheraff: We know this can't fail since all procedure calls are wrapped inside a try/catch themselves

iamnafets:
This is epic @Sheraff and @KATT ! Was literally starting to put this together last night. Nice win for performance but also opens up some new options for streaming through the stack.

helmturner:
I have a question the answer to which may break me...

Does the use of streaming here only affect batched requests by returning individual requests through the stream? For example, if I'm using this link and send a singular request that calls a procedure that returns a ReadableStream of Server-Sent Events, will the `data` property returned by the hook be populated _before_ or _after_ all chunks have been streamed? If before, would the value be the chunk itself?

I ask because I was attempting to implement the GPT-4 API with streaming responses and, to avoid waiting actual **minutes** for a response, I'm streaming the chunks over SSE via a NextJS/Vercel Edge Route and concatenating them in component state. 

The reason I say it might break me... I literally just implemented a link and an adapter for exactly my use case. When I installed `@trpc/server` as a peer-dep, I noticed the new version introduced a type error about a missing `eagerGeneration` key, which is how I found out this shipped the same day 😵‍💫

Sheraff:
@helmturner

> Or, put another way - is each chunk a procedure result, or a piece of a procedure result?

Each chunk is a procedure result, sorry... You can see the chunks in the network tabs of your devtools: we're doing a form of "line delimited json" so each line is a chunk.

But we've marked this link as "unstable" because, even though it works and the functionality is here to stay, we're still open to making changes to the underlying implementation so we're open to good ideas.

There were talks of implementing streaming *within individual procedures*, but we couldn't think of a clean way to present the data client-side. Do we want promises inside the returned object? If it's a "stream of events" (like in the case you're describing) do we want a promise inside a promise inside a promise... And then you have to `use()` them in your react components. 

> I noticed the new version introduced a type error about a missing `eagerGeneration` key

Could you tell me more about this type error? `eagerGeneration` is indeed a new key, but we were trying to ship this feature without any breaking change so if you had an error we might have failed on that front and I'd love to rectify that.

ItsWendell:
@Sheraff this breaks on React Native (without polyfills) since TextDecoder is not available in the hermes engine.

Sheraff:
> @Sheraff this breaks on React Native (without polyfills) since TextDecoder is not available in the hermes engine.

Good to know. I have a few questions then:
- Does tRPC as a whole still work normally (if you don't import the `unstable_httpBatchStreamLink` and use another terminating link instead)?
- Is there another native API that exists in place of `TextDecoder`?
- Did you try it with a polyfill then? Did that work ok?

Right now I'm thinking all I can do is mention this in the docs, and maybe provide a hook to replace `TextDecoder` (the same way there are hooks to replace `fetch` for example) for those who don't want to add globals to their runtime. But I'm open to ideas.

ItsWendell:
@Sheraff `const textDecoder = new TextDecoder();` is defined in the global scope within parseJsonStream, I think this causes the whole app to crash with the following error:

```
ReferenceError: Property 'TextDecoder' doesn't exist, js engine: hermes 
    at RootRouteNodeProvider (http://192.168.2.25:19000/index.bundle?platform=android&dev=true&hot=false:118657:23)
    at ContextNavigator (http://192.168.2.25:19000/index.bundle?platform=android&dev=true&hot=false:123320:24)
    at RNCSafeAreaProvider
    at SafeAreaProvider (http://192.168.2.25:19000/index.bundle?platform=android&dev=true&hot=false:112450:24)
    at RNGestureHandlerRootView
    at GestureHandlerRootView
    at GestureHandler
    at ExpoRoot (http://192.168.2.25:19000/index.bundle?platform=android&dev=true&hot=false:123307:24)
    at App
    at withDevTools(App) (http://192.168.2.25:19000/index.bundle?platform=android&dev=true&hot=false:105050:27)
    at RCTView
    at View (http://192.168.2.25:19000/index.bundle?platform=android&dev=true&hot=false:48075:43)
    at RCTView
    at View (http://192.168.2.25:19000/index.bundle?platform=android&dev=true&hot=false:48075:43)
    at AppContainer (http://192.168.2.25:19000/index.bundle?platform=android&dev=true&hot=false:47952:36)
    ```
    https://github.com/trpc/trpc/blob/816da0c71b58d251297b6fd7ed651641409ffcaa/packages/client/src/links/internals/parseJSONStream.ts#L61

To answers your questions:
- So no, TRPC as a whole does not work normally. The encoder was probably placed in the global scope to hopefully win some performance by not having to initialize it again, moving it into the function that actually uses it would fix it. Not sure what the exact performance impact it but it's probably neglable.
- I'm not sure what the best 'native' approach would be for this API, from a quick google search, I don't think React Native / Hermes has something (simple) built-in to do text encoding / decoding.
- Did not try it (yet) with a polyfill.

**Comment by Sheraff** [comment]
> seems to be causing the entire app to crash

Damn I'm sorry. It is declared outside of a function, but still in a module that shouldn't be imported (and can even be tree-shaken out) if you don't use it. So I'm surprised the consequences are so radical, but I'll try and fix this asap.

**Comment by ItsWendell** [comment]
@Sheraff Thanks for your quick replies! I made a pull requests for a potential (temporary) solution here: https://github.com/trpc/trpc/pull/4480 that would fix it.

I'm not sure why it's not being tree-shaken, might not happen properly in development, we were importing TRPC in our Provider like this:

```
import {
  createTRPCProxyClient,
  httpBatchLink,
  httpLink,
  inferRouterProxyClient,
} from "@trpc/client";
```

helmturner:
@Sheraff The type error is specific to my custom adapter. I was importing some types from the HTTP utils, while others were customized - so the new required property caused a mismatch between an imported interface and a custom interface.

I have the adapter up at github.com/alecvision/trpc-sse-adapter and the link will follow soon.

It's largely based off of the HTTP link and fetch adapter, but I'm using subscription procedures and applying a bit of a hack to route requests around the websocket link (pattern-matching procedures, which requires the hack to be applied in both the server _and_ client configs).

The only clean API I could think of involves adding a new procedure type with the exact same semantics as a subscription, but that uses a header or HTTP method (I'm using both, and I picked PATCH). Requests that include the correct method or header are then routed through a link/adapter like the one I hacked together.

Unfortunately, while I'm pretty handy with TypeScript, the types involved in procedures would take more time to grock than I can afford ATM. If you can provide any insight on how I might implement a new procedure type, I'd be happy to give it a shot and put in a PR!

Sheraff:
Reported issues have been fixed in the latest release https://github.com/trpc/trpc/releases/tag/v10.29.1

Issue #3297, opened 2022-11-27
feat: Support RSC & App Layouts


KATT:
### Describe the feature you'd like to request

We should make official support for Next 13 app layouts & RSC.

### Describe the solution you'd like to see

Being able to transparently use tRPC identically in RSC-components and in `"use client"`-components. Unsure how they can be hydrated.

This is very open-ended and hard to answer without tonnes of exploration. We've done some in https://github.com/trpc/next-13

### Describe alternate solutions

n/a

### Additional information

https://github.com/trpc/next-13

### 👨‍👧‍👦 Contributing

- [ ] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

toddgeist:
tldr; Have `createTRPCNext` return a Provider like the `creatTRPCReact` version, and have it follow the `srr:false` config

We have settled on using the Next13 `app` directory but not using any new data-fetching stuff yet. We are still trying to use TRPC for client-side data fetching. This seems to work so far, although getting set up is a bit tricky.  

The main idea is to think of it as using plain node or express on the back end and a separate react app on the front. There are two configs for TRPC, one for the Next API handler and one for the client. This works, but since we are running next, those client components may run on the server on the first-page load, which breaks stuff. The workaround is to only do the TRPC queries inside dynamically imported components with `ssr:false`

Al this works, but it is a bit of a pain, and I feel some tweaks could make this setup easier. 

And most importantly, I think this is a worthwhile goal. The ultimate expression of what TRPC needs to be in the new RSC paradigm is an important question, but it will take some time to figure out. But in the meantime, if we could make this approach easier, it would give more people an opportunity to try out TRPC in next13 and provide feedback.

I don't feel I know the codebase well enough to tackle a PR on my own, but perhaps with some guidance I could get something together.

Thought?

AlanMorel:
Hello, I was wondering if there's been any progress on this issue.

With Next.js working towards stabilizing the app directory, more people will be switching over and with that wishing to use tRPC with it.

Cheers!

spencerbart:
Cheering on this development! I'm excited to see it soon.

ansh:
I stopped using the `app` directory because of this issue

akomm:
I don't say that everything is fine and perfect, just try to figure out what we have right now and what **probably** works:

**Server:** 
The API itself isn't a problem, remained  the same within `pages/api`. 

**Client:**
[...]

PR #4359, opened 2023-05-12
feat(`next`): experimental server actions

KATT:
#3297


packages/next/src/app-dir/server.ts
```diff
@@ -32,3 +38,48 @@ export function experimental_createTRPCNextAppDirServer<
...
+
+export function experimental_createServerActionHandler<
+  TInstance extends AnyTRPCInstance,
+>(
+  _t: AnyTRPCInstance,
+  opts: {
+    createContext: () => MaybePromise<TInstance['_config']['$types']['ctx']>;
+  },
+) {
+  // TODO allow this to take a `TRouter` in addition to a `AnyProcedure`
```
juliusmarminge, line 64: Why?

KATT: Make a full router into actions would be cool

juliusmarminge: And then you'd pass it into the hook (or just <form action={}/>) like `actions.post.create`

Yea that'd be cool :)

PR #4371, opened 2023-05-16
fix(`react-query`): query and mutation defaults being overwritten in hook wrappers

Haaxor1689:
Closes #

## 🎯 Changes

Wrapper around useMutation hook was not honoring options set by `getMutationDefaults` hook at all. Default options are now loaded as well and used as a fallback to wherever `opts` are used.

## ✅ Checklist


KATT (commented):
Shouldn't this "just work" automatically by react-query? What is exactly being overwritten?

Haaxor1689:
No because for example with onSuccess handler the trpc wrapper adds it's own onSuccess where it calls the onSuccess from options passed to the wrapper so it isn't overwritten. But if there was onSuccess defined in global settings, then the wrapper wasn't checking for it, thus not calling it.

KATT (approved):

PR #4374, opened 2023-05-17
fix: relax input type conditions for exposing useInfiniteQuery

lithdew:
## 🎯 Changes

Suppose we define a router procedure's input with Zod as `z.object({ cursor: z.string().optional() }).default({})`. The inferred input type, which may be checked using `inferRouterInputs<Router>["procedure"]`, erroneously is the type union `void | {cursor?: string} | undefined`.

The error is caused by `inferProcedureInput` union-ing the input type with void should the input potentially be undefined.

useInfiniteQuery at the moment is only exposed for procedures whose inputs solely extend `{cursor?: any}`. Procedures whose inputs are potentially undefined, that guarantee** after parsing to set defaults to the `cursor` field however should also expose useInfiniteQuery.

This PR relaxes the input types of useInfiniteQuery to allow inputs that are potentially undefined by exposing useInfiniteQuery for procedures whose inputs extends the type `{cursor?: any} | void`.

## ✅ Checklist


KATT (changes requested):
I want some tests on this - you can see #4378 for reference.

Since it's only typescript-stuff, you can wrap it in `ignoreErrors(() => ` and add some tests type tests.

lithdew:
> I want some tests on this - you can see #4378 for reference.
> 
> Since it's only typescript-stuff, you can wrap it in `ignoreErrors(() => ` and add some tests type tests.

Wrote some type-checking tests to ensure that `useInfiniteQuery` is exposed on procedures whose inputs are optional.

packages/tests/server/react/infiniteQuery.test.tsx
```diff
@@ -391,4 +392,24 @@ describe('Infinite Query', () => {
...
+
+    assertType<Extends<Input, { cursor?: string }>>(false);
+    assertType<Extends<Input, { cursor?: string } | void>>(true);
+    assertType<Extends<Input, { cursor?: string } | undefined>>(true);
+    assertType<Extends<Input, { cursor?: string } | void | undefined>>(true);
+
+    // Assert 'useInfiniteQuery' is exposed in 'trpc.paginatedPosts'.
+
+    expectTypeOf(trpc.paginatedPosts.useInfiniteQuery).toBeFunction();
+  });
```
KATT, line 414:
```suggestion
```



  test('useInfiniteQuery() is **not** exposed if there is not cursor', () => {
    ignoreErrors(async () => {
      // @ts-expect-error 'cursor' is required
      factory.trpc.postById.useInfiniteQuery;
      const ssg = createServerSideHelpers({
        router: factory.appRouter,
        ctx: {},
      });

      // good
      await ssg.paginatedPosts.fetchInfinite({ limit: 1 });

      // @ts-expect-error 'cursor' is required
      await ssg.postById.fetchInfinite({ limit: 1 });
    });
  });

KATT (dismissed):

PR #4379, opened 2023-05-17
chore: add PR naming convention

KATT:
OK naming:

- chore: x
- fix(`x`)
- chore(`x`)
- feat(`x`)

### DIFF
--- a/.github/workflows/pr-naming.yml
+++ b/.github/workflows/pr-naming.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: deepakputhraya/action-pr-title@master
         with:
           allowed_prefixes: 'feat,fix,chore' # title should start with the given prefix
-          regex: '/^(chore|docs|(feat|fix|chore)\(`[^`]*`\)): /' # Regex the title should match.
+          regex: '^(chore|docs|(feat|fix|chore)\(`[^`]*`\)): ' # Regex the title should match.
           prefix_case_sensitive: true # title prefix are case insensitive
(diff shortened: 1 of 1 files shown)


bautistaaa (approved):
oh hell yea

.github/workflows/pr-naming.yml
```diff
@@ -0,0 +1,20 @@
+name: Pull Request naming convention
+# This workflow is triggered on pushes to the repository.
+on:
+  pull_request:
```
juliusmarminge, line 4: I think this must be a pull_request_rather if the workflow uses a secret and should be exposed to forks

juliusmarminge: We have something dimilar in t3: https://github.com/t3-oss/create-t3-app/blob/next/.github/workflows/pr-labeler.yml

KATT: it seems to work 🤷 

try renaming the PR

juliusmarminge: Im not on a form 😎

KATT: ohhhhh,

bautistaaa: the docs seem diff for both but i almost wonder should we use the one you're using on T3. seems more popular 🤣

KATT: https://github.com/trpc/trpc/pull/4380

PR #4403, opened 2023-05-22
feat(`httpBatchLink`): alternative DX `streamRequester` for batch streaming

Sheraff:
Proposal for https://github.com/trpc/trpc/pull/4347

## 🎯 Changes

Avoid duplicating `httpBatchLink` logic into `httpBatchStreamLink` by simply allowing for the single part that is different (the `fetch` method) to be passed as an option in the instantiation of the link.

Allows for
- DRYer code
- tree shaking

The new option is called `requester` and defaults to the "normal" `batchRequester`:
```ts
httpBatchLink({
  requester: unstable_streamRequester,
})
```

**Another possibility** (though not what is implemented here) would be to have a `loader` option (instead of the `requester` option) which would include the `validate` method in addition to the `fetch` method. While it's not required in this case, it might be a good way to future-proof this option.

## ✅ Checklist


www/docs/client/links/httpBatchLink.md
```diff
@@ -144,6 +144,46 @@ export const trpc = createTRPCNext<AppRouter>({
...
+      // 👇 enable streaming mode
+      requester: unstable_streamRequester,
+    }),
+  ],
+});
+```
+
+Using the `streamRequester` will:
+
+- Cause the requests to be sent with a `X-Trpc-Batch-Mode: stream` header
```
KATT, line 174: I think we can just do `trpc-xxx` :D 

CleanShot 2023-05-23 at 09 24 07@2x

PR #4410, opened 2023-05-24
fix(`next)`: rm conditional hook in `withTRPC`

KATT:


mozzius (approved):
🤦‍♀️ oops - I must have missed that it was a hook. Sorry!

PR #4417, opened 2023-05-25
fix: exclude `*.test.*`-files in build outputs

KATT:


package.json
```diff
@@ -22,6 +22,7 @@
     "lint-prune": "! ts-prune | grep -v \"used in module\"",
     "clean": "find . -name node_modules -o -name .turbo -o -name .next -o -name dist -type d -prune | xargs rm -rf",
     "codegen:override-prisma": "tsx scripts/addPrismaOverrides.ts",
+    "codegen:entrypoints": "turbo --filter=\"@trpc/*\" codegen:entrypoints",
```
KATT, line 25: I am stupid and don't understand turbo completely, but in order to trigger this I had to do `p codegen:entrypoints -- no-cache`.

Not sure if this should even be here or if it should just be magically triggered 😅

juliusmarminge: it should be triggered automatically on build

juliusmarminge: maybe package.json should be in the outputs too actually

juliusmarminge, line 25:
```suggestion
```

turbo.json
```diff
@@ -1,5 +1,6 @@
 {
   "$schema": "https://turborepo.org/schema.json",
+  "globalDependencies": ["tsconfig.json", "scripts/entrypoints.ts"],
```
KATT, line 3: cc @juliusmarminge this seems to do it - looks good to you?

juliusmarminge: sure!

juliusmarminge: im still getting the dev errors though 
CleanShot 2023-05-25 at 14 10 55@2x

KATT: yeah, I tried to fix that but got deralied and fixed this instead :D

PR #4438, opened 2023-05-28
fix(`client`): remove properties from `ResponseEsque` that are not used

tjenkinson:
## 🎯 Changes

Removes all properties from the fetch response type that are not actually used. This is so that the `fetch` ponyfill also doesn't have to implement unnecessary parts.

Specifically I'm using this library in a Cloudflare worker to make requests to a Cloudflare Durable object worker, and to do that I'm providing the durable object fetch function. This works perfectly 🎉  except it hits a type error though because the `Response` type in`@cloudflare/workers-types` ([visible here](https://www.npmjs.com/package/@cloudflare/workers-types?activeTab=code)) is 

```ts
declare class Response extends Body {
  constructor(body?: BodyInit | null, init?: ResponseInit);
  static redirect(url: string, status?: number): Response;
  static json(any: any, maybeInit?: ResponseInit | Response): Response;
  clone(): Response;
  readonly status: number;
  readonly statusText: string;
  readonly headers: Headers;
  readonly ok: boolean;
  readonly redirected: boolean;
  readonly url: string;
  readonly webSocket: WebSocket | null;
  readonly cf?: any;
}
```

which is missing `type` which (without this change) is required, but not used.

Here is the error:

image

Bug fix

## ✅ Checklist


KATT (approved):

KATT (approved):

huw:
Thanks mate, you’re a hero

PR #4017, opened 2023-03-18
New `www` page to explain mocking

jharlowuk:
Hey folks! Thank you so much for creating tRPC, I'm sure you don't need more compliments but I am using it pretty much everyday and completely adore what it's done for my productivity 😄 

## 🎯 Changes
- This PR aims to add a new page called `Mocking` that goes through the steps needed to mock tRPC calls from components being rendered in isolated environments (tests, Storybook, etc.)
- The page goes through why mocking at the Service Worker level is worthwhile, how to create a `TRPCMockedProvider` component that can be used in various context, and how to write performant, easy to execute frontend unit tests on React components that contain tRPC queries and mutations

## 👀 Why I wrote this
In setting up my own tRPC client and server in Next.js, getting set up with unit tests has taken a large amount of time. I understand the appeal of having a local backend to test on, however, I felt like there are enough people asking for mocking and mocking is a standard and good enough practice that having a mocking page would be valuable to those looking to do it. 

I've also tried to make the page focused enough on "mocking" nomenclature that I don't think someone looking specifically to an integration or E2E test would find this page first. I hope this is acceptable! If not, I'm happy to find somewhere else to post this ❤️ 

## ✅ Checklist
- [X] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).

PR #4451, opened 2023-05-30
docs: mocking

c-ehrlich:
Closes #

## 🎯 Changes

Adds a page about mocking to the docs

Based on https://github.com/trpc/trpc/pull/4017

## ✅ Checklist

Todo for a future PR: either move the reference repo into the tRPC examples, or modify one of the existing examples to include a mock


KATT (commented):
Since this relies on an external package, maybe we should instead link to docs there?

c-ehrlich:
> Since this relies on an external package, maybe we should instead link to docs there?

I can see both sides on this. Not sure what's better

The case in favor of including this:
- more visibility / discoverability
- could also add other testing docs (for example how to unit test a procedure with `createCaller` and monkey patched context)

The case against:
- external package
- API probably still unstable

Happy to close this and open a PR on `msw-trpc` instead if you think it should go there :)

sachinraja:
I agree with Alex, if we're putting it in our docs then we're essentially saying we'll keep it maintained like a core package.

PR #4455, opened 2023-05-31
chore: revert pnpm updrade as it breaks some example projects

KATT:
new lockfiles from #4437 breaks deployments e.g.

https://github.com/trpc/examples-next-prisma-starter

#4456  - Would like for renovate to stop updating pnpm and node automatically, it's a bit more involved than what it is capable of

### DIFF
--- a/.github/workflows/subtree.yml
+++ b/.github/workflows/subtree.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - maybe
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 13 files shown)


.github/workflows/subtree.yml
```diff
@@ -70,3 +71,4 @@ jobs:
           path: 'examples/${{ matrix.path }}'
           deploy_key: ${{ secrets.TRPC_DEPLOY_TOKEN }}
           force: true # will force push to the downstream repository
+          branch: main # will push to the main branch of the downstream repository
```
KATT, line 74: Allows us to test subtree on branches without having to clean them up afterwards

PR #4397, opened 2023-05-19
chore: fix next-app-dir example

KATT:
Closes #4396

## 🎯 Changes

Not working. Just demoing

PR #4462, opened 2023-06-01
fix: lint error on main

juliusmarminge:
not sure how it got there and why it didn't give error when you added this plugin in #4445

also wasn't sure what the best approach was - either disable the rule in tests or just disable this one line. i opted for the latter


juliusmarminge:
moved this commit to #4397 to unblock me :)

Issue #4460, opened 2023-06-01
QoL: fix vendored jest_dom types


juliusmarminge:
### Provide environment information

main

### Describe the bug

<img width="1021" alt="CleanShot 2023-06-01 at 09 30 02@2x" src="https://github.com/trpc/trpc/assets/51714798/6ae4d244-3bcd-49fe-b47e-b4394c66b070">

### Link to reproduction

https://github.com/trpc/trpc

### To reproduce

`pnpm build` in root

### 👨‍👧‍👦 Contributing

- [ ] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

KATT:
I reverted the changes in #4463 to do with the jest helpers

For reference and more context - I want us to have all the test dependencies in the root of our project so we can throw test files in any package without headaches

varugasu:
Is this still an issue? I just ran `pnpm build` and no warnings

PR #4463, opened 2023-06-01
fix(`server`): fix minor build warnings due to usage of private fields

kamilogorek:
ref https://github.com/trpc/trpc/issues/4460 and fixes the additional unrelated issues of:

```
@trpc/server:build: (!) Unresolved dependencies
@trpc/server:build: https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
@trpc/server:build: @swc/helpers/src/_class_private_field_get.mjs (imported by src/adapters/node-http/content-type/form-data/streamSlice.ts)
@trpc/server:build: @swc/helpers/src/_class_private_field_init.mjs (imported by src/adapters/node-http/content-type/form-data/streamSlice.ts)
@trpc/server:build: @swc/helpers/src/_class_private_field_set.mjs (imported by src/adapters/node-http/content-type/form-data/streamSlice.ts)
```

It was complaining because vendored type is not used anywhere outside of tests, and the tests package has no `build` step.


juliusmarminge:
Woo can we update SWC now too then? 

CleanShot 2023-06-01 at 10 59 34@2x

KATT (dismissed):

KATT (approved):

PR #4480, opened 2023-06-03
fix: TextDecoder in parseJsonStream's global scope causes unsupported…

ItsWendell:
… environments to crash

This is a hotfix to prevent React Native clients from crashing, by putting the `const textDecoder = newTextDecoder()` within the parseJsonStream function itself we prevent it from always being initialized, and people who use React Native can opt out of using this batch stream link.

Closes #

## 🎯 Changes

## ✅ Checklist


KATT:
Hey @ItsWendell, I think this should be fixed in the latest release, sorry for the issue! Open a bug if it doesn't work :)

See #4481

PR #4481, opened 2023-06-03
fix(`server` + `client`): Post release `httpBatchStreamLink` issues

Sheraff:
### `TextDecoder` doesn't exist in the global scope for React Native https://github.com/trpc/trpc/pull/4347#issuecomment-1575105895

Commit: https://github.com/trpc/trpc/pull/4481/commits/f877b24ccf38f76f365b66d8e850ca464a643694

- Added a `textDecoder` option to the instantiation of `httpBatchStreamLink`:
  ```ts
  unstable_httpBatchStreamLink({
    url: 'http://localhost:3000',
    textDecoder: new TextDecoder(),
    // ^? textDecoder: { decode: (input: Uint8Array) => string }
  })
  ```
  This is not required to be provided, but allows for overriding the native API (like is already done for `fetch`).

- The `TextDecoder` API is now accessed from the global scope `global.TextDecoder`
  
  ❓️❓️Which is the most supported one? `global` or `globalThis`? ❓️❓️

- The `TextDecoder` API is now accessed from inside a function, so should not crash the client, *even if it wasn't tree-shaken out**

### on Next.js 12, `httpBatchStreamLink` works like a regular `httpBatchLink`

Commit: https://github.com/trpc/trpc/pull/4481/commits/cbe8398c0f3b4bc08284a3ab1024b0c915909c7c

On some servers, "pushing data to a stream" is not sufficient to "stream a response". We might have several cases like this (different adapters, different uses...), but so far I've only encountered 1.

On Next.js <13 and on Express with a `compress()` link, the response isn't streamed because it's being gzipped. Both of these use the `nodeHTTPRequestHandler` and in both of these cases, there is a method added to the `res` object: [`res.flush()`](https://expressjs.com/en/resources/middleware/compression.html).

- add optional `flush` key on `NodeHTTPResponse` type (which is accurate: sometimes it's there sometimes not)
- optional chaining call of `.flush()` on every `.write()` for a streamed response
- rearrange adapters so that the "setting up a stream" phase happens in `onHead` instead of first `onChunk`. This was not really necessary for this fix, but seems more correct and might make future similar fixes easier.

### the new version introduced a type error about a missing `eagerGeneration` key https://github.com/trpc/trpc/pull/4347#issuecomment-1574669152

Commit: https://github.com/trpc/trpc/pull/4481/commits/4e8c26027ccd80b1b57ba680f11144dd4d99a80b

Version 10.29 added a new `eagerGeneration` key in the arguments passed to `ResponseMetaFn` to indicate that this function is being called *before* knowing the response body. This key is always provided by `resolveHTTPResponse` but some people seem to have been using the type externally. So we're making it optional.

## ✅ Checklist


packages/client/src/links/internals/httpUtils.ts
```diff
@@ -138,6 +138,7 @@ export const jsonHttpRequester: Requester = (opts) => {
 export type HTTPRequestOptions = HTTPBaseRequestOptions &
   ContentOptions & {
     headers: () => HTTPHeaders | Promise<HTTPHeaders>;
+    TextDecoder?: TextDecoder;
```
kamilogorek, line 141: We may use `TextDecoderEsque` here as well, just for the consistency's sake.

packages/client/src/links/httpBatchStreamLink.ts
```diff
@@ -7,13 +7,16 @@ import { streamingJsonHttpRequester } from './internals/parseJSONStream';
 import { Operation } from './types';
 
 const streamRequester: RequesterFn = (requesterOpts) => {
+  const textDecoder =
```
kamilogorek, line 10: I'd use a helper like we do for `fetch` https://github.com/trpc/trpc/blob/df2d2fb3cbbe5736ee6e6594d7b2bd01bd6c112d/packages/client/src/getFetch.ts#L12-L28 so it's more clear what went wrong in case this fails.

kamilogorek:
> ❓️❓️Which is the most supported one? global or globalThis? ❓️❓️

We use `globalThis` in other places.

packages/client/src/links/internals/httpUtils.ts
Sheraff: good catch

KATT (dismissed):

packages/client/src/links/HTTPBatchLinkOptions.ts
```diff
@@ -13,6 +14,12 @@ export interface HTTPBatchLinkOptions extends HTTPLinkBaseOptions {
     | ((opts: {
         opList: NonEmptyArray<Operation>;
       }) => HTTPHeaders | Promise<HTTPHeaders>);
+
+  /**
+   * Used for `httpBatchStreamLink`. Will default to the webAPI `TextDecoder`,
+   * but you can use this option if your client runtime doesn't provide it.
+   */
+  textDecoder?: TextDecoderEsque;
```
KATT, line 22: I would much prefer if this was part of **only** the stream options, so we'd have a `HTTPBatchStreamLinkOptions`

KATT (approved):

PR #4482, opened 2023-06-03
feat(react-query): allow for external apps in `createServerSideHelpers`

EmericW:
Closes #

## 🎯 Changes

Currently the `createServerSideHelpers` function has a required `router` property that expects the TRPC router to be provided. While this works in a NextJS project (Where the back-end is implemented in NextJS as well), this will not work for projects with a separate front-end and back-end.

This PR provides a solution by giving the implementor the option to provide an instance of the untyped client instead of the router.

## ✅ Checklist


EmericW:
Hi everybody,

I know this implementation is a bit rough around the edges right now, but I would like to know if a feature like this would even be considered.
Given some feedback/pointers I would love to finish this.

Things I know that need to be changes:
 - No breaking changes. Right now the addition of the required `type` property is a breaking change. Maybe we could export a new `createServerSideHelpers` variant all together?
 - Some tests seem to fail, but at first glance they seem to be related to the addition of the required `type` property.
 - Documentation

I'm guessing this has not yet been implemented for performance reasons. I know it is generally discouraged to make http requests from SSG/SSR functions to Vercel API's since you could possible hit a double cold start that way.
But that is not entirely relevant when working with an external API which could be hosted on a regular server.

packages/react-query/src/server/types.ts
```diff
@@ -1,14 +1,26 @@
...
 import {
   AnyRouter,
   ClientDataTransformerOptions,
   inferRouterContext,
 } from '@trpc/server';
 import { CreateTRPCReactQueryClientConfig } from '../shared';
 
-interface CreateSSGHelpersOptionsBase<TRouter extends AnyRouter> {
+interface CreateRouterSSGHelpersOptionsBase<TRouter extends AnyRouter> {
+  type: 'router';
```
KATT, line 10:
```suggestion
```

packages/react-query/src/server/types.ts
```diff
@@ -1,14 +1,26 @@
...
   router: TRouter;
   ctx: inferRouterContext<TRouter>;
   transformer?: ClientDataTransformerOptions;
 }
-export type CreateSSGHelpersOptions<TRouter extends AnyRouter> =
-  CreateSSGHelpersOptionsBase<TRouter> & CreateTRPCReactQueryClientConfig;
+
+interface CreateClientSSGHelpersOptionsBase<TRouter extends AnyRouter> {
+  type: 'client';
```
KATT, line 17:
```suggestion
```

KATT (dismissed):
I like this feature and am happy to get it in

What we need to do to finish it:

1. It can't be a breaking change so we need to remove the `type` discriminator or make it optional for the current default cause
2. We have to add tests - especially around how it works with e.g. errors


**Edit:** I missed your comment and it seems like you already knew of the things that needed fixing :)

EmericW:
I have remove the breaking change. This is the simplest I could come up with without allowing for invalid configs like: 
```
{
   router: appRouter,
   client: untypedClient,
   transformer: SuperJSON,
}
```

EmericW:
Another consideration:

With the current implementation you have to provide the AppRouter type twice.
Screen Shot 2023-06-09 at 17 40 53

If you don't provide it to the `createServerSideExternalHelpers` function, you get type errors when trying to access your procedures. Unfortunately I don't know of any reasonable methods to make this generic required.

Screen Shot 2023-06-09 at 18 08 18

Maybe we could change the API to expect the `CreateTRPCClientOptions` directly, instead of  an `TRPCUntypedClient `.

This doesn't solve the optional generic type issues. But it might lower the complexity of the whole thing a bit.

Screen Shot 2023-06-09 at 18 07 31

KATT:
You should be able to pass in a [`createTRPCProxyClient`](https://trpc.io/docs/client/vanilla/setup#3-initialize-the-trpc-client) instead and infer it from that.

e.g.


```ts
const client = createTRPCProxyClient<AppRouter>({
  links: [
    httpBatchLink({
      url: 'http://localhost:3000/api/trpc',
    }),
  ],
  transformer: SuperJSON,
});

const ssg = createServerSideExternalHelpers({
  client,
});
```

EmericW:
Would that be possible? I think we won't have access to the "flat" client methods we need in https://github.com/EmericW/trpc/blob/90619f4b3a6c1fc8e8b23477a77eb7038a84461c/packages/react-query/src/ssg/ssg.ts#L62 for example.

KATT:
> Would that be possible? I think we won't have access to the "flat" client methods we need in https://github.com/EmericW/trpc/blob/90619f4b3a6c1fc8e8b23477a77eb7038a84461c/packages/react-query/src/ssg/ssg.ts#L62 for example.

The prox could have a hidden `.__rawClient` or something that we can access internally and typecast

KATT:
That branch has now been merged to main so you'll be able to use that function now

EmericW:
My man! will continue the work on this soon.

EmericW:
FYI: After some manual testing on how errors are handled I drew the following conclusions:

 1. A logical error thrown inside the router is handled in the same way as with the `createServerSideInternalHelper`. The error is serialised in the output of the `dehydrate` call.
 2. A network error while performing the remote fetch is handled in the same way as when performing a query using the proxy client in `getStaticProps`. It just throws and stops the function call.

Based on these finding I don't think we need to implement any special error handling.

EmericW:
I added [a few more commits](https://github.com/EmericW/trpc/commits/feature/external-app-ssg-helpers) to my fork updating the documentation, but they don't seem to be reflected here.

KATT:
moved to #4547

PR #4489, opened 2023-06-06
[RFC] feat(`react-query` `server` `client`) Streaming results from mutations and queries

### LINKED ISSUE #4477: feat: Streaming Mutations / Queries

```typescript
publicProcedure
     for(let x=0x;<10;x++) {
       yield x;
       await sleep(100);
     }
   });
```

### Additional information

### 👨‍👧‍👦 Contributing

## Funding

<picture>
  Fund with Polar
</picture>
</a>

```[tasklist]
### Tasks
```

iamnafets:
Closes #4477 

## 🎯 Changes

This change (1) introduces a `queryGenerator` and `mutationGenerator` method to the client, and (2) adds a streamingData state to useQuery and useMutation. The developer experience is that you can now implement your query or mutation function with an AsyncGenerator (e.g. `async function*`) and retrieve every emitted event on the client side.

Caveats:
* This is currently a proof of concept. It's working (deploying to production soon), but probably has lots of rough edges.
* I have not yet implemented code for the wslink etc as I'd like to get some agreement on the API first.
* I've abused `next-minimal-starter`, I'll fix that (among other things) before requesting a proper review.

Notes:
* I like the idea of having an async generator implementation of `query` and `mutation`. It forces the developer to think properly about the boundaries of the request as opposed to callbacks or observables which seem like they can be persistent but actually are not.
* I'm totally abusing @Sheraff 's streaming HTTP link here by making the observables properly observable and having multiple returns for a given call.
* When streaming is not supported by the link, it seems sensible to just fall back to the last element passed over the wire. This enables a graceful fallback for cases where you're using streaming to progressively improve the UI (e.g. OpenAI APIs) but the fidelity of the final message is most important.
* I experimented with a version of `useQuery` that simply overrode the data through `setQueryData` each time it had a new chunk. This blew up the `loading` semantics and felt rather hacky so the `streamingData` state is a bit cleaner.

## ✅ Checklist


Sheraff:
- Could you give a description of the resulting API? I feel a little overwhelmed to have to read a 5k LOC changes PR to understand it. (like what is the shape of the request, how is it streamed back, what happens on the client to make the data useable by the user?)
- Could you try running `pnpm format:write` so that changes that are just formatting issues don't appear in the PR? That'll trim down what I need to focus on
- On the client side, during my streaming PR we debated on the use of async generators / iterators and it was concluded that they're fine on the server, but should be avoided on the client for compatibility reasons
- this PR adds a lot of client-side code, and there is a goal to try and keep it small where possible. Maybe you should consider a separate link so that loading all of this code is optional?

sachinraja:
I really wish you hadn't done this much work already because I don't think a generator is the right API for streaming. Something like [what SvelteKit does](https://svelte.dev/blog/streaming-snapshots-sveltekit) would probably be better.

iamnafets:
> I really wish you hadn't done this much work already because I don't think a generator is the right API for streaming. This also complicates the code a lot right now and in the future (adds 2 new procedure types!). Something like [what SvelteKit does](https://svelte.dev/blog/streaming-snapshots-sveltekit) would probably be better.

I'm not too worried about it. This is mostly a draft. If you have a better idea on what the API should be, feel free to comment on the attached feature issue.

KATT:
Superseded #4911

PR #4496, opened 2023-06-09
fix: dev script terminating with rollup3 & turbo

juliusmarminge:
Ref:
- https://github.com/vercel/turbo/issues/4983 
- https://github.com/rollup/rollup/pull/5027 (this is the patch i made)
- https://github.com/rollup/rollup/pull/5004

Before:

https://github.com/trpc/trpc/assets/51714798/a3685bf5-b3e4-47cb-8699-e901471123f5

After

Also seems like Rollup has changed it's behavior in terms of waiting for deps to finish, cause now the client shows errors for a while until the server dev is finished, but then works as expected. Seems like a bad change cause that was probably the only thing that it did better than tsup before 🤷🏼‍♂️ 

Oh well...

https://github.com/trpc/trpc/assets/51714798/96a8a24c-3653-4237-b6c4-434817d6d9e9


juliusmarminge:
This is hacky - reverting to Rollup 2 instead

PR #4497, opened 2023-06-09
chore(deps): revert to rollup 2

juliusmarminge:
Rollup 3 seems to be broken...

See #4496


kamilogorek:
Tested locally, fixes the issue

PR #4514, opened 2023-06-12
Load env configs file using Next.js loadEnvConfig function

Elvincth:
## 🎯 Changes

Load env configs file using Next.js loadEnvConfig function
so what we can auto load the env configs from e.g. .env.local, .env.development.local etc.

## ✅ Checklist


juliusmarminge:
Why is this needed?

Elvincth:
> Why is this needed?

When moved to custom server it lost its ability to auto load next.js custom env files. With this it can fix this issue.

juliusmarminge:
I'm not sure about this. We are already using environment variables in this example and they're working fine.

Also, if they should be explicitely loaded anywhere it should be in the custom server, not a trpc file

PR #4516, opened 2023-06-12
feat(`client`): unable to unwrap a proxy client to an untyped client

KATT:
Needed for #4482 to be nice.

Allows you to turn a proxy client into an untyped client.

Would love review from @sachinraja here :) especially if it doesn't pass all tests lol

### DIFF
--- a/packages/client/src/createTRPCClientProxy.ts
+++ b/packages/client/src/createTRPCClientProxy.ts
@@ -113,6 +113,9 @@ export function createTRPCClientProxy<TRouter extends AnyRouter>(
     if (client.hasOwnProperty(key)) {
       return (client as any)[key as any];
     }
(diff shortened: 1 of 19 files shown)


packages/client/src/createTRPCClientProxy.ts
```diff
@@ -66,9 +68,12 @@ type DecorateProcedure<TProcedure extends AnyProcedure> =
 /**
  * @internal
  */
-type DecoratedProcedureRecord<TProcedures extends ProcedureRouterRecord> = {
+type DecoratedProcedureRecord<
+  TRouter extends AnyRouter,
```
KATT, line 72: We seem to need to be able to keep track of the original router within the generics in order to unwrap it later

sachinraja (approved):

PR #4528, opened 2023-06-14
fix(server): allow overriding root context in nested middleware

### LINKED ISSUE #4527: bug: cannot override root context with nested middleware
### Describe the bug

When nesting middlewares, if one middleware changes the root context and one middleware is nested after that, the first middleware's changes are overwritten by the root context. This doesn't make sense because this isn't the order that tRPC executes the middlewares in. The root context is executed first and then middlewares can alter/extend that.

### Link to reproduction

https://stackblitz.com/edit/github-rh1vjf?file=src%2Fpages%2Fapi%2Ftrpc%2F[trpc].ts

### To reproduce

Look at the difference in context type between the different procedures on StackBlitz.

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

sachinraja:
Closes #4527
Closes T-108

## 🎯 Changes

Merge the root context into the actual output of the middlewares only when necessary (function inputs like opts.ctx, used in t.middleware, t.use, t.procedure.query, t.procedure.mutation, t.procedure.subscription). This allows the proper order of overrides from root context -> middlewares instead of each middleware including the root context in it's `_ctx_out` type and preventing overriding the root context with nested middleware.

### DIFF
--- a/packages/tests/server/regression/issue-4527-nested-middleware-root-context.test.ts
+++ b/packages/tests/server/regression/issue-4527-nested-middleware-root-context.test.ts
@@ -36,7 +36,7 @@ test('root context override on nested middlewares', () => {
     )
     .query(async (opts) => {
       expectTypeOf<{
-        req: Request | undefined;
+        req?: Request;
         apiKey: string;
       }>(opts.ctx);
 
@@ -52,7 +52,7 @@ test('root context override on nested middlewares', () => {
     )
     .query(async (opts) => {
       expectTypeOf<{
-        req: Request | undefined;
+        req?: Request;
         apiKey: string;
       }>(opts.ctx);
 
--- a/packages/tests/server/regression/issue-4527-nested-middleware-root-context.test.ts
+++ b/packages/tests/server/regression/issue-4527-nested-middleware-root-context.test.ts
@@ -1,0 +1,55 @@
+import { initTRPC, TRPCError } from '@trpc/server/src';
+import { z } from 'zod';
+
+test('root context override on nested middlewares', () => {
+  const t = initTRPC
+    .context<{
+      apiKey?: string | null;
+      req?: Request;
+    }>()
+    .create();
+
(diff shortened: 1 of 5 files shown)


packages/tests/server/regression/issue-4527-nested-middleware-root-context.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  protectedApiProcedure
+    .use(formDataMiddleware)
+    .input(
+      z.object({
+        somehash: z.string(),
+      }),
+    )
+    .query(async (opts) => {
+      expectTypeOf<string>(opts.ctx.apiKey);
+
```
juliusmarminge, line 39:
```suggestion
      expectTypeOf<Request | undefined>(opts.ctx.req);

```

for good measure 🤷🏼

packages/tests/server/regression/issue-4527-nested-middleware-root-context.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  // root context -> formDataMiddleware -> enforceApiKey
+  protectedApiFormDataProcedure
+    .input(
+      z.object({
+        somehash: z.string(),
+      }),
+    )
+    .query(async (opts) => {
+      expectTypeOf<string>(opts.ctx.apiKey);
+
```
juliusmarminge, line 52:
```suggestion
      expectTypeOf<Request | undefined>(opts.ctx.req);

```

for good measure 🤷🏼

juliusmarminge (approved):

packages/tests/server/regression/issue-4527-nested-middleware-root-context.test.ts
sachinraja: good call :+1:

bautistaaa (approved):

PR #4530, opened 2023-06-15
feat: streamable procedure support

### LINKED ISSUE #4477: feat: Streaming Mutations / Queries

helmturner:
revert(server): roll back node adapter to 0aaf147;
revert(server): roll back fetch adapter to 0aaf147;
revert(server): roll back fastify adapter to ababea8;

Closes #4477

## 🎯 Changes

- Reverts adapters to pre-httpBatchLink versions and extracts response-body creation logic into separate modules
- Enables ReadableStream bodies to be created by and returned from `resolveHTTPResponse`
- Sets the stage for eagerly streaming queries 

## ✅ Checklist


helmturner:
On further, further investigation... I'm not actually reading out of order, at least in most cases. Everything works pretty well, actually. There are plenty of successful tests for both batch and single procedure calls. Save for some edge cases and some adapter issues, looks like this is working? Merging main brought it to 50 failing tests, but I made some small changes and cut that to 36 failing tests - a personal best, so far!

packages/server/src/http/buildResponse.ts
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        const chunks = getRemainingChunks(envelope.result.data);
+        //XXX
+        if (style !== 'event-stream' && style !== 'json-stream') {
+          // We're not streaming, so we need to resolve the generator and return the complete result
+          let done;
+          let value;
+
+          while (!done) ({ done, value } = await chunks.next());
```
KATT, line 197: I think we should just throw here instead and have the rule if you request an async generator, you have to use a transport/http link that supports it

helmturner: It seems intuitive that returning a generator from a tRPC resolver wouldn't work without additional configuration. I guess my thought was preserving compatibility in the case you have multiple clients in heterogenous runtimes with varying degrees of streaming support, e.g. mobile and web.

My toy use-case for the SSE side of things was the GPT4 API, but blocking on a mobile device until the response completes is a pretty crap experience in that case, so it's probably an unrealistic edge-case unless there's some other scenario where blocking for an otherwise-streamed response is acceptable UX.

KATT:
#4911

PR #4542, opened 2023-06-19
Select-transform-2

Sheraff:
Closes #

## 🎯 Changes

## ✅ Checklist

### DIFF
--- a/packages/react-query/src/createTRPCReact.tsx
+++ b/packages/react-query/src/createTRPCReact.tsx
@@ -109,7 +109,7 @@ export type DecorateProcedure<
        */
       useSuspenseQuery: <
         TQueryFnData extends inferTransformedProcedureOutput<TProcedure> = inferTransformedProcedureOutput<TProcedure>,
-        TData extends inferTransformedProcedureOutput<TProcedure> = inferTransformedProcedureOutput<TProcedure>,
+        TData = TQueryFnData,
       >(
         input: inferProcedureInput<TProcedure>,
         opts?: Omit<
--- a/packages/react-query/src/createTRPCReact.tsx
+++ b/packages/react-query/src/createTRPCReact.tsx
@@ -45,6 +45,39 @@ import {
 } from './shared/hooks/types';
 import { CreateTRPCReactOptions } from './shared/types';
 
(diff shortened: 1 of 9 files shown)


packages/react-query/src/shared/hooks/types.ts
```diff
@@ -47,9 +48,11 @@ export interface TRPCUseQueryBaseOptions {
   trpc?: TRPCReactRequestOptions;
 }
 
-export interface UseTRPCQueryOptions<TPath, TInput, TOutput, TData, TError>
-  extends UseQueryOptions<TOutput, TError, TData, [TPath, TInput]>,
-    TRPCUseQueryBaseOptions {}
+export interface UseTRPCQueryOptions<TPath, TInput, TOutput, TData, TError, TQueryData>
+  extends Omit<UseBaseQueryOptions<TOutput, TError, TData, TQueryData, [TPath, TInput]>, 'placeholderData'>,
+  TRPCUseQueryBaseOptions {
+    placeholderData?: TOutput | PlaceholderDataFunction<TOutput>;
```
Sheraff, line 54: I tried `TOutput`, `TData` and `TQueryData` here, seems to yield the same behavior (or not work at all, can't remember which)

packages/react-query/src/createTRPCReact.tsx
```diff
@@ -54,31 +87,41 @@ export interface ProcedureUseQuery<
 > {
   <
     TQueryFnData extends inferTransformedProcedureOutput<TProcedure> = inferTransformedProcedureOutput<TProcedure>,
-    TData extends inferTransformedProcedureOutput<TProcedure> = inferTransformedProcedureOutput<TProcedure>,
+    TData = TQueryFnData,
+    TQueryData extends inferTransformedProcedureOutput<TProcedure> = TQueryFnData,
```
Sheraff, line 91: This exposes a 3rd generic that people probably shouldn't use. We might want to hide it (which I think is what RQ does with `UseQueryOptions` wrapping `UseBaseQueryOptions`). Or we can just name it `__BadIdeaDontUseThisType__`

juliusmarminge:
Supserseded by #4543

PR #4547, opened 2023-06-22
feat(`react-query`): allow for external apps in `createServerSideHelpers`

KATT:
Supersedes #4482

Docs: https://www-git-allow-external-apps-in-createserverside-trpc.vercel.app/docs/client/nextjs/server-side-helpers

> ## 🎯 Changes
> Currently the `createServerSideHelpers` function has a required `router` property that expects the TRPC router to be provided. While this works in a NextJS project (Where the back-end is implemented in NextJS as well), this will not work for projects with a separate front-end and back-end.
> 
> This PR provides a solution by giving the implementor the option to provide an instance of the untyped client instead of the router.

### DIFF
--- a/packages/react-query/src/server/index.ts
+++ b/packages/react-query/src/server/index.ts
@@ -1,7 +1,7 @@
 export { createServerSideHelpers } from './ssgProxy';
 export type {
   /**
-   * @deprecated this exported type is will be removed in the next major version
+   * @deprecated this exported is planned to be removed in the next major version
    */
   CreateServerSideHelpersOptions as CreateSSGHelpersOptions,
 } from './types';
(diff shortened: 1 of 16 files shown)


packages/react-query/src/server/index.ts
```diff
@@ -1,4 +1,9 @@
 export { createServerSideHelpers } from './ssgProxy';
-export type { CreateSSGHelpersOptions } from './types';
+export type {
+  /**
+   * @deprecated this exported type is will be removed in the next major version
```
KATT, line 4:
```suggestion
   * @deprecated this exported is planned to be removed in the next major version
```

PR #4557, opened 2023-06-25
Add 'override' to 'cause' in TRPCError

anri-asaturov:
Closes #

## 🎯 Changes

I was getting the following error with "@trpc/server@npm:10.32.0" and TypeScript 5

```
../../node_modules/@trpc/server/src/error/TRPCError.ts:23:19 - error TS4114: This member must have an 'override' modifier because it overrides a member in the base class 'Error'.

23   public readonly cause?: Error;
                     ~~~~~
```

so I fixed it by adding the `override` modifier, not sure why hasn't anyone complained yet.

## ✅ Checklist

### DIFF
--- a/packages/server/src/error/TRPCError.ts
+++ b/packages/server/src/error/TRPCError.ts
@@ -20,7 +20,7 @@ export function getTRPCErrorFromUnknown(cause: unknown): TRPCError {
 }
 
 export class TRPCError extends Error {
-  public readonly cause?: Error;
+  public override readonly cause?: Error;
   public readonly code;
 
   constructor(opts: {


juliusmarminge:
> @trpc/server@npm:10.32.0

Is this using Deno or? Can you provide a reproduction example first cause I've never seen this error before and noone has reported it.

kamilogorek:
This will happen if you set `noImplicitOverride: true` in your `tsconfig`.

packages/server/src/error/TRPCError.ts
```diff
@@ -20,7 +20,7 @@ export function getTRPCErrorFromUnknown(cause: unknown): TRPCError {
 }
 
 export class TRPCError extends Error {
-  public readonly cause?: Error;
+  public override readonly cause?: Error;
```
KATT, line 23: What happens in older TS versions when the `override`-keyword is encountered?

anri-asaturov: I can't reproduce it outside of the legacy project I'm working on, it's probably some kind of a weird combination of dependencies, can't update them all (especially transitive ones) in this repo. I'm going to go ahead and cancel this PR, sorry for disturbance.

anri-asaturov:
> This will happen if you set `noImplicitOverride: true` in your `tsconfig`.

Oh, you were right, thank you, I didn't notice it was on.

Issue #4589, opened 2023-07-04
bug: fetch binding doesn't support datadog/browser-sdk


skovhus:
### Provide environment information

```
System:
    OS: macOS 13.4
    CPU: (8) arm64 Apple M1 Pro
    Memory: 113.39 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.3.1 - ~/.nvm/versions/node/v20.3.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.19.0/bin/yarn
    npm: 9.6.7 - ~/.nvm/versions/node/v20.3.1/bin/npm
    pnpm: 8.6.2 - /opt/homebrew/bin/pnpm
    Watchman: 2023.06.12.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 114.0.5735.198
    Safari: 16.5
  npmPackages:
    @tanstack/react-query: 4.24.10 => 4.24.10 
    @trpc/client: 10.20.0 => 10.20.0 
    @trpc/react-query: 10.20.0 => 10.20.0 
    @trpc/server: 10.20.0 => 10.20.0 
    react: 17.0.2 => 17.0.2 
    typescript: 5.0.2 => 5.0.2 
```

### Describe the bug

The way we do [bind fetch to window](https://github.com/trpc/trpc/blob/main/packages/client/src/getFetch.ts) in order to be able to pass it around in an object isn't compatible with DataDog RUM monitoring (and potentially a lot of other monitoring solutions). 

I'm unsure if this is something that can be fixed by the DataDog RUM library (see instrumentation code [here](https://github.com/DataDog/browser-sdk/blob/main/packages/core/src/browser/fetchObservable.ts) and [here](https://github.com/DataDog/browser-sdk/blob/main/packages/core/src/tools/instrumentMethod.ts)), but it seems like we fairly easily could avoid the binding.

### Link to reproduction

n/a

### To reproduce

It requires a lot of setup if we want to test DataDog end to end. But we could extract the instrumentation part.

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

PR #4590, opened 2023-07-04
fix(client): enable instrumenting fetch

skovhus:
Closes https://github.com/trpc/trpc/issues/4589

## 🎯 Changes

This is a suggested fix to enable third party (like DataDog) to instrument fetch. We do this by avoiding to bind window or globalThis to fetch.

If we go this direction then why not just get rid of `resolveHTTPLinkOptions`?

## ✅ Checklist


skovhus:
@KATT thanks. Any thoughts on my comment?

> If we go this direction then why not just get rid of resolveHTTPLinkOptions?

KATT:
Always a good idea to get rid of stuff :)

PR #4595, opened 2023-07-05
Update Docs: Move context within trpc

francistogram:
## 🎯 Changes

Just an update to the [example docs](https://trpc.io/docs/client/links/splitLink#2-perform-request-without-batching)

The second example is right with `context` within the `trpc` object and the first example doesn't work

image

## ✅ Checklist

### DIFF
--- a/www/docs/client/links/splitLink.mdx
+++ b/www/docs/client/links/splitLink.mdx
@@ -58,8 +58,10 @@ const client = createTRPCProxyClient<AppRouter>({
 
 ```ts title='client.ts'
 const postResult = proxy.posts.query(null, {
-  context: {
-    skipBatch: true,
+  trpc: {
+    context: {
+      skipBatch: true,
+    },
   },
 });
 ```


kamilogorek:
This works exactly as expected for me.

```js
import {
  createTRPCProxyClient,
  httpBatchLink,
  httpLink,
  splitLink,
} from '@trpc/client';
import type { AppRouter } from './router';

async function main() {
  const url = 'http://127.0.0.1:8000/trpc';

  const proxy = createTRPCProxyClient<AppRouter>({
    links: [
      splitLink({
        condition(op) {
          console.log(op.context); // {mkey: "wat"}
          return op.context.skipBatch === true;
        },
        true: httpLink({
          url,
        }),
        false: httpBatchLink({
          url,
        }),
      }),
    ],
  });

  proxy.hello.query(null, {
    context: { mkey: 'wat' },
  });
}

main().catch((err) => console.error(err));
```

Can you provide a repro using https://stackblitz.com/?

williameliasson-dev (approved):

PR #4618, opened 2023-07-12
Multivalue header and cookie support aws lambda

### LINKED ISSUE #4546: feat: Add support for multiValueHeaders and cookies in aws-lambda adapter

For the aws-lambda adapter, multiple cookies cannot be returned currently when using a Rest API - using a comma-separated `Set-Cookie` header returns the first cookie in the header, instead you should use either the `multiValueHeaders` or `cookies` property if using a V2-based API gateway (HTTP-API or WS API).

To add support for this, we will need to:

1. Replace the existing comma-separated array joining to instead return two objects: `headers` and `multiValueHeaders` if we are using a v1 response format (for v2 we can just return headers).
1. To support passing `cookies` back when using the V2 response spec, we will need to augment the `HTTPResponse` type by adding an optional `cookies?: string[]` type.

1. If using the v1 lambda response spec:
    1. The multi-value headers can then be passed into the existing response [here](https://github.com/trpc/trpc/blob/main/packages/server/src/adapters/aws-lambda/index.ts#L59) as another field.

1. If using the v2 lambda response spec:
    1. V2-based APIs support comma-separated headers, so the current solution will work
    1. We should also pass `cookies` which can be returned via APIGateway [within this response](https://github.com/trpc/trpc/blob/main/packages/server/src/adapters/aws-lambda/index.ts#L66)

I have attempted to return a `Set-Cookie` header as an comma-separated string and an array however only one cookie is currently sent back.

### Additional information

The specification for the lambda response format can be found [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html).

### 👨‍👧‍👦 Contributing

## Funding

<a href="https://polar.sh/trpc/trpc/issues/4546">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/4546/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

thomasedw:
Closes #4546

## 🎯 Changes

This PR adds support for multiValueHeaders and cookies to be returned from the AWS lambda adapter. 

To do this I have created a new type `AwsLambdaHTTPResponse` which extends from `HTTPResponse` and added a new generic into the type signature for `resolveHTTPResponse` so you can inject any other required attributes to return. This allows other adapters to insert their own values if required in the future, while maintaining the adapter/framework agnostic design as much as possible.

I have updated the documentation for the lambda adapter to show an example of settings cookies / multiValueHeaders. This is my first PR in this project, happy for any feedback or suggested changes.

## ✅ Checklist


Sheraff:
That looks relatively clean but I'm not sure I fully understand the need for all these changes.
- is the main goal to be able to "accumulate" multi-value headers over many procedures? like specifically `Set-Cookie` can be a list and we don't want only the last value to be taken into account. Is that it?
- are `multiValueHeaders` and `cookies` aws-specific terms?
- shouldn't we just have a better accumulator in `initResponse` inside `resolveHTTPResponse`?
  ```ts
  for (const [key, value] of Object.entries(meta.headers ?? {})) {
    headers[key] = value;
  }
  ```

Sorry if these are naive questions, you've probably thought about the problem more than me and I'm trying to understand why this solution was chosen.

thomasedw:
@Sheraff not naive questions, will try my best to answer:

In the case where you want to send back multiple `Set-Cookie` headers in a lambda function, depending on which type of API Gateway you're using you'll either need to use `MultiValueHeaders` (for Rest APIs - v1 format) or `cookies` (for HTTP/WS APIs - v2 format), otherwise the values will not be sent back correctly from the lambda to API gateway. 

The specific payload format can be found [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html), afaik these are AWS specific terms, but I attempted to make the `ResponseMeta` approach generic in the event that other adapters can also expect custom properties to be returned.

The better accumulator approach was what I was planning to implement initially and then return these directly, however that would have meant these properties could be returned for adapters that don't expect them, because this only affects the AWS lambda adapter, I wanted to avoid changing the outputs of other adapters if possible.

packages/server/src/http/resolveHTTPResponse.ts
```diff
@@ -351,11 +361,13 @@ export async function resolveHTTPResponse<
...
+      const { status, headers, ...otherMetadata } = headResponse;
 
       return {
-        status: headResponse.status,
-        headers: headResponse.headers,
+        status,
+        headers,
         body,
+        ...otherMetadata,
       };
```
Sheraff, line 371: Why do we need to destructure `status` and `headers` from `headResponse` only to re-spread them together in the same object?

Sheraff:
A few misc comments: 
- does this work with batching if several procedures set the `set-cookies` header? Headers support in general is a gray area in trpc (because of batching), but if they're explicitly mentioned in the docs, we should also know the limits
- for now AWS is the only adapter that doesn't support streaming (because I don't know AWS, so I had to leave this adapter out), however the types are changing for all adapters with this PR and changes weren't propagated to `unstable_onHead` (method called to provide response headers in adapters that support streaming). Is there a specific reason why this wouldn't be appropriate and we want to leave it out or should we also change the type of `unstable_onHead`?

packages/server/src/http/resolveHTTPResponse.ts
thomasedw: I did this mostly for readability, but can spread the full `headResponse` if that'd be preferable.

thomasedw:
I checked with batching and it doesn't merge headers across requests - instead it returns the last returned header from the batched requests. For steaming these types should be available too, I'll check out those types you pointed out and change them, thanks!

KATT (changes requested):
Failing CI, please re-request review again when it doesn't

KATT:
Feel free to re-open later. Doing cleanup & don't want PRs gathering dust

PR #4625, opened 2023-07-16
feat(server): better way to export the fetch request handler

Bekacru:
Closes #

## 🎯 Changes

This PR proposes a more concise syntax for the `fetchRequestHandler` export in `fetchRequestHandler.ts`. I made the change in a new function called routeFetchRequestHandler. However, it's worth considering adding this new function to the main fetchRequestHandler function (if this gonna be merged) in some way to avoid breaking current apps that use it. With this version, you no longer need to export GET and POST with as, and you don't need to pass the request object to the handler. Instead, the routeFetchRequestHandler function returns an object with GET and POST properties, each of which is an async function that calls fetchRequestHandler with the appropriate request method.

exmaple usage:
```ts
const {GET, POST} = routeFetchRequestHandler({
  endpoint: "/api/trpc",
  router: appRouter,
  createContext: () => createTRPCContext(),
});


```

## ✅ Checklist

### DIFF
--- a/packages/server/src/adapters/fetch/fetchRequestHandler.ts
+++ b/packages/server/src/adapters/fetch/fetchRequestHandler.ts
@@ -114,8 +114,8 @@ export async function fetchRequestHandler<TRouter extends AnyRouter>(
   return promise;
 }
 
-export async function routeFetchRequestHandler<TRouter extends AnyRouter>(
-  opts: FetchHandlerRequestOptions<TRouter>,
+export function routeFetchRequestHandler<TRouter extends AnyRouter>(
+  opts: Omit<FetchHandlerRequestOptions<TRouter>, 'req'>,
 ) {
   return {
     GET: async (req: Request) => await fetchRequestHandler({ ...opts, req }),
--- a/packages/server/src/adapters/fetch/fetchRequestHandler.ts
+++ b/packages/server/src/adapters/fetch/fetchRequestHandler.ts
@@ -113,3 +113,12 @@ export async function fetchRequestHandler<TRouter extends AnyRouter>(
 
   return promise;
 }
+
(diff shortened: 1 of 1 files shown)


juliusmarminge:
Could potentially make an overload that checks if the `fetchRequestHandler` has gotten a `req` param passed to it and infer which method to use from that.

However, I'm unsure if this optimization is worth it just to reduce your export by 1 line for specific frameworks. If I'm not mistaken, some framework even uses lower case `get` and `post`, which would make this unusable to them and just cause confusion which to use?

Bekacru:
> Could potentially make an overload that checks if the `fetchRequestHandler` has gotten a `req` param passed to it and infer which method to use from that.
> 
> However, I'm unsure if this "optimization" is worth it just to reduce your export by 1 line for specific frameworks. If I'm not mistaken, some framework even uses lower case `get` and `post`, which would make this unusable to them and just cause confusion which to use?

yeah we could infer the type and overload. Is the fetchRequestHandler used elsewhere other than next? and if it's I'd assume the majority will be using next so it might make sense to have a nice differnt export like the one I did maybe with better name. tho it might not worth the confusion as u said.

juliusmarminge:
> > Could potentially make an overload that checks if the `fetchRequestHandler` has gotten a `req` param passed to it and infer which method to use from that.
> 
> > 
> 
> > However, I'm unsure if this "optimization" is worth it just to reduce your export by 1 line for specific frameworks. If I'm not mistaken, some framework even uses lower case `get` and `post`, which would make this unusable to them and just cause confusion which to use?
> 
> 
> 
> yeah we could infer the type and overload. Is the fetchRequestHandler used elsewhere other than next? 

Yes, see here for all the different example use cases we have for the fetch adapter: https://trpc.io/docs/server/adapters/fetch

PR #4648, opened 2023-07-23
fix: use ts option noPropertyAccessFromIndexSignature

maritz:
## 🎯 Changes

Enabling typescript compile option `noPropertyAccessFromIndexSignature` for server package.

Feel free to just drop this PR if you want, since it's obviously a stylistic "issue".

### Why though?

I use nx in my workspaces and some of the default generators set this option by default, so I've been using it for a bit now.
@trpc/server so far is the only dependency that has caused my builds to fail because it does not use this option and there is no good option in Typescript to say "use this strict stylistic option only for my code" as far as I know.

I've fixed the problem for myself using `patch-package` and thought to myself I could just offer the patch here as well.

## ✅ Checklist


packages/server/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "include": ["src"],
   "compilerOptions": {
     "lib": ["ES2020", "DOM", "DOM.Iterable"],
-    "target": "ES2020"
+    "target": "ES2020",
+    "noPropertyAccessFromIndexSignature": true
```
juliusmarminge, line 7: if we enable it it should be enabled for all packages, imo

juliusmarminge (changes requested):
i don't really care whether it's on or not. generally, I recommend turning on `skipLibCheck` in your own packages to prevent typescript from checking external libs.

if we enable it though, it should be enabled in all packages, and not just the server one.

KATT (changes requested):
Sorry, but I hate this rule & don't wanna read code this way 😅 

you should probably not typecheck dependencies


in your tsconfig:

- `"skipLibCheck": true,`
- `"exclude": ["node_modules"]`

PR #4713, opened 2023-08-12
feat(`server`): add standalone middleware (experimental)

jussisaurio:
## 🎯 Changes

Proposal/feat: `standaloneMiddleware`

```
  const projectAccessMiddleware = standaloneMiddleware<{
    ctx: { allowedProjects: string[] }; // defaults to 'object' if not defined
    input: { projectId: string }; // defaults to 'unknown' if not defined
  }>().create((opts) => {
    if (!opts.ctx.allowedProjects.includes(opts.input.projectId)) {
      throw new TRPCError({
        code: 'FORBIDDEN',
        message: 'Not allowed',
      });
    }

    return opts.next();
  });
```
    
Enables creation of middlewares that are not bound to any particular
tRPC instance; instead they can define a type of `Context`/`Input` that they
expect, and as long as the preceding middleware chain provides `Context` and `Input`
types that are assignable to the middleware's required context type,
the middleware can be used in the chain. The required context defaults to `object` and the required input defaults to `unknown`.
    
This commit provides a few tests that demonstrate the usage of
standalone middlewares in two different tRPC instances.

## ✅ Checklist

### DIFF
--- a/packages/server/src/core/index.ts
+++ b/packages/server/src/core/index.ts
@@ -17,7 +17,11 @@ export type {
   ProcedureOptions,
 } from './procedure';
 export type { inferParser } from './parser';
-export { createInputMiddleware, createOutputMiddleware } from './middleware';
+export {
+  createInputMiddleware,
+  createOutputMiddleware,
+  standaloneMiddleware,
+} from './middleware';
 export type { MiddlewareFunction, MiddlewareBuilder } from './middleware';
 export { initTRPC } from './initTRPC';
 export * from './types';
--- a/packages/server/src/core/index.ts
+++ b/packages/server/src/core/index.ts
@@ -20,7 +20,7 @@ export type { inferParser } from './parser';
 export {
   createInputMiddleware,
   createOutputMiddleware,
-  standaloneMiddleware,
+  experimental_standaloneMiddleware,
 } from './middleware';
 export type { MiddlewareFunction, MiddlewareBuilder } from './middleware';
 export { initTRPC } from './initTRPC';
--- a/packages/server/src/core/middleware.ts
+++ b/packages/server/src/core/middleware.ts
@@ -11,6 +11,8 @@ import {
 import { ProcedureParams } from './procedure';
 import { ProcedureType } from './types';
 
+
+
 /**
  * @internal
  */
@@ -200,13 +202,14 @@ export function createMiddlewareFactory<
 export const standaloneMiddleware = <
   TCtx extends {
     ctx?: object;
+    meta?: object;
     input?: unknown;
   },
 >() => ({
   create: createMiddlewareFactory<
     RootConfig<{
-      ctx: TCtx extends { ctx: infer T } ? T : object;
-      meta: object;
+      ctx: TCtx extends { ctx: infer T extends object } ? T : object;
+      meta: TCtx extends { meta: infer T extends object } ? T : object;
       errorShape: object;
       transformer: object;
     }>,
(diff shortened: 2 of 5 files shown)


jussisaurio:
It might be more composable for the "restricted middleware" to require just `T` instead of `Context & T` so that earlier middlewares can override parts of `Context` if the "restricted middleware" does not need it

jussisaurio:
Completely rewrote this PR in favor of the `standaloneMiddleware` approach outlined above. Rewrote initial PR message.

packages/server/src/core/middleware.ts
```diff
@@ -180,14 +186,34 @@ export function createMiddlewareFactory<TConfig extends AnyRootConfig>() {
...
+  create: createMiddlewareFactory<
+    RootConfig<{
+      ctx: TCtx extends { ctx: infer T } ? T : object;
+      meta: object;
+      errorShape: object;
+      transformer: object;
+    }>,
+    TCtx extends { input: infer T } ? T : unknown
+  >(),
+});
```
juliusmarminge, line 215: what's the rationale for this being a builder? why can it not just be `standaloneMiddleware((opts) => { ... })`

jussisaurio: Hey, thanks for the question. Your guess is right: tRPC typings are quite far in the weeds and since this was my first tour around the codebase, using a builder here was the approach that helped get the types working without a lot of type noise. I tried the direct callback approach with types that I assumed were equivalent, but ran into errors that I would need some time to figure out.

I'm still not completely happy with the TSC error messages (i.e. in cases where an error is expected) even for the builder version, but at least I got it to do what I intended it to do.

juliusmarminge: Yes i think @Nick-Lucas tried this before and eventually gave up, but this is definetely something we want to have

jussisaurio: What's the way forward with this feature? I'll be honest and say I currently don't have a very good idea of why wrapping this in `.create()` results in valid types but using a direct callback doesn't.

I dug up [this old discussion](https://github.com/trpc/trpc/issues/4191) which you were probably referring and definitely agree that this would up the flexibility as far as 3rd party libs etc are concerned.

However I also kind of don't want to integrate a feature where it's unclear why it works one way but doesn't work the other way. I'll probably have to spend more time understanding exactly how all the ProcedureBuilder/MiddlewareFunction/MiddlewareBuilder/ProcedureParams etc. interact together, but right now it's really hard to hold all of it in my head.

jussisaurio: I think the issue is that with the builder version, `_ctx_out` can be inferred from the provided callback implementation, but in the non-builder version the callback's type has to be annotated, and I can't see a way of doing it so that `_ctx_out` gets a valid type that is inferred from the implementation.

packages/server/src/core/index.ts
```diff
@@ -17,7 +17,11 @@ export type {
   ProcedureOptions,
 } from './procedure';
 export type { inferParser } from './parser';
-export { createInputMiddleware, createOutputMiddleware } from './middleware';
+export {
+  createInputMiddleware,
+  createOutputMiddleware,
+  standaloneMiddleware,
```
KATT, line 23: - Prefix with `experimental_` 
- Add a doc section under https://trpc.io/docs/server/middlewares

KATT (commented):
This is great! I'm not confident with the exact API but I definitely agree with the _intention_ - we should have a way of defining standalone middleware based on a fuzzy-matched context/input/etc

Could you prefix it with `experimental_` and ad docs? I'm happy to merge it today

bautistaaa:
wow this is awesome! and not too much noise with the types

KATT (approved):

chinanderm:
This is a fantastic feature.

PR #4715, opened 2023-08-13
server: add support for next valibot version

### LINKED ISSUE #4737: bug: Using valibot validation creates type errors

  System:
    OS: macOS 13.0.1
    CPU: (8) arm64 Apple M1
    Memory: 59.02 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 20.5.0 - ~/.volta/tools/image/node/20.5.0/bin/node
    Yarn: 4.0.0-rc.43 - ~/.volta/tools/image/yarn/4.0.0-rc.43/bin/yarn
    npm: 9.8.0 - ~/.volta/tools/image/node/20.5.0/bin/npm
    pnpm: 8.6.10 - ~/.volta/bin/pnpm
  Browsers:
    Chrome: 113.0.5672.126
    Edge: 114.0.1823.79
    Firefox Nightly: 89.0a1
    Safari: 16.1
  npmPackages:
    @trpc/client: ^10.37.1 => 10.37.1 
    @trpc/server: ^10.37.1 => 10.37.1

### Describe the bug

Using a simple valibot validator for validation outputs a type error

### Link to reproduction

https://stackblitz.com/edit/js-yma3yp?file=index.ts

### To reproduce

1) Install `@trpc/server@10.37.1` and `valibot@0.13.0`
2) Create a procedure using a valibot input validator
```ts
t.procedure.input(string()).query(({ input }) => `Hello, ${input}!`);
```
image

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

fabian-hiller:
Closes #4737

## 🎯 Changes

In the next version of Valibot, the internal API will change. Therefore I have extended the implementation.

I had to add an import statement. Since `valibot` is currently only a dev dependency, I am not sure if this is problematic.

## ✅ Checklist

### DIFF
--- a/packages/server/src/core/internals/getParseFn.ts
+++ b/packages/server/src/core/internals/getParseFn.ts
@@ -21,14 +21,6 @@ export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
     return parser.parse.bind(parser);
   }
 
-  if (typeof parser._parse === 'function') {
-    // ParserValibotEsque (>= v0.13.X)
-    return async (value) => {
-      const { parseAsync } = await import('valibot');
-      return parseAsync(parser, value);
-    };
-  }
-
   if (typeof parser.validateSync === 'function') {
     // ParserYupEsque
     return parser.validateSync.bind(parser);
@@ -44,6 +36,14 @@ export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
     return (value) => {
       parser.assert(value);
       return value as TType;
+    };
+  }
+
+  if (typeof parser._parse === 'function') {
+    // ParserValibotEsque (>= v0.13.X)
+    return async (value) => {
+      const { parseAsync } = await import('valibot');
+      return parseAsync(parser, value);
     };
   }
 
--- a/packages/server/src/core/internals/getParseFn.ts
+++ b/packages/server/src/core/internals/getParseFn.ts
@@ -1,3 +1,4 @@
+import { parseAsync } from 'valibot';
 import { Parser } from '../parser';
 
 export type ParseFn<TType> = (value: unknown) => Promise<TType> | TType;
@@ -17,8 +18,13 @@ export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
 
   if (typeof parser.parse === 'function') {
     // ParserZodEsque
-    // ParserValibotEsque
+    // ParserValibotEsque (<= v0.12.X)
     return parser.parse.bind(parser);
+  }
+
+  if (typeof parser._parse === 'function') {
+    // ParserValibotEsque (>= v0.13.X)
+    return (value) => parseAsync(parser, value);
   }
 
   if (typeof parser.validateSync === 'function') {
--- a/packages/server/src/core/internals/getParseFn.ts
+++ b/packages/server/src/core/internals/getParseFn.ts
@@ -1,4 +1,3 @@
-import { parseAsync } from 'valibot';
 import { Parser } from '../parser';
 
 export type ParseFn<TType> = (value: unknown) => Promise<TType> | TType;
@@ -24,7 +23,10 @@ export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
 
   if (typeof parser._parse === 'function') {
     // ParserValibotEsque (>= v0.13.X)
-    return (value) => parseAsync(parser, value);
+    return async (value) => {
+      const { parseAsync } = await import('valibot');
+      return parseAsync(parser, value);
+    };
   }
 
   if (typeof parser.validateSync === 'function') {


packages/server/src/core/internals/getParseFn.ts
```diff
@@ -1,3 +1,4 @@
+import { parseAsync } from 'valibot';
```
juliusmarminge, line 1: Will this treeshake away or will it cause runtime errors cause it can't be found for projects not using Valibot?

fabian-hiller: I do not know. That's why I mentioned it in the description.

fabian-hiller: I can imagine that it could cause problems. An alternative would be a dynamic import. However, I have no experience with it.

juliusmarminge: It might mean that the API must change for valibot users to

```ts
import * as v from "valibot";

const myProc = publicProceudre
  .input(v.parseAsync(
    v.object({ ... })
  .mutation(...)
```

fabian-hiller: I think this doesn't work because the value being validated is the 2 parameter of the parse function.

fabian-hiller: Another alternative would be to offer an adapter that can import Valibot.

fabian-hiller: Sorry, I didn't look closely at your code example yesterday. My previous statements are wrong. I think your idea is good. I'll check it in the next days and get back to you.

packages/server/src/core/internals/getParseFn.ts
```diff
@@ -17,10 +18,15 @@ export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
 
   if (typeof parser.parse === 'function') {
     // ParserZodEsque
-    // ParserValibotEsque
+    // ParserValibotEsque (<= v0.12.X)
     return parser.parse.bind(parser);
   }
 
+  if (typeof parser._parse === 'function') {
```
juliusmarminge, line 25: Wait what is the change you're making? Why can't we just call this _parse method?

fabian-hiller: We can call `._parse`. However, `._parse` no longer throws a `ValiError` but returns an object containing either the data under `output` or all the problems found under `issues`.

https://github.com/fabian-hiller/valibot/blob/main/library/src/methods/parse/parse.ts

fabian-hiller: We have revised the internal error handling and thus massively improved the performance.

fabian-hiller: I could, of course, throw a standard `Error` instead of a `ValiError`. But then, all validation information would be lost, which is a problem when you want to find out why the validation failed. I can't throw a `ValiError` because I would have to import it.

juliusmarminge: hmm okay so I think either you need to add a method to the validator that throws, or we'll have to accept a worse DX for users by letting them call `parseAsync` in a custom validator function as i show yesterday `.input(raw => parseAsync(MY_VALIBOT_VALIDATOR, raw)`

fabian-hiller: Alternatively I could try to import `ValiError` dynamically. If someone uses Valibot, it must be installed. However, I have no experience with dynaic imports and therefore do not know how they behave in the build step.

fabian-hiller: I have updated the code. It seems to work. At least no error occurs during the tests. I tested the following with the current version of Valibot to verify:

```ts
  if (typeof parser.parse === 'function') {
    // ParserValibotEsque (<= v0.12.X)
    if (typeof parser.schema === 'string') {
      console.log('VALIBOT 🔥🔥🔥🔥🔥🔥🔥');
      return async (value) => {
        const { parseAsync } = await import('valibot');
        return parseAsync(parser, value);
      };
    }
    // ParserZodEsque
    return parser.parse.bind(parser);
  }
```

fabian-hiller: screenshot

juliusmarminge: The only problem i can see is a separate treatment of valibot and any other library that (in the future) puts a `_parse` on their objects will not work since that import wont work

fabian-hiller: I understand. However, this generally applies to the current implementation of tRPC. For example, this argument could also be made with the current implementation of `ParserScaleEsque`. If tRPC wants to avoid this in the long run, individual adapters, a standardization, or an additional layer like [`TypeSchema`](https://github.com/decs/typeschema) are required.

For a quick solution, I suggest moving the code for Valibot >= v0.13.X to the end. If it should collide with another library in the near future, more conditions can be added as each schema from different libraries has different properties.

juliusmarminge: True. What do you say @KATT & @sachinraja, is the dynamic import okay?

KATT: No, we I don't think we should have dynamic imports, not sure if that will even work everywhere -- we need a parse function

fabian-hiller: According to my research, `import()` is a standard supported by any environment, including Node.js and Deno. However, as described, I have no practical experience with it yet.

fabian-hiller:
@KATT theoretically, I can do that. Practically, however, it contradicts the philosophy and API design of Valibot. Because then the design would no longer be completely modular and we would increase the bundle size for all users.

KATT:
> @KATT theoretically, I can do that. Practically, however, it contradicts the philosophy and API design of Valibot. Because then the design would no longer be completely modular and we would increase the bundle size for all users.

I understand your point of view, but I also don't see how Valibot is at all useful without the ability to parse stuff, so it feels like a thing that people always include anyway?

It makes it very hard to integrate without having a hard dependency - dynamic imports don't work seamlessly everywhere AFAIK so I don't think this would "just work".

fabian-hiller:
Since the selling point of Valibot is a modular design with a small bundle size starting at less than 300 bytes, adding the `parse` function will increase the initial bundle size by 200 bytes from one of the next versions. Also, it is questionable why we only add `parse` and not other methods like `safeParse`, which increases the bundle size even more.

If someone only uses our `is` type guard or the `safeParse` method, those 200 bytes are useless and make a difference for Valibot's goal.

What is wrong with adding Valibot as a dependency and importing the `parse` method? Since it is the server package of tRPC, it doesn't make any difference, does it? Is there any other way to make the `issues` returned by `._parse` accessible?

For example, an alternative would be to implement a `ParseError` class in tRPC that we can use to throw an error with parse issues:

```ts
export type Issues = [Record<string, any>, ...Record<string, any>[]];

export class ParseError extends Error {
  issues: Issues;
  constructor(issues: Issues) {
    super(issues[0].message);
    this.name = 'ParseError';
    this.issues = issues;
  }
}
```

fabian-hiller:
If there is no other way, I will investigate implementing a `.parse` method to weigh the pros and cons for Valibot.

fabian-hiller:
> No, we I don't think we should have dynamic imports, not sure if that will even work everywhere

@KATT do you know of any server-side environment where dynamic imports cause problems? According to my research, it is an established standard supported by all environments, including Node.js and Deno, for many years. Since the dynamic import only applies when Valibot is used, it should not affect tRPC. In my tests, no problems occurred when, for example, Zod was used and Valibot was not installed because the dynamic import was not executed in this case.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import

fabian-hiller:
To continue working on my bachelor's thesis, I must release the new version of Valibot on Monday or Tuesday. I hope we find a solution so that tRPC is still supported.

fabian-hiller:
I couldn't wait any longer and released the new version v0.13.0. I hope we can find a solution to make it work with tRPC.

juliusmarminge:
Sorry for not responding. I'm on vacation this week and just wanted to keep away entirely from Github.

The best thing for now might be to urge users on the newer version to do

```ts
.input((raw) => parseAsync(string(), raw))
```

Until we fully decide if we can have dynamic imports or not. As I've been away I havent done any research on the support that `import()` has, but it seems like you have and concluded it's good. I'll research once i get back home

kamilogorek:
This might be problematic for tools like https://github.com/vercel/ncc or https://github.com/vercel/nft which are used to bundle/deploy node.js apps in various environments.
When using dynamic import only in a specific branch, static analysis won't be able to decide whether a given dependency should be bundled or not. This might have an outcome of always trying to bundle `valibot` no matter what.

KATT:
Conclusion: We won't merge this right now. Feel free to do a PR to the official docs using @decs' library that uses this proposed approach.

We're not sure if we want to have dynamic `import`s in tRPC right now and I'm going to holiday for two weeks now and prefer not thinking about merging risky stuff today :)

fabian-hiller:
I will create a PR next week that will update the tRPC docs to the new Valibot version.

PR #4732, opened 2023-08-21
Feat(server): narrow the type of 'meta' when runtime meta value injected

jussisaurio:
## 🎯 Changes

Feat(server): narrow the type of 'meta' when runtime meta value injected

Currently tRPC middlewares always assume that the type of `opts.meta`
is `Meta | undefined` even when a procedure injects metadata using
`.meta()` or `defaultMeta` is provided to `initTRPC().create()`.

This commit modifies the behavior so that whenever either `procedure.meta()`
or `defaultMeta` is provided, the type of Meta is narrowed in downstream
handlers.

The `RootConfig` created in `initTRPC` is now changed
from `Meta` to `Partial<Meta>`, since this facilitates removing the type:

`_meta: TConfig['_meta'] | undefined`

from the definition of `MiddlewareFunction` (this is needed so that middlewares
aren't always stuck assuming `opts.meta` is a union of `Meta` and `undefined`.

Note: this would have the potential to break users' typechecks if they are doing sth
like:

`if (opts.meta && !opts.meta.foo) { throw new Error("bang!"); }`

because with the changed types, `opts.meta` is Partial<Meta>. To combat this
I changed the definition of `MiddlewareFunction` to `Meta | undefined` ONLY when `meta`
has not been injected by a procedure or by `defaultMeta`, which should preserve
backwards compatibility, as without runtime injection, the runtime value of Meta
will be `undefined`.

## ✅ Checklist

### DIFF
--- a/packages/server/src/core/initTRPC.ts
+++ b/packages/server/src/core/initTRPC.ts
@@ -21,7 +21,7 @@ import {
 } from './internals/config';
 import { mergeRouters } from './internals/mergeRouters';
 import { createBuilder } from './internals/procedureBuilder';
-import { PickFirstDefined, ValidateShape } from './internals/utils';
+import { Overwrite, PickFirstDefined, ValidateShape } from './internals/utils';
 import { createMiddlewareFactory } from './middleware';
 import { createRouterFactory } from './router';
 
@@ -103,7 +103,9 @@ function createTRPCInner<TParams extends PartialRootConfigTypes>() {
 
     type $Config = RootConfig<{
       ctx: $Context;
-      meta: $Meta;
+      meta: TOptions['defaultMeta'] extends infer T extends object
+        ? Overwrite<Partial<$Meta>, T>
+        : Partial<$Meta>;
       errorShape: $ErrorShape;
       transformer: $Transformer;
     }>;
--- a/packages/server/src/core/internals/procedureBuilder.ts
+++ b/packages/server/src/core/internals/procedureBuilder.ts
@@ -130,7 +130,11 @@ export interface ProcedureBuilder<TParams extends ProcedureParams> {
   /**
    * Add a meta data to the procedure.
    */
-  meta(meta: TParams['_meta']): ProcedureBuilder<TParams>;
+  meta<TMeta extends Partial<TParams['_meta']>>(
+    meta: TMeta,
+  ): ProcedureBuilder<
+    Overwrite<TParams, { _meta: Overwrite<TParams['_meta'], TMeta> }>
+  >;
   /**
    * Add a middleware to the procedure.
    */
--- a/packages/server/src/core/middleware.ts
+++ b/packages/server/src/core/middleware.ts
@@ -136,7 +136,11 @@ export type MiddlewareFunction<
     path: string;
     input: TParams['_input_out'];
     rawInput: unknown;
-    meta: TParams['_meta'] | undefined;
+    // Always infer opts.meta as potentially undefined in middlewares if it hasn't been injected by a previous middleware or by defaultMeta
+    // Allowing meta to be undefined is for backwards compatibility reasons
+    meta: Record<PropertyKey, never> extends TParams['_meta']
+      ? TParams['_meta'] | undefined
+      : TParams['_meta'];
     next: {
       (): Promise<MiddlewareResult<TParams>>;
       <$Context>(opts: { ctx: $Context }): Promise<
--- a/packages/server/src/core/types.ts
+++ b/packages/server/src/core/types.ts
@@ -15,8 +15,9 @@ export type inferRouterContext<TRouter extends AnyRouter> =
   inferRouterDef<TRouter>['_config']['$types']['ctx'];
 export type inferRouterError<TRouter extends AnyRouter> =
   inferRouterDef<TRouter>['_config']['$types']['errorShape'];
-export type inferRouterMeta<TRouter extends AnyRouter> =
-  inferRouterDef<TRouter>['_config']['$types']['meta'];
+export type inferRouterMeta<TRouter extends AnyRouter> = Required<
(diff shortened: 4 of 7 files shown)


jussisaurio:
forgot to set this as draft (on the run now)

the runtime implementation is still a bit wrong

jussisaurio:
This is ready for review now as far as I am concerned. Apologies for the hassle earlier.

KATT (commented):
I think we should get rid of the default meta - you can just do a reusable base procedure for that (right?)

jussisaurio:
> I think we should get rid of the default meta - you can just do a reusable base procedure for that (right?)

Probably yeah. While we're on the subject of breaking changes:

I also got to thinking what the point of declaring `Meta` (the type) in `initTRPC` was in the first place, if it could always be inferred from `procedure.meta()` injections, but I guess I had misunderstood the use cases for `meta` a bit -- didn't realize that decorating meta _after_ middlewares was a thing (example from docs):

```
const authProcedure = publicProcedure
  .use(authMiddleware)
  .meta({
    authRequired: true;
    role: 'user'
  });
// ^ Meta: { authRequired: true, role: 'user' }

const adminProcedure = authProcedure
  .meta({
    role: 'admin'
  });
// ^ Meta: { authRequired: true, role: 'admin' }
```

I dunno how much the above is used, but if you imagine a world where:

- [standaloneMiddleware](https://github.com/trpc/trpc/pull/4713) eventually replaces `t.middleware`; effectively t.middleware is just `standaloneMiddleware<{ ctx: Context, meta: Meta }>()` which can be re-exported the same way people are currently exporting `export const middleware = t.middleware`,
- `meta` is always injected in procedures before the middlewares that use it, and the type is inferred from the value

Then in theory you'd have no explicit need to declare `Meta` at the `initTRPC` level:

```
const t = initTRPC.context<Context>.create();

const publicProcedure = t.procedure.meta({ foo: 'bar' })
// imagine that standaloneAuthMiddleware requires 'authRequired' and 'role'
const authProcedure = publicProcedure.meta({ authRequired: true, role: 'user' }).use(standaloneAuthMiddleware);
const adminProcedure = publicProcedure.meta({ authRequired: false, role: 'admin' }).use(standaloneAuthMiddleware);

// imagine that fooHandlerMiddleware requires `foo: string`
const fooProcedure = adminProcedure.use(fooHandlerMiddleware);
```

However enforcing that meta flows down from handler to handler via inference instead of explicit declaration @ `initTRPC` level might get rid of some useful things users are currently doing:

```
// For example if a middleware in 'authProcedure' required 'role' but somehow wouldn't have that property injected before those middlewares, then this kind of thing wouldn't work:
const adminProcedure = authProcedure
  .meta({
    role: 'admin'
  });
```
, and there might be something else crucial here that I'm simply not seeing. The way I'm using tRPC is certainly a small subset of the total aggregate use cases.

Just food for thought, might be totally off base here

GeorgeIpsum:
Hi all! Not sure if this is the right place to comment, but I think this feature is useful to a usecase at my org. I wanted to add my 2 cents to why being able to narrow the Meta type further could be helpful:

My org's codebase uses Prisma to handle db ORM; we've defined an "rbacProcedure" for procedures that require certain access controls on the User in our db. Our "rbacProcedure" has a middleware that introspects the meta object supplied to the procedure to determine whether or not a user has the proper access controls:

```typescript
export const rbacProcedure = authProcedure.use(async ({ next, ctx, meta }) => {
  const { rbac } = meta;

  const user = await ctx.prisma.user.findUnique({
    // ...
  });

  for (const [key, requiredPerm] of Object.entries(rbac)) {
    const controlKey = key as UserControlKey;
    const userPerm = user.accessControl[controlKey];

    if (!evalRole(userPerm, requiredPerm)) {
      throw new TRPCError({
        code: "UNAUTHORIZED",
        message: `Role check failed for user with email ${user.pii}`,
      });
    }

    return next({
      ctx: {
        ...ctx,
        user,
      },
    });
  }
}
```

Another pattern that has emerged in our codebase is that since we're querying for the user in this access control, we also allow for adding other fields that can be included on the query:

```typescript
  const { rbac, userFields } = meta;

  const user = await ctx.prisma.user.findUnique({
    include: { ...userFields }
    // ...
  });
```

When we type meta right now, we use Prisma's generated Prisma.UserInclude type:
```typescript
export interface Meta {
  rbac?: AtLeastOne<UserControls>;
  userFields?: Prisma.UserInclude;
}
```

Herein lies the problem. The resulting `user` object is not typesafe because when our query takes in `userFields` as a `Prisma.UserInclude` type, the resulting user type will include ALL possible relations instead of the specific ones we're asking for in `userFields` when we call our rbacProcedure:
```typescript
const appRouter = t.router({
  checkResources: rbacProcedure.meta({
    rbac: { resourceA: "WRITE", resourceB: "READ" },
    userFields: { relationA: true, relationB: { include: { /**...*/ } } }
    // in our rbacProcedure middleware we'd like our meta.userFields to be the exact type of the above
  }).query(/** etc */)
)}
```

I don't think necessarily that the solution being proposed solves this (or if we're egregiously misusing the meta field), but we like the ergonomics of being able to declare what "resources" we'd like on the user object directly in our procedure definitiion. If nothing else I'd like for the Meta interface we supply to allow some kind of generic injection such as the following:

```typescript
const userFieldsForProc = {
  resourceA: true
  // ...
} satisfies Prisma.UserInclude;

const someRbacProcedure = rbacProcedure.meta<{ userFields: typeof userfieldsForProc }>(userFieldsForProc).query(/** ... */);
```

I have no idea if this is the right place to posit this question. Is this possible with tRPC currently or with some type of addition to this PR (if so would be more than happy to contribute), or is it totally out of scope for the project?

Thanks for reading!

KATT:
This is not great for TS performance so I'm not very sure about it yet - we might consider it for next major

PR #4751, opened 2023-08-26
fix(`@trpc/server`): rearrange the order of context generation in `resolveHTTPResponse` method

maticzav:
This PR fixes an issue where context would be empty in `onError` and `errorFormatter` functions when calling the TRPC server with invalid JSON data.

We encountered this error while testing by calling an endpoint like `http://localhost:3001/trpc/brands.getDefault?input=QUERY`. Because "QUERY" is not a valid JSON, the default `getJsonContentTypeInputs` (i.e. `getInputs`) function would throw. Since context currently only gets generated *after* the input is parsed, it wouldn't be created correctly.

> We use context to communicate through the application the configuration parameters. One of the configurations is the production/development stage which we use to determine whether error should be transformed. Because of the bug, the transformer wouldn't work as expected and essentially leaked information about our file system in the stack trace.

```ts
const t = initTRPC.context<Context>().create({
  transformer: superjson,
  errorFormatter: ({ shape, ctx }) => {
    if (ctx?.config.isProduction) {
      return {
        message: 'Internal server error',
        code: -32603 as any,
        data: {},
      }
    }

    return shape
  },
})
```

I believe this is not intentional and this PR fixes the bug by changing the order of context generation and input parsing. This way `caughtErrorToData` gets the correct context value even if `getInputs` throws on invalid JSON.

### DIFF
--- a/.changeset/twelve-rocks-look.md
+++ b/.changeset/twelve-rocks-look.md
@@ -1,0 +1,5 @@
+---
+'@trpc/server': patch
+---
+
+Fixed a bug where the context could be empty in `errorFormatter` and `onError` functions when the requested endpoint url contained invalid JSON data.
--- a/.changeset/twelve-rocks-look.md
+++ b/.changeset/twelve-rocks-look.md
@@ -1,5 +1,0 @@
----
-'@trpc/server': patch
----
-
-Fixed a bug where the context could be empty in `errorFormatter` and `onError` functions when the requested endpoint url contained invalid JSON data.
--- a/packages/server/src/http/resolveHTTPResponse.ts
+++ b/packages/server/src/http/resolveHTTPResponse.ts
@@ -289,6 +289,13 @@ export async function resolveHTTPResponse<
   try {
     // we create context first so that (unless `createContext()` throws)
     // error handler may access context information
+    //
+    // this way even if the client sends malformed input that might cause an exception:
+    //  - `opts.error` has value,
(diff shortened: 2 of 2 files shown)


Sheraff:
@maticzav Could you leave a comment in the code to explain why `ctx` needs to be created before it's used? (explaining the code path that would result in an issue if not created before `getInputs` is called)

Sheraff (commented):
Could you leave a comment in the code to explain why ctx needs to be created before it's used? (explaining the code path that would result in an issue if not created before getInputs is called)

maticzav:
@Sheraff I updated the code and added a comment that seemed in line with the structure of other comments. Let me know if that's what you had in mind! 🙂

Sheraff:
> @Sheraff I updated the code and added a comment that seemed in line with the structure of other comments. Let me know if that's what you had in mind! 🙂

That's great, very clear, thanks. And good catch.

Sheraff (approved):

.changeset/twelve-rocks-look.md
juliusmarminge, line 1: We dont use changesets, can you remove this?

maticzav: Yes!

juliusmarminge (approved):

KATT:
FYI, this will be removed in the next major as we have a new `info`-object passed to the `createContext` which contains info about the procedures that will be ran and their inputs

PR #4756, opened 2023-08-27
feat: allow passing listening listener parameter to standalone server

### LINKED ISSUE #4740: feat: add a listener to standalone server

I use tRPC as a helper tool to manage databases yielding in my test backbone.
To run starting setup [I need to wait](https://github.com/luixo/receipt-app/pull/2/files?diff=split&w=1#diff-6a0dad1153b6b44d0d07bf1db78d7c7b26021aae5940150b440f2d558094b88aR29-R32) until the server is bootstrapped, but I have no option for that with a standalone server package.

Add a listener passed to the `http` server instance. The patch [can be found](https://github.com/luixo/receipt-app/pull/2/files?diff=split&w=1#diff-76e074d8439d47f302e7eb34678ef817ac4af0cf44e9b0d378535e5370d3399d) in my repo.

Use external methods of checking server listening

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

## Funding

<a href="https://polar.sh/trpc/trpc/issues/4740">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/4740/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

sachinraja:
Closes #4740

### DIFF
--- a/packages/server/src/adapters/standalone.ts
+++ b/packages/server/src/adapters/standalone.ts
@@ -45,8 +45,12 @@ export function createHTTPServer<TRouter extends AnyRouter>(
 
   return {
     server,
-    listen: (port?: number, hostname?: string) => {
-      server.listen(port, hostname);
+    listen: (
+      port?: number,
+      hostname?: string,
+      listeningListener?: () => void,
+    ) => {
+      server.listen(port, hostname, listeningListener);
       const actualPort =
         port === 0 ? ((server.address() as any).port as number) : port;
 


kamilogorek:
I'm fine with the change, but keep in mind that this is actually already possible to achieve this without any API changes.

```js
const { server, listen } = createHTTPServer();
server.on('listening', () => console.log('hi there'));
listen();
```

I'd also rework it to `server.listen({ port, host: hostname}, listeningListener);` instead, as currently it'll use 4th variant of the API, which might be confusing to some (due to possible `backlog` argument), where `server.listen(options, callback)` is always clear - https://nodejs.org/api/net.html#serverlisten

luixo:
> I'd also rework it to server.listen({ port, host: hostname}, listeningListener); instead, as currently it'll use 4th variant of the API, which might be confusing to some (due to possible backlog argument), where server.listen(options, callback) is always clear - https://nodejs.org/api/net.html#serverlisten

Can we do that? Using a union of options to pass would be just great.
I'm up to take a PR if you need a helping hand.

sachinraja:
I don't think we need to complicate that function, personally I don't think we should be exposing a `listen` function at all.

PR #4787, opened 2023-09-11
fix(`server`): allow adding charset after `application/json`

### LINKED ISSUE #4783: bug: charset=utf8 results in empty input

Redacted due to privacy and it's not relevant to the bug

### Describe the bug

I'm using `trpc-openapi` and `adapters/aws-lambda`.

Some of our customers are using Zapier to communicate with our API.
Zapier uses the header `Content-Type: application/json; charset=utf-8` - note the `charset=utf-8`.

When `charset=utf-8` exists, the input to the relevant procedure is `{}`

I can confirm the lambda receives the header `Content-Type: application/json; charset=utf-8`. And I've wrapped the handler to convert it to `Content-Type: application/json` - and then that works as expected

[See thread on Discord](https://discord.com/channels/867764511159091230/1148232196507832390)

### Link to reproduction

-

### To reproduce

Deploy a Lambda, with tRPC. Use `curl` to hit the API with the header `Content-Type: application/json; charset=utf-8`

### Additional information

I've had a look at the repo, I can find string comparisons to `'application/json'`

- https://github.dev/trpc/trpc/blob/aa1a3daf58254fa9f0160847874632eb3c8faaad/packages/server/src/adapters/node-http/content-type/json/getPostBody.ts#L16-L17
- https://github.dev/trpc/trpc/blob/aa1a3daf58254fa9f0160847874632eb3c8faaad/packages/server/src/adapters/node-http/content-type/json/index.ts#L7

However both are in `node-http` and I don't think they're relevant to the lambda adapter.

Open to making a PR, but I'll need some guidance

### 👨‍👧‍👦 Contributing

## Funding

<a href="https://polar.sh/trpc/trpc/issues/4783">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/4783/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

KATT:
Couldn’t reproduce #4783, but this seems like a good measure

Closes #4783

### DIFF
--- a/packages/server/src/adapters/node-http/content-type/json/getPostBody.ts
+++ b/packages/server/src/adapters/node-http/content-type/json/getPostBody.ts
@@ -13,7 +13,8 @@ export async function getPostBody(opts: {
         ok: true,
         data: req.body,
         // If the request headers specifies a content-type, we assume that the body has been preprocessed
-        preprocessed: req.headers['content-type'] === 'application/json',
+        preprocessed:
+          !!req.headers['content-type']?.startsWith('application/json'),
       });
       return;
     }
--- a/packages/server/src/adapters/node-http/content-type/json/index.ts
+++ b/packages/server/src/adapters/node-http/content-type/json/index.ts
@@ -4,7 +4,7 @@ import { getPostBody } from './getPostBody';
 
 export const nodeHTTPJSONContentTypeHandler = createNodeHTTPContentTypeHandler({
   isMatch(opts) {
-    return opts.req.headers['content-type'] === 'application/json';
+    return !!opts.req.headers['content-type']?.startsWith('application/json');
   },
   getBody: getPostBody,
   getInputs: getJsonContentTypeInputs,
--- a/packages/server/src/adapters/fetch/fetchRequestHandler.ts
+++ b/packages/server/src/adapters/fetch/fetchRequestHandler.ts
@@ -25,10 +25,9 @@ export async function fetchRequestHandler<TRouter extends AnyRouter>(
     query: url.searchParams,
     method: opts.req.method,
     headers: Object.fromEntries(opts.req.headers),
-    body:
-      opts.req.headers.get('content-type') === 'application/json'
-        ? await opts.req.text()
-        : '',
+    body: opts.req.headers.get('content-type')?.startsWith('application/json')
+      ? await opts.req.text()
+      : '',
   };
 
   let resolve: (value: Response) => void;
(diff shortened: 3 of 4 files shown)


packages/server/src/adapters/node-http/content-type/json/getPostBody.ts
```diff
@@ -13,7 +13,8 @@ export async function getPostBody(opts: {
         ok: true,
         data: req.body,
         // If the request headers specifies a content-type, we assume that the body has been preprocessed
-        preprocessed: req.headers['content-type'] === 'application/json',
+        preprocessed:
+          !!req.headers['content-type']?.startsWith('application/json'),
```
kamilogorek, line 17: `startsWith` already returns a `boolean`, no need to coerce.

Sheraff (approved):

packages/server/src/adapters/node-http/content-type/json/index.ts
```diff
@@ -4,7 +4,7 @@ import { getPostBody } from './getPostBody';
 
 export const nodeHTTPJSONContentTypeHandler = createNodeHTTPContentTypeHandler({
   isMatch(opts) {
-    return opts.req.headers['content-type'] === 'application/json';
+    return !!opts.req.headers['content-type']?.startsWith('application/json');
```
kamilogorek, line 7: Same here

packages/server/src/adapters/node-http/content-type/json/getPostBody.ts
kamilogorek: Oh, its possibly `undefined` 😢

Sheraff: It's optional chaining, might return undefined. I assumed that's why it was coerced.

juliusmarminge: optional chainging would return undefined, guessing that's why the boolean coercion is there

kamilogorek (approved):

packages/server/src/adapters/node-http/content-type/json/getPostBody.ts
kamilogorek: That response speed. Gotta go fast

sachinraja:
I think #4783 is an issue specific to aws lambda. So I don’t think this should close it

KATT:
> I think #4783 is an issue specific to aws lambda. So I don’t think this should close it

As far as tRPC goes, AFAICT, does everything we "can do" from our side. If something else is up, I think it's an issue outside of tRPC

PR #4788, opened 2023-09-11
fix(`client`): skip calling `.abort()` on AbortController after procedure completion

### LINKED ISSUE #4750: bug: AbortController's cancel() called after every outgoing request finishes
For background, we use TRPC for server <> server communication, which is not the primary use case, but it does work really well. This post concerns the TRPC client's HTTP link, and talks about **outgoing** requests to other TRPC services from a client perspective. When we initialise the `HTTPLink`, we pass in an `AbortController` from the `adobe-fetch` library. We use `adobe-fetch` because it has the best HTTP/2 support.

We're seeing sporadic `AbortError`s being thrown sometimes (on the client side) when we try to make requests from one service to another. It's very unclear to me why this is happening at the moment, I'm investigating it.

I noticed something curious. The abort() method on the AbortController (passed into the HTTP link as an argument) seems to be being called at the end of every outgoing request, regardless of whether or not that request was successful.

Screenshot 2023-08-26 at 16 50 19

Looking in the code here.
https://github.com/trpc/trpc/blob/main/packages/client/src/links/internals/httpUtils.ts#L200-L202

```
export function httpRequest(
  opts: HTTPRequestOptions,
): PromiseAndCancel<HTTPResult> {
  const ac = opts.AbortController ? new opts.AbortController() : null;
  const meta = {} as HTTPResult['meta'];

  const promise = new Promise<HTTPResult>((resolve, reject) => {
    fetchHTTPResponse(opts, ac)
      .then((_res) => {
        meta.response = _res;
        return _res.json();
      })
      .then((json) => {
        meta.responseJSON = json;
        resolve({
          json: json as TRPCResponse,
          meta,
        });
      })
      .catch((err) => {
        reject(TRPCClientError.from(err, { meta }));
      });
  });
  const cancel = () => {
    ac?.abort();
  };
  return { promise, cancel };
}
```

The `cancel` function here ultimately ends up being given to the observable as a cleanup function [here](https://github.com/trpc/trpc/blob/main/packages/client/src/links/httpLink.ts#L76-L78C13)

I didn't expect this, I'm sorry if I'm missing something, but I would only expect `cancel()` to be invoked when the requester actually wants to cancel a request that is currently in flight, and not every time after a request has completed.

Is this behaviour intentional? What is the rationale behind this approach? I can't be certain this has something to do with the errors we're seeing but it feels like it might.

Thanks!

_Originally posted by @kieranm in https://github.com/trpc/trpc/discussions/4749_

## Funding

<a href="https://polar.sh/trpc/trpc/issues/4750">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/4750/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

KATT:
Closes #4750

## 🎯 Changes

- Skip calling `.abort()` on AbortController after procedure completion
- Tweaked some turbo-stuff to 
  - make sure `postinstall` includes everything for `test-watch` to work 
  - reinforced a the subtle "convention" that `:`-tasks means something that is run in parallel best on a `run-p x:*`-command

## ✅ Checklist


packages/tests/server/regression/issue-4750-no-cancel.test.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  await proxy.q
+    .query(undefined, {
+      signal: abortController.signal,
+    })
+    .catch(() => {
+      /// ..
+    });
+
+  expect(abortController.signal).toBe(false);
```
kamilogorek, line 56: I believe you wanted to assert on `.aborted` here? If so, it passes just fine when changed.

KATT: haha yes, this test is whack

CONTRIBUTING.md
```diff
@@ -30,17 +30,15 @@ In another terminal, you can for instance navigate to `examples/next-prisma-star
...
-
-First, run
+Open one terminal and run:
 
 ```bash
-pnpm test
+pnpm dev
 ```
 
-in the project root directory. This will make sure all the build artifacts and codegen has run. After that, you can run the tests in watch mode using
```
KATT, line 41: lol, i didn't even know this

CONTRIBUTING.md
```diff
@@ -30,17 +30,15 @@ In another terminal, you can for instance navigate to `examples/next-prisma-star
...
+In a second terminal, you can run the tests in watch mode using
 
-```
+```bash
```
kamilogorek, line 41: #protip ```shell also works :)

juliusmarminge (approved):

packages/client/src/links/internals/httpUtils.ts
```diff
@@ -194,11 +196,14 @@ export function httpRequest(
         });
       })
       .catch((err) => {
+        done = true;
         reject(TRPCClientError.from(err, { meta }));
       });
```
kamilogorek, line 201: You can also use `.finally(() => { done = true; });` to have everything in one place.

packages/client/src/links/internals/httpUtils.ts
```diff
@@ -194,11 +196,14 @@ export function httpRequest(
...
   });
   const cancel = () => {
-    ac?.abort();
+    if (!done) {
```
kamilogorek, line 204: That works as a workaround, but I wonder why observable is even cancelling it in the first place? Or is it just a generic "cleanup" function?

KATT: Yeah, there might be something to what you're saying here

packages/client/src/links/internals/httpUtils.ts
KATT: Unfortunately I can't because calling `reject()`/`resolve()` causes a race condition that gets to the cleanup function before `.finally()` is executed

packages/client/src/links/internals/httpUtils.ts
KATT: Two years since I wrote the observable-stuff, my memory is rust, but I think it's a generic cleanup, but could prob be improved

PR #4801, opened 2023-09-17
feat: Add transparency about httpBatchStreamLink to docs

zokins:
Added transparency for not using `httpBatchStreamLink` if you need to alter the response headers inside procedures

## 🎯 Changes
Added note in docs to ensure people understand when to not use the batch stream

## ✅ Checklist


www/docs/client/links/httpBatchStreamLink.md
```diff
@@ -15,6 +15,10 @@ We have prefixed this as `unstable_` as it's a new API, but you're safe to use i
 
 > All usage and options are identical to [`httpBatchLink`](./httpBatchLink.md).
 
+:::note
+If you require the ability to change/set response headers (which includes cookies) from within your procedures, make sure to use `httpBatchLink` instead! [Read more](https://trpc.io/docs/client/links/httpBatchLink).
+:::
```
juliusmarminge, line 20: Should state the reason you have to do this is cause streaming doesn't support setting headers after the response has started streaming

zokins: > Should state the reason you have to do this is cause streaming doesn't support setting headers after the response has started streaming

Thanks for pointing that out Julius, I've added a snippet to the end :)

kamilogorek (approved):

www/docs/client/links/httpBatchStreamLink.md
```diff
@@ -15,6 +15,10 @@ We have prefixed this as `unstable_` as it's a new API, but you're safe to use i
 
 > All usage and options are identical to [`httpBatchLink`](./httpBatchLink.md).
 
+:::note
+If you require the ability to change/set response headers (which includes cookies) from within your procedures, make sure to use `httpBatchLink` instead! This is due to the fact that `unstable_httpBatchStreamLink` does not support setting headers once the stream has begun. [Read more](https://trpc.io/docs/client/links/httpBatchLink).
```
juliusmarminge, line 19: nit, ubt this sort of makes it sound like it's a limitation on the link itself, but it's the opposite - streaming in general doesn't support setting headers after the stream has begun

PR #4805, opened 2023-09-19
chore: next merge 2023 09 19

juliusmarminge:


packages/client/src/TRPCClientError.ts
```diff
@@ -97,11 +99,22 @@ export class TRPCClientError<TRouterOrProcedure extends ErrorInferrable>
       }
       return cause;
     }
+    if (isTRPCErrorResponse(cause)) {
+      return new TRPCClientError(cause.error.message, {
+        ...opts,
+        result: cause as any,
```
juliusmarminge, line 105: added this `as any` which wasn't on main to get the types not to complain about the subclass instantiation thing

PR #4843, opened 2023-09-26
chore: bump react-query to rc

juliusmarminge:
bumps rq to rc


packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -484,6 +485,8 @@ export function createRootHooks<
   ): UseTRPCSuspenseInfiniteQueryResult<unknown, TError, unknown> {
     const hookResult = useInfiniteQuery(path, input, {
       ...opts,
+      // FIXME: How should we handle suspense hooks?
+      // @ts-expect-error - whatever?
```
juliusmarminge, line 489: tests are passing - are we safe to use this option even if it's not public?

or should we use the built-in suspense hooks? how do we do with the prefetching logic we have in "our normal" hooks?

juliusmarminge: cc @KATT @TkDodo

KATT: Let's go for it I think -- WDYT @TkDodo?

KATT (approved):

packages/react-query/src/shared/hooks/createHooksInternal.tsx
TkDodo: I would encourage you to please use the built-in `useSuspenseQuery`, `useSuspenseInfiniteQuery` and `useSuspenseQueries` hook. It's an implementation detail that they are now implemented with `suspense:true`, because I didn't get around to changing that, but it very well might change in the future in a minor version.

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -484,6 +485,8 @@ export function createRootHooks<
...
       suspense: true,
       enabled: true,
       throwOnError: true,
```
TkDodo, line 492: FYI, we also don't default to `throwOnError: true`. details on that here: https://tanstack.com/query/v5/docs/react/guides/suspense#throwonerror-default

KATT: Let's use the same behavior as RQ here then!

packages/react-query/src/shared/hooks/createHooksInternal.tsx
KATT: Done!

PR #4858, opened 2023-09-28
idea: replacement for superjson

KATT:
Closes #

## 🎯 Changes

Idea of a replacement for superjson


Sheraff:
Would this system be exposed to developers? (for example if trpc requests are intercepted in a service worker, one needs a way to parse the responses)

KATT:
> Would this system be exposed to developers? (for example if trpc requests are intercepted in a service worker, one needs a way to parse the responses)

I was thinking I could make this as a package similar to superjson

KATT:
https://github.com/KATT/tupleson

PR #3271, opened 2022-11-26
test: port inference, links, outputParser, withTRPC to v10

neo773:
Closes #

## 🎯 Changes

This PR ports  `inference.test.ts`, `links.test.ts`, `outputParser.test.ts`, `withTRPC.test.tsx` to v10
## ✅ Checklist

- [x] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [ ] If necessary, I have added documentation related to the changes made.
- [x] I have added or updated the tests related to the changes made.

PR #4868, opened 2023-10-01
chore: use vitest expect-type only

mmkal:
Closes #

## 🎯 Changes

Replace copy-pasted expect-type from https://github.com/mmkal/expect-type/pull/16 with the prerelease version published to npm.

Creating as a draft initially so I can make sure this doesn't regress anything. I ran `pnpm test-ci` with and without this change a bunch of times and the timing was pretty much the same, and it passes.

Note: I don't actually see how `inferenceUtils.ts` is ever referenced but I'm guessing there's some tsconfig setup making its exports available globally. So I just replaced its contents with `export * from 'expect-type'`.

CC @KATT and @neo773 who I think copy-pasted it originally in #3271? Anything I should check before declaring #mmkal/expect-type#16 ready for release? I want to make sure this repo and vitest are both happy with it and can stop maintaining forks.

## ✅ Checklist


mmkal:
Oh, it looks like vitest exports their forked `expectTypeOf` implementation and that's what's being used here - no wonder there was so little difference! There _might_ be a bug in their implementation too, but will do some digging.

mmkal:
Updated to remove the expect-type dependency completely for now. If trpc is using vitest it should just use the re-exported version from there.

KATT (approved):

KATT (approved):

KATT (dismissed):

juliusmarminge (approved):

PR #4880, opened 2023-10-05
feat(client): accept URL object param

cometkim:
This makes `HTTPLink` and `HTTPBatchLink` accept `URL` object for the `url` param

### DIFF
--- a/packages/client/src/links/internals/httpUtils.ts
+++ b/packages/client/src/links/internals/httpUtils.ts
@@ -38,7 +38,7 @@ export function resolveHTTPLinkOptions(
   opts: HTTPLinkBaseOptions,
 ): ResolvedHTTPLinkOptions {
   return {
-    url: opts.url.toString(),
+    url: opts.url.toString().replace(/\/$/, ''), // Remove any trailing slashes
     fetch: opts.fetch,
     AbortController: getAbortController(opts.AbortController),
   };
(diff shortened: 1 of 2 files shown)


juliusmarminge (dismissed):
Seems reasonable to me, beware though that you can't init a relative URL with the URL class (for client side requests for example), so the "classic" `getBaseURL` would need to return something from window.location instead of just empty string if you wanna put it through the URL constructor

```ts
// before
function getBaseUrl() {
  if (typeof window !== 'undefined') {
    return "";
  }
  if (process.env.VERCEL_URL) {
    return `https://${process.env.VERCEL_URL}`;
  }
  return `http://localhost:${process.env.PORT ?? 3000}`;
}

// use URL as `getBaseURL() + "/api/trpc"`

// after
function getBaseUrl() {
  if (typeof window !== 'undefined') {
    return window.location.host;
  }
  if (process.env.VERCEL_URL) {
    return `https://${process.env.VERCEL_URL}`;
  }
  return `http://localhost:${process.env.PORT ?? 3000}`;
}

// use URL as `new URL("/api/trpc", getBaseURL())`
```

juliusmarminge:
Added a test and also removed any trailing slashes as I noticed the URL object appends that sometimes, meaning we'd get a base url like `http://localhost:3000/api/trpc/` and then any request would go to `http://localhost:3000/api/trpc//hello` which would cause a 404.

Anything that can go wrong here @KATT ?

KATT (approved):

PR #4887, opened 2023-10-06
fix(`server`): don't destroy underlying tcp socket when maxBodySize exceeded

### LINKED ISSUE #4886: bug: PAYLOAD_TOO_LARGE error is never returned to client when body size exceeds maxBodySize

```
  System:
    OS: macOS 13.5.2
    CPU: (10) arm64 Apple M1 Max
    Memory: 7.06 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.5.1 - ~/Library/Caches/fnm_multishells/26427_1696576131210/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 9.8.0 - ~/Library/Caches/fnm_multishells/26427_1696576131210/bin/npm
  Browsers:
    Chrome: 117.0.5938.149
    Safari: 16.6
  npmPackages:
    @trpc/client: 10.38.5 => 10.38.5 
    @trpc/server: 10.38.5 => 10.38.5 
    typescript: 5.1.6 => 5.1.6
```

### Describe the bug

When using the tRPC Express adapter and specifying the `maxBodySize` parameter, the `PAYLOAD_TOO_LARGE` error is actually not propagated to the client. Instead, the `fetch` promise rejects without a response and TRPCClient reports `Error: TRPCClientError: Unexpected end of JSON input`

I added the following log to the tRPC module `nodeHTTPRequestHandler`:

```
        opts.req.on('error', err => {
            console.log('Request error', opts.path, err)
        })
```

And this error callback gets called with an Aborted error (`ECONNRESET`). What seems to happen is that the request stream is closed before the headers or the response body can be sent back to the client.

This line in `getPostBody` explicitly destroys the underlying request TCP socket when the `maxBodySize` is exceeded:

https://github.com/trpc/trpc/blob/3f6f3b73d9720b92aab5f0d72d61c874e6a4772b/packages/server/src/adapters/node-http/content-type/json/getPostBody.ts#L31

Commenting out `// req.socket.destroy()` fixes the issue and the `PAYLOAD_TOO_LARGE` error is properly propagated to the client.

I'm unsure why `req.socket.destroy()` is explicitly called, but there's probably a reason?

### Link to reproduction

https://github.com/jussisaurio/trpc-body-size-repro

### To reproduce

1. Create a tRPC server with `maxBodySize` and set it to a low value that you can easily exceed with a manual test
2. Call the endpoint with a payload exceeding `maxBodySize`

Expected: PAYLOAD_TOO_LARGE error
Actual: Unexpected end of JSON input

### Additional information

Willing to file a PR as soon as I understand why `req.socket.destroy()` is called in the first place

### 👨‍👧‍👦 Contributing

jussisaurio:
Closes #4886

## 🎯 Changes

Issue #4886 describes the bug. I could see no reason to hang up the underlying TCP socket in this case so I just made a PR.

Wrote a failing test first to verify the bug, then changed the implementation.

## ✅ Checklist

### DIFF
--- a/packages/server/src/adapters/node-http/content-type/json/getPostBody.ts
+++ b/packages/server/src/adapters/node-http/content-type/json/getPostBody.ts
@@ -28,7 +28,6 @@ export async function getPostBody(opts: {
           ok: false,
           error: new TRPCError({ code: 'PAYLOAD_TOO_LARGE' }),
         });
-        req.socket.destroy();
       }
     });
     req.on('end', () => {
--- a/packages/tests/server/adapters/__router.ts
+++ b/packages/tests/server/adapters/__router.ts
@@ -27,4 +27,7 @@ export const router = t.router({
       message: 'Unexpected error',
     });
   }),
+  exampleMutation: t.procedure
+    .input(z.object({ payload: z.string() }))
+    .mutation(() => 'ok'),
 });
--- a/packages/tests/server/adapters/express.test.tsx
+++ b/packages/tests/server/adapters/express.test.tsx
@@ -35,6 +35,7 @@ async function startServer() {
     '/trpc',
     trpcExpress.createExpressMiddleware({
       router,
+      maxBodySize: 10, // 10 bytes,
       createContext,
     }),
   );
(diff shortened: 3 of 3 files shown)


KATT (approved):
thank you! ❤️

PR #4911, opened 2023-10-12
feat: add experimental support for iterators and streams

### LINKED ISSUE #1853: update request handling in resolveHttpResponse to follow Fetch API
follow up for #1852

We should change the input to `resolveHTTPResponse` to accept Fetch API objects by default. All other handlers should then be updated to map to the Fetch API.

Right now the Fetch adapter has to adapt objects to what tRPC accepts:
https://github.com/trpc/trpc/blob/456aad652c9e98e77044012f47177a620176e009/packages/server/src/adapters/fetch/fetchRequestHandler.ts#L23-L28

<sub>[TRP-66](https://linear.app/trpc/issue/TRP-66/update-request-handling-in-resolvehttpresponse-to-follow-fetch-api)</sub>

## Funding

<a href="https://polar.sh/trpc/trpc/issues/1853">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/1853/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

### LINKED ISSUE #4477: feat: Streaming Mutations / Queries

KATT:
- Closes #4477
- Closes #1853

tl;dr is that queries and mutations can return generator functions, and you can "just" use them

https://github.com/trpc/trpc/blob/77f23db4f5b54ed97e01fd50cff98602d2b6ca27/packages/tests/server/tupleson.test.ts#L383-L399

## 🎯 Changes

Proof of concept of using tupleson to handle streaming data - tl;

- [x] Make new `resolveHTTPResponse`-function that returns a fetch `Response`
- [x] Use tupleson to serialize promises on the backend
- [x] Create a `httpTuplesonLink` that handles this type of response from the server
- [x] Make procedure as a generator function possible
- [x] Make tupleson configurable in `initTRPC` and figure out a way to reuse same configuration for both client and server (transformers will be unnecessary as we can configure them all in tson)


packages/client/src/links/httpTuplesonLink.ts
KATT, line 1: This is not done at all - just a stub

packages/client/src/links/httpTuplesonLink.ts
```diff
@@ -0,0 +1,144 @@
...
+      getUrl,
+      getBody,
+    },
+    ac,
+  );
+  const cancel = () => ac?.abort();
+  const textDecoder = new TextDecoder();
+  const promise = responsePromise.then(async (res) => {
+    const stringIterator = mapIterable(
+      readableStreamToAsyncIterable(res.body as any),
```
KATT, line 86: Not sure what is different from our tests in tupleson, but `res.body` doesn't have `.getReader()`


@Sheraff - any idea?

packages/client/src/links/httpTuplesonLink.ts
```diff
@@ -0,0 +1,144 @@
...
+      (v) => textDecoder.decode(v as any),
+    );
+
+    type FIXME = any;
+
+    const output = await opts.parseAsync<FIXME>(stringIterator);
```
KATT, line 92: We should probably make `parseAsync` work with readable streams 🤷

packages/server/src/http/resolveHTTPResponse.ts
```diff
@@ -453,3 +482,174 @@ export async function resolveHTTPResponse<
...
+      procedureCalls.map((p) =>
+        p.then((result) => {
+          // transform
+          return transformTRPCResponse(router._def._config, result);
+        }),
+      ),
+    );
+
+    // TODO - maybe tupleson should return a ReadableStream??
+    const stream = asyncIterableToReadableStream(tson);
```
KATT, line 621: We should probably make tson return a stream

packages/client/src/links/httpTuplesonLink.ts
Sheraff: It depends whether it's a response from Web API or a response from Node IIRC. It's been a while so I don't really remember but i had left these comments in https://github.com/trpc/trpc/blob/main/packages/client/src/links/internals/parseJSONStream.ts

```ts
  /**
   * As given by `(await fetch(url)).body`
   */
  readableStream: NodeJS.ReadableStream | WebReadableStreamEsque;
```

```ts
  // we handle 2 different types of streams, this if where we figure out which one we have
  if ('getReader' in readableStream) {
    await readStandardChunks(readableStream, onChunk);
  } else {
    await readNodeChunks(readableStream, onChunk);
  }
```

```ts
/**
 * Handle NodeJS stream
 */
function readNodeChunks(
  stream: NodeJS.ReadableStream,
  onChunk: (chunk: Uint8Array) => void,
) {
  return new Promise<void>((resolve) => {
    stream.on('data', onChunk);
    stream.on('end', resolve);
  });
}

/**
 * Handle WebAPI stream
 */
async function readStandardChunks(
  stream: WebReadableStreamEsque,
  onChunk: (chunk: Uint8Array) => void,
) {
  const reader = stream.getReader();
  let readResult = await reader.read();
  while (!readResult.done) {
    onChunk(readResult.value);
    readResult = await reader.read();
  }
}
```

KATT: Nice, the nodejs one is an async iterable so i could just do `for await (....)`

packages/server/src/adapters/node-http/nodeHTTPRequestHandler.ts
```diff
@@ -123,14 +79,24 @@ export async function nodeHTTPRequestHandler<
...
+
+    if (!opts.res.statusCode || opts.res.statusCode === 200) {
+      opts.res.statusCode = res.status;
+    }
+    opts.res.statusMessage = res.statusText;
+    for (const [key, value] of res.headers.entries()) {
+      opts.res.setHeader(key, value);
+    }
+
+    if (res.body) {
```
Akxe, line 94: A check of `Symbol.asyncIterator in res.body`, would at least assert that the body is an `AsyncIterable`.

packages/server/src/http/internals/types.ts
```diff
@@ -9,8 +9,31 @@ import { TRPCResponse } from '../../rpc';
...
 
+/**
+ * @deprecated use `Headers`
+ */
 export type HTTPHeaders = Dict<string[] | string>;
 
+export function httpHeadersToFetchHeaders(headers: HTTPHeaders): Headers {
+  const newHeaders = new Headers();
+  for (const [key, value] of Object.entries(headers)) {
+    if (!value) {
```
Akxe, line 20: I am not sure about this check... what if someone uses the presence of a header rather than its value... people are wierd

packages/tests/server/tupleson.test.ts
```diff
@@ -402,6 +403,65 @@ describe('with transformer', () => {
...
+describe.todo('with tupleson transformer', () => {
+  const orderedResults: number[] = [];
+  const tuplesonOptions: TsonOptions = {
+    nonce: () => '__tson',
+    types: [tsonBigint],
+  };
+  const ctx = konn()
+    .beforeEach(() => {
+      const t = initTRPC.create({
+        transformer: 'tupleson', // <------- help, idk if this is the right thing
```
Akxe, line 415: To me, it depends, if `tupleson` is the default anyway, the string is fine, if not I would opt for a reference from the `tupleson` package in order to make it tree shakable

The tree shake will, however, only fork if the default is different from `tupleson`, or if this is a required property and thus none is the default...

helmturner: > To me, it depends, if `tupleson` is the default anyway, the string is fine, if not I would opt for a reference from the `tupleson` package in order to make it tree shakable
> 
> The tree shake will, however, only fork if the default is different from `tupleson`, or if this is a required property and thus none is the default...

@Akxe
We've been leaning towards making Tupleson *the* transformer moving forward, deprecating the need for this field. It would make TSON a hard requirement for future implementations (but we maintain backward compat for a bit).

If users want to customize the transform logic, they do so by configuring TSON (which allows an alternative to the default `JSON` from stringifying/parsing).

Essentially, we would be moving the 'transformer' concept to be an implementation detail of Tupleson, instead of an implementation detail of tRPC.

That was an idea, at least. Thoughts?

Akxe: I do like the idea of using TSON. My point was merely informative :)

Akxe: I do not want a string to be there. It means that even if you were to replace it, the TSON dependency is forced upon the consumer. 

From my point of view; no choice or choice in the form of a `function` or a `Class`.

Why string? What would you put there otherwise?

XantreGodlike:
@KATT What are the progress on it? I think streaming is really useful

KATT:
> @KATT What is a progress on it? I think streaming is really useful

@XantreGodlike it's not high on the priority list for us right now as we don't _need it_ ourselves. Our main priority at the moment is finishing the breaking changes we have for v11.

If you work for a company that would like to fund the development of this, we may reconsider, feel free to contact us if so.

XantreGodlike:
> > @KATT What is a progress on it? I think streaming is really useful
> 
> @XantreGodlike it's not high on the priority list for us right now as we don't _need it_ ourselves. Our main priority at the moment is finishing the breaking changes we have for v11.
> 
> If you work for a company that would like to fund the development of this, we may reconsider, feel free to contact us if so.

Nope, it's for hobby purposes

JonathonRP:
@KATT any progress on this vs unstable_streamingLink?

rdewolff:
Looking at the stream support 👀

lacherogwu:
Looking forward for this as its very useful for download big files

hansoksendahl:
Looking forward to this for LLM applications so I can stream the LLM output without having to write a separate end-point through the framework I am using.

lrwlf:
Is there anything we can help? Looking forward to be merged soon

PR #4916, opened 2023-10-13
Rename all useContext instances to useUtils

Nick-Lucas:
useContext is a special hook name in react, and all our types and variables call this value "utils", both of which can cause confusion.

Deprecating useContext and adding useUtils makes a lot of sense (to me) to tidy up this part of the API for v11

## 🎯 Changes

## ✅ Checklist


packages/react-query/src/shared/proxy/utilsProxy.ts
```diff
@@ -215,7 +215,7 @@ type DecorateProcedure<
 type DecorateRouter = {
   /**
    * Invalidate the full router
-   * @link https://trpc.io/docs/v10/useContext#query-invalidation
+   * @link https://trpc.io/docs/v10/useUtils#query-invalidation
```
juliusmarminge, line 218: I know we dont have docs for v11 yet but this link will be broken?

v10/useUtils

KATT: we shouldn't change v11 docs yet because we haven't generated a new version for v11, but good point

Nick-Lucas:
Closing for now as there's not a strong consensus that this is worth it and the right way forward

PR #4932, opened 2023-10-17
refactor: rename and deprecate proxy named stuff

juliusmarminge:
Part of https://github.com/trpc/trpc/issues/3496

Removes `Proxy` from names, re-exports old name marked as deprecated

went a bit crazy and did it for our tests too 😅


examples/deno-deploy/src/client.ts
juliusmarminge, line 1: idk how to make deno resolve local package... reverting here

kamilogorek: You need to update import maps for that. There's no "nice" way to make it work locally for monorepo _and_ remotely at the same time without it. It requires docs to be updated to reflect that.

Also not from that example readme is not true anymore:
> Note: Deno Deploy does https://github.com/denoland/deploy_feedback/issues/314.

KATT (approved):
If we wanna make sure the old alias works we could omit changing any tests as part of this

KATT:
((you can do something like `git checkout origin/main packages/tests/server/**/*`))

juliusmarminge:
> If we wanna make sure the old alias works we could omit changing any tests as part of this

You can see all the tests passing here

https://github.com/trpc/trpc/actions/runs/6549848029/job/17787672965

PR #4946, opened 2023-10-21
feat(server): allow a mix of text and files with fileUploadHandler

jussisaurio:
## 🎯 Changes

…dler

Sending a FormData object with a mix of files and text fields was not possible when using `experimental_createFileUploadHandler`, because the handling of all fields was baked into the `UploadHandler` business logic, and logic for handling non-file fields was only implemented for the memory upload handler. Ergo, the file upload handler just ignored all fields that were not files.

This PR does a few things:

1. Moves the logic for handling non-file fields to `parseMultipartFormData`, and only file fields are passed to the upload handler. Now non-file fields are handled identically regardless of upload handler type.
2. Like with other body parsers, adds `maxBodySize` parameter to the function `parseMultipartFormData` that defaults to Infinity. Throws when combined size of all fields and files exceeds the maximum and unlinks any files that were already uploaded.
3. Exposes `NodeOnDiskFile` as `experimental_NodeOnDiskFile` because `zfd.file()` schema from `zod-form-data` throws when `NodeOnDiskFile` is returned.
4. Exposes `MaxBodySizeExceededError`

## ✅ Checklist

### DIFF
--- a/packages/server/src/adapters/node-http/content-type/form-data/index.ts
+++ b/packages/server/src/adapters/node-http/content-type/form-data/index.ts
@@ -8,23 +8,36 @@
 /**
  * @see https://github.com/remix-run/remix/blob/0bcb4a304dd2f08f6032c3bf0c3aa7eb5b976901/packages/remix-server-runtime/formData.ts
  */
+import * as fs from 'fs/promises';
 import { Readable } from 'node:stream';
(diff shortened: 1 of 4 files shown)


KATT (approved):
Thanks for this!

I don't have much time to scrutinize the code but seeing this is an experimental feature and you have been _very_ thorough with tests and your motivation for this, I trust it

Thanks

PR #4949, opened 2023-10-21
AWS Lambda Batch Streaming

Nick-Lucas:
_nevermind_ API Gateway doesn't support the streaming features, which is like 90% of the reason to use them. Weird decisions from AWS.

https://aws.amazon.com/blogs/compute/introducing-aws-lambda-response-streaming/

https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html

## 🎯 Changes

## ✅ Checklist


Nick-Lucas:
_nevermind_ API Gateway doesn't support the streaming features, which is like 90% of the reason to use them. Weird decisions from AWS.

Issue #544, opened 2021-06-16
Consider SSE to enable server->client http subscriptions


delaneyj:
Reading through the #532 PR it occurred to me that the JSON-RPC notification system only works one way, from client to server.  Most of the time you want it to be from server to client.  Although nothing about the spec says which side is which, in a browser this basically means the browser is the client.

One interesting approach could be to use [Server Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events).  This would allow for stateless load balance-able subscriptions that could work with HTTP2.  From what I'm seeing though I'm not sure if it would technically fall under a JSON-RPC complaint spec; though you probably can treat as a server to client notification.  Nice thing is this is already built into browsers with [polyfills available for node](https://www.npmjs.com/package/eventsource) and most languages.

I think you'd create/cancel as you do currently but the client would also fetch on a new route /subscription/3nxu4ex234.  Know this is separate from the websocket implementation but SSE isn't a well known.  The major downside of the SSE spec as it stands is the text only protocol (compared to binary for websockets) but in a JSON-RPC context the point is moot.

<sub>[TRP-63](https://linear.app/trpc/issue/TRP-63/consider-sse-to-enable-server-client-http-subscriptions)</sub>

## Funding
* You can sponsor this specific effort via a [Polar.sh](https://polar.sh) pledge below
* We receive the pledge once the issue is completed & verified

<a href="https://polar.sh/trpc/trpc/issues/544">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/544/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/trpc/trpc/issues/544/pledge.svg">
</picture>
</a>

KATT:
SSE should be possible to add alongside WebSockets. I prefer WS myself as I don't have to think about max concurrent connection limit and my application doesn't have user numbers that make it hard to scale anyway.

The way our HTTP subscriptions work right now is through long polling and you pass a cursor of the current state and server let's the request hang until there's new data / times out. Inspect https://chat.trpc.io/ to see how the long polling works.

delaneyj:
[...]

PR #4960, opened 2023-10-26
Feat: Streaming

helmturner:
Closes #4530, #4477, #544, #1853

## 🎯 Changes

TBD

## ✅ Checklist


packages/server/src/transformer.ts
```diff
@@ -86,9 +89,113 @@ export type DefaultDataTransformer = CombinedDataTransformer & {
...
+    tsonBigint,
+    tsonDate,
+    tsonRegExp,
+    tsonSet,
+    tsonMap,
+    tsonUndefined,
+    tsonNumberGuard,
+    tsonURL,
+    tsonSymbol,
+  ] as const;
```
KATT, line 161: I am not sure we should be adding a bunch of transformers automatically as it is quite a big change,

I'd rather have only support for Promise and AsyncIterable and let the user define the others

packages/server/src/transformer.ts
```diff
@@ -86,9 +89,113 @@ export type DefaultDataTransformer = CombinedDataTransformer & {
...
+              },
+            ]),
+      ],
+    });
+  }
+
+  return {
+    input: _createTson('input'),
+    output: _createTson('output'),
+  };
```
KATT, line 200: Won't creating the transformer here _and_ in the resolveHttp create two different instances that have different nonce and contains two different "heads"?

Will likely not cause any issues, but it'd be nicer if they'd use the same instance

packages/server/src/http/buildResponse.ts
KATT, line 1: I fail to see why we need this big complex file - I think the complexity here should be handled by tupleson for any streaming responses

E.g. https://github.com/trpc/trpc/blob/dd01f7f873ed1ee820c15c91fee0c19d08b9833b/packages/server/src/http/resolveHTTPResponse.ts#L601-L613

KATT (commented):
Some random thoughts/comments

packages/server/src/http/buildResponse.ts
helmturner: Yep, that's cruft that hasn't been removed yet. Just there for reference.

packages/server/src/transformer.ts
helmturner: So, if I understand, were basically just defaulting to JSON for sync types, since tupleson only tuplefies non-json-serializables? By the way, I missed that little detail for way too long, and may have made some assumptions that rely on ALL serialized values being wrapped in a tuple (with nonce)

packages/server/src/transformer.ts
helmturner: Hmm, did I miss some lazy initialization of the revolvers? I didn't realize I was instantiating twice. Can you link the line?

KATT:
#4911

PR #5003, opened 2023-11-05
fix(next): omit "fetch" from options

withyellow:
## 🎯 Changes

Use `fetch` from options when it's defined.
In the current implementation, `opts.fetch` it typed but never gets used.

## ✅ Checklist

### DIFF
--- a/packages/next/src/app-dir/links/nextHttp.ts
+++ b/packages/next/src/app-dir/links/nextHttp.ts
@@ -1,4 +1,5 @@
 import {
+  getFetch,
   httpBatchLink,
   HTTPBatchLinkOptions,
   httpLink,
@@ -33,12 +34,12 @@ export function experimental_nextHttpLink<
       const revalidate = requestRevalidate ?? opts.revalidate ?? false;
 
       const linkFactory = opts.batch ? httpBatchLink : httpLink;
-      const fetchFn = opts.fetch ?? fetch;
+
       const link = linkFactory({
         headers: opts.headers as any,
         url: opts.url,
         fetch: (url, fetchOpts) => {
-          return fetchFn(url, {
+          return getFetch(opts.fetch)(url, {
             ...fetchOpts,
             // cache: 'no-cache',
             next: {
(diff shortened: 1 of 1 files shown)


packages/next/src/app-dir/links/nextHttp.ts
```diff
@@ -33,11 +33,12 @@ export function experimental_nextHttpLink<
       const revalidate = requestRevalidate ?? opts.revalidate ?? false;
 
       const linkFactory = opts.batch ? httpBatchLink : httpLink;
+      const fetchFn = opts.fetch ?? fetch;
       const link = linkFactory({
         headers: opts.headers as any,
         url: opts.url,
         fetch: (url, fetchOpts) => {
-          return fetch(url, {
+          return fetchFn(url, {
```
kamilogorek, line 41: There's `getFetch` util from `@trpc/client` that you can use here.

juliusmarminge: We need next’s polyfilled fetch in order to use their cache system though right? So I dont see why you would provide a custom fetch function when using this link

withyellow:
``` typescript
next: {
  revalidate,
  tags: [cacheTag],
}
```

Maybe take more control of the `next` params.

Or just omit  `fetch` from `opts`.

withyellow:
The main problem is that there's a `opts.fetch` type definition now, but it's not really being used in requests, which can be misleading. I think we should either use `opts.fetch` or omit it from `opts`.

packages/next/src/app-dir/links/nextHttp.ts
juliusmarminge: I’d say omit it from types, unless you can provide a concrete example of when you’d want to overwrite the next options (as modifying the cachetag will have to be changed in the revalidate() too which is bound to the client and not the link)

withyellow: Done.

juliusmarminge (approved):

PR #5017, opened 2023-11-09
fix(server): inference fix for inputs with middleware

KATT:
Closes #

## 🎯 Changes

## ✅ Checklist


packages/tests/server/regression/issue-5017-inference-middleware.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      type Output = AppRouterOutputs['str'];
+      expectTypeOf<Input>().toBeString();
+      expectTypeOf<Output>().toBeString();
+    }
+    {
+      type Input = AppRouterInputs['strWithMiddleware'];
+      //   ^?
+      type Output = AppRouterOutputs['strWithMiddleware'];
+      expectTypeOf<Input>().toBeString();
+      expectTypeOf<Output>().toBeString();
```
KATT, line 40: This is the one that's failing

jussisaurio: I think the proper solution would be a sane implementation of `Overwrite` that only does things you'd expected. Example pseudo:

```
Given A and B,

- If they are both objects, overwrite A's keys with B
- If only A is object, return A
- If only B is object, return B
- If A extends any, return A
- If B extends any, return B
- Otherwise return never
```

But are there any other practical usecases for Overwrite?

jussisaurio: I think it makes more sense for an intuitive meaning of "Overwrite" so that key-wise overwrite happens when both are objects, and otherwise `TWith` just replaces `TType` entirely.

PR #5026, opened 2023-11-09
chore(v11): remove some deprecated stuff

juliusmarminge:
Relates to #3496

### DIFF
--- a/packages/react-query/src/server/ssgProxy.ts
+++ b/packages/react-query/src/server/ssgProxy.ts
@@ -17,7 +17,7 @@ import {
   AnyRootConfig,
   AnyRouter,
   callProcedure,
-  ClientDataTransformerOptions,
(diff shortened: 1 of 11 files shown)


packages/react-query/src/server/ssgProxy.ts
```diff
@@ -42,7 +42,7 @@ import {
 interface CreateSSGHelpersInternal<TRouter extends AnyRouter> {
   router: TRouter;
   ctx: inferRouterContext<TRouter>;
-  transformer?: ClientDataTransformerOptions;
+  transformer?: DataTransformerOptions;
```
juliusmarminge, line 45: I'm not completely sure of the differences between all the different DataTransformer types to be fair, so this should probably be validated @KATT

KATT (approved):

PR #5067, opened 2023-11-20
fix(server): fetch adapter endpoint path with trailing slash causes 404

maddsua:
Extends [PR 4893](https://github.com/trpc/trpc/pull/4893)

This was discussed in private, but just for the reference what is it all about:

Having a trailing slash in endpoint path caused 404's.

Now it doesn't =)

Closes #4893

### DIFF
--- a/packages/server/src/adapters/fetch/fetchRequestHandler.ts
+++ b/packages/server/src/adapters/fetch/fetchRequestHandler.ts
@@ -20,7 +20,7 @@ export async function fetchRequestHandler<TRouter extends AnyRouter>(
   };
 
   const url = new URL(opts.req.url);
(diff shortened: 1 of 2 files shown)


KATT (dismissed):

juliusmarminge (approved):

KATT (approved):

tclain-pillar:
Seems this would warrant talking to downstream communities.

We use the hono (very thin wrapper around the fetch adapter) (that use trailing comma) for instance and this broke our production release gloriously (range update on CI :-( ) for a couple of hours.

we had to migrate to aws lambda adpter in emergency, but this still seems broken on my test locally.

tclain-pillar:
Currently the only thing working is this local patch for us:

```ts
  const path = url.pathname.slice(
    opts.endpoint.length + (opts.endpoint.endsWith("/") ? 1 : 2)
  );
```

KATT:
Feel free to do a PR, this PR should guide on where to look :)

PR #5072, opened 2023-11-21
chore: run `pnpm prune`

KATT:
Closes #

## 🎯 Changes

ran `pnpm dedupe`


KATT:
Some subtle breaking change within miniflare, can't spend more time on it

PR #5088, opened 2023-11-23
chore: tweak autofixer & reduce test flakiness

KATT:
Closes #

## 🎯 Changes

- No double job declaration
- Only run autofix if any linting fails
- Add retries to flaky playwright


examples/next-prisma-starter/vitest.config.ts
```diff
@@ -8,5 +8,6 @@ export default defineConfig({
     alias: {
       '~/': fileURLToPath(new URL('./src/', import.meta.url)),
     },
+    setupFiles: ['dotenv/config'],
```
KATT, line 11: Vitest failed **sometimes** based on bad env vars

Not sure how it ever worked

PR #5099, opened 2023-11-23
feat(server + client): add method override option

### LINKED ISSUE #3910: feat: add HTTP method override option

Make it possible to handle biiiiig input payloads on queries by allowing queries to be sent through HTTP `POST`.

- `allowOverride: boolean` on HTTP adapters
  - With this option enabled, we can ignore the checks that make sure that queries are called with GET and mutations as POST
- A custom link that allows you to remap all queries on the upstream as a `POST` _or_ allow `httpBatchLink` to have some option for it

#3590

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

<sub>[TRP-58](https://linear.app/trpc/issue/TRP-58/feat-add-http-method-override-option)</sub>

## Funding

<a href="https://polar.sh/trpc/trpc/issues/3910">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/3910/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

KATT:
Closes #3910

## 🎯 Changes

Add method override.

## 🚧  Checklist

 - [x] Add some failing tests
 - [x] Add some more tests
 - [x] Add working implementation
   - [x] Client
   - [x] Server
 - [x] Update docs
 - [ ] clarify necessity of _procedureType query parameter (see PR comments)
 - [ ] check formDataRequester (see PR comment)

### DIFF
--- a/packages/client/src/links/httpFormDataLink.ts
+++ b/packages/client/src/links/httpFormDataLink.ts
@@ -1,5 +1,5 @@
 import { httpLinkFactory } from './httpLink';
-import { GetBody, httpRequest, Requester } from './internals/httpUtils';
+import { GetBody, httpRequest, isPOST, Requester } from './internals/httpUtils';
 
 const getBody: GetBody = (opts) => {
   if (!('input' in opts)) {
@@ -12,7 +12,7 @@ const getBody: GetBody = (opts) => {
 };
 
 const formDataRequester: Requester = (opts) => {
-  if (opts.type !== 'mutation') {
+  if (!isPOST(opts)) {
     // TODO(?) handle formdata queries
     throw new Error('We only handle mutations with formdata');
   }
--- a/packages/tests/server/methodOverride.test.ts
+++ b/packages/tests/server/methodOverride.test.ts
@@ -174,32 +174,6 @@ test('everything as POST', async () => {
   expect(req).toMatchInlineSnapshot();
 });
 
-test('use method override when not allowed', async () => {
-  const t = await setup({
-    methodOverride: {
-      enabled: false,
-    },
-    linkOptions: {
-      unstable_methodOverride: 'POST',
-    },
-  });
-
-  const err = await waitError(() =>
-    t.client.q.query({
-      who: 'test1',
-    }),
-  );
-
-  expect(err).toMatchInlineSnapshot();
-
-  expect(t.requests).toHaveLength(1);
-  const req = t.requests[0]!;
-
-  expect(req.method).toBe('POST');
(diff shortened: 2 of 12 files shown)


jussisaurio:
Had a go at this. Probably needs more tests still. Ended up with `_procedureType` for the query parameter name.

jussisaurio:
Sort of skeptical whether the `GET` override would ever be used... might simplify things to call the config option `alwaysUsePOST: boolean` or something

packages/client/src/links/httpFormDataLink.ts
```diff
@@ -12,7 +12,7 @@ const getBody: GetBody = (opts) => {
 };
 
 const formDataRequester: Requester = (opts) => {
-  if (opts.type !== 'mutation') {
+  if (!isPOST(opts)) {
```
jussisaurio, line 15: I didn't check whether this makes sense btw

KATT:
I realised that could just skip the override query param too since we can never have collisions on the same key for queries or mutations

jussisaurio:
> I realised that could just skip the override query param too since we can never have collisions on the same key for queries or mutations

Do you mean we skip the query param and if `methodOverride: true` on the server we just try to find the procedure in both `queries` and `mutations` ?

packages/tests/server/methodOverride.test.ts
```diff
@@ -0,0 +1,439 @@
...
+    await t.client.q.query({
+      who: 'test1',
+    }),
+  ).toBe('hello test1');
+
+  expect(t.requests).toHaveLength(1);
+  const req = t.requests[0]!;
+
+  expect(req.method).toBe('POST');
+  expect(req.url).toContain('_procedureType=query'); // maybe this should be _type=query? or instead?
```
KATT, line 174: > > I realised that could just skip the override query param too since we can never have collisions on the same key for queries or mutations
> 
> Do you mean we skip the query param and if `methodOverride: true` on the server we just try to find the procedure in both `queries` and `mutations` ?

yeah, I mean could just skip this part :)

jussisaurio: yep. "downside" is that then you have to check the `def` for both `query` and `mutation` but that's not a huge deal, just a bit less explicit

an alternative API to this whole thing would be to add `httpMethod?: 'GET' | 'POST'` to `ProcedureOptions`, since if we'll try to find the procedure both as a query and mutation on the server anyway, `methodOverride` doesn't need to be a global client-side setting and can be done per procedure call

PR #5101, opened 2023-11-24
fix: remove more internal exports

KATT:
Closes #

## 🎯 Changes

Remove a bunch of external imports

ref #3496:
> [ ] Look through `@internal`-marked types/fns and see if we can avoid exposing them in the bundle altogether
> * Overall, we should try to get rid of `export * from X` and have index files that only export exactly what we want
> * Once we've done this, we need to look at third party libraries using trpc and make sure that they have a migration path - maybe we start exporting some stuff specifically for plugins

## ✅ Checklist


packages/client/src/createTRPCUntypedClient.ts
```diff
@@ -1,4 +1,4 @@
-import { AnyRouter } from '@trpc/server';
+import { AnyRouter } from '@trpc/server/unstableInternalsExport';
```
juliusmarminge, line 1: Should be public imo

packages/client/src/links/wsLink.ts
```diff
@@ -7,8 +6,13 @@ import {
   TRPCRequestMessage,
   TRPCResponseMessage,
 } from '@trpc/server/rpc';
-import { MaybePromise } from '@trpc/server/unstableInternalsExport';
-import { transformResult } from '../shared/transformResult';
+import {
+  AnyRouter,
+  inferRouterError,
```
juliusmarminge, line 11: Public?

packages/next/src/app-dir/links/nextCache.ts
```diff
@@ -1,8 +1,12 @@
 // import "server-only";
 
 import { TRPCClientError, TRPCLink } from '@trpc/client';
-import { AnyRouter, callProcedure, inferRouterContext } from '@trpc/server';
 import { observable } from '@trpc/server/observable';
+import {
+  AnyRouter,
+  callProcedure,
```
juliusmarminge, line 7: We should have a public way of doing this as it's nice for RSC

packages/client/src/createTRPCUntypedClient.ts
KATT: Should we rename it to `TRPCRouter`?

KATT: 1. rename it
2. deprecate and alias old

packages/next/src/app-dir/links/nextCache.ts
KATT: Don't you use `router.createCaller()` for that?

KATT: That said, we should deprecate `router.createCaller` and create a `createCaller(x)` instead

packages/server/src/index.ts
```diff
@@ -16,3 +16,11 @@ export type {
...
   inferAsyncReturnType,
 } from './types';
+
+export type {
+  AnyRouter as AnyTRPCRouter,
+  /**
+   * @deprecated use `AnyTRPCRouter` instead
+   */
+  AnyRouter,
+} from './core';
```
KATT, line 26: cc @juliusmarminge

packages/client/src/index.ts
KATT, line 1: This still needs cleanup

packages/next/src/app-dir/links/nextCache.ts
juliusmarminge: createCaller doesnt play very nice as a client if you have async context.

createTRPCClient with custom link using callProcedure is the best I've found

packages/client/src/createTRPCUntypedClient.ts
juliusmarminge: Yea sounds good

packages/client/src/links/wsLink.ts
KATT: Can change it if people ask for it

packages/next/src/app-dir/links/nextCache.ts
KATT: Could you add an example of this within the tRPC repo somewhere so I can understand it better?

juliusmarminge: This shows pretty well why it's not the best ergonomics when using createCaller.

https://github.com/t3-oss/create-t3-turbo/blob/eae39060ee1ea1bad06842c3a3ce7a80463405c5/apps/nextjs/src/trpc/server.ts#L63-L72

The two accomplishes the exact same thing but the client doesn't require nested awaits.

We could probably make something that's even better though, like allowing createCaller to accept a function, and then that'd be called at the same time as the actual procedure. Since you can wrap it in cache, it'd still be de-duped.

```ts
const trpc = appRouter.createCaller(createContext);

const post = await trpc.post.all();
```

juliusmarminge: although you'd still miss things like having the loggerLink etc

KATT:
I think #5101 is a better approach - closing, might re-open later

PR #5107, opened 2023-11-26
ref: Change all export * to specific exports

kamilogorek:
This PR transforms all `export * from 'x'` to listing all exports explicitly.
The end goal is to reduce the overall API surface, and this is just a pre-required step to make it easier to reason about all exports and see at hand what exactly is being exported.

It exports _everything_ that was already exported. No more, no less, and this was the exact goal.


sachinraja (dismissed):

juliusmarminge (approved):

KATT (commented):
Could you get it up to date?

Maybe we should rename so nothing is called `index.ts`? 🤔 The index files only exists as index files for the sake of entrypoints in the package.json but we never want to import from them internally

kamilogorek:
Is this still relevant? if so, I can get it updated 👀

juliusmarminge:
> Is this still relevant? if so, I can get it updated 👀

yes, that'd be great. i did for the new core package but didn't touch the other ones

PR #5142, opened 2023-12-12
feat: bun adapter

### LINKED ISSUE #5140: feat: Native Bun.serve adapter

A trpc adapter to integrate with Bun's fast serve for both http and web sockets 

I'm new to both Bun and TRPC, but something like this:

```ts
import * as trpcBun from '@trpc/server/adapters/bun';

const server = Bun.serve(trpcBun({ port: 3000 }));

console.log(`Listening on localhost: ${server.port}`);
```

Using any of the other adapters will work in Bun, however the performance claims by using Bun.serve are something that may be worth looking into.

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

## Funding

<a href="https://polar.sh/trpc/trpc/issues/5140">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/5140/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

cah4a:
Closes #5140

It's hard to build the thing correctly.
Bun types mess type-checking when included globally.
Or not emitted to d.ts if ignored via `// @ts-ignore`.

Does somebody have ideas?

### DIFF
--- a/packages/server/src/adapters/bun/createBunHttpHandler.ts
+++ b/packages/server/src/adapters/bun/createBunHttpHandler.ts
@@ -1,0 +1,24 @@
+import type {Server} from "bun";
+import {fetchRequestHandler} from "../fetch";
+import type { AnyRouter, inferRouterContext } from "../../core";
+import type { HTTPBaseHandlerOptions } from "../../http";
+
+export type BunHttpHandlerOptions<TRouter extends AnyRouter> = HTTPBaseHandlerOptions<TRouter, Request> & {
+    endpoint: string;
+    createContext?: (opts: { req: Request }) => inferRouterContext<TRouter> | Promise<inferRouterContext<TRouter>>;
+};
+
+export function createBunHttpHandler<TRouter extends AnyRouter>(opts: BunHttpHandlerOptions<TRouter>) {
+    return (request: Request, server: Server) => {
+        if (
+            server.upgrade(request, {data: {req: request}})
+        ) {
+            return new Response(null, {status: 101});
+        }
+
+        return fetchRequestHandler({
+            ...opts,
+            req: request,
+        });
+    }
+}
--- a/packages/server/src/adapters/bun/createBunWSHandler.ts
+++ b/packages/server/src/adapters/bun/createBunWSHandler.ts
@@ -1,0 +1,279 @@
+import type {ServerWebSocket, WebSocketHandler} from "bun";
+import {
+    JSONRPC2,
+    parseTRPCMessage,
+    TRPCClientOutgoingMessage,
+    TRPCResponseMessage,
+} from "../../rpc";
+import {
+    AnyRouter,
+    callProcedure,
+    getTRPCErrorFromUnknown,
+    inferRouterContext,
+    TRPCError,
+} from "../..";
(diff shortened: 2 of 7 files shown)


packages/server/src/adapters/bun/createBunWSHandler.ts
```diff
@@ -0,0 +1,279 @@
...
+                            error,
+                            type: "unknown",
+                            path: undefined,
+                            input: undefined,
+                            ctx,
+                        }),
+                    });
+
+                    // close in next tick
+                    (global.setImmediate ?? global.setTimeout)(() => {
```
Jarred-Sumner, line 84: bun does implement both `setImmediate` and `setTimeout`

packages/server/src/adapters/bun/createBunWSHandler.ts
```diff
@@ -0,0 +1,279 @@
...
+            client.data.handleRequest = async (msg: TRPCClientOutgoingMessage) => {
+                const {id, jsonrpc} = msg;
+                /* istanbul ignore next -- @preserve */
+                if (id === null) {
+                    throw new TRPCError({
+                        code: "BAD_REQUEST",
+                        message: "`id` is required",
+                    });
+                }
+                if (msg.method === "subscription.stop") {
```
Jarred-Sumner, line 114: by the way, bun has builtin support for pub/sub. it might be able to be used here. it would be more efficient than a userland implementation, assuming it's feature complete enough to be used

https://bun.sh/docs/api/websockets#pub-sub

cah4a: Don't see how this could help.
Every connection should have its own topic, which doesn't sound for me.

packages/server/src/adapters/bun/createBunHttpHandler.ts
```diff
@@ -0,0 +1,24 @@
...
+    endpoint: string;
+    createContext?: (opts: { req: Request }) => inferRouterContext<TRouter> | Promise<inferRouterContext<TRouter>>;
+};
+
+export function createBunHttpHandler<TRouter extends AnyRouter>(opts: BunHttpHandlerOptions<TRouter>) {
+    return (request: Request, server: Server) => {
+        if (
+            server.upgrade(request, {data: {req: request}})
+        ) {
+            return new Response(null, {status: 101});
```
brandon-schabel, line 16: based on buns docs, I believe it automatically returns 101 if the connection was successful:

https://bun.sh/guides/websocket/upgrade

From your point of view do you think this if statement is required?

cah4a: By my point of view this statement is required by type signature specified [here](https://github.com/oven-sh/bun/blob/b6f60680f699f63c37e416b54ef1ebd3a14aaff5/packages/bun-types/bun.d.ts#L2105).

cah4a:
https://github.com/cah4a/trpc-bun-adapter

juliusmarminge:
> https://github.com/cah4a/trpc-bun-adapter

Aweesome! Feel free to add this to https://trpc.io/awesome

PR #5150, opened 2023-12-14
feat: allow trpc with fastify to use only one path

vafanassieff:
## 🎯 Changes

`@fastify/websocket` create a single `WebSocketServer` for the `fastify` instance.

In the `fastify` documentation they suggest that if you want to use multiple websocket endpoint you should do like this :

```
  fastify.get('/', { websocket: true }, (connection /* SocketStream */, req /* FastifyRequest */) => {
    connection.socket.on('message', message => {
      connection.socket.send('hi from server')
    })
  })
```

Since `trpc` use the `wss.on('connection')` handler, it catches all the websocket connections and not only the one on `trpc` prefix.

My goal was to use the websocket for `/trpc` and also for `/`.

Without this all the request from `/` would be passed to the trpc handler.

## ✅ Checklist


packages/server/src/adapters/ws.ts
```diff
@@ -38,10 +39,14 @@ export type CreateWSSContextFnOptions = NodeHTTPCreateContextFnOptions<
 export function applyWSSHandler<TRouter extends AnyRouter>(
   opts: WSSHandlerOptions<TRouter>,
 ) {
-  const { wss, createContext, router } = opts;
+  const { wss, createContext, router, prefix } = opts;
 
   const { transformer } = router._def._config;
   wss.on('connection', async (client, req) => {
+    if (prefix && req.url !== prefix) {
```
juliusmarminge, line 46: should this be strict equality? can the url never have query params or similar?

vafanassieff: Indeed `req.url` contain the query params, we could do `req.url.startsWith` or `req.url.include`

juliusmarminge:
Converted to draft. Feel free to mark ready for review when done

juliusmarminge:
Would be nice with a test for this as well

vafanassieff:
I added a test @juliusmarminge, let me know if you have more questions

KATT (changes requested):
Please change base branch to `next` - we're not doing changes on v10 except bug fixes

vafanassieff:
@KATT It's done

KATT (approved):

Issue #5194, opened 2023-12-20
bug: TRPCError class, public readonly cause?: Error; this line should have override keyword according to error


me-imfhd:
### Provide environment information

  System:
    OS: Linux 6.2 Ubuntu 22.04.3 LTS 22.04.3 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
    Memory: 1.40 GB / 7.66 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 20.6.1 - ~/.nvm/versions/node/v20.6.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v20.6.1/bin/yarn
    npm: 10.1.0 - ~/.nvm/versions/node/v20.6.1/bin/npm
    pnpm: 8.8.0 - ~/.nvm/versions/node/v20.6.1/bin/pnpm
  Browsers:
    Chrome: 119.0.6045.199
  npmPackages:
    @turbocell/api: workspace:* => 0.0.0 
    @turbocell/db: workspace:* => 0.0.0 
    @turbocell/eslint-config: workspace:* => 0.0.0 
    @turbocell/trpc-express: workspace:* => 0.0.0 
    @turbocell/tsconfig: workspace:* => 0.0.0 
    @types/bcrypt: ^5.0.2 => 5.0.2 
    @types/body-parser: ^1.19.3 => 1.19.5 
    @types/cookie-parser: ^1.4.4 => 1.4.6 
    @types/cors: ^2.8.14 => 2.8.17 
    @types/express: ^4.17.19 => 4.17.21 
    @types/express-session: ^1.17.8 => 1.17.10 
    @types/formidable: ^3.4.3 => 3.4.5 
    @types/jsonwebtoken: ^9.0.3 => 9.0.5 
    @types/node: 20.8.0 => 20.8.0 
    @types/swagger-ui-express: ^4.1.6 => 4.1.6 
    bcrypt: ^5.1.1 => 5.1.1 
    body-parser: ^1.20.2 => 1.20.2 
    concurrently: ^8.2.2 => 8.2.2 
    cookie-parser: ^1.4.6 => 1.4.6 
    cors: ^2.8.5 => 2.8.5 
    dotenv: ^16.3.1 => 16.3.1 
    esbuild: ^0.14.38 => 0.14.54 
[...]

PR #5195, opened 2023-12-20
fix(server + client): add `noImplicitOverride` & add explicit overrides in `TRPCError` / `TRPCClientError`

me-imfhd:
…for typescript above v4.3

Closes #https://github.com/trpc/trpc/issues/5194

Please let me know, if i am doing something i should not, or there is any better way to fix this i would do so on being asked, i am just trying to fix this issue, so users can use this on config where noImplicitOverride is set to true in thier codebase

## 🎯 Changes
added override keyword which is supposed to fix this issue
developer can now add [noImplicitOverride](https://www.typescriptlang.org/tsconfig/noImplicitOverride.html) compiler option on and it should not throw the error the issue is talking about
it is a feature/bug for typescript versions above 4.3, if user has enabled the noimplicitOverride keyword, he might get to face this error

## ✅ Checklist


kamilogorek:
Tests are failing due to `This member cannot have an 'override' modifier because it is not declared in the base class 'Error'.`, because for some reason, our tests `tsconfig.json` differs from the ones we use for client or server.
Packages use either `esnext` or es2023`, yet tests override lib to use `es2020`.

@trpc/trpc-core does anyone know the historical reason for this? If we decide to change it to align with the build config, we also need to remove 2 `ts-expect-error` from `packages/tests/server/TRPCError.test.ts` on lines `37` and `66`.

me-imfhd:
@kamilogorek any solution you propose i should get tests passing?

KATT:
> Tests are failing due to `This member cannot have an 'override' modifier because it is not declared in the base class 'Error'.`, because for some reason, our tests `tsconfig.json` differs from the ones we use for client or server. Packages use either `esnext` or es2023`, yet tests override lib to use `es2020`.
> 
> @trpc/trpc-core does anyone know the historical reason for this? If we decide to change it to align with the build config, we also need to remove 2 `ts-expect-error` from `packages/tests/server/TRPCError.test.ts` on lines `37` and `66`.

cc @kamilogorek I added a `@ts-ignore` for now and I want to get it merged - if you have an idea how to update the config so it doesn't appear, please submit a PR 🙏 🙏

PR #5209, opened 2023-12-23
docs: Fix cache-control example

tmokmss:
## 🎯 Changes
Reading [this doc](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#directives), I think the sample code for `cache-control` should be inverted to achieve the intended behavior.

> cache request for 1 day + revalidate once every second

## ✅ Checklist


KATT:
I think it's correct as-is. Max-age is the max allowed age for it to be fetched in cache

tmokmss:
ahh I was confused a bit. Thanks for the clarification!

PR #5241, opened 2023-12-27
fix: cache artifcats in entrypoints generation script

juliusmarminge:
ensures turbo cache outputs are updated as new entrypoints are added


KATT (approved):
lgtm

Issue #5004, opened 2023-11-05
feat: publishing AppRouter without hassle


olehmisar:
### Describe the feature you'd like to request

I am using trpc in bunch of projects that are spread across npm packages. E.g., I have a server `my-server` and a client `my-client`. In `my-client` code I am importing `AppRouter` like this:
```ts
import type { AppRouter } from `my-server`;
```
I want to be able to publish `AppRouter` types on NPM without exposing any sensitive information like my server source code or `Context` of the `AppRouter`.

### Describe the solution you'd like to see

When generating .d.ts files via `tsc` or `vite`, .d.ts files contain all imported type definitions. Since `AppRouter` depends on `Context`, and `Context` usually contains all backend services (e.g., `UsersService`, `PaymentService`, etc), all the backend type declarations are also included in .d.ts files.

I see at least two solutions:
- Somehow replace `Context` object with `unknown` when generating .d.ts files
- If `appRouter` is defined in `src/router.ts`, only include `dist/router.d.ts` in package.json `files` field.

I hope trpc provides a recommended way to publish `AppRouter`, so developers do not need to think about it themselves.

### Describe alternate solutions

Don't provide a recommended way of publishing `AppRouter` and leave it to devs.

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

KATT:
#4709 does this - implemented in next major (which you can use today)

olehmisar:
According to the description of the linked PR, context object is still included in the router type(although context is no longer present in procedure types). Is it possible to remove context from .d.ts altogether?

KATT:
Off the top of my head, the `@trpc/client` needs

- error formatter
- transformer (or at least knowing if there is one)

The rest could technically be omitted with regards to the client. 

However, in order to call `router.createCaller()` we'd need the `Context`.... potentially we could have a way to strip out stuff from routers that'd still work to create clients from.


Feel free to hack on this, know it's quite time consuming though, but you have pretty free reign to challenge how we do things seeing that it's a major change upcoming anyway

matannahmani:
is this now available?

deldrid1:
I still don't see a very easy way to export the types to a file and then import them somewhere else.

[...]

PR #5242, opened 2023-12-28
feat: add support for `satisfies` when creating routers

KATT:
Idea for #5004

## 🎯 Changes

Allow using `satisfies` when creating router.

```ts
const appRouter = {
    hello: t.procedure.query(() => 'hello world'),
} satisfies TRPCRouter;
```

- This would prevent context etc from being visible when publishing on npm
- The client only technically really needs the error shape and to know if there's a transformer


juliusmarminge:
Completed in another PR

PR #5259, opened 2024-01-04
feat: add fetchOptions to experimental_nextHttpLink

rmarscher:
With createTRPCNext, I pass a custom fetch to httpBatchLink to enable `mode: 'cors'` and `credentials: 'include'`. The fetch option is currently omitted in `experimental_nextHttpLink`.

I understand that this is an experimental API, so if you just want to close this PR and create an alternative solution, that works. I thought it might be a little more helpful to open a PR rather than an issue/discussion. Thank you!

## 🎯 Changes

Add `fetchOptions` to `NextLinkBaseOptions` which allow overriding the options being passed to fetch.

## ✅ Checklist

^^ Let me know if you'd like me to update any tests or docs.

### DIFF
--- a/packages/next/src/app-dir/links/nextHttp.ts
+++ b/packages/next/src/app-dir/links/nextHttp.ts
@@ -11,6 +11,7 @@ import { generateCacheTag } from '../shared';
 interface NextLinkBaseOptions {
   revalidate?: number | false;
   batch?: boolean;
+  fetchOptions?: RequestInit;
 }
 
 interface NextLinkSingleOptions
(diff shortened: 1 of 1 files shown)


rmarscher:
I think I should probably just close this. It seems like more work is going to need to be done to have isomorphic queries inside components being rendered for SSR.

I think the ideal scenario is to be able to use a query hook inside components with normal client fetch settings and include additional options for how they should be run and cached when they are executed on the server during SSR. If the query hooks are configured to run on the server, the client could use the data from the server without making another immediate request.

I see the NextJS docs mention issues with using fetch with React `use` hooks - so it may require more upstream changes for an isomorphic query hook to be possible. https://nextjs.org/docs/app/building-your-application/data-fetching/fetching-caching-and-revalidating#fetching-data-on-the-client-with-third-party-libraries

PR #5267, opened 2024-01-06
feat: add ability to have a parser as a callback

### LINKED ISSUE #1963: [Feature Request] Add a way to access the 'ctx' in the 'input' definition
I want to transform a _slug_ or _id_ before passing it to the `resolve` function. Currently you cannot access the router `ctx` inside the `input` function which has the downside that you have to do transformations inside the `resolve` function (if the transformation requires the `ctx` object). 

For example: You have a user profile page _(profile/{userId})_ that uses trpc to get the profile data of a user. You pass this _userId_ to trpc and want to convert the _userId_ into a _user object_ by utilizing some kind of `UserRepository` that is available on your `ctx` object. The `UserRepository` is an instance of a service, and cannot be accessed without the `ctx` object as you do not allow singletons. 

By allowing the `input` to access the `ctx` we can utilize Zod's [transform](https://github.com/colinhacks/zod#transform) to convert a _slug_ or _id_ **before** it is passed to the `resolve` function. This way we can keep our `resolve` functions as clean as possible and keep any sort of input validation & transformation inside our `input` function.

According to [the docs](https://trpc.io/docs/router) we can currently pass a _Zod_, _Yup_ or _superstruct_ object as input. I am not aware how this type is checked internally but if they all provide an `object` instead of a `function` we could allow a `function` as input with the `ctx` as an argument. 

The end-result could look like:

```ts
.query('profile', {
  input: ({ ctx }) => 
      yup.object({
          user: yup.number().required(),
      })
      .transform((userId) => ctx.userRepository.findById(userId))
      .refine((user) => !!user, {
           message: 'User could not be found.',
       }),
      
  resolve({ input }) {
    return {
      username: input.user.username,
    };
  },
});

```

lxia1220:
Closes #1963 

Continous from #5234

## 🎯 Changes

 - ParserCustomValidatorEsque signature is changed
 ```ts
.input(({input}) => Parser | value))
.output(({output}) => Parser | value))
```

usage) 
```ts
.input(z.number())
.input(({ctx, input}) => z.number()))
.input(({ctx, input}) => input as string))
 ```
 
## Todo

 - [x] change output cb's value name input -> output
 - [ ] need to clean up
 - [ ] ~codemod for ParserCustomValidatorEsque sigs~

## ✅ Checklist

### DIFF
--- a/packages/server/src/core/internals/getParseFn.ts
+++ b/packages/server/src/core/internals/getParseFn.ts
@@ -2,13 +2,10 @@ import { Parser } from '../parser';
 
 export type ParseFn<TType> = (value: unknown) => Promise<TType> | TType;
 
-export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
+export function getParseFnInner<TType>(
+  procedureParser: Parser,
+): ParseFn<TType> | null {
   const parser = procedureParser as any;
-
-  if (typeof parser === 'function') {
-    // ParserCustomValidatorEsque
-    return parser;
-  }
 
   if (typeof parser.parseAsync === 'function') {
     // ParserZodEsque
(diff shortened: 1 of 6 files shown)


lxia1220:
need to clean up & change output cb's value name `input` -> `output`

PR #5275, opened 2024-01-07
refactor: flatten(?) core package, remove `@trpc/client/shared`

juliusmarminge:
aims to flatten stuff out a bit, better Separation of Concerns (tm), colocate a few stuff that was spread out for (imo) no reason

removes `@trpc/client/shared` entrypoint in favor of exporting transformResult from core lib ( don't think it makes much sense to do this for react-query as it's used by @trpc/next and moving these stuff to core doens't make more sense

### TODO / FIXME

- [x] flatten out observable module, might just slap everything from nested internal to a single file? https://github.com/trpc/trpc/pull/5275/commits/0d4533838a1fdb1c7324d51aab1b3280dff7da1a
- [ ] (maybe) remove `src/index.ts` in favor of more granular export paths (we can import from `@trpc/core/serialize`, `@trpc/core/error`, `@trpc/core/http` etc etc. 
- [x] Look over what needs to be exported. Check what's only used in core and try removing and see if declaration support breaks https://github.com/trpc/trpc/pull/5275/commits/9edd52e3e84aaff86dd3b809672cd3b4ffb26447
- [ ]


juliusmarminge:
Current dependencies on/for this PR:
* `next`
  * **PR #5275** <a href="https://app.graphite.dev/github/pr/trpc/trpc/5275?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/trpc/trpc/5275?utm_source=stack-comment).

packages/core/src/core/inference.ts
juliusmarminge, line 1: this file may not be needed, could put `inferXProcedure` in procedure.ts, `inferRouterX` in router.ts etc etc

KATT, line 1: Doesn't this file create circular dependencies? 🤔

packages/core/package.json
```diff
@@ -44,6 +44,11 @@
       "require": "./dist/index.js",
       "default": "./dist/index.js"
     },
+    "./error": {
+      "import": "./dist/error/index.mjs",
+      "require": "./dist/error/index.js",
+      "default": "./dist/error/index.js"
+    },
```
KATT, line 51: Is this entrypoint needed? Can't we just let treeshaking do its thing?

packages/core/src/core/inference.ts
```diff
@@ -0,0 +1,89 @@
...
+      : TRouterOrProcedure extends AnyProcedure
+      ? TType extends 'input'
+        ? inferProcedureInput<TRouterOrProcedure>
+        : inferTransformedProcedureOutput<
+            TRouter['_def']['_config'],
+            TRouterOrProcedure
+          >
+      : never
+    : never;
+};
```
KATT, line 79: I'm not sure about this `inference.ts`-file.

This "concern" for instance feels like it should be grouped with _procedures_ and not with "inference"

I feel like the domain of which things should be grouped should be "domain"-driven more than that this is the collection of type helpers

makes sense?

packages/next/src/app-dir/create-action-hook.tsx
```diff
@@ -4,7 +4,7 @@ import type {
   TRPCRequestOptions,
 } from '@trpc/client';
 import { createTRPCUntypedClient, TRPCClientError } from '@trpc/client';
-import { transformResult } from '@trpc/client/shared';
+import { transformResult } from '@trpc/core';
```
KATT, line 7: 👌

packages/tests/server/react/polymorphism.subtyped-factory.tsx
```diff
@@ -3,7 +3,7 @@
 //  which can be used with existing components, but has extra data for other use cases
 //
 import type { AnyRootConfig, createBuilder } from '@trpc/core';
-import type { createRouterFactory } from '@trpc/core/router';
+import type { createRouterFactory } from '@trpc/core/core/router';
```
KATT, line 6: should this be
```suggestion
import type { createRouterFactory } from '@trpc/core';
```

packages/core/src/observable/index.ts
```diff
@@ -1,16 +1,10 @@
...
   Observer,
-  OperatorFunction,
-  Subscribable,
-  SubscriptionLike,
   TeardownLogic,
-  UnaryFunction,
   Unsubscribable,
   UnsubscribeFn,
 } from './types';
-export { map, share, tap } from './operators';
```
KATT, line 15: How come you prefer to slap all operators in one file? Separated feels cleaner to me

No strong preference but I sorta just copied https://github.com/ReactiveX/rxjs/tree/master/packages/rxjs/src/internal/operators

KATT (approved):
LGTM. No strong preference on this stuff - if it feels more logical to you, it def will be others. Good work!!

I feel like having a folder called `packages/core/src/core` feels a bit off, I don't know what "core" means. I'm fine just having files in the root `core/src` too

packages/core/src/core/inference.ts
juliusmarminge: Yes, i went back and forth on this one and i think you’re right

packages/core/src/observable/index.ts
juliusmarminge: I didnt feel like the observable/internals felt necessary, and since we only had 3 it made sense to just squash it into a single file but can def revert if you prefer

packages/core/package.json
juliusmarminge: Sort of just followed what you had with observable and rpc and they had their own entrypoint

juliusmarminge:
> I feel like having a folder called packages/core/src/core feels a bit off, I don't know what "core" means. I'm fine just having files in the root core/src too

I’ll revert that commit then. I guess i menat the core stuff that makes up your api (router, procedures, middlewares etc)

packages/core/src/observable/index.ts
KATT: i love the internals-folder being removed, i am neutral towards combining the operators

PR #5112, opened 2023-11-27
RFC: Middleware Appliers for Third Party Middlewares

Nick-Lucas:
A problem we continue to have with the experimental standalone middlewares, and also t.middleware, is that they have caused many bugs with type inference over time, and so compose unreliably. It's not clear that all problems are solvable. 

This is an approach which draws inspiration from other libraries like Tanstack Router, and simply doesn't try to fight the typescript type system by making everything into a function call of known type

I copied over the minimal example project for this prototype so the diff looks bigger than it should, focus on:

* `AcmeMiddleware.ts`: https://github.com/trpc/trpc/pull/5112/files#diff-860c9b23a905a4f6494e0cf370a317bde88b6e702ba3cb26db9ba79115fb1b78
* `trpc.ts`: https://github.com/trpc/trpc/pull/5112/files#diff-9cfcb5afca1d1d2518a9465e6aba2367ea9e39aa0236e19a634a7b1a13854341
* `index.ts`: https://github.com/trpc/trpc/pull/5112/files#diff-ebbbc5e7b43d53fbdc9a56f8f3125a6470d7dcfc7ce4114a4402d63f3e591f99
* the shared `Next` type: https://github.com/trpc/trpc/pull/5112/files#diff-28374e196350161477845434683132162955a507c4d6ff3bebad3774ebb0749d

## 🎯 Changes

What changes are made in this PR? Is it a feature or a bug fix?

## ✅ Checklist

- [ ] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [ ] If necessary, I have added documentation related to the changes made.
- [ ] I have added or updated the tests related to the changes made.

PR #5296, opened 2024-01-10
V11 nick experimentation on work codebase

Nick-Lucas:
Closes #

## 🎯 Changes

## ✅ Checklist

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -11,7 +11,11 @@ export {
 } from './error/TRPCError';
 
 export { initTRPC } from './initTRPC';
-export type { MiddlewareFunction, MiddlewareBuilder } from './middleware';
+export type {
+  AnyMiddlewareFunction,
+  MiddlewareFunction,
+  MiddlewareBuilder,
+} from './middleware';
 export {
   createInputMiddleware,
   createOutputMiddleware,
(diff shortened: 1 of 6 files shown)


packages/core/src/index.ts
```diff
@@ -11,7 +11,11 @@ export {
 } from './error/TRPCError';
 
 export { initTRPC } from './initTRPC';
-export type { MiddlewareFunction, MiddlewareBuilder } from './middleware';
+export type {
+  AnyMiddlewareFunction,
+  MiddlewareFunction,
+  MiddlewareBuilder,
+} from './middleware';
```
Nick-Lucas, line 18: Need at least AnyMiddlewareFunction in order to continue rolling the Middleware Appliers pattern in userland. Though export just the `Next` function type as in #5112 would be the productionised approach.

packages/react-query/src/shared/polymorphism/queryLike.ts
```diff
@@ -20,6 +23,14 @@ export type QueryLike<
...
     opts?: InferQueryOptions<TConfig, TProcedure, any>,
   ) => InferQueryResult<TConfig, TProcedure>;
+
+  useSuspenseQuery: (
+    variables: inferProcedureInput<TProcedure>,
+    opts?: InferQueryOptions<TConfig, TProcedure, any>,
+  ) => UseTRPCSuspenseQueryResult<
+    inferProcedureOutput<TProcedure>,
+    TRPCClientErrorLike<TConfig>
+  >;
```
Nick-Lucas, line 33: Since suspense:true is gone on useQuery now, my own codebase needs this to continue using suspense in polymorphic components.

Longer term I do think we should fine a way to drop complexity of the polymorphism types and give them a rebrand at bear minimum, since they've been historically a bit fragile

PR #5315, opened 2024-01-11
chore: make examples point at `@next`

KATT:
`npm:pkg-name@your-range-here` is to make manypkg not complain


juliusmarminge:
Will this still respect `prefer-workspace-pacakges` ?

KATT:
> Will this still respect `prefer-workspace-pacakges` ?

@juliusmarminge yeah it will but it's blocked by a bug in manypkg right now

juliusmarminge (approved):

PR #5335, opened 2024-01-14
chore: export from main entrypoint instead of sub-paths

juliusmarminge:
- this makes it so the core package only has a single entrypoint and not needing to worry about different module resolution techniques having to resolve as `@trpc/core/dist/rpc` or `@trpc/core/rpc` (coming from exports condition) or `@trpc/core/rpc/index.d.ts` (coming from our generated entrypoint barrel-folders)


packages/server/src/rpc.ts
KATT, line 1: Maybe this could/should be `import {} from '../@trpc-core-unstable-do-not-import-this-please'` like in #5331? 

Unsure if when this is used TypeScript will be able to "see" the types needed from `@trpc/core` (which only seems to be `ProcedureType` & `CombinedDataTransformer`

KATT (approved):
If this doesn't work, maybe it works in combination with #5331?

KATT:
just seem to be `httpLink` that needs updating
```
@***/client:build: [!] (plugin rpt2) Error: src/links/httpLink.ts:3:28 - error TS2307: Cannot find module '@***/core/observable' or its corresponding type declarations.
@***/client:build: 
@***/client:build: 3 import { observable } from '@***/core/observable';
@***/client:build:                              ~~~~~~~~~~~~~~~~~~~~~~~
@***/client:build: src/links/httpLink.ts:34:21 - error TS7006: Parameter 'observer' implicitly has an 'any' type.
@***/client:build: 
@***/client:build: 34         observable((observer) => {
@***/client:build:                        ~~~~~~~~
@***/client:build: 
```

PR #5350, opened 2024-01-17
fix: remove named imports from react

juliusmarminge:
Fixes:

```sh
You're importing a component that needs createContext. It only works in a Client Component but none of its parents are marked with "use client", so they're Server Components by default.
Learn more: https://nextjs.org/docs/getting-started/react-essentials

   ╭─[/Users/julius/dev/trpc/packages/react-query/dist/utilsProxy-5402e992.mjs:1:1]
 1 │ import { createTRPCClientProxy } from '@trpc/client';
 2 │ import { createFlatProxy, createRecursiveProxy } from '@trpc/core';
 3 │ import { createContext } from 'react';
   ·          ─────────────
 4 │
 5 │ /**
 6 │  * To allow easy interactions with groups of related queries, such as
   ╰────

Maybe one of these should be marked as a client entry with "use client":
  ../../trpc/packages/react-query/dist/utilsProxy-5402e992.mjs
  ../../trpc/packages/react-query/dist/server/index.mjs
  ./src/app/layout.tsx
 ⚠ Fast Refresh had to perform a full reload due to a runtime error.
 ⨯ ../../trpc/packages/react-query/dist/utilsProxy-659744c7.mjs
ReactServerComponentsError:
```

### DIFF
--- a/packages/react-query/src/internals/context.tsx
+++ b/packages/react-query/src/internals/context.tsx
@@ -20,7 +20,7 @@ import type {
   TRPCUntypedClient,
 } from '@trpc/client';
 import type { AnyRouter, DistributiveOmit } from '@trpc/core';
-import { createContext } from 'react';
+import * as React from 'react';
 import type { ExtractCursorType } from '../shared';
 import type { TRPCQueryKey } from './getQueryKey';
 
@@ -226,4 +226,4 @@ export interface TRPCQueryUtils<TRouter extends AnyRouter> {
     queryKey: TRPCQueryKey,
   ) => InfiniteData<unknown> | undefined;
 }
-export const TRPCContext = createContext(null as any);
+export const TRPCContext = React.createContext?.(null as any);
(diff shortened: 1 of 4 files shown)


packages/react-query/src/internals/context.tsx
```diff
@@ -226,4 +226,4 @@ export interface TRPCQueryUtils<TRouter extends AnyRouter> {
     queryKey: TRPCQueryKey,
   ) => InfiniteData<unknown> | undefined;
 }
-export const TRPCContext = createContext(null as any);
+export const TRPCContext = React.createContext?.(null as any);
```
juliusmarminge, line 229: unsupported side-effect ^^

VictorGlindasPaf:
I'm curious as to why removing named exports is needed?
Locally the issue is resolved by just converting `createContext(null)` to `createContext?.(null)` in the built `.mjs` file.

packages/react-query/src/internals/context.tsx
KATT: Nice trick!

KATT (approved):

PR #5355, opened 2024-01-18
fix(server): delete the `@trpc/core`-package

KATT:
`@trpc/core` was a nice idea but the ecosystem doesn't seem ready for it. 

Especially with the combination of pnpm monorepos + TypeScript where it's tough not to get _"The inferred type of 'xxxx' cannot be named without a reference to '.pnpm/@trpc+core@[.....]'. This is likely not portable. A type annotation is necessary"_-style errors.

## 🎯 Changes

- Move `@trpc/core` back into the server-package
- Move `@trpc/server/observable` out of core again since it's so isolated it could even be its own dedicated package
- Add new `@trpc/server/unstable-core-do-not-import`-export which should dissuade people not to use things from that entry point

## References

- Fixes https://github.com/trpc/tmp-trpc-declaration-repro
- https://github.com/trpc/tmp-trpc-declaration-repro/pull/7


packages/client/src/TRPCClientError.ts
```diff
@@ -1,10 +1,14 @@
...
+import type {
+  inferErrorShape,
+  Maybe,
+  TRPCErrorResponse,
+  TRPCInferrable,
+} from '@trpc/server/unstable-core-do-not-import';
 import {
   getCauseFromUnknown,
   isObject,
   type DefaultErrorShape,
```
juliusmarminge, line 10: thy is this not getting moved by the linter to the type imports?

packages/client/turbo.json
```diff
@@ -1,7 +1,5 @@
 {
   "$schema": "https://turborepo.org/schema.json",
   "extends": ["//"],
-  "pipeline": {
-    "codegen-entrypoints": { "outputs": ["package.json", "links/**"] }
-  }
+  "pipeline": { "codegen-entrypoints": { "outputs": ["package.json"] } }
```
juliusmarminge, line 4: why was this changed?

packages/server/src/@trpc/server/rpc.ts
juliusmarminge, line 1: these file paths looks very strange..
CleanShot 2024-01-19 at 13 13 48

juliusmarminge (approved):

packages/client/turbo.json
juliusmarminge: should there not be a `/links` sub-entrypoint anymore?

packages/server/src/@trpc/server/rpc.ts
KATT: I just want them to be clear to guide adapters -- people generally look at the adapters for reference. Maybe I overdid it here?

juliusmarminge: I’d lean that way and just use the index route. As long as we have named exports in /index.ts we can make sure to only use publically available stuff anyways

KATT: Alright, I'll just merge this for now so we can draw a line in the sand, but let us do that in a follow-up

packages/client/src/TRPCClientError.ts
KATT: unsure, but it'll get fixed automatically once prettier updates :D

packages/server/src/@trpc/server/rpc.ts
juliusmarminge: Or maybe even make new packages for each? `@trpc/adapter-next` etc?

KATT: Yeah, that'd be even nicer - so much boilerplate though 😅

PR #5383, opened 2024-01-24
docs: optional req res for inner context fixes #5344

### LINKED ISSUE #5344: bug: Context & InnerContext sample not working

v10.45

### Describe the bug

I tried to follow the instructions located at this page
https://trpc.io/docs/server/context#inner-and-outer-context

However, I am having difficulty accessing the res & req objects on my API procedures.

Searching through GH, I found an earlier issue
https://github.com/trpc/trpc/pull/3325#issuecomment-1335060254

I have tried this one too,
image

But here, you can see that the res & req objects are also undefined. How did this happen? Am I missing something here?

### Link to reproduction

N/A

### To reproduce

Just follow the samples on the documentation

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

## Funding

<a href="https://polar.sh/trpc/trpc/issues/5344">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/5344/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

aludvigsen:
Closes #5344

## 🎯 Changes

Code sample update in docs. Provide `req` and `res` to context when using outer/inner context.

### DIFF
--- a/www/docs/server/context.md
+++ b/www/docs/server/context.md
@@ -174,6 +174,7 @@ interface CreateInnerContextOptions extends Partial<CreateNextContextOptions> {
  */
 export async function createContextInner(opts?: CreateInnerContextOptions) {
   return {
+    ...opts,
     prisma,
     session: opts.session,
   };


Nick-Lucas:
I think you need to read that section again, because both req/res and the session already end up on the context in that example, this change does nothing

aludvigsen:
But the types don't end up in the context when using inner and outer context. Under the example code, it says, "It is important to infer your Context from the inner context, as only what is defined there is really always available in your procedures."
If you don't pass the 'opts' in the inner context, you won't have the correct types for the context. Req/res will not be undefined, but missing (in the types).

PR #5388, opened 2024-01-25
refactor Observable.pipe type to use variadic tuple types

trvswgnr:
I refactored the `Observable.pipe` type in the `@trpc/server` package. This replaces manual function overloads with [variadic tuple types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#variadic-tuple-types) (introduced in TS 4.0) so an arbitrary number of operator functions to be passed to `pipe`, while still being type-safe.

## 🎯 Changes

- Refactored `Observable.pipe` type to use variadic tuple types for its operator arguments.
- Implemented conditional types to infer the output value and error types based on the passed operator functions.
- Removed the multiple overloaded definitions of `pipe`.

## 🚀 Benefits

- Better type inference, allowing an arbitrary number of operator functions to be passed without needing more overloads.
- Backward compatibility and no writing manual type definitions if more are needed.
- More ✨scalable✨ and ✨maintainable✨.

## ⚠️ Considerations

I ran/passed all tests, but I don't see any type-level tests for this. Anything that relies on specific overload signatures of `Observable.pipe` should probably be reviewed.

## ✅ Checklist


KATT (changes requested):
CI fails 😅

Nick-Lucas:
This might have a fairly negative impact on compiler performance, IIRC recursive types are quite expensive to compute and/or cache, and the explicit+horrible version is actually very performant for TS

Has been a while but I was researching around how to solve this problem early last year and determined that explicit was best for libraries

Issue #4129, opened 2023-04-03
feat: Lazy load routers


zomars:
### Describe the feature you'd like to request

Basically `next/dynamic` but for the backend. 

Recently our tRPC router has been growing significantly for many reasons but mostly due to heavy third party SDKs. The problem with this is that even if we're not calling that specific procedure we still load ALL the router dependencies when calling any procedure.

This impacts performance and cold boots significantly.

### Describe the solution you'd like to see

Splitting big tRPC routers into lazy loadable chunks could help loading only what is needed when a procedure is called. I would imagine some pseudo code like this:

![image](https://user-images.githubusercontent.com/3504472/229645485-315a4d62-2044-49ef-b420-0b33bf7fd503.png)

### Describe alternate solutions

Other aternatives involve moving all the procedure code into a separate file a use imports for that [as done in here](https://github.com/calcom/cal.com/pull/7889/commits/6ce440851e172db3f4c7ae720e8396c9201eb316).

Also [lazy loading only third party libraries](https://github.com/calcom/cal.com/pull/7889/commits/a2ff4d9a0cac732af3cd842988feaad471187488) (like Stripe, Google, etc.) seems to help. But it would be a much nicer DX being able to simply lazy load full routers and support it natively.

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

<sub>[TRP-19](https://linear.app/trpc/issue/TRP-19/feat-lazy-load-routers)</sub>

## Funding
* You can sponsor this specific effort via a [Polar.sh](https://polar.sh) pledge below
* We receive the pledge once the issue is completed & verified

<a href="https://polar.sh/trpc/trpc/issues/4129">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/4129/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/trpc/trpc/issues/4129/pledge.svg">
</picture>
</a>

KATT:
Cool. I'm optimistic we could support this. As a workaround, anyone who approaches this problem should be able to achieve the same perf gains using the SOA-approach: https://github.com/trpc/trpc/tree/main/examples/soa


## An outline of how this could work internally:

**Rough how it currently works:**

- We store a dictionary with all procedures in a flat record, e.g. `{ 'post.byId': () => { /* ... */ `
[...]

PR #5404, opened 2024-01-26
feat(server): short-hand routers and spread support

### LINKED ISSUE #3630: feat: add shorthand `{}` for sub-routers

I don't wanna write `router({ })` for my sub-routers

These should be equivalent

```ts
const appRouter = router({
  foo: router({
    bar: publicProcedure.query(() => '...'),
  })
})
```

```ts
const appRouter = router({
  foo: {
    bar: publicProcedure.query(() => '...'),
  }
})
```

Not doing it

### Additional information

If it creates tech debt, don't do it.

### 👨‍👧‍👦 Contributing

<sub>[TRP-49](https://linear.app/trpc/issue/TRP-49/feat-add-shorthand-for-sub-routers)</sub>

## Funding

<a href="https://polar.sh/trpc/trpc/issues/3630">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/3630/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

KATT:
- Allow for creating routers that are just objects
- Will simplify `.d.ts`-files for people who use `satisfies TRPCRouterRecord`

Closes #3630

Could help make #4129 easier

### How it looks like

This will allow you to define routers like this:

```ts

const post = {
  byId: t.procedure.query(() => '..'),
} satisfies RouterRecord; // <--  ✨ using satisfies for sub-routers could(?) be less heavy for the compiler and `.d.ts`-files



const appRouter = t.router({
  post,
  // ✨ short-cut routers by doing a `{}`
  deeply: {
    nested: {
      greeting: t.query(() => '...'),
    },
  },
});

```


packages/client/src/createTRPCClient.ts
```diff
@@ -78,12 +78,14 @@ type DecorateProcedure<
...
 type DecoratedProcedureRecord<
   TRouter extends AnyRouter,
-  TProcedures extends ProcedureRouterRecord,
+  TRecord extends RouterRecord,
 > = {
-  [TKey in keyof TProcedures]: TProcedures[TKey] extends AnyRouter
-    ? DecoratedProcedureRecord<TRouter, TProcedures[TKey]['_def']['record']>
-    : TProcedures[TKey] extends AnyProcedure
-    ? DecorateProcedure<TRouter['_def']['_config']['$types'], TProcedures[TKey]>
+  [TKey in keyof TRecord]: TRecord[TKey] extends infer $Value
```
juliusmarminge, line 83: why do you need this `extends infer` here?

packages/tests/server/mergeRouters.test.ts
```diff
@@ -17,6 +18,26 @@ test('mergeRouters', async () => {
...
+    foo: t.procedure.query(() => 'foo'),
+  } satisfies TRPCRouterRecord;
+  const router2 = {
+    bar: t.procedure.query(() => 'bar'),
+  } satisfies TRPCRouterRecord;
+
+  const merged = t.router({
+    ...router1,
+    ...router2,
+  });
```
juliusmarminge, line 34: niceee - can we deprecate this monstosity then?

https://github.com/trpc/trpc/blob/next/packages/server/src/unstable-core-do-not-import/router.ts#L272-L309

or maybe just make it an alias?

juliusmarminge (commented):
Great work, only nitpick I'd point out is that there's a lot of these:

```ts
  TRouter['_def']['_config']['$types'],
  TRouter['_def']['record']
```

don't we have inference helpers for those?

```ts
  inferRootTypes<TRouter>
  ...
```

?

packages/client/src/createTRPCClient.ts
KATT: I just think it makes the "looping" clearer

KATT: Only need to grab `TRecord[Key]` once, it's like a declaring a variable

juliusmarminge: > Only need to grab `TRecord[Key]` once, it's like a declaring a variable

aah i see, how is the compiler performance comparatively?

KATT:
> Great work, only nitpick I'd point out is that there's a lot of these:
> 
> ```ts
>   TRouter['_def']['_config']['$types'],
>   TRouter['_def']['record']
> ```
> 
> don't we have inference helpers for those?
> 
> ```ts
>   inferRootTypes<TRouter>
>   ...
> ```
> 
> ?

I'd be fine to change that, I'd also be happy if we changed the router so the grabbing would just be e.g. 

- `TRouter['_def']['$types']`
- `TRouter['_def']['_config']['$types']`
- `TRouter[RootTypesSymbol]`
- `TRouter['_config']['$types']`

I think that sort of approach would be even nicer than using a helper - it feels like an unnecessary abstraction

packages/client/src/createTRPCClient.ts
KATT: No idea tbh 🙈

packages/tests/server/mergeRouters.test.ts
KATT: yes that'd be nice

KATT: we could alias it

juliusmarminge (approved):
i see that you can't spread actual routers:

```ts
// router/post.ts
import { createTRPCRouter, protectedProcedure, publicProcedure } from "../trpc";

export const authRouter = createTRPCRouter({
  getSession: publicProcedure.query(({ ctx }) => {
    return ctx.session;
  }),
  getSecretMessage: protectedProcedure.query(() => {
    // testing type validation of overridden next-auth Session in @acme/auth package
    return "you can see this secret message!";
  }),
});

// root.ts
import { authRouter } from "./router/auth";
import { postRouter } from "./router/post";
import { createTRPCRouter } from "./trpc";

export const appRouter = createTRPCRouter({
  auth: authRouter,
  post: postRouter,
  ...authRouter, // <-- doesn't work
});

// export type definition of API
```

how do you make the typesystem / proxy understand spread? if we to alias mergeRouters to this implementation we should support spreading routers and not just records?

CleanShot 2024-01-30 at 08 42 50@2x

KATT:
> i see that you can't spread actual routers:
> 
> ```ts
> // router/post.ts
> import { createTRPCRouter, protectedProcedure, publicProcedure } from "../trpc";
> 
> export const authRouter = createTRPCRouter({
>   getSession: publicProcedure.query(({ ctx }) => {
>     return ctx.session;
>   }),
>   getSecretMessage: protectedProcedure.query(() => {
>     // testing type validation of overridden next-auth Session in @acme/auth package
>     return "you can see this secret message!";
>   }),
> });
> 
> // root.ts
> import { authRouter } from "./router/auth";
> import { postRouter } from "./router/post";
> import { createTRPCRouter } from "./trpc";
> 
> export const appRouter = createTRPCRouter({
>   auth: authRouter,
>   post: postRouter,
>   ...authRouter, // <-- doesn't work
> });
> 
> // export type definition of API
> export type AppRouter = typeof appRouter;
> ```
> 
> how do you make the typesystem / proxy understand spread? if we to alias mergeRouters to this implementation we should support spreading routers and not just records?

No idea, but I think most people use sub-routers with the other way, so might be a non-issue

PR #5421, opened 2024-02-01
fix(react-query): infinite queries invalidation

### LINKED ISSUE #5412: bug: Cursor not reset when invalidating data in an infiniteQuery

  System:
    OS: macOS 14.0
    CPU: (12) arm64 Apple M2 Max
    Memory: 3.99 GB / 96.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.7.0 - ~/.nvm/versions/node/v20.7.0/bin/node
    npm: 10.1.0 - ~/.nvm/versions/node/v20.7.0/bin/npm
    bun: 1.0.7 - ~/.bun/bin/bun
  Browsers:
    Chrome: 118.0.5993.70
    Safari: 17.0

### Describe the bug

When invalidating an infinite query, the cursor of the infinite query is not being reset.

### Expected
- Page 0: Cursor 0
- Page 1: Cursor 1
- Page 2: Cursor 2
- (invalidate)
- Page 0: Cursor 0
- Page 1: Cursor 1
- Page 2: Cursor 2

### Actual
- Page 0: Cursor 0
- Page 1: Cursor 1
- Page 2: Cursor 2
- (invalidate)
- Page 0: Cursor 3
- Page 1: Cursor 4
- Page 2: Cursor 5

The useQuery docs state that:

> When an infinite query becomes stale and needs to be refetched, each group is fetched sequentially, starting from the first one. This ensures that even if the underlying data is mutated, we're not using stale cursors and potentially getting duplicates or skipping records. If an infinite query's results are ever removed from the queryCache, the pagination restarts at the initial state with only the initial group being requested.

Therefore, I would expect that all the previous pages would be fetched using the previous page params.

### Link to reproduction

https://stackblitz.com/edit/github-6aj2fp?file=src%2Fpages%2F_app.tsx,src%2Fpages%2Fapi%2Ftrpc%2F[trpc].ts,src%2Fpages%2Findex.tsx

### To reproduce

- Fetch data using an infinite query
- Invalidate (e.g. `trpc.myData.invalidate()`)
- The data begins to refetch at the last cursor instead of setting the cursor to `undefined`

### Additional information

Maybe this is a misunderstanding on my part, but I thought it worth filing a big / raising this question.

My use case: I need an infinite query and that query needs to be refetched whenever my data is invalidated.

### 👨‍👧‍👦 Contributing

KATT:
Closes #5412

## 🎯 Changes

Fix infinite queries invalidation

## ✅ Checklist

### DIFF
--- a/packages/react-query/src/internals/getClientArgs.ts
+++ b/packages/react-query/src/internals/getClientArgs.ts
@@ -9,12 +9,7 @@ export function getClientArgs<TOptions>(
   pageParam?: any,
 ) {
   const path = queryKey[0];
-  let input = queryKey[1]?.input;
-  if (pageParam) {
-    input = {
-      ...(input ?? {}),
-      cursor: pageParam,
-    };
-  }
+  const input = queryKey[1]?.input;
+  if (pageParam) (input as any).cursor = pageParam;
   return [path.join('.'), input, (opts as any)?.trpc] as const;
 }
--- a/packages/react-query/src/internals/getClientArgs.ts
+++ b/packages/react-query/src/internals/getClientArgs.ts
@@ -9,7 +9,12 @@ export function getClientArgs<TOptions>(
   pageParam?: any,
 ) {
   const path = queryKey[0];
-  const input = queryKey[1]?.input;
-  if (pageParam) (input as any).cursor = pageParam;
+  let input = queryKey[1]?.input;
+  if (pageParam) {
+    input = {
+      ...(input ?? {}),
+      cursor: pageParam,
+    };
+  }
   return [path.join('.'), input, (opts as any)?.trpc] as const;
 }
--- a/packages/react-query/src/shared/hooks/createHooksInternal.tsx
+++ b/packages/react-query/src/shared/hooks/createHooksInternal.tsx
@@ -503,26 +503,6 @@ export function createRootHooks<
     useSuspenseInfiniteQuery,
   };
 }
-/* istanbul ignore next */
-/**
- * Hack to infer the type of `createReactQueryHooks`
- * @link https://stackoverflow.com/a/59072991
- */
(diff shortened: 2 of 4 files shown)


packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -503,26 +503,6 @@ export function createRootHooks<
...
-
-type returnTypeInferer<TType> = TType extends (
-  a: Record<string, string>,
-) => infer U
-  ? U
-  : never;
-type fooType<TRouter extends AnyRouter, TSSRContext = unknown> = GnClass<
-  TRouter,
-  TSSRContext
->['fn'];
```
KATT, line 525: Unrelated but is nice to get rid of this, our TS req is high enough not to require this

packages/react-query/src/internals/getClientArgs.ts
```diff
@@ -9,7 +9,12 @@ export function getClientArgs<TOptions>(
   pageParam?: any,
 ) {
   const path = queryKey[0];
-  const input = queryKey[1]?.input;
-  if (pageParam) (input as any).cursor = pageParam;
```
KATT, line 13: cc @juliusmarminge this caused a bug report a year later 🐛  😅 

let's never mutate any non-local variable

KATT: took me an hour to find lol, even went on a detour testing if react-query did what it should

martenbjork:
Thanks for fixing this ❤️ I spent all Monday wondering what the heck I was doing wrong. Glad — for my sanity's sake — to see that a mutated variable was the root cause and not the demise of my brain. Cheers, the fix is really high impact for us and works like a charm. 🥰

PR #5441, opened 2024-02-04
chore(server): rename variable and add comment

KATT:
Closes #

## 🎯 Changes

Got confused staring at our own code for `ProcedureResolver` so added a comment

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
+++ b/packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
@@ -74,13 +74,13 @@ type ProcedureResolver<
   TContextOverrides,
   TInputOut,
   TOutputParserIn,
-  TOutputInferred,
+  $Output,
 > = (opts: {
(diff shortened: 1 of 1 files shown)


packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
```diff
@@ -73,12 +73,15 @@ type ProcedureResolver<
   _TMeta,
   TContextOverrides,
   TInputOut,
-  TOutputIn,
-  TOutputOut,
+  TOutputParserIn,
+  $Output,
```
Nick-Lucas, line 77: Was the consistency not beneficial before? TOutputParserIn is nice an explicit, any reason the Out needs to have a different scheme?

KATT: I just like having the `$`-prefix when it's an inferred type

Nick-Lucas: image

Nick-Lucas (approved):

PR #5451, opened 2024-02-06
feat(react-query): next-app-router example with prefetch helpers

juliusmarminge:
More integrated version of https://github.com/t3-oss/create-t3-turbo/pull/877

adds a wrapped proxy around createCaller that puts the promise into a server-side QueryClient. devs can then choose to wrap some client component that needs some server prefetched data  with the `<HydrateClient>` component. no network requests will be made from client components until the prefetched data goes stale:

https://utfs.io/f/8278fdc7-572e-4b14-9f10-5f9fece747bb-yl67w4.15.58.mp4

you choose if you want to await the promise or not depending on if you also need to use the data in the RSC. if you void it, rendering wont be blocked until a `useSuspenseQuery` consumes the promise

This gives the same behavior as the `ReactQueryStreamedProvider`, but you can do authed procedures since there's no fetching in client componetns, even on the server, so no need to hack the RSC headers into the CC which causes some weird security concerns

https://github.com/trpc/trpc/assets/51714798/9bbc5b68-8588-4785-a6e2-d87edb0f068d


juliusmarminge:
@KATT the linter is superborked here - it fixes things that's not supposed to be fixed and the fixees causes type-errors. any idea what's up with that?

examples/next-app-router/src/app/page.tsx
```diff
@@ -0,0 +1,22 @@
...
+  await trpc.post.list();
+
+  return (
+    <main className="container max-w-xl py-8">
+      <Suspense>
+        <div className="absolute right-8 top-8">
+          <UserButton />
+        </div>
+      </Suspense>
+      <HydrateClient>
```
KATT, line 17: good that this is contextual

if you have many of these, i guess they keep on sending stuff doubly? it would stack up quite a lot of it was used a lot

maybe we could keep track of stuff that's already been sent to the client so we don't send same query times across many hydration boundaries

KATT (approved):

examples/next-app-router/src/app/page.tsx
juliusmarminge: > if you have many of these, i guess they keep on sending stuff doubly? it would stack up quite a lot of it was used a lot

i would actually have thought RQ would handle this internally. isn't this an issue no matter where you use them?

TkDodo: there is no deduplication with `HydrationBoundary`. Every boundary will send its content to the client.

@ephem we talked about this some time ago I think ?

TkDodo: I think there is if you use the same `queryClient` maybe ?

juliusmarminge: it should be a single query client per request https://github.com/trpc/trpc/pull/5451/files#diff-8ef8de9fb13401a32e846aeaedf32d659dd5f222706751573fb792d04d36e66cR24

TkDodo: I think we've moved away from this recommendation precisely because of this:

https://tanstack.com/query/v5/docs/framework/react/guides/advanced-ssr#alternative-use-a-single-queryclient-for-prefetching

> In the example above, we create a new queryClient for each Server Component that fetches data. This is the recommended approach, but if you want to, you can alternatively create a single one that is reused across all Server Components:

also:

> As a future improvement, we might look into creating a dehydrateNew() function (name pending) that only dehydrate queries that are new since the last call to dehydrateNew(). Feel free to get in touch if this sounds interesting and like something you want to help out with!

Is this you getting in touch to implement that feature 😅 ?

juliusmarminge: aha ;)

dBianchii:
This is great. Having to manually pass around initialData to client components, or doing manual prefetching and dehydration does feel very repetitive. Hope this can be merged soon.

masterjanic:
Would love to have this feature as well, it would save me so much time ❤️

packages/react-query/src/rsc.tsx
```diff
@@ -0,0 +1,72 @@
...
+      // @ts-expect-error - ??
+      (acc, key) => acc[key],
+      caller,
+    ) as unknown as DecorateProcedure<AnyProcedure>;
+
+    const [input] = args;
+    const promise = proc(input);
+
+    void getQueryClient().prefetchQuery({
+      queryKey: getQueryKeyInternal(path, input, 'query'),
```
juliusmarminge, line 54: should make this dynamic and not have ´query` type hard-coded

413n:
Love this feature!
May I ask if you have already thought if this prefetching helpers could be mixed with the new `experimental_caller` on the procedures? Since presumably there is no `appRouter` with the `experimental_caller`, I was curious to know if this could be feasible

juliusmarminge:
> Love this feature!
> 
> May I ask if you have already thought if this prefetching helpers could be mixed with the new `experimental_caller` on the procedures? Since presumably there is no `appRouter` with the `experimental_caller`, I was curious to know if this could be feasible

You need the router for the client component to be able to fetch

packages/react-query/src/rsc.tsx
TkDodo: pretty sure it doesn't work with infinite queries right now

juliusmarminge: Yea. That's what i gotta fix

KATT (approved):
This is great!!!!!!!!!

Should we combine all "kitchen sink"-y things inside the [existing `next-app-dir`-project](https://github.com/trpc/trpc/tree/next/examples/.experimental/next-app-dir)-project perhaps? I'm personally almost loosing track of all of our parallel experimental approaches 😅

juliusmarminge:
> This is great!!!!!!!!!
> 
> 
> 
> Should we combine all "kitchen sink"-y things inside the [existing `next-app-dir`-project](https://github.com/trpc/trpc/tree/next/examples/.experimental/next-app-dir)-project perhaps? I'm personally almost loosing track of all of our parallel experimental approaches 😅 

Yea probably a good idea.

patches/@tanstack__query-core@5.40.0.patch
juliusmarminge, line 1: this patch is implemented upstream here: https://github.com/TanStack/query/pull/7538

KATT (approved):

examples/.experimental/next-app-dir/src/app/rsc-rq-prefetch/page.tsx
```diff
@@ -0,0 +1,18 @@
+import { HydrateClient, trpc } from '~/trpc/rq-server';
+import React from 'react';
+import { Post } from './post';
+
+export const dynamic = 'force-dynamic';
+
+export default function Home() {
+  void trpc.getLatestPost.prefetch();
```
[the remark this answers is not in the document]
juliusmarminge: `void` will ignore errors?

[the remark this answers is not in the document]
juliusmarminge: Then the query will be retried when it mounts on the client depending on your retry options

Issue #5455, opened 2024-02-06
feat: better `cacheTag` control in next app dir layouts


dalechyn:
### Describe the feature you'd like to request

<details>
  <summary>TLDR</summary>
  
  `ctx` in queries is often being used as inputs when some are defined in the middlewares.

   This is not taken in account with experimental app dir support, as the `cacheTag` only caches the inputs, but never the context.

   Therefore, I am suffering a bug when a query that just uses `ctx` to retrieve user-specific data from `ctx.session`, is being cached with no inputs `[]`, and I am able to see the cached data of another user.
</details>

# Introduction

Experimental tRPC's NextJS AppRouter layout support involves caching of query inputs, and is implemented in the next way:
https://github.com/trpc/trpc/blob/b72c7aa729cdd303ea86dc4b04153224ed0a27c6/packages/next/src/app-dir/links/nextCache.ts#L34-L43

# Problem
While it works for most of the cases, problems arise when your query logic depends on trpc's `middleware`s.

Here's an example implementation of a *protected procedure* taken from the tRPC's documentation:
```typescript
import { initTRPC, TRPCError } from '@trpc/server';

export const t = initTRPC.context<Context>().create();

// you can reuse this for any procedure
export const protectedProcedure = t.procedure.use(async function isAuthed(
  opts,
) {
  const { ctx } = opts;
  // `ctx.user` is nullable
  if (!ctx.user) {
    //     ^?
    throw new TRPCError({ code: 'UNAUTHORIZED' });
  }

  return opts.next({
    ctx: {
      // ✅ user value is known to be non-null now
      user: ctx.user,
      // ^?
    },
  });
});

t.router({
  // this is accessible for everyone
  hello: t.procedure
    .input(z.string().nullish())
    .query((opts) => `hello ${opts.input ?? opts.ctx.user?.name ?? 'world'}`),
  admin: t.router({
    // this is accessible only to admins
    secret: protectedProcedure.query((opts) => {
      return {
        secret: 'sauce',
      };
    }),
  }),
});
```
As you can see, the `ctx.user?.name` in this example is **implicitly** becoming an input, because you use it in the query implementation, yet it's not being included in the `cacheKey`!

In this particular case, which is taken from the docs, the consequence is that the first user querying the `hello` query, would make it cache the response `hello alice` (i.e), then when the second user queries the same procedure (bob i.e.), he would see `hello alice` as it was cached with a `cacheKey` of `[]` as inputs to the functions simply don't exist.

[...]

PR #4567, opened 2023-06-27
createAction from a "client"

juliusmarminge:
Closes #

## 🎯 Changes

What changes are made in this PR? Is it a feature or a bug fix?

- allows you to pass in a router instance to the `experimental_createServerActionHandler` and then creating actions from your client like `createAction(api.createPost)`

Needs #4375

## ✅ Checklist

- [ ] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [ ] If necessary, I have added documentation related to the changes made.
- [ ] I have added or updated the tests related to the changes made.

PR #5458, opened 2024-02-07
feat(next/app-dir): `contextCache` to control context values to add to `cacheTag`

dalechyn:
Closes: #5455

## 🎯 Changes
This change allows a user to specify `contextCache` method in the `experimental_nextCacheLink`, which lets one select the values from the context that should be considered when a `cacheTag` is generated.

It also changes the way `cacheTag` is generated, taking care of the fact that the `cacheTag` cannot be longer then 256 characters, thus instead of stringifying input values, it will form a sha-256 hash of those.

## ✅ Checklist

> there are no test cases covering this and I haven't added any.


dalechyn:
@KATT, `noCache` middleware is not implemented as discussed in #5455, and I also don't know how to deal with other places where `generateCacheTag` is used as such don't have access to the context as opposed to `experimental_nextCacheLink`'s `createContext()` function that exposes one.

I need some help figuring this part out. Anyway, the commit should not break those as the `context` parameter is also marked with `any` type, therefore is optional to be passed.

packages/next/package.json
```diff
@@ -112,5 +111,8 @@
   },
   "funding": [
     "https://trpc.io/sponsor"
-  ]
+  ],
+  "dependencies": {
+    "@noble/hashes": "^1.3.3"
```
juliusmarminge, line 116: Use builtin crypto module?

juliusmarminge: this code will only ever be run serverside, no?

and the crypto module is implemented in all major players including node, bun, workers, edge etc etc

dalechyn: I'm not sure to be honest, i'm not that deep in the codebase, but I trust you on that.

Will change it to use crypto module instead!

dalechyn:
@juliusmarminge, ci fails as webpack cannot import `node:crypto`.

also found this thing: https://nextjs.org/docs/messages/node-module-in-edge-runtime, that might indicate that it would not be possible to use `node:crypto` in edge environment.

packages/next/src/app-dir/links/nextCache.ts
```diff
@@ -33,18 +38,22 @@ export function experimental_nextCacheLink<TRouter extends AnyRouter>(
...
+              input,
+              opts.cacheContext?.(ctx),
+            );
+            // Let per-request revalidate override global revalidate
+            const requestRevalidate =
+              typeof context['revalidate'] === 'number' ||
+              context['revalidate'] === false
+                ? context['revalidate']
+                : undefined;
+            const revalidate = requestRevalidate ?? opts.revalidate ?? false;
```
juliusmarminge, line 55: This can still be hoisted one level up?

dalechyn: Agree, good catch!

packages/next/src/app-dir/links/nextCache.ts
```diff
@@ -18,6 +18,11 @@ import { generateCacheTag } from '../shared';
 
 type NextCacheLinkOptions<TRouter extends AnyRouter> = {
   router: TRouter;
+  /**
+   * define which values from the context should be considered into the cache
+   * key
+   */
+  cacheContext?: (ctx: inferRouterContext<TRouter>) => any[];
```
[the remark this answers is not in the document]
dalechyn: I suggest making this property `| undefined` instead of partial so that user is explicitly passing `undefined` if he doesn't want to manipulate the cache key.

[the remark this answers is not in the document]
dalechyn: included in the example!

examples/.experimental/next-app-dir/src/trpc/server-invoker.ts
```diff
@@ -21,15 +21,9 @@ export const api = experimental_createTRPCNextAppDirServer<typeof appRouter>({
...
-              session: await auth(),
-              headers: {
-                cookie: cookies().toString(),
-                'x-trpc-source': 'rsc-invoke',
-              },
-            };
-          },
+          // include the user id in the cache key
+          cacheContext: (ctx) => [ctx.session?.user.id],
+          createContext,
```
juliusmarminge, line 26: we don't need it on the link then right? if it's required higher up?

dalechyn: you're right, we have to hoist it down somehow. couldn't find a fast way to do that.

juliusmarminge: iirc it'll be on the runtime

packages/next/src/app-dir/links/nextCache.ts
```diff
@@ -45,6 +49,12 @@ export function experimental_nextCacheLink<TRouter extends AnyRouter>(
         const promise = opts
           .createContext()
```
juliusmarminge, line 50: I think you can get this function from the runtime arg on line 36

dalechyn:
@juliusmarminge i need some guidance here, i'm not sure wether I'm on the right or wrong direction. Starting to have a real type battle here.

dalechyn:
Screenshot 2024-02-11 at 14 37 55
Just tried it all out, caching by hash seems working for experimental app dir example, even when a function has no input it includes whatever I pass to the `cacheContext` fn to the cache tag.

413n:
Really interesting solution! Good job @dalechyn! I'm encountering actually this exact problem of wrong cache keys and cache shared between users requests.

I would like to share my use case that maybe could be useful to consider now or maybe it's off-topic, not sure. 

I think it was some months ago that @juliusmarminge you created an experimental PR (https://github.com/trpc/trpc/pull/4567) that added the createAction function where you could you pass the mutation and also the api paths to revalidate. I did found it really good and so when the PR was closed I copied the whole /app-dir folder into my api package.

To fix the cache shared between users I added manually the userId into the cacheTag (as you did @dalechyn).
But I also removed the inputs from the cacheTag because it would create inconsistencies. Let me explain.
- I have an "Add friend" page where you could search by `username` and the input would be taken from the `SearchParams`, so the input query could be anything, like "tes", "testus" or "testuser".
  - The cacheTag would be: `<userId>.friends.search?input='testus'`
- When a user is found then there is a mutation that sends a friend request. Now, I would need to revalidate the `friends.search` procedure cache to reload the results in it and show that the request was sent.

**Here it comes the problem**: if I have defined the `input` in the `cacheTag` then how could I revalidate each result ignoring the input? If I do `<userId>.friends.search` (without the `?input`) it will not revalidate all the procedure results but only the procedure path without any input. Or at least looking at the implementation revalidateTag, this is not possible since it checks the cacheTag with `.includes` and not for example with `.startsWith` (https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/spec-extension/revalidate.ts#L76).
Therefore I removed the inputs from the cacheTag logic to be able to correctly revalidate also this case.

What do you think? Am I missing something?

PR #5474, opened 2024-02-12
Add batchInterval option to dataLoader

stereosteve:
## 🎯 Changes

Adds an option to `httpBatchLink` to configure a `batchInterval` in milliseconds.

This is an option that exists in [dataloader](https://github.com/graphql/dataloader?tab=readme-ov-file#batch-scheduling) and can be helpful if rendering of items in a list is spread out across multiple event loop ticks.

## ✅ Checklist

### DIFF
--- a/packages/client/src/links/HTTPBatchLinkOptions.ts
+++ b/packages/client/src/links/HTTPBatchLinkOptions.ts
@@ -6,6 +6,7 @@ import type { HTTPHeaders, Operation } from './types';
 export type HTTPBatchLinkOptions<TRoot extends AnyClientTypes> =
   HTTPLinkBaseOptions<TRoot> & {
     maxURLLength?: number;
+    batchInterval?: number;
     /**
      * Headers to be set on outgoing requests or a callback that of said headers
      * @link http://trpc.io/docs/client/headers
--- a/packages/client/src/links/internals/createHTTPBatchLink.ts
+++ b/packages/client/src/links/internals/createHTTPBatchLink.ts
@@ -78,7 +78,10 @@ export function createHTTPBatchLink(
         return { validate, fetch };
       };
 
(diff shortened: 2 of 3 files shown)


juliusmarminge (changes requested):
Looks good, mind adding a test and perhaps a sentence in the docs?

packages/client/src/links/HTTPBatchLinkOptions.ts
```diff
@@ -6,6 +6,7 @@ import type { HTTPHeaders, Operation } from './types';
 export type HTTPBatchLinkOptions<TRoot extends AnyClientTypes> =
   HTTPLinkBaseOptions<TRoot> & {
     maxURLLength?: number;
+    batchInterval?: number;
```
KATT, line 9: This could be

```ts
HTTPLinkBaseOptions<TRoot> & DataLoaderOptions & { maxURLLength?: '..' }
```
that way, it inherits whatever docs and options we add on dataloader

packages/client/src/links/internals/createHTTPBatchLink.ts
```diff
@@ -78,7 +78,10 @@ export function createHTTPBatchLink(
         return { validate, fetch };
       };
 
-      const query = dataLoader<Operation, HTTPResult>(batchLoader('query'));
+      const query = dataLoader<Operation, HTTPResult>(
+        batchLoader('query'),
+        opts.batchInterval,
```
KATT, line 83:
```suggestion
        opts
```

KATT (changes requested):
I wonder what the desired behavior should be when triggering another fetch while one is pending - should we further debounce it? That's what we did in #914 

- [ ] Needs JSDoc about the option
- [ ] Should be an options object
- [ ] Needs docs  in https://trpc.io/docs/client/links/httpBatchLin

PR #5497, opened 2024-02-19
Export `AnyTRPCProcedureBuilder ` from `@trpc/server`

ymc9:
Closes #

## 🎯 Changes

## ✅ Checklist

- [ x] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [x ] If necessary, I have added documentation related to the changes made.
- [x ] I have added or updated the tests related to the changes made.

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
+++ b/packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
@@ -93,10 +93,10 @@ export type AnyProcedureBuilder = ProcedureBuilder<
   any,
   any,
   any,
-  any,
-  any,
-  any,
-  any
+  UnsetMarker,
+  UnsetMarker,
+  UnsetMarker,
+  UnsetMarker
 >;
 
 /**
--- a/packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
+++ b/packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
@@ -325,9 +325,9 @@ export function createBuilder<TContext, TMeta>(
   const builder: AnyProcedureBuilder = {
     _def,
     input(input) {
-      const parser = getParseFn(input as Parser);
+      const parser = getParseFn(input);
       return createNewBuilder(_def, {
-        inputs: [input as Parser],
+        inputs: [input],
         middlewares: [createInputMiddleware(parser)],
       });
     },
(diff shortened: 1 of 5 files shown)


ymc9:
An E2E test case is failing, but it doesn't seem to relate to the changes made. I can't reproduce the failure locally either.

packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
```diff
@@ -93,10 +93,10 @@ export type AnyProcedureBuilder = ProcedureBuilder<
...
   any,
   any,
-  any,
-  any,
-  any,
-  any
+  UnsetMarker,
+  UnsetMarker,
+  UnsetMarker,
+  UnsetMarker
```
Nick-Lucas, line 99: So I'm basically 100% sure this is going to cause regressions which out test suite may not be capturing yet (because we never broke it before)

Here's the reason: https://trpc.io/docs/server/validators#input-merging

and here's the scenario:

image

It's totally valid to set up an input or output and then create a router with that procedure, so UnsetMarker wouldn't be the root type in that situation

packages/tests/server/react/polymorphism.factory.tsx
```diff
@@ -33,10 +29,6 @@ export type FileExportStatusType = z.infer<typeof FileExportStatus>;
 // Dependencies
 //
 
-type BaseProcedure<TRoot extends AnyRootTypes> = ReturnType<
```
Nick-Lucas, line 36: Interestingly what was here before actually does produce a type with typeof unsetMarker there, so inter-compat with .input merging wouldn't work in production today. Maybe we can find a compromise then

Nick-Lucas: I tried export this from @trpc/server basically and found the same problem popped up with procedure.input() being passed through
```
// type $RootTypes = ReturnType<typeof initTRPC.create>['_config']['$types'];
// export type AnyTRPCProcedureBuilder = ReturnType<
//   typeof createBuilder<$RootTypes['ctx'], $RootTypes['meta']>
// >;
```

packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
Nick-Lucas: Okay so as per https://github.com/trpc/trpc/pull/5497/files#r1494523733 I think what we could do is leave _this_ AnyProcedureBuilder be

If you copy it over to packages/react-query/src/shared/polymorphism and export it from there though, I think we can live with its shortcomings as a limitation and open it up later. But we shouldn't make the rest of the codebase live with UnsetMarker hard-coded

PR #5536, opened 2024-03-04
feat(server): add `partykit` adapter

overra:
> [!NOTE]
> Work in progress: currently adding a partykit example to test with.

Closes #

## 🎯 Changes

This adds an adapter for `partykit`.

### Exports

- `tRPCPartyServer` as an extension of `Party.Server`, to be used in its place
- `applyParty` accepts an options object with: instance (instance of tRPCPartyServer), router, prefix, createContext, createWssContext, onError

## ✅ Checklist

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -68,6 +68,11 @@
       "require": "./dist/adapters/node-http/index.js",
       "default": "./dist/adapters/node-http/index.js"
     },
+    "./adapters/partykit": {
+      "import": "./dist/adapters/partykit/index.mjs",
+      "require": "./dist/adapters/partykit/index.js",
+      "default": "./dist/adapters/partykit/index.js"
+    },
     "./adapters/standalone": {
       "import": "./dist/adapters/standalone.mjs",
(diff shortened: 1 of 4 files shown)


KATT (commented):
Hey!

Thanks so much for this PR!

For now, we don't want to add more adapters to tRPC itself - they're quite a bit of work to maintain. Instead, please create a userland package called `trpc-partykit` or something of the like.

If we can get the team at @partykit to maintain and/or sponsor maintenance of this, we'd be open to adding it as a first-class citizen.

overra:
> For now, we don't want to add more adapters to tRPC itself - they're quite a bit of work to maintain. Instead, please create a userland package called `trpc-partykit` or something of the like.

I didn't think about the additional maintenance overhead, and I appreciate the way you addressed this. I noticed the closed bun adapter PR w/ a userland package link, and wasn't sure what prompted them to go that route. 

> If we can get the team at @partykit to maintain and/or sponsor maintenance of this, we'd be open to adding it as a first-class citizen.

👀 Either way, I'm down! I'm just excited to combine the two.

PR #5537, opened 2024-03-04
feat(next/app-dir): `contextCache` to control context values to add to `cacheTag`

dalechyn:
Dup of #5458

## 🎯 Changes

## ✅ Checklist


juliusmarminge (commented):
Reverted a bunch of unrelated changes

Looks good and works as expected. One thought though, we might have some procedures that are fine to cache between users. Should it be possible to filter based on the procedure too?

dalechyn:
> Reverted a bunch of unrelated changes
> 
> Looks good and works as expected. One thought though, we might have some procedures that are fine to cache between users. Should it be possible to filter based on the procedure too?

good catch.

I can imagine the selector returning an array of procedures with the context selection to cache.

```ts
contextSelector: (ctx) => [{ procedure: 'greeting', select: [ctx.session?.user.id, ctx._userIdMock] }]
```

Such probably would work and is possible to make typesafe pretty easily.

But I don't think it is an ideal approach though. One would have to update such for every procedure he has and would have to keep them updated if the context changes, in a totally separate location.

I think an ideal DX would be to select the context at the procedure declaration level:
Taking the `greeting` declaration from the example: https://github.com/trpc/trpc/blob/955edd79adaf441bfabc1f45bb83fc74355568b5/examples/.experimental/next-app-dir/src/server/routers/_app.ts#L28-L39

It would make so much more sense to have the context selection defined at that level.
```ts
greeting: publicProcedure
  .input(
    z.object({
      text: z.string(),
    }),
  )
  /* `context` is an optional procedure building block that narrows down the context
   * to pass only the values the procedure depends on, and we "might" be able to generate
   * the `cacheTag` at this point specifically.
   * If user doesn't use the `context` function, we might either give the full unselected context
   * or don't define the `ctx` on the query level at all – can't use without selection.
   */
  .context(ctx => ({ trpcSource: ctx.headers?.['x-trpc-source'] }))
  .query(async ({input, ctx: { trpcSource }) => {
    console.log('request from', trpcSource);
    return `hello ${input.text} - ${Math.random()}`;
  }),
```

That would however introduce a breaking change, but the migration would be super easy. Just throw in `.context(ctx=>ctx)` everywhere.

juliusmarminge:
We could pass the `op` to the selector fn, so it'd be like

```ts
contextSelector: (ctx, op) => {
  if (op.path.startsWith("something") return [];
  return [ctx.user.id]
}
```

dalechyn:
> We could pass the `op` to the selector fn, so it'd be like
> 
> ```ts
> contextSelector: (ctx, op) => {
>   if (op.path.startsWith("something") return [];
>   return [ctx.user.id]
> }
> ```

i think this would work for the first iteration

examples/.experimental/next-app-dir/src/trpc/server-invoker.ts
```diff
@@ -1,35 +1,31 @@
...
+import { createContext } from './shared-server';
 
 /**
  * This client invokes procedures directly on the server without fetching over HTTP.
  */
 export const api = experimental_createTRPCNextAppDirServer<typeof appRouter>({
   config() {
     return {
+      createContext: () => createContext('invoke'),
+      contextSelector: (ctx, callOpts) => {
```
juliusmarminge, line 15: I think this example should show a "private by default" and opt-in to sharing cache for specific procedures thst doesnt have sensitive data

packages/client/src/internals/getAbortController.ts
```diff
@@ -8,11 +8,10 @@ export function getAbortController(
     return customAbortControllerImpl;
   }
 
-  // eslint-disable-next-line @typescript-eslint/prefer-optional-chain
   if (typeof window !== 'undefined' && window.AbortController) {
     return window.AbortController;
   }
-  // eslint-disable-next-line @typescript-eslint/prefer-optional-chain
```
juliusmarminge, line 15: ???

packages/client/src/links/internals/getTextDecoder.ts
```diff
@@ -7,11 +7,10 @@ export function getTextDecoder(
     return customTextDecoder;
   }
 
-  // eslint-disable-next-line @typescript-eslint/prefer-optional-chain
   if (typeof window !== 'undefined' && window.TextDecoder) {
     return new window.TextDecoder();
   }
-  // eslint-disable-next-line @typescript-eslint/prefer-optional-chain
```
juliusmarminge, line 14: ???

packages/next/src/app-dir/server.ts
```diff
@@ -26,31 +28,46 @@ import { cache } from 'react';
...
+>(opts: CreateTRPCNextAppRouterServerOptions<TRouter>) {
+  const getClient = cache(async (callOpts: ProxyCallbackOptions) => {
     const config = opts.config();
-    return createTRPCUntypedClient(config);
+    const client = createTRPCUntypedClient(config);
+    const runtime = client.runtime as NextAppDirRuntime<TRouter>;
+
+    const ctx = await config.createContext();
+    runtime.ctx = ctx;
+    runtime.cacheTagSeparators = config.contextSelector?.(ctx, callOpts) ?? [];
```
juliusmarminge, line 51: I dont think we can cache this anymore when it's procedure-specific, it should be invoked every call i think

juliusmarminge (changes requested):

packages/client/src/internals/getAbortController.ts
dalechyn: `pnpm lint-fix` did that, i undoed the changes and a workflow applied formatting once again 😵‍💫

packages/client/src/links/internals/getTextDecoder.ts
dalechyn: https://github.com/trpc/trpc/commit/68429bba8efcf25905acd8cf1789466b88df329e

examples/.experimental/next-app-dir/src/trpc/server-invoker.ts
dalechyn: Good thoughts, will implement

examples/.experimental/next-app-dir/src/trpc/server-invoker.ts
```diff
@@ -13,9 +13,8 @@ export const api = experimental_createTRPCNextAppDirServer<typeof appRouter>({
     return {
       createContext: () => createContext('invoke'),
       contextSelector: (ctx, callOpts) => {
-        if (callOpts.path[0] === 'privateGreeting')
-          return [ctx.session?.user.id, ctx._userIdMock];
-        return [];
+        if (!['privateGreeting'].includes(callOpts.path[0])) return [];
```
juliusmarminge, line 16: a comment would be nice - i don't know if even i would know what this does if i revisit this example in a month or 2 ;)

packages/next/src/app-dir/server.ts
```diff
@@ -25,35 +26,42 @@ import {
...
 export function experimental_createTRPCNextAppDirServer<
   TRouter extends AnyRouter,
->(opts: CreateTRPCNextAppRouterOptions<TRouter>) {
-  const getClient = cache(() => {
+>(opts: CreateTRPCNextAppRouterServerOptions<TRouter>) {
+  return createRecursiveProxy(async (callOpts) => {
     const config = opts.config();
-    return createTRPCUntypedClient(config);
-  });
+    const client = createTRPCUntypedClient(config);
```
juliusmarminge, line 47: this can still be deduped as it was before, no?

pnpm-lock.yaml
juliusmarminge, line 1: revert lock changes

juliusmarminge (commented):
Changes overall looks good to me.

While we are considering deprecating all of this stuff in favor of the experimental_caller() approach (#5589), we could merge this in the interrim after resolving the few nits I left

packages/next/src/app-dir/server.ts
dalechyn: done!

PR #5542, opened 2024-03-04
Add example of validation using @robolex/sure

nicu-chiciuc:
## 🎯 Changes

Updates `www/docs/server/validators.md` with example of using `@robolex/sure`

Add basic snapshot test in `packages/tests/server/validators.test.ts`

## ✅ Checklist

### DIFF
--- a/www/docs/server/validators.md
+++ b/www/docs/server/validators.md
@@ -402,7 +402,7 @@ export const err = (schema) => (input) => {
 };
 ```
 
-```ts twoslash
+```ts
 import { err, object, string } from '@robolex/sure';
 import { initTRPC } from '@trpc/server';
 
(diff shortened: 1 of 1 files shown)


packages/tests/package.json
```diff
@@ -13,6 +13,7 @@
     "@fastify/busboy": "^2.0.0",
     "@fastify/websocket": "^7.1.2",
     "@miniflare/core": "^2.9.0",
+    "@robolex/sure": "^0.11.2",
     "@tanstack/react-query": "^5.0.0",
```
KATT, line 17: Please don't add this library as an officially supported one before it reaches `1.0.0`, I don't want us to maintain random breaking changes

If you want, just remove it from the test suite and have it in the docs without using twoslash

KATT (changes requested):
Thanks! 🙏

packages/tests/package.json
nicu-chiciuc: Got it, removed the tests and twoslash

juliusmarminge (approved):

PR #5579, opened 2024-03-22
chore: bump pnpm

KATT:
bumps pnpm and fix the lockfile that is regenerated on every install


pnpm-lock.yaml
```diff
@@ -1256,8 +1256,8 @@ importers:
         specifier: ^2.8.8
         version: 2.8.8
       prisma:
-        specifier: https://registry.npmjs.com/prisma/-/prisma-5.8.1.tgz?id=%40examplesnext-websockets-starter
-        version: '@registry.npmjs.com/prisma/-/prisma-5.8.1.tgz?id=%2525252525252525252525252525252525252525252525252525252525252525252525252525252540examplesnext-websockets-starter'
```
KATT, line 1260: this stopped happening on every install when dropping the `@` sign in the examples

likely a bug in pnpm

PR #5584, opened 2024-03-26
feat(server): add `experimental_caller()`

KATT:
Closes #5568

## 🎯 Changes

Adds `experimental_caller`


juliusmarminge (approved):
🚀

juliusmarminge:
I don't think this quite solves the problem if the action is invoked through a form though right? FormData transformation isn't handled anywhere. I guess I could fix that with a middleware?

KATT:
> I don't think this quite solves the problem if the action is invoked through a form though right? FormData transformation isn't handled anywhere. I guess I could fix that with a middleware?

formdata transformation is handled already in `formDataToObject` https://github.com/trpc/trpc/tree/next/packages/next/src/app-dir

PR #5587, opened 2024-03-26
feat(next app dir): rethrow built-in Next errors

KATT:
Closes #

## 🎯 Changes

- Adds ability to automatically rethrow built-in Next.js errors in `experimental_createServerActionHandler()` 
- adds an `onError`-callback
- adds support for `experimental_caller` in server actions (I think!?)
- makes `createContext` in server actions optional if you have no context defined

### DIFF
--- a/packages/next/src/app-dir/server.ts
+++ b/packages/next/src/app-dir/server.ts
@@ -155,13 +155,15 @@ export function experimental_createServerActionHandler<
           rawInput = transformer.input.deserialize(rawInput);
         }
 
-        const data = await proc({
-          input: undefined,
-          ctx,
-          path: 'serverAction',
-          getRawInput: async () => rawInput,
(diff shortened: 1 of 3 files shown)


packages/next/src/app-dir/server.ts
```diff
@@ -64,6 +67,15 @@ export function experimental_createTRPCNextAppDirServer<
...
 
+/**
+ * Rethrow errors that should be handled by Next.js
+ */
+const rethrowNextErrors = (error: TRPCError) => {
+  const { cause } = error;
+  if (isRedirectError(cause) || isNotFoundError(cause)) {
+    throw error.cause;
+  }
+};
```
KATT, line 78: cc @juliusmarminge you probably want some variant of this if you're throwing errors in RSC

PR #5593, opened 2024-03-27
return redirect

juliusmarminge:
cause next's way of throwing causes the compiler to lie to us and easy to cause typemismatch on frontend

This way we get the proper returntype on the client, namely `undefined`
CleanShot 2024-03-27 at 12 25 35

The main difference is that you have to **return** the redirect since it doesn't throw inside the action.

### DIFF
--- a/packages/server/src/adapters/next-app-dir/redirect.ts
+++ b/packages/server/src/adapters/next-app-dir/redirect.ts
@@ -22,6 +22,7 @@ export class TRPCRedirectError extends TRPCError {
  * This provides better typesafety than the `next/navigation`'s `redirect()` since the action continues
  * to execute on the frontend even if Next's `redirect()` has a return type of `never`.
  * @public
+ * @remark You should only use this if you're also using `nextAppDirCaller`.
  */
 export const redirect = (url: URL | string, redirectType?: RedirectType) => {
   // We rethrow this internally so the returntype on the client is undefined.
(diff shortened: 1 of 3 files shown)


packages/server/src/adapters/next-app-dir/redirect.ts
```diff
@@ -1,27 +1,30 @@
...
 import { TRPCError } from '../../@trpc/server';
 
 /**
  * @internal
  */
 export class TRPCRedirectError extends TRPCError {
   public readonly args;
-  constructor(...args: Parameters<typeof __redirect>) {
-    const [url] = args;
+  constructor(url: URL | string, redirectType?: RedirectType) {
```
KATT, line 9: why this change?

packages/server/src/adapters/next-app-dir/redirect.ts
```diff
@@ -1,27 +1,30 @@
...
+ * This provides better typesafety than the `next/navigation`'s `redirect()` since the action continues
+ * to execute on the frontend even if Next's `redirect()` has a return type of `never`.
  * @public
+ * @remark You should only use this if you're also using `nextAppDirCaller`.
  */
-export const redirect: typeof __redirect = (...args) => {
-  throw new TRPCRedirectError(...args);
+export const redirect = (url: URL | string, redirectType?: RedirectType) => {
+  // We rethrow this internally so the returntype on the client is undefined.
+  return new TRPCRedirectError(url, redirectType) as unknown as undefined;
```
KATT, line 29: I like this

KATT (approved):

packages/server/src/adapters/next-app-dir/redirect.ts
KATT: nice

PR #5616, opened 2024-04-08
fix(client): createInnerProxy exclude apply bind and call

bluecco:
Closes #

## 🎯 Changes

bug fix: check `.call` and `apply` method for `createInnerProxy`

followup of: https://github.com/trpc/trpc/pull/3973

## ✅ Checklist

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/createProxy.ts
+++ b/packages/server/src/unstable-core-do-not-import/createProxy.ts
@@ -20,13 +20,20 @@ function createInnerProxy(callback: ProxyCallback, path: string[]) {
     },
     apply(_1, _2, args) {
       const lastOfPath = path[path.length - 1];
-      const isCallBindApply =
-        lastOfPath === 'apply' ||
-        lastOfPath === 'bind' ||
-        lastOfPath === 'call';
+
+      const isCall = lastOfPath === 'call';
+      if (isCall) {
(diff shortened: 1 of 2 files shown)


KATT (commented):
What does this fix? None of our stuff relies on `.bind()` afaik

bluecco:
hi @KATT, this prevent `trpc` to break when using polyfill for async or `regenerator-runtime` package
in that case, `.call` is used to invoke functions.

never had `.bind` too, is to cover all the cases possible (was also mentioned here [https://github.com/trpc/trpc/pull/3973#issuecomment-1468611461](#3973) but never answered)

KATT:
Ok, I'm not sure this is working properly though, please add tests 🙏

bluecco:
@KATT added tests and updated the logic in `createProxy.ts`
image

`.call` needs to send an array as args (otherwise it would fail in `createTRPCClient.ts`)

also removed the check for `.bind()`, as you said nothing relies on that (checked codebase and did multiple tests, ok to remove it)

KATT (approved):

KATT:
If codecov has gone down, does the test actually test the new thing? 👀

KATT:
cc @bluecco I think I fixed codecov reporting. I can't update the PR as you haven't allowed edits from contributors

PR #5653, opened 2024-04-21
fix: better error for invalid content type

juliusmarminge:
improves error for easier debuggability when no content type handler is found

### DIFF
--- a/packages/server/src/adapters/content-handlers/selectContentHandlerOrUnsupportedMediaType.ts
+++ b/packages/server/src/adapters/content-handlers/selectContentHandlerOrUnsupportedMediaType.ts
@@ -12,11 +12,10 @@ export function selectContentHandlerOrUnsupportedMediaType<
 
   for (const h of handlers) {
     const match = h.isMatch(opts);
+    received = match.received;
     if (match.match) {
       handler = h;
       break;
-    } else {
-      received = match.received;
     }
   }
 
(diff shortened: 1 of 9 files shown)


packages/tests/server/adapters/standalone.test.ts
```diff
@@ -207,7 +207,7 @@ test('force content-type on mutations', async () => {
             "path": "mut",
             "stack": "[redacted]",
           },
-          "message": "Invalid Content-Type header. This request may not be supported by your tRPC Adapter, or possibly by tRPC at all",
+          "message": "Invalid Content-Type header 'undefined'. This request may not be supported by your tRPC Adapter, or possibly by tRPC at all",
```
KATT, line 210: Maybe another error than "undefined" for missing ones? 👀

KATT (approved):

packages/server/src/adapters/content-handlers/selectContentHandlerOrUnsupportedMediaType.ts
```diff
@@ -1,18 +1,30 @@
...
+  let received: Maybe<string> = undefined;
+
+  for (const h of handlers) {
+    const match = h.isMatch(opts);
+    received = match.received;
+    if (match.match) {
+      handler = h;
+      break;
+    }
+  }
```
Nick-Lucas, line 20: As discussed in discord, I think best to do this in a way which is fairly linear. iterating over and taking a match object from the last in the chain seems to create complexity

We probably want to already know the header at this level in the stack and just ask the content handler if it can handle that type?

PR #5654, opened 2024-04-21
chore: better error for mismatched content-type

juliusmarminge:
Closes #5653

- improves error by providing the received content type to the error
- uses `Headers` (Node18+) whose `.get()` is case-insensitive

### DIFF
--- a/packages/server/src/adapters/content-handlers/selectContentHandlerOrUnsupportedMediaType.ts
+++ b/packages/server/src/adapters/content-handlers/selectContentHandlerOrUnsupportedMediaType.ts
@@ -6,6 +6,8 @@ import type {
 
 interface MinimalHandlerOpts {
   req: {
+    // TODO: This could probably only take Headers and do the conversion higher up,
+    //       but that's a bigger refactor for another day
     headers: HTTPHeaders | Headers;
   };
 }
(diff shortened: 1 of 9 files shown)


packages/server/src/adapters/content-handlers/selectContentHandlerOrUnsupportedMediaType.ts
```diff
@@ -1,20 +1,35 @@
 import { TRPCError } from '../../@trpc/server';
-import type { BaseContentTypeHandler } from '../../@trpc/server/http';
+import type {
+  BaseContentTypeHandler,
+  HTTPHeaders,
+} from '../../@trpc/server/http';
+
+interface MinimalHandlerOpts {
+  req: {
+    headers: HTTPHeaders | Headers;
```
juliusmarminge, line 9: This could probably only take `Headers` and do the conversion higher up, but that's a bigger refactor for another day

packages/server/src/adapters/content-handlers/selectContentHandlerOrUnsupportedMediaType.ts
```diff
@@ -1,20 +1,35 @@
...
+  };
+}
 
 export function selectContentHandlerOrUnsupportedMediaType<
-  THandlerOpts,
+  THandlerOpts extends MinimalHandlerOpts,
   THandler extends BaseContentTypeHandler<THandlerOpts>,
 >(handlers: THandler[], opts: THandlerOpts) {
-  const handler = handlers.find((handler) => handler.isMatch(opts));
+  const contentType = new Headers(opts.req.headers as HeadersInit).get(
```
juliusmarminge, line 17: this cast is fine, idk why it complains that you can't pass a string[] in ther...

CleanShot 2024-04-21 at 11 02 14@2x

Nick-Lucas, line 17: Yeah this is way better

packages/server/src/adapters/content-handlers/selectContentHandlerOrUnsupportedMediaType.ts
```diff
@@ -1,20 +1,35 @@
...
+    'content-type',
+  );
+
+  const error = new TRPCError({
```
Nick-Lucas, line 21: Maybe we should still only create this when we actually need it? Unnecessary to create it for every request

packages/server/src/adapters/content-handlers/selectContentHandlerOrUnsupportedMediaType.ts
Nick-Lucas: Yeah that's fine, limited number of these types right now so we can wrangle it here

packages/server/src/adapters/content-handlers/selectContentHandlerOrUnsupportedMediaType.ts
Nick-Lucas: Does Lambda Adapter definitely behave okay? That has mixed case headers

packages/server/src/adapters/content-handlers/selectContentHandlerOrUnsupportedMediaType.ts
```diff
@@ -1,18 +1,42 @@
...
+  const contentType = headers.get('content-type');
+
+  if (contentType === null) {
     return [
       undefined,
       new TRPCError({
         code: 'UNSUPPORTED_MEDIA_TYPE',
         message:
-          'Invalid Content-Type header. This request may not be supported by your tRPC Adapter, or possibly by tRPC at all',
+          'No Content-Type header detected on the incoming request. This request may not be supported by your tRPC Adapter, or possibly by tRPC at all',
```
juliusmarminge, line 28: ref https://github.com/trpc/trpc/pull/5653#discussion_r1573672026

Nick-Lucas (approved):

packages/server/src/adapters/content-handlers/selectContentHandlerOrUnsupportedMediaType.ts
juliusmarminge: if it adheres to the current type (Either `Record<string, string>` or `Record<string, string[]>` it should work. Otherwise we have invalid types somewhere else

PR #5668, opened 2024-04-24
feat(react-query): add support for global onSuccess for mutation

tomer-yechiel:
Closes #

## 🎯 Changes

fix the issue that global onSuccess doesnt get called

for example when constructing query client the following
```
new QueryClient({
      defaultOptions: {
        mutations:{
          onSuccess: () => console.log('')
        }
      },
    })
```

this is a bug fix breaking change.

please let me know if it acceptable breaking change and if so i will add documentations and tests

## ✅ Checklist


KATT (commented):
this looks good, but could you add a failing test case?

tomer-yechiel:
> this looks good, but could you add a failing test case?

done

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -241,7 +241,10 @@ export function createRootHooks<
     const queryClient = useQueryClient();
 
     const mutationKey = [path];
-    const defaultOpts = queryClient.getMutationDefaults(mutationKey);
+
+    const defaultOpts = queryClient.defaultMutationOptions(
+      queryClient.getMutationDefaults(mutationKey),
+    );
```
KATT, line 247: cc @TkDodo 

this looks very odd to me, is this really how to fetch mutation options? it feels like we're fetching defaults twice

KATT (commented):
looks good @tomer-yechiel, do we wanna do the same for queries?

tomer-yechiel:
> looks good @tomer-yechiel, do we wanna do the same for queries?

i dont think there is a similar callback for queries

packages/react-query/src/shared/hooks/createHooksInternal.tsx
tomer-yechiel: queryClient.defaultMutationOptions can accept defaults as a parameter and it merge the parameters it accept with the global default giving precedence to the accepted parameters

KATT:
I'm happy to merge this as soon as the tests pass

```

[1] server/react/queryClientDefaults.test.tsx:142:50 - error TS2554: Expected 0 arguments, but got 1.
[1] 
[1] 142       expect(onSuccessLocalSpy).toHaveBeenCalled(1);
[1]                                                      ~
```

tomer-yechiel:
> I'm happy to merge this as soon as the tests pass
> 
> ```
> 
> [1] server/react/queryClientDefaults.test.tsx:142:50 - error TS2554: Expected 0 arguments, but got 1.
> [1] 
> [1] 142       expect(onSuccessLocalSpy).toHaveBeenCalled(1);
> [1]                                                      ~
> ```

done.

KATT (approved):

packages/react-query/src/shared/hooks/createHooksInternal.tsx
TkDodo: `queryClient.defaultMutationOptions` looks up the default options internally and merges with the passed options, so this shouldn't be necessary. What we do internally is this:

```
options: client.defaultMutationOptions(options),
```

so if you only want the default options, without passing anything, you'd need to do:

```
const defaultOpts = queryClient.defaultMutationOptions()
```

tomer-yechiel: > so if you only want the default options, without passing anything, you'd need to do:
> 
> ```
> const defaultOpts = queryClient.defaultMutationOptions()
> ```

i dont have much experience with react query, isn't this going to break users with defaults on the mutationKey?

PR #5670, opened 2024-04-25
Export `TRPCBuilder` class from `@trpc/server` package

### LINKED ISSUE #5666: feat: Expose `TRPCBuilder` class or interface

I'd like to be able to use TRPC in vanilla JavaScript using JSDoc and TypeScript for type-checking. To define a context type, the TRPC API expects that we use type parameters like so:

```typescript
const t = initTRPC.context<MyAppContext>().create();
```

This is convenient when using TypeScript but not usable in JavaScript to get correctly-typed contexts. In many cases, the absence of type parameters in some APIs can be compensated for by specifying the type of the variable that gets the return value. e.g.:

```typescript
// TypeScript library
declare function example<ExplicitType>(): MyCollection<ExplicitType>

// JavaScript call site
/** @type {MyCollection<ExplicitType>} */
const collection = example();
```

Unfortunately, the type returned by `initTRPC.context()` (`TRPCBuilder<...>`) is not exported so that's not an option and if I instead annotate the type returned by `.create()` after it, the return type of `create` is an object literal that's even more verbose so annotating that type would be extremely cumbersome and defeats the purpose of the existing API.

Export the `TRPCBuilder` class or at least its type as that would make it trivial to annotate the context type with JSDoc like so:

```javascript
import { initTRPC } from "@trpc/server";

/** @type {import("@trpc/server").TRPCBuilder<{ ctx: MyAppContext }>} */
const init = initTRPC.context();

const t = init.create(); // explicitly typed context for procedures, middleware and subscriptions.
```

Explicitly define and export the type returned by `TRPCBuilder#create()` so the initial workaround I described could work 🤷?

### Additional information

This was requested before in #4145 but was closed because no example use case was provided. Some example was provided later but the thread was locked shortly after. Since type-checking JavaScript with JSDoc annotations leverages TypeScript's excellent type checker without build tools, this seems like a valid use case to support.

### 👨‍👧‍👦 Contributing

opiation:
Closes #5666

## 🎯 Changes

This change exports the `TRPCBuilder` class from `@trpc/server` package.

I don't know what the reasoning is specfically for keeping the entire `TRPCBuilder` class private, but given that exposing instances of that class to the user is required in order to use it, keeping the class private seems like it might have been to coerce users into using the `initTRPC....` functions instead of instantiating the class directly.

For this reason, I exported only the type of the class so that the type could be used to explicitly define context and/or meta types in JavaScript land...

```js

/** @type {import("@trpc/server").TRPCBuilder<{ ctx: MyContext }>} */
const init = initTRPC.context();

const t = init.create();
// Define procedures on `t` like usual with a context of type
// `MyContext`...
````

Since the main purpose of exporting the `TRPCBuilder` class is to allow JS users to specify the builder type returned from `initTRPC` in their code, it seemed reasonable to colocate the type with the `initTRPC` singleton.

If exporting the whole class would be preferable for some reason, that seems entirely feasible and does not expose any new API surface area since the class has no constructor.

## ✅ Checklist

- [x] ~~I have added or updated the tests related to the changes made.~~

### DIFF
--- a/packages/server/src/@trpc/server/index.ts
+++ b/packages/server/src/@trpc/server/index.ts
@@ -35,6 +35,7 @@ export {
   type MutationProcedure as TRPCMutationProcedure,
   type QueryProcedure as TRPCQueryProcedure,
   type SubscriptionProcedure as TRPCSubscriptionProcedure,
+  type TRPCBuilder,
 } from '../../unstable-core-do-not-import';
 
 export type {
--- a/packages/server/src/unstable-core-do-not-import/initTRPC.ts
+++ b/packages/server/src/unstable-core-do-not-import/initTRPC.ts
@@ -149,3 +149,4 @@ class TRPCBuilder<TContext extends object, TMeta extends object> {
  * @link https://trpc.io/docs/v11/quickstart
  */
 export const initTRPC = new TRPCBuilder();
+export type { TRPCBuilder };


KATT (approved):
We try to only export what's needed by the community, that way we can refactor other stuff whenever we want without risk of breaking people's apps

opiation:
> We try to only export what's needed by the community, that way we can refactor other stuff whenever we want without risk of breaking people's apps

That was my assumption as well. Thanks for clarifying and exporting this 🙏!

PR #5691, opened 2024-04-28
chore: move complexity from `resolveResponse`

KATT:
- move some complexity from `resolveResponse` to the content type handlers
- basically, content types does the request planning and returns the calls that are made and getters for their inputs

not sure if you prefer this @Nick-Lucas


packages/server/src/unstable-core-do-not-import/http/types.ts
```diff
@@ -60,7 +62,17 @@ export type ProcedureCall = {
...
   isBatchCall: boolean;
-  calls: ProcedureCall[];
+  calls: {
+    path: string;
+    /**
+     * Read the raw input (deduped and memoized)
+     */
+    getRawInput: () => Promise<unknown>;
+    /**
+     * Get already parsed inputs - won't trigger reading the body or parsing the inputs
```
KATT, line 72:
```suggestion
     * Get already parsed input for the procedure - won't trigger reading the body or parsing the inputs
```

Lilja:
As written in discord, but moving to github for posterity:

I tried the latest build and it does not work for me. The response time was increased to 600-700ms(from 200-300ms) and did not return any json data in the response.

`11.0.0-alpha-tmp-04-26-resolveResponse-and-multi.428` breaks my fastify backend

`11.0.0-alpha-tmp-04-26-resolveResponse-and-multi.427` works all fine an dandy

PR #5711, opened 2024-05-09
feat(server + client): subscriptions over HTTP

KATT:
- Fork of #5700
- Ref #544 (we'll probably make this into SSE)

## 🎯 Changes

Adds a `httpSubscriptionLink`


KATT:
See #5713

PR #5713, opened 2024-05-10
feat(server + client): add support for server-sent events (SSE)

### LINKED ISSUE #544: Consider SSE to enable server->client http subscriptions

KATT:
- Fork of #5700
- Closes #544

## 🎯 Changes

Adds a `httpSubscriptionLink` using server-sent events (SSE)

**Docs:**  https://www-git-05-10-subscriptions-sse-trpc.vercel.app/docs/client/links/httpSubscriptionLink
## ✅ Checklist


Sheraff (commented):
This looks dope but I don't feel qualified to give much of a green flag here, having never played with SSE

packages/server/src/unstable-core-do-not-import/stream/sse.ts
```diff
@@ -0,0 +1,258 @@
...
+          }, 1000).unref()
+        : null;
+
+      return {
+        promise: deferred.promise,
+        clear: () => {
+          if (timeout) {
+            clearTimeout(timeout);
+          }
+        },
```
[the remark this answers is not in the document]
KATT: Honestly, I am unsure

helmturner:
Pulled this down to take a look, but apparently I can't get the install to succeed? Been a while since I've messed around in this codebase, but the error I'm seeing is one I recently saw on another codebase (and never did find a solution for):

>└─ Failed in 1.2s at /Users/alec/dev/trpc-2
examples/next-prisma-starter postinstall$ pnpm generate
[12 lines collapsed]
│ import { PrismaClient } from '@prisma/client/edge'
│ const prisma = new PrismaClient()
│ ```
│ See other ways of importing Prisma Client: http://pris.ly/d/importing-client
│ ┌────────────────────────────────────────────────────────────────┐
│ │  Supercharge your Prisma Client with global database caching,  │
│ │  scalable connection pooling and real-time database events.    │
│ │  Explore Prisma Accelerate: https://pris.ly/cli/-accelerate    │
│ │  Explore Prisma Pulse: https://pris.ly/cli/-pulse              │
│ └────────────────────────────────────────────────────────────────┘

Any idea what that's about?

KATT:
> Pulled this down to take a look, but apparently I can't get the install to succeed? Been a while since I've messed around in this codebase, but the error I'm seeing is one I recently saw on another codebase (and never did find a solution for):
> 
> > └─ Failed in 1.2s at /Users/alec/dev/trpc-2
> > examples/next-prisma-starter postinstall$ pnpm generate
> > [12 lines collapsed]
> > │ import { PrismaClient } from '@prisma/client/edge'
> > │ const prisma = new PrismaClient()
> > │ ```
> > │ See other ways of importing Prisma Client: http://pris.ly/d/importing-client
> 
> Any idea what that's about?
> 
> EDIT: I think this line is the key: `│ ERROR: "codegen:tests" exited with 1.`

@helmturner try `pnpm clean && pnpm i`

helmturner:
I'm totally blocked. I've read some of it, but every time I use the `turbo` cli I get an error (which means I can't even get past install because of the post-intstall script).

```sh
tsx scripts/addPrismaOverrides.ts

  × failed to execute local turbo process
  ╰─▶ Malformed Mach-o file (os error 88)

 ELIFECYCLE  Command failed with exit code 1.
  × failed to execute local turbo process
  ╰─▶ Malformed Mach-o file (os error 88)

 ELIFECYCLE  Command failed with exit code 1.
 ELIFECYCLE  Command failed with exit code 143.
ERROR: "codegen:tests" exited with 1.
 ELIFECYCLE  Command failed with exit code 1.
```

@KATT Any idea what that is? It happens any time I try to use the `turbo` cli. I've tried uninstalling/reinstalling turbo, but no luck.

examples/next-prisma-sse-subscriptions/src/server/routers/post.ts
```diff
@@ -0,0 +1,194 @@
...
+
+        lastEventId = id;
+      };
+
+      // if someone is typing, emit event immediately
+      yield* maybeYield(currentlyTyping);
+
+      for await (const [who] of dbEvents.toIterable('isTypingUpdate')) {
+        yield* maybeYield(who);
+      }
```
BeBoRE, line 192: Are you using this just to test out iterables or does this pattern have benefits over using observable and eventListeners in this situation?

Adrastopoulos:
Would it be possible to pass in our own Eventsource implementation so we can customize initial headers sent, among other things? https://github.com/EventSource/eventsource?tab=readme-ov-file#extensions-to-the-w3c-api

packages/server/src/unstable-core-do-not-import/stream/sse.test.ts
```diff
@@ -139,3 +141,162 @@ test('e2e, server-sent events (SSE)', async () => {
...
+      i++;
+      yield {
+        id: i,
+        data: i,
+      } satisfies SSEvent;
+    }
+    while (true) {
+      // yield 2 events at a time to test if the client will get both without reconnecting in between
+      yield* yieldEvent();
+      yield* yieldEvent();
```
Sheraff, line 159: Just out of curiosity, why are you forwarding the yield here? Since `yieldEvent` only yields once, this looks equivalent to a regular function call, isn't it?
```suggestion
    function yieldEvent() {
      i++;
      return {
        id: i,
        data: i,
      } satisfies SSEvent;
    }
    while (true) {
      // yield 2 events at a time to test if the client will get both without reconnecting in between
      yield yieldEvent();
      yield yieldEvent();
```

(TBH I've never actually had a use-case for forwarding an iterator with `yield*` so I might be missing a nuance here)

packages/server/src/unstable-core-do-not-import/stream/sse.ts
```diff
@@ -147,9 +154,19 @@ export function sseStreamProducer(opts: {
       }
 
       stream.controller.enqueue(chunk);
+
+      if (opts.emitAndEndImmediately) {
+        // end the stream in the next tick so that we can send a few more events from the queue
+        setTimeout(maxDurationPromise.resolve, 1);
```
Sheraff, line 160: I'm not sure I understand the `1` here. Would this be a use-case for `setImmediate`? Or just `setTimeout(..., 0)`?

packages/server/src/unstable-core-do-not-import/stream/sse.test.ts
KATT: Just felt cool to use that feature haha

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -60,14 +79,15 @@ export function unstable_httpSubscriptionLink<
                 eventSource,
               },
             });
-            // console.log('started', new Date());
-            eventSource.removeEventListener('open', onStarted);
+
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            eventSource!.removeEventListener('open', onStarted);
```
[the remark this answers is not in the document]
KATT: I think this matches behavior of wss but I am not sure we need onStarted at all in the first place

www/docs/community/awesome-trpc.mdx
```diff
@@ -71,27 +71,27 @@ A collection of resources on tRPC.
 
 ## 🍀 Starting points, example projects, etc
 
-| Description                                                                                             | Link                                                            |
-| ------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------- |
-| **Recommended:** Starter project with Prisma, Next.js, tRPC, E2E-testing                                | https://github.com/trpc/examples-next-prisma-starter            |
-| **create-t3-turbo** - Clean and simple starter repo using the T3 Stack along with Expo React Native     | http://github.com/t3-oss/create-t3-turbo                        |
-| **create-t3-app** - Scaffold a starter project using the T3 Stack (Next.js, tRPC, Tailwind CSS, Prisma) | https://create.t3.gg                                            |
-| WebSockets Starter Project                                                                              | https://github.com/trpc/examples-next-prisma-starter-websockets |
```
juliusmarminge, line 79: i'd keep this reference

juliusmarminge (dismissed):
lots of unfamiliar stuff so my review should be taken lightly but couldn't find anything obvious that stands out

examples/next-sse-chat/src/server/routers/channel.ts
```diff
@@ -0,0 +1,133 @@
...
+        const id = Object.keys(who).sort().toString();
+        if (lastEventId === id) {
+          return;
+        }
+        yield {
+          id,
+          data: Object.keys(who).filter(
+            (user) => user !== opts.ctx.session?.user?.name,
+          ),
+        } satisfies SSEvent;
```
KATT, line 119: What would you think about letting you `yield` anything and have a special function when you wanna yield `SSEvent`?

e.g.

```ts
yield { foo: "bar" }
```

then `data` would simply be `"bar"`

and doing eg

```ts
yield sseEvent({
  id: 1,
  foo: "bar",
})
```
if you wanna `yield` a "proper" SSE event

JonathonRP: I like that though instead of sseEvent, how about just sse or ssEvent or serverSentEvent?

KATT: sure, not bothered about the naming

PR #5731, opened 2024-05-22
New subscription-based adapters

domlen2003:
Closes #

## 🎯 Changes
- implements a base utils for subscription-based adapters
- implements the old `ws` adapter on top of the base utils
- adds new `cloudflare-durable-objects` adapter

## ✅ Checklist


domlen2003:
Somehow it fails on client build, where I changed nothing?

domlen2003:
okay i think there is actually some help needed with setting up the build for cloudflare worker types

domlen2003:
Also im currently running into a wall with DO I/O: `Cannot perform I/O on behalf of a different Durable Object.`
Any ideas?

KATT:
This is very exciting - could you share some details on how we'd test this? I've never actually used the CF stuff here but I'm open to adding an adapter 



Got limited time over the next week to play with this

KATT:
Also as an FYI, we got #5713 pending which adds support for SSE and for subscriptions to be generator functions rather than our own observable thing. I don't know if CF has something to support SSE well but that'd be quite exciting

domlen2003:
Switching from the obervables would be a huge improvement imho, currently I recreate the subscription by sending it an empty dummy message. This is actually not optimal as it breaks the guaranty of the subscription message to me non null (if specified by the router).

domlen2003:
> could you share some details on how we'd test this?

Well cloudflare has an (experimantal?) [vitest integration](https://developers.cloudflare.com/workers/testing/vitest-integration/) tho so far i wasn't successfull in integrating it with the current test harness. 

Also i have quite limited experience with rollup and struggle with setting up the 'cloudflare:workers' imports correctly.

BeBoRE:
CloudFlare Workers do support SSE, you have to pass a ReadSteam inside of the response object that you return and you can write to this. CloudFlare limits workers to 50ms of CPU time, if you exceed this limit the worker gets terminated.

https://developers.cloudflare.com/workers/runtime-apis/streams/
https://blog.cloudflare.com/workers-ai-streaming/

Working demo:
https://github.com/BeBoRE/cloudflare-sse/
https://cloudflare-sse.bebore.workers.dev/

domlen2003:
Okay thats interesting, but would we be able to hold it for longer in a Durable Object?

dinwwwh:
CF hibernation is the most powerful (it delivers messages in real-time) and cost-efficient method. Any other method will be limited in features or costly in terms of CPU time.

domlen2003:
Yes thats my opinion too, I would recommend focussing on the hibernation websocket API since this would be the most cost effective. 
SSE on workers would probably be a nice to have but not as important.

BeBoRE:
From the code that I have read in #5713, SSE subscriptions should work out of the box with CloudFlare workers, however I couldn't get the repo to install to test it properly.

domlen2003:
It seems i have a working example rn. It actually survives hibernation and resusbcribes. Tho it's completely untested.

KATT:
Will close this as I'm cleaning up issues and it looks abandoned

If you wanna pick this up, please ping me on Discord and we can figure how to test it etc

PR #5741, opened 2024-05-26
feat(react-query): treat async iterables as arrays

KATT:
Closes #

## 🎯 Changes

When using streaming and react-query, it doesn't make sense to deal with async iterables.

Instead, here, we aggregate and update the query cache as results are propagated

Thread on discord that prompted this: https://discord.com/channels/867764511159091230/1243613771981652070

**Docs:** https://www-git-05-26-iterable-query-trpc.vercel.app/docs/client/react/useQuery#streaming

## ✅ Checklist


packages/tests/server/react/useQuery.test.tsx
```diff
@@ -194,6 +205,76 @@ describe('useQuery()', () => {
...
+        Object {
+          "data": undefined,
+          "status": "pending",
+        },
+        Object {
+          "data": Array [
+            1,
+            2,
+          ],
+          "status": "success",
```
KATT, line 257: this should be "pending"

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -181,7 +182,36 @@ export function createRootHooks<
...
+
+                const aggregate: unknown[] = [];
+                for await (const value of result) {
+                  aggregate.push(value);
+
+                  query.setState({
+                    data: [...aggregate],
+                  });
+                }
+                return aggregate;
```
KATT, line 212: cc @TkDodo does this look correct?

packages/tests/server/react/useQuery.test.tsx
```diff
@@ -194,6 +205,88 @@ describe('useQuery()', () => {
...
+        Object {
+          "data": Array [
+            1,
+            2,
+            3,
+          ],
+          "status": "success",
+        },
+      ]
+    `);
```
KATT, line 278: All the different states

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -181,7 +182,36 @@ export function createRootHooks<
...
+              }
```
KATT, line 213: cc @TkDodo does this look good to you?

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -181,7 +182,36 @@ export function createRootHooks<
...
+
+              if (isAsyncIterable(result)) {
+                const queryHash = hashKey(queryKey);
+
+                const queryCache = queryClient.getQueryCache();
+
+                const query = queryCache.build(queryClient, {
+                  queryKey,
+                  queryHash,
+                });
```
Sheraff, line 197: i've never used the `.build()` method, but for all the methods I've used, I think I never needed both the `queryHash` *and* the `queryKey`. 

Does this still work with only the `queryHash`?
```suggestion
                const query = queryCache.build(queryClient, {
                  queryHash,
                });
```

KATT: On my phone but think they were both required

Sheraff: You're right, because `.build` can "create" a query, even if it does not yet exist in the cache.

Are there methods in tRPC that could kick off a `tanstack/query` query without going through the cache manager? Like if we use `useQuery`, it immediately creates the query.

- if there are such methods, then `.build()` seems indeed appropriate
- if there aren't, maybe `const query = queryCache.get(queryHash)` could suffice

Also, maybe it doesn't matter and TkDodo can tell us more!

packages/react-query/src/shared/hooks/createHooksInternal.tsx
Sheraff, line 212: Maybe we shouldn't mutate the state once it's be handed over to tanstack/query
```suggestion
                let aggregate: unknown[] = [];
                for await (const value of result) {
                  aggregate = [...aggregate, value];

                  query.setState({
                    data: aggregate,
                  });
                }
                return aggregate;
```

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -181,7 +182,36 @@ export function createRootHooks<
...
+
+                query.setState({
+                  data: [],
+                  status: 'pending',
+                });
```
Sheraff, line 202: Would love TkDodo's opinion on this. When I was trying to think of this DX I couldn't decide between the possible "is streaming" states
- `status: 'pending', fetchStatus: 'fetching', data: <defined>`
- `status: 'success', fetchStatus: 'fetching', data: <defined>`
- `status: 'success', fetchStatus: 'idle', data: <defined>`

KATT: Hmm you might have a point - I rarely think of fetchStatus

packages/tests/server/react/useQuery.test.tsx
```diff
@@ -194,6 +205,88 @@ describe('useQuery()', () => {
...
+      // initial
+      'pending',
+      // waiting 3 values
+      'pending',
+      'pending',
+      'pending',
+      // done iterating
+      'success',
+    ]);
+  });
```
Sheraff, line 289: My proposal for that test would be to also try observing the `fetching` state in addition to the `loading` state, as it could help inform the various possible `setState` used in the queryFn

```suggestion
  test('iterable', async () => {
    const { client, App } = ctx;
    const states: Partial<QueryState>[] = [];
    function MyComponent() {
      const query1 = client.post.iterable.useQuery(undefined, {
        trpc: {
          context: {
            stream: 1,
          },
        },
      });
      states.push({
        status: query1.status,
        data: query1.data,
        fetchStatus: query1.fetchStatus
      });

      expectTypeOf(query1.data!).toMatchTypeOf<number[]>();

      return (
        <pre>
          {query1.status}:{query1.isFetching ? 'fetching' : 'notFetching'}
        </pre>
      );
    }

    const utils = render(
      <App>
        <MyComponent />
      </App>,
    );
    await waitFor(() => {
      expect(utils.container).toHaveTextContent(`success:notFetching`);
    });

    expect(states).toMatchInlineSnapshot(`
      Array [
        Object {
          "data": undefined,
          "fetchStatus": "fetching",
          "status": "pending",
        },
        Object {
          "data": Array [],
          "fetchStatus": "fetching",
          "status": "pending",
        },
        Object {
          "data": Array [
            1,
          ],
          "fetchStatus": "fetching",
          "status": "pending",
        },
        Object {
          "data": Array [
            1,
            2,
          ],
          "fetchStatus": "fetching",
          "status": "pending",
        },
        Object {
          "data": Array [
            1,
            2,
            3,
          ],
          "fetchStatus": "idle",
          "status": "success",
        },
      ]
    `);
    expect(states.map((s) => `${s.status}-${s.fetchStatus}`)).toEqual([
      // initial
      'pending-fetching',
      // waiting 3 values
      'pending-fetching',
      'pending-fetching',
      'pending-fetching',
      // done iterating
      'success-idle',
    ]);
  });
```

packages/react-query/src/shared/hooks/createHooksInternal.tsx
KATT: That's why I already make a copy when setting the state like

```ts
query.setState({
  data: [...aggregate],
});
```

Sheraff: fair enough, i must have misread

packages/react-query/src/shared/hooks/createHooksInternal.tsx
KATT: cc @TkDodo wdyt is best here?

TkDodo: in `pending` state, data should be `undefined`. That's how we have it typed, `pending` means "no data": https://github.com/TanStack/query/blob/8f55e36a7c94242176875bbb18c63ab6991c6af0/packages/query-core/src/types.ts#L651-L658

if `data` is defined, this is probably the most correct representation:

```
status: 'success', fetchStatus: 'fetching', data: <defined>
```

PR #5757, opened 2024-06-01
chore: support next15 in actions

juliusmarminge:
The helpers we imported from the internal dist path are removed in Next.js 15 RC making the current action stuff incompatible.

This little workaround should fix this. We can clean it up once 15 goes stable and we can use that as a peer dependency for these new stuff I think

### DIFF
--- a/packages/server/src/adapters/next-app-dir/rethrowNextErrors.ts
+++ b/packages/server/src/adapters/next-app-dir/rethrowNextErrors.ts
@@ -5,6 +5,7 @@ import { TRPCRedirectError } from './redirect';
 /**
  * @remarks The helpers from `next/dist/client/components/*` has been removed in Next.js 15.
  * Inlining them here instead...
+ * @see https://github.com/vercel/next.js/blob/5ae286ffd664e5c76841ed64f6e2da85a0835922/packages/next/src/client/components/redirect.ts#L97-L123
  */
 const REDIRECT_ERROR_CODE = 'NEXT_REDIRECT';
 function isRedirectError(error: unknown) {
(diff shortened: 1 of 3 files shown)


packages/server/src/adapters/next-app-dir/rethrowNextErrors.ts
```diff
@@ -1,24 +1,60 @@
-import { isNotFoundError } from 'next/dist/client/components/not-found';
-import { isRedirectError } from 'next/dist/client/components/redirect';
-import {
-  notFound as __notFound,
-  redirect as __redirect,
-} from 'next/navigation';
+import * as nextNavigation from 'next/navigation.js';
```
juliusmarminge, line 1: without the `.js` we get this error:

CleanShot 2024-06-01 at 14 35 03@2x

packages/server/src/adapters/next-app-dir/rethrowNextErrors.ts
```diff
@@ -1,24 +1,60 @@
...
 import type { TRPCError } from '../../@trpc/server';
 import { TRPCRedirectError } from './redirect';
 
 /**
- * Rethrow errors that should be handled by Next.js
+ * @remarks The helpers from `next/dist/client/components/*` has been removed in Next.js 15.
```
juliusmarminge, line 6: CleanShot 2024-06-01 at 14 37 06@2x

packages/server/src/adapters/next-app-dir/rethrowNextErrors.ts
```diff
@@ -1,24 +1,60 @@
...
+  );
+}
 
+const NOT_FOUND_ERROR_CODE = 'NEXT_NOT_FOUND';
+function isNotFoundError(error: unknown) {
+  if (typeof error !== 'object' || error === null || !('digest' in error)) {
+    return false;
+  }
+  return error.digest === NOT_FOUND_ERROR_CODE;
+}
```
KATT, line 35: Maybe we should link to where they are defined in Next.js so it'll be easier to keep them up-to-date

KATT (approved):

packages/server/src/adapters/next-app-dir/rethrowNextErrors.ts
juliusmarminge: Added

juliusmarminge: although we shouldn't have to update them as newer versions of next will just use `unstable_rethrow`. they're there just as a fallback for Next 14 users

PR #5775, opened 2024-06-05
chore: loosen some generic types

KATT:
Closes #

## 🎯 Changes

- Allows `T` / `$`  as a valid generic argument without a suffix
- Also allows e.g. `T1` / `T2` / `T3`


juliusmarminge:
doesn't this make the entire rule redundant?

KATT:
Nah, you can't do single character `U` etc

PR #5814, opened 2024-06-20
feat(react-query): add support for mutations to `createTRPCQueryUtils` and improve offline support

### LINKED ISSUE #5800: feat(react-query): `createTRPCQueryUtils` should support mutations

```
  System:
    OS: macOS 14.5
    CPU: (10) arm64 Apple M1 Max
    Memory: 4.26 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.5.1 - ~/.nvm/versions/node/v20.5.1/bin/node
    npm: 9.8.0 - ~/.nvm/versions/node/v20.5.1/bin/npm
    bun: 1.0.25 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 126.0.6478.61
    Edge: 126.0.2592.56
    Safari: 17.5
  npmPackages:
    @tanstack/react-query: ^5.39.0 => 5.45.0
    @trpc/client: next => 11.0.0-rc.403+ab71a1be6
    @trpc/react-query: next => 11.0.0-rc.403+ab71a1be6
    @trpc/server: next => 11.0.0-rc.403+ab71a1be6
    next: ^14.2.1 => 14.2.4
    react: ^18.3.0 => 18.3.1
    typescript: ^5.4.2 => 5.4.5
```

### Describe the bug

The type of `createTRPCQueryUtils({ queryClient, client }).anyRouter.anyMutationProcedure` is always `never`.

If this is intentional, I believe it's not ideal, because there are some `queryClient` methods that can be useful for mutations, such as [queryClient.setMutationDefaults](https://tanstack.com/query/v5/docs/reference/QueryClient/#queryclientsetmutationdefaults), which is crucial for implementing proper offline mutation persistence (see https://github.com/trpc/trpc/discussions/5797)

### Link to reproduction

https://github.com/simonecervini/trpc-issue-mutations-create_trpc_query_utils

### To reproduce

Open `/src/app/page.tsx` and check the types of `ThisIsCorrectlyTyped` and `ThisIsNotCorrectlyTyped`

### Additional information

Probably related: https://discord-questions.trpc.io/m/1236883887871754260

### 👨‍👧‍👦 Contributing

## Funding

<a href="https://polar.sh/trpc/trpc/issues/5800">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/5800/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

simonecervini:
Closes #5800

## 💡 Motivation

This PR is largely motivated by the need to improve offline support in tRPC. According to my research, there were two main missing pieces:

- a type-safe way to call `utils.anyRouter.anyMutationProcedure.setMutationDefaults()`
- a type-safe way to call the canonical mutation function (the one defined implicitly by tRPC) when setting `mutationFn` with `setMutationDefaults()`

I decided to go further and implement all the missing `QueryClient` methods useful to interact with mutations, and I added specific tests for offline scenarios. 

## 🎯 Changes

This PR adds to `createTRPCQueryUtils` all the utility functions for managing mutations.

There are 5 mutation methods in the `QueryClient` class:
- [isMutating](https://tanstack.com/query/latest/docs/reference/QueryClient/#queryclientismutating)
- [getMutationDefaults](https://tanstack.com/query/latest/docs/reference/QueryClient/#queryclientgetmutationdefaults)
- [setMutationDefaults](https://tanstack.com/query/latest/docs/reference/QueryClient/#queryclientsetmutationdefaults)
- [getMutationCache](https://tanstack.com/query/latest/docs/reference/QueryClient/#queryclientgetmutationcache)
- [resumePausedMutations](https://tanstack.com/query/latest/docs/reference/QueryClient/#queryclientresumepausedmutations)

The last two methods, `getMutationCache` and `resumePausedMutations`, don’t accepts arguments and can be called directly from the `QueryClient` instance, so we don’t need to implement them in `createTRPCQueryUtils`.

## ✅ Checklist


packages/react-query/src/internals/context.tsx
```diff
@@ -240,5 +241,18 @@ export interface TRPCQueryUtils<TRouter extends AnyRouter> {
   getInfiniteQueryData: (
     queryKey: TRPCQueryKey,
   ) => InfiniteData<unknown> | undefined;
+
+  /**
+   * @link https://tanstack.com/query/latest/docs/reference/QueryClient/#queryclientsetmutationdefaults
+   */
+  setMutationDefaults: (
+    mutationKey: string[][],
```
simonecervini, line 249: Not sure about this type. `string[][]` should be better than `MutationKey` from TanStack Query, which is defined as `ReadonlyArray<unknown>`. But I am not sure what the relationship between this type and `TRPCQueryKey` should be, assuming there should be a relationship

packages/react-query/src/shared/proxy/utilsProxy.ts
```diff
@@ -296,6 +301,8 @@ export const getQueryType = (
     case 'setInfiniteData':
       return 'infinite';
 
+    case 'setMutationDefaults': // This sounds wrong
```
simonecervini, line 304: Not sure about this

simonecervini: It should be good. When the input of `getQueryKey` is a mutation procedure the `QueryType` used is `"any"`

packages/react-query/src/internals/context.tsx
simonecervini: Probably `[string[]]` is a better choice here (= `TRPCQueryKey` without the optional second value of the tuple)

We can also define it as `[TRPCQueryKey[0]]` or define a new alias

simonecervini: Updated in https://github.com/trpc/trpc/pull/5814/commits/790ec8e0e541f651dc13abfd407976de12823a5d

juliusmarminge (commented):
Hey, appreciate the time you put into this PR

However, I struggle to find the value added with this. What is trpc specific here? The wrappers you added here doesn't use the procedure input/output in any way afaict, so why can't you just use the `queryClient.isMutating()` or similar directly here?

simonecervini:
Hey @juliusmarminge 

`isMutating()` is arguably the most useless function added. I included it just to make the API complete, and you’re right about that.

However, `getMutationDefaults` and `setMutationDefaults` actually use the procedure input/output, and the tRPC wrapper around React Query solves multiple problems. For example:

1. With this implementation, you cannot call mutation utility functions for query procedures or query utility functions for mutation procedures. Without tRPC, you can call `queryClient.getMutationDefaults(['post.list'])` without getting type errors, which is very problematic in an offline-first app with dozens of `setMutationDefaults` calls outside React for almost every mutation procedure.

2. There’s not an easy way, without tRPC-specific wrappers, to call the canonical mutation function when setting the default mutation function (required for offline persistence). Instead of:
```ts
utils.post.create.setMutationDefaults(({canonicalMutationFn}) => ({ mutationFn: canonicalMutationFn }))
```
you should need to write something like this:
```ts
const mutationKey = getQueryKey(procedure)[0];
queryClient.setMutationDefaults(mutationKey, {
  mutationFn: (input: unknown) =>
    trpcClient.mutation(mutationKey.join("."), { input }),
});
```
And you don't get errors if `procedure` isn't a mutation procedure.

3. All the callbacks are type-safe:
Screenshot 2024-06-24 alle 15 44 59

packages/react-query/src/internals/context.tsx
```diff
@@ -240,5 +241,27 @@ export interface TRPCQueryUtils<TRouter extends AnyRouter> {
   getInfiniteQueryData: (
     queryKey: TRPCQueryKey,
   ) => InfiniteData<unknown> | undefined;
+
+  /**
+   * @link https://tanstack.com/query/latest/docs/reference/QueryClient/#queryclientsetmutationdefaults
+   */
+  setMutationDefaults: (
+    mutationKey: [string[]],
```
juliusmarminge, line 249: iirc our mutationKeys use the same structure?

```suggestion
    mutationKey: TRPCQueryKey,
```

packages/react-query/src/internals/context.tsx
```diff
@@ -240,5 +241,27 @@ export interface TRPCQueryUtils<TRouter extends AnyRouter> {
...
+      | MutationOptions
+      | ((args: {
+          canonicalMutationFn: (input: unknown) => Promise<unknown>;
+        }) => MutationOptions),
+  ) => void;
+
+  /**
+   * @link https://tanstack.com/query/latest/docs/reference/QueryClient#queryclientgetmutationdefaults
+   */
+  getMutationDefaults: (mutationKey: [string[]]) => MutationOptions | undefined;
```
juliusmarminge, line 260:
```suggestion
  getMutationDefaults: (mutationKey: TRPCQueryKey) => MutationOptions | undefined;
```

packages/react-query/src/internals/context.tsx
```diff
@@ -240,5 +241,27 @@ export interface TRPCQueryUtils<TRouter extends AnyRouter> {
...
+
+  /**
+   * @link https://tanstack.com/query/latest/docs/reference/QueryClient#queryclientismutating
+   */
+  isMutating: (filters: { mutationKey: [string[]] }) => number;
```
juliusmarminge, line 265:
```suggestion
isMutating: (filters: { mutationKey: TRPCQueryKey }) => number;
```

packages/react-query/src/shared/proxy/utilsProxy.ts
```diff
@@ -340,6 +365,10 @@ function createRecursiveUtilsProxy<TRouter extends AnyRouter>(
       },
       getData: () => context.getQueryData(queryKey),
       getInfiniteData: () => context.getInfiniteQueryData(queryKey),
+      setMutationDefaults: () =>
+        context.setMutationDefaults([queryKey[0]], input),
```
juliusmarminge, line 369: what's this destructuring (`[queryKey[0]]`) needed for?

packages/react-query/src/utils/createUtilityFunctions.ts
```diff
@@ -145,5 +145,31 @@ export function createUtilityFunctions<TRouter extends AnyRouter>(
     getInfiniteQueryData: (queryKey) => {
       return queryClient.getQueryData(queryKey);
     },
+
+    setMutationDefaults: (mutationKey, options) => {
+      const path = mutationKey[0];
```
juliusmarminge, line 150: Again, I'm not very familiar with these functions but why don't we use the full key?

packages/react-query/src/internals/context.tsx
simonecervini: > iirc our mutationKeys use the same structure?

Yeah, I wasn’t sure about that. I wrote something here: https://github.com/trpc/trpc/pull/5814#discussion_r1648223193

`TRPCQuery` is fine because it’s defined as
```ts
export type TRPCQueryKey = [
  string[],
  { input?: unknown; type?: Exclude<QueryType, 'any'> }?,
]
```
But if I recall correctly, the second element of the tuple (the object) isn’t meaningful for mutations and isn’t used at all. You can check this in the `useMutation` definition (`createHooksInternal.tsx`) where the mutationKey is just `[path]`.

So it’s just a matter of semantics/preferences. If you prefer `TRPCQuery`, it’s perfectly fine.

packages/react-query/src/utils/createUtilityFunctions.ts
simonecervini: Because we need the path to effectively call the mutation function on the tRPC client. A similar thing is done in the `useMutation` definition. This is basically just the inverse formula of:

```ts
const mutationKey = [path];
```

Screenshot 2024-07-05 alle 15 05 41

packages/react-query/src/shared/proxy/utilsProxy.ts
simonecervini: This is directly connected to https://github.com/trpc/trpc/pull/5814#discussion_r1666738105. We can avoid it if we decide to use `TRPCQueryKey` as the mutation key in all these added functions.

packages/react-query/src/utils/createUtilityFunctions.ts
juliusmarminge: Oh right... read that one incorrectly 🫠

packages/react-query/src/internals/context.tsx
juliusmarminge: At the very least I think it should be a named type `TRPCMutationKey` placed next to `TRPCQueryKey` with the difference in semantics laid out

packages/tests/server/react/offline.test.tsx
juliusmarminge, line 1: Out of scope of this PR but I think a trpc offline example would be cool to have

packages/react-query/src/internals/context.tsx
simonecervini: Ok perfect! We can define it as:
```ts
export type TRPCMutationKey = [
  TRPCQueryKey[0]
];
```
to highlight how the two types are related, or simply:
```ts
export type TRPCMutationKey = [
  string[]
];
```

Do you think we should also update the return type of the `getQueryKey` function? We could use `TRPCMutationKey` when the input extends `DecoratedMutation<any>` and `TRPCQueryKey` otherwise. However, in that case, the name of the function itself, `getQUERYKey`, might be confusing

juliusmarminge: I think 2 functions would be best

packages/tests/server/react/offline.test.tsx
simonecervini: Yeah, I thought the same while working on this PR 😁 I could work on this in the future. I’m currently working on integrating an offline-first mobile app for a web application based on tRPC, and I think I might have more to share in the coming months

packages/react-query/src/internals/context.tsx
simonecervini: Okay, everything's clear. I’ll proceed then

simonecervini: Which option do you prefer:

- Export `getMutationKey` and `TRPCMutationKey` from `@trpc/react-query/getQueryKey`
- Export `getMutationKey` from `@trpc/react-query/getMutation` (and add a new documentation page)
- Deprecate `@trpc/react-query/getQueryKey` and export `getQueryKey`, `getMutationKey`, `TRPCQueryKey`, and `TRPCMutationKey` from a new module `@trpc/react-query/getKey` (or a better name if available)

juliusmarminge: The former sounds good to me but no strong opinion, and also re-export it from the main entrypoint for public consumption

simonecervini: Added everything in https://github.com/trpc/trpc/pull/5814/commits/5322e8875b0d36dba466b0a99f6500c01fe1a34a

packages/react-query/src/shared/proxy/utilsProxy.ts
simonecervini: Replaced `[queryKey[0]]` with `getMutationKeyInternal(path)`.

`createRecursiveUtilsProxy` is still very "query-centric" (for good reasons). I think this is the best we can do without doing a bigger refactoring

juliusmarminge (approved):

PR #5819, opened 2024-06-22
fix(sse): `sse()`-id not propagated

KATT:
I caused a bug where  `id` in `sse({})`  wasn't propagated. Fixed it and added tests that the right thing is actually propagated

### DIFF
--- a/examples/next-sse-chat/.env.example
+++ b/examples/next-sse-chat/.env.example
@@ -1,1 +1,3 @@
-POSTGRES_URL=postgres://postgres:adminadmin@0.0.0.0:5432/db
+POSTGRES_URL=postgres://postgres:adminadmin@0.0.0.0:5432/sse-chat
+NEXTAUTH_URL=http://localhost:3000
+NEXTAUTH_SECRET=secret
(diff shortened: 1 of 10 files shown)


examples/next-sse-chat/.env.example
```diff
@@ -1 +1,3 @@
-POSTGRES_URL=postgres://postgres:adminadmin@0.0.0.0:5432/db
\ No newline at end of file
+POSTGRES_URL=postgres://postgres:adminadmin@0.0.0.0:5432/sse-chat
```
KATT, line 1:
```suggestion
POSTGRES_URL=postgres://postgres:@0.0.0.0:5432/sse-chat
```

packages/tests/server/httpSubscriptionLink.test.ts
```diff
@@ -200,6 +200,11 @@ test('disconnect and reconnect with an event id', async () => {
     expect(onData.mock.calls.length).toBeGreaterThan(5);
   });
 
+  expect(onData.mock.calls[0]![0]).toEqual({
+    data: 0,
+    id: '0',
```
KATT, line 205: this was missing

PR #5828, opened 2024-06-24
fix: prefetch in RSC

juliusmarminge:
Fixes 2 bugs:

- `_def` sometimes being undefined. Fixed by adding a flat proxy and making the `_def` a synchronous getter.
image

- race condition when queries are fast and resolves before hitting the network boundary resulting in serialized data being put in the query cache (upstream fix https://github.com/TanStack/query/pull/7615/files)
image

## Fixed result:

https://github.com/trpc/trpc/assets/51714798/ac9ce585-ba05-4f6a-89d5-7ea42a44290b


patches/@tanstack__query-core@5.45.0.patch
juliusmarminge, line 1: https://github.com/TanStack/query/pull/7615

packages/server/src/unstable-core-do-not-import/router.ts
```diff
@@ -276,40 +276,41 @@ export function createCallerFactory<TRoot extends AnyRootTypes>() {
...
+            const fullPath = [path, ...opts.path].join('.');
+
+            const procedure = _def.procedures[fullPath] as AnyProcedure;
+
+            let ctx: Context | undefined = undefined;
+            try {
+              return Promise.resolve(
+                isFunction(ctxOrCallback) ? ctxOrCallback() : ctxOrCallback,
+              ).then((resolvedContext) => {
+                ctx = resolvedContext;
```
juliusmarminge, line 293: i don't think this is actually necessary anymore

KATT (approved):

packages/react-query/package.json
```diff
@@ -64,14 +64,14 @@
   },
   "dependencies": {},
   "peerDependencies": {
-    "@tanstack/react-query": "^5.40.0",
+    "@tanstack/react-query": "^5.49.2",
```
KATT, line 67: nice

juliusmarminge: fuck I hate that manypkg bumps peerDep range. We can keep it at 5.40 assuming most ppl wont use this RSC void prefetch thing?

KATT: i think we should bump it and i'm fine with bumping this with any minor/patch release

PR #5862, opened 2024-07-05
Add `usePrefetchQuery `

DogPawHat:
## 🎯 Changes

Adding `usePrefetchQuery`, `usePrefetchInfiniteQuery` from Tanstack Query into `@trpc/react-query`

These hooks are mostly useful for doing component level prefetching near the top of a component tree above a Suspense boundary. I've added the documentation in a `Prefetching` sub-section under the Suspense docs, with a caveat that prefetching and render-as-you-fetch is dependent on what framework and router you are using (e.g. prefetching in server-side (`getServerSideProps`) > prefetching in Tanstack/React Router loaders > component prefetching with `usePrefetchQuery` > regular fetch-as-you-render)

## ✅ Checklist


DogPawHat:
The existing `TRPCFetchQueryOptions` interface doesn't currently have the `trpc` property defined, so I removed it from the body of the implemtation here. This is a pretty big difference vs vanilla `usePrefetchQuery` so I'd wonder if an extension of that interface would be needed here.

www/docs/client/react/usePrefetchQuery.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+</details>
+
+```tsx title='components/MyComponent.tsx'
+import { trpc } from '../utils/trpc';
+
+function SuspendingComponent() {
+  // The query is already prefetched
+  const helloSuspense = trpc.hello.useSuspenceQuery();
```
juliusmarminge, line 66:
```suggestion
  const helloSuspense = trpc.hello.useSuspenseQuery();
```

www/docs/client/react/usePrefetchQuery.md
```diff
@@ -0,0 +1,83 @@
...
+function useQuery(
+  input: TInput,
+  opts?: TRPCFetchQueryOptions;
+)
+
+interface TRPCFetchQueryOptions
+  extends FetchQueryOptions { }
+```
+
+Since `TRPCFetchQueryOptions` extends @tanstack/react-query's `FetchQueryOptions`, you can use any of their options here such as `enabled`, `refetchOnWindowFocus`, etc.
```
juliusmarminge, line 22: is this just copy-pasted from useQuery docs, might need an update

www/docs/client/react/usePrefetchQuery.md
```diff
@@ -0,0 +1,83 @@
+---
```
juliusmarminge, line 1: in general this page probably needs a short explanation on where and how to use this, it doesn't have to repeat a bunch of stuff from the `useQuery` docs?

Maybe just say like:

`unlike [useQuery](...), `usePrefetchQuery` does not return data ... It is useful when using Suspense in SPAs to fire off a query before a suspense boundary with a children `useSuspenseQuery`... Read more at Tanstack's docs...`

DogPawHat:
I've put the docs for these under the Suspence section with some caveats that I think are nessary to mention with these hooks (namely that server side hydration and/or route level prefetching is superior to component level prefetch at a high level.)

I also closed #5863 in favour of introducing `usePrefetchInfiniteQuery` here. I still need to write a test for it though, hoping to add it in the next day or so.

DogPawHat:
Tests have been updated

packages/react-query/src/createTRPCReact.tsx
```diff
@@ -108,6 +120,14 @@ export type MaybeDecoratedInfiniteQuery<TDef extends ResolverDef> =
           TRPCClientErrorLike<TDef>,
           TDef['input']
         >;
+        usePrefetchInfiniteQuery: (
+          input: Omit<TDef['input'], 'cursor' | 'direction'>,
```
juliusmarminge, line 124:
```suggestion
          input: Omit<TDef['input'], ReservedInfiniteQueryKeys> | SkipToken,
```

DogPawHat: Suggestion accepted and committed

juliusmarminge (approved):

PR #5865, opened 2024-07-05
feat: ability to cancel streamed responses + add `TRPCRequestInfo.signal`

### LINKED ISSUE #5851: bug: Stream Link does not abort

```
  System:
    OS: Windows 11 10.0.22631
    CPU: (32) x64 AMD Ryzen 9 7950X 16-Core Processor
    Memory: 46.84 GB / 63.65 GB
  Binaries:
    Node: 20.13.1 - C:\P\apps\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 10.5.2 - C:\P\apps\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (126.0.2592.81)
    Internet Explorer: 11.0.22621.3527
  npmPackages:
    @trpc/client: next => 11.0.0-rc.436+fbb4a2fb4
    @trpc/server: next => 11.0.0-rc.436+fbb4a2fb4
    typescript: ^5.5.0 => 5.5.3
```

### Describe the bug

The AbortController signal seems ignored by Stream Links with async generators.

We'd expect an interruption on the server side, and on the client side, but packets continue to be produced by the async generator, flow to the client, and the client keeps awaiting and receiving them.

I don't know what the correct behavior would be, but right now the signal seems to be completely ignored on both ends.

### Link to reproduction

https://github.com/enricoros/trpc-repro-stream-unabort

### To reproduce

In summary:
1. The client creates a Stream Link, and queries with a `{signal: ac.signal}`, like this (minimal):
```typescript
const trpc = createTRPCClient<AppRouter>({
  links: [unstable_httpBatchStreamLink({url: 'http://localhost:3000',})],
});

const iterable = await trpc.examples.iterable.query(undefined, { signal: ac.signal });
```

2. The server implements a procedure with an async generator, like this (minimal):
```typescript
    iterable: publicProcedure.query(async function* () {
      for (let i = 0; i < 8; i++) {
        await new Promise((resolve) => setTimeout(resolve, 500));
        yield i;
      }
    }),
```

3. We would expect both the server and the client to be aborted, but execution continues on both.
This is the output:
image

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

## Funding

<a href="https://polar.sh/trpc/trpc/issues/5851">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/5851/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

KATT:
Closes #5851

## 🎯 Changes

- Ability to cancel streamed responses
  - When `signal.abort()` is called on `.query(x, { signal })`
  - When `return;` is done when consuming a streamed response
- Add `signal` to `TRPCRequestInfo` to be able to cancel downstream `fetch()` / db calls as a response to the browser cancelling the request

###  Cancellation when stopping an async iterable

This is done by handling `return()` - when you do e.g.

```ts
for await (const item of iterable) {
  if (item === 1) break; // <-- triggers `return()`
  console.log(item)
}
```

We handle this in the `jsonl.ts` - if there are no remaining things to iterate, we cancel the request through an abort controller

###  Refactor

> Ignore whitespace when reviewing 

- Removed cancellation from dataloader and replaced with using abort signals in most places
- Changed a lot of abort logic to use abort controllers
- Removed AbortControllerEsque etc - I don't see why we need it when fetch is both in browsers and in node
- A bunch of signatures changed from `{ cancel: () => void; promise: Promise<T>` -> `Promise<T>` which made me to simplify a bunch of code

## ✅ Checklist


packages/server/src/unstable-core-do-not-import/stream/jsonl.ts
```diff
@@ -490,38 +509,64 @@ export async function jsonlStreamConsumer<THead>(opts: {
...
+                    maybeAbort();
+                    throw (
+                      opts.formatError?.({ error: data }) ??
+                      new AsyncError(data)
+                    );
+                }
+              },
+              return: async () => {
+                wrapper.returned = true;
+                maybeAbort();
```
KATT, line 562: This is sort of the key in it all

PR #5875, opened 2024-07-07
feat(websockets): add support for `lastEventId`

KATT:
Closes #

Adds support for lastEventId for WebSockets, the same way we have in SSE

Basically, it allows the client to automatically track the latest sent event id and it'll send it to the server on any disconnection

**📚  Docs:** https://www-git-07-06-lasteventid-trpc.vercel.app/docs/websockets#automatic-tracking-of-id-using-tracked-recommended

### 🎯 Changes

- Deprecates `sse({id, data})` in favor of `tracked(id, data)`
- Adds support for auto-tracking IDs in `wsLink`


packages/server/src/adapters/ws.ts
```diff
@@ -277,13 +304,21 @@ export function getWSConnectionHandler<TRouter extends AnyRouter>(
               break;
             }
 
+            const result: TRPCResultMessage<unknown>['result'] = {
+              type: 'data',
+              data: next.value,
+            };
+
+            if (isSSEMessageEnvelope(next.value)) {
```
juliusmarminge, line 312: Mixin of SSE stuff in WS?

packages/tests/server/websockets.test.ts
```diff
@@ -1258,6 +1292,74 @@ describe('lazy mode', () => {
...
+        expect(onStarted).toHaveBeenCalledTimes(1);
+      });
+      ctx.nextIterable();
+      await waitFor(() => {
+        expect(onData).toHaveBeenCalledTimes(1);
+      });
+      expect(onData.mock.calls[0]![0]).toEqual({
+        id: '1',
+        data: {
+          id: '1',
```
juliusmarminge, line 1331: Double id does feel a bit weird

packages/tests/server/websockets.test.ts
```diff
@@ -48,6 +56,31 @@ function factory(config?: {
...
+
+        while (true) {
+          from++;
+          await iterableDeferred.promise;
+          const msg: Message = {
+            id: String(from),
+            title: 'hello ' + from,
+          };
+
+          yield sse({
```
juliusmarminge, line 78: Again, mixing in sse into websockets feels a bit off. Seems like there's some common parts here that can use a unified name?

packages/tests/server/websockets.test.ts
```diff
@@ -48,6 +56,31 @@ function factory(config?: {
...
+            id: String(from),
```
juliusmarminge, line 79: If this is the source of that double id it feels like it maybe should be a first argument like you proposed on discord

KATT:
cc @juliusmarminge i switched the helper in #5904

PR #5886, opened 2024-07-09
Subscription connection state

### LINKED ISSUE #5786: feat: Make useSubscription return info about the subscription

It's currently pretty difficult to tell whether a subscription is actually connected and listening for events. To track the status of the connection you would have to handle these events yourself. It would be nice if there was an easier way to track the status of the subscription.

Have the useSubscription hook return the status of the connection.

```tsx
const { status } = api.message.useSubscription();

if (status !== 'connected') return <p>You are currently not connected to the server :(</p>;

return <p>Connection active :D</p>;
```

Currently to show the user they are not connected to the service I am tracking the online status of the device, but this is only loosely correlated to actual status of the connection. This means that it will always show that the service is available when the device is online even while our service is down.

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

## Funding

<a href="https://polar.sh/trpc/trpc/issues/5786">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/5786/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

### LINKED ISSUE #5871: feat: access subscription connection status from tRPC client

My app uses tRPC's subscriptions via websockets. I would like a way to display an indicator if the client was disconnected from the websocket (ex. went offline).

Trying to use `onOpen` and `onClose` in `createWSClient()` for this kinda works, but it seemed like gracefully disconnecting from the subscriptions was triggering `onClose`, and I couldn't figure out how to distinguish from the disconnect code whether it was due to an error or a graceful disconnect.

Some way to access a general status of subscriptions from the tRPC client (ex. via React context).

Maybe calling `.useSubscription()` could return an object with the status of that specific subscription.

N/A

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

## Funding

<a href="https://polar.sh/trpc/trpc/issues/5871">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/5871/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

BeBoRE:
Closes #5871 
Closes #5786

## 🎯 Changes

Lets useSubscription return a result containing the state of the connection, most recently received data (useful for myself), and utilities to better recover from errors.

## ✅ Checklist

- [x] Verify old state callbacks (onStarted, onError) still function the same
- [x] Verify there are no unnecessary re-renders
- [x] Remove debug console logs

## 📋 Tasks

- [x] Extend what the SubscriptionObservable can emit (connection state)
- [x] Implement the specification as described in the UseSubscriptionResult typedef
- [x] Extend httpSubscriptionLink to conform to new observable events
- [x] Make state only change when watched properties change
- [x] Extend wsLInk to conform to new observable events


packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -87,12 +90,16 @@ export function unstable_httpSubscriptionLink<
                 eventSource,
               },
             });
-
-            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-            eventSource!.removeEventListener('open', onStarted);
```
BeBoRE, line 92: I deleted this line since there was already an eventListener for this setup right? Why is there multiple?

packages/client/src/internals/TRPCUntypedClient.ts
```diff
@@ -30,6 +30,7 @@ export interface TRPCSubscriptionObserver<TValue, TError> {
   onStarted: (opts: { context: OperationContext | undefined }) => void;
   onData: (value: TValue) => void;
   onError: (err: TError) => void;
+  onConnecting: () => void;
```
KATT, line 33: Maybe we should have an onState or something that could deprecate all the others

packages/client/src/links/httpSubscriptionLink.ts
KATT: This removes it after it has started

packages/server/src/unstable-core-do-not-import/rpc/envelopes.ts
```diff
@@ -99,6 +99,7 @@ export type TRPCClientOutgoingMessage =
 
 export interface TRPCResultMessage<TData>
   extends JSONRPC2.ResultResponse<
+    | { type: 'connecting'; data?: never }
```
KATT, line 102: I don't think this should be in the rpc spec since it doesn't affect actual communication between server/client

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -306,11 +307,22 @@ export function createRootHooks<
...
     opts: UseTRPCSubscriptionOptions<unknown, TError>,
-  ) {
+  ): UseTRPCSubscriptionResult<unknown, TError> {
     const enabled = opts?.enabled ?? input !== skipToken;
     const queryKey = hashKey(getQueryKeyInternal(path, input, 'any'));
     const { client } = useContext();
 
+    const [subscriptionState, setSubscriptionState] = React.useState<
+      UseTRPCSubscriptionResult<unknown, TError>
+    >({
```
KATT, line 317: Look at the types of React query - we should likely mimic those

Example

- error state is guaranteed to have error
- [...]

packages/client/src/links/httpSubscriptionLink.ts
BeBoRE: Oh wow I really am dyslexic

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -319,33 +331,76 @@ export function createRootHooks<
...
+                ...prev,
+                state: 'connected',
+                isConnecting: false,
+                isConnected: true,
+              }));
             }
           },
           onData: (data) => {
             if (!isStopped) {
+              setSubscriptionState((prev) => ({
```
KATT, line 363: I'm not sure we wanna update data for every event - it can become incredibly expensive for high throughput events, likely better to force developers to handle data themselves

packages/react-query/src/shared/hooks/types.ts
```diff
@@ -139,9 +139,22 @@ export interface UseTRPCMutationOptions<
 export interface UseTRPCSubscriptionOptions<TOutput, TError> {
   enabled?: boolean;
   onStarted?: () => void;
+  onConnecting?: () => void;
   onData: (data: TOutput) => void;
   onError?: (err: TError) => void;
 }
+
+export type SubscriptionState = 'connecting' | 'connected' | 'stopped';
```
KATT, line 147: - idle
- pending
- error

packages/react-query/src/shared/hooks/types.ts
```diff
@@ -139,9 +139,22 @@ export interface UseTRPCMutationOptions<
...
+
+export interface UseTRPCSubscriptionResult<TOutput, TError> {
```
KATT, line 149: We need separate interfaces for different states (like tanstack query)

packages/react-query/src/shared/hooks/types.ts
KATT: Maybe connecting as well, as a pre-state of pending

packages/react-query/src/shared/hooks/createHooksInternal.tsx
KATT: If we update data of every event, we need to have a smart handler that won't update data unless it's used - not looked at the magic that RQ does for that

BeBoRE: I probably agree, I'll reconsider, removing it for now

BeBoRE:
I have a few points to think on:

1. What does the error state mean? Does it include that the connection has been interrupted or does it only mean the subscription has completely failed? If connection interruptions don't count as errors how should this be distinguished.
2. What does stopped mean? Is this when the subscription has completely failed. Or does it just mean that the subscription isn't running due to the subscription having ended succesfully or the subscription ending due to an exception.
3. What does completed mean? Is this when the subscription has ended succesfully?
4. You mentioned adding a retry callback, what should the behaviour of the retry be if the subscription is still running or reconnecting?
5. Should we add an end callback aswell to end the subscription, or should this be handled by the enabled option?
6. (How) should we distinguish connecting vs reconnecting vs retry?

packages/react-query/src/shared/hooks/types.ts
BeBoRE: Idle means subscription is not enabled?

KATT:
> I have a few points to think on:
> 
> 1. What does the error state mean? Does it include that the connection has been interrupted or does it only mean the subscription has completely failed? If connection interruptions don't count as errors how should this be distinguished.

I think loss of connection that _is_ recoverable isn't an error - it's more of a temporary loss of connection that should move from maybe `pending` -> `connecting`

> 2. What does stopped mean? Is this when the subscription has completely failed. Or does it just mean that the subscription isn't running due to the subscription having ended succesfully or the subscription ending due to an exception.

stopped should probably not be a status - instead we'd use `idle` if it's _disabled_ through a `skipToken`

> 3. What does completed mean? Is this when the subscription has ended successfully?

yeah.

honestly, I don't think this belongs in subscriptions at all - we should remove it. subscriptions should be infinite and not finite state

> 4. You mentioned adding a retry callback, what should the behaviour of the retry be if the subscription is still running or reconnecting?

no matter the state - trashing whatever is going on and restarting it

it should probably consider `lastEventId` but having it overridable through arguments

> 5. Should we add an end callback aswell to end the subscription, or should this be handled by the enabled option?

as i mentioned above, I am not sure "end" belongs in the context of subscriptions - it was made before we had streamed RPC-calls

> 6. (How) should we distinguish connecting vs reconnecting vs retry?

i kinda think of states as

| state  | description |
| ------------- | ------------- |
|  idle | disabled subscription / initial state  |
| connecting  | we're connecting or reconnecting to get subscriptions |
| pending  | actively listening |
| error  | unrecoverable error until manual retry through a call |

i don't think "completed" belongs anymore tbh

BeBoRE:
Gotha, I don't see/have a use-case for finite subscriptions either, you can accomplish the same goal better with SSE query.

BeBoRE:
Updated the typedefs based on your feedback any thoughts on this?
https://github.com/trpc/trpc/pull/5886/files#diff-74a113df0ef51986032f3866be5f2067b6114ade631d41ac09e46ab3083d2258

packages/react-query/src/shared/hooks/types.ts
```diff
@@ -138,10 +140,344 @@ export interface UseTRPCMutationOptions<
...
+    errorUpdatedAt: Date.now(),
+  };
+};
+
+export type UseTRPCSubscriptionResult<TInput, TError> =
+  | TRPCSubscriptionIdleResult<TInput, TError>
+  | TRPCSubscriptionStartingResult<TInput, TError>
+  | TRPCSubscriptionPendingResult<TInput, TError>
+  | TRPCSubscriptionReconnectingResult<TInput, TError>
+  | TRPCSubscriptionErrorResult<TInput, TError>;
```
KATT, line 479: Niiiiiiceeeee

cc @juliusmarminge what do you think of this approach to `.useSubscription()` results?

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -93,6 +101,34 @@ export function unstable_httpSubscriptionLink<
...
+          });
+
+          eventSource.addEventListener('error', () => {
+            if (eventSource?.readyState === EventSource.CLOSED) {
+              return;
+            }
+
+            const error = TRPCClientError.from(
+              new Error(`EventSource Network Error`),
+            );
```
KATT, line 121: We can probably use some sort of value from the callback here

KATT (commented):
This stuff is looking very promising!

examples/next-sse-chat/src/app/channels/[channelId]/chat.tsx
```diff
@@ -31,6 +38,68 @@ const listWithAnd = (list: string[]) => {
   return `${list.slice(0, -1).join(', ')}, and ${list.at(-1)}`;
 };
 
+function SubscriptionStatus(props: {
```
Comment (KATT) [inline examples/next-sse-chat/src/app/channels/[channelId]/chat.tsx:41]: cc @BeBoRE I added this component

packages/client/src/links/httpSubscriptionLink.ts
BeBoRE: Couldn't find anything in the typedef of the event, but I haven't looked that hard into it, console does say CONNECTION_DISCONNECTER or something similar, so there might be something to extract?

examples/next-sse-chat/src/app/channels/[channelId]/chat.tsx
```diff
@@ -31,6 +38,68 @@ const listWithAnd = (list: string[]) => {
...
+  subscription: UseTRPCSubscriptionResult<
+    unknown,
+    {
+      message: string;
+    }
+  >;
```
Comment (BeBoRE) [inline examples/next-sse-chat/src/app/channels/[channelId]/chat.tsx:47]: We should probably extend the input inference utility to do this easier

packages/client/src/links/httpSubscriptionLink.ts
KATT: Yeah I realised we don't serialize errors here & it'd be quite involved to do

examples/next-sse-chat/src/app/channels/[channelId]/chat.tsx
Comment (BeBoRE) [reply examples/next-sse-chat/src/app/channels/[channelId]/chat.tsx:41]: Looks really good, nice showcase of the feature

BeBoRE:
I implemented a way of tracking the properties that are being watched, if those don't change the hook doesn't cause a re-render. I am also now tracking the last piece of data that was received, which is useful for my use-case.

BeBoRE:
The wsLink seems a lot more complex, I'll have to analyse that a lot more to understand how to make that work. Complexity probably stems from the fact that it doesn't only handle one subscription, but multiple on the same connection.

BeBoRE:
I also seem to be getting this error sometimes, but not consistently. Is this a known issue?
max listeners error

packages/client/src/links/httpSubscriptionLink.ts
BeBoRE: The Event that gets emitted can be of type ErrorEvent. Resolving this :)

BeBoRE:
> The wsLink seems a lot more complex, I'll have to analyse that a lot more to understand how to make that work. Complexity probably stems from the fact that it doesn't only handle one subscription, but multiple on the same connection.

@KATT Do you mind implementing this yourself? I'll probably not be available much the next few days.

packages/react-query/src/shared/hooks/createHooksInternal.tsx
BeBoRE: Re-added this :)

packages/client/src/links/httpSubscriptionLink.ts
BeBoRE: Nevermind, when using Chrome it just seems to be an Event and not an ErrorEvent, haven't tested on other browsers

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -334,21 +421,62 @@ export function createRootHooks<
...
+
       return () => {
         isStopped = true;
         subscription.unsubscribe();
+
+        updateSubscriptionState(getIdleResult(effectRestart));
       };
+
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [queryKey, enabled]);
```
BeBoRE, line 477: Is there a reason why input and path are not in this list?

BeBoRE:
I extended the wsLink with the new connection state events, very crudely needs a lot more testing. It seems that when using WebSockets the browser doesn't recognize connection issues for a long time. The starting connecting state and pending state work quite well however.

BeBoRE:
I believe the only thing missing now are tests, unless there needs to be more changes to the specifications.

PR #5893, opened 2024-07-13
fix(client) `throwIfAborted` usage for React Native

jikyu:
Closes #

## 🎯 Changes

`AbortSignal.throwIfAborted()` is NOT available on React Native and causes all tRPC procedures to fail on React Native. Since there is no good widely maintained polyfill for `AbortController / AbortSignal`, we could just use a straighforward implementation instead of using `throwIfAborted` directly (+ lightweight polyfill for `AbortError`)

This error was introduced with `11.0.0-rc.455` tag here:

https://github.com/trpc/trpc/pull/5865/files#diff-d2f09745539c905e09fb5b0eabdfddcca87cb2496a3bc9ddcd585c89e4620e8eR163

Alternative approach to this PR is bring back the deleted `AbortControllerInstanceEsque` polyfill and add our implementations there.

## ✅ Checklist

### DIFF
--- a/packages/client/src/links/internals/httpUtils.ts
+++ b/packages/client/src/links/internals/httpUtils.ts
@@ -161,7 +161,7 @@ export type HTTPRequestOptions = ContentOptions &
 
 export async function fetchHTTPResponse(opts: HTTPRequestOptions) {
   if (opts.signal?.aborted) {
-    throw new Error(opts.signal.reason || 'Aborted')
+    throw new Error('AbortError')
   }
 
   const url = opts.getUrl(opts);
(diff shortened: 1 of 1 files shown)


packages/client/src/links/internals/httpUtils.ts
```diff
@@ -154,13 +154,27 @@ export const jsonHttpRequester: Requester = (opts) => {
...
+
 export type HTTPRequestOptions = ContentOptions &
   HTTPBaseRequestOptions & {
     headers: () => HTTPHeaders | Promise<HTTPHeaders>;
   };
 
 export async function fetchHTTPResponse(opts: HTTPRequestOptions) {
-  opts.signal?.throwIfAborted();
+  if (opts.signal?.aborted) {
+    throw new AbortError();
```
KATT, line 175: Prioritise using throwIfAborted if it's available on object 🙏

PR #5909, opened 2024-07-22
feat: support typeschema spec

lobomfz:
## 🎯 Changes

This is a barebones implementation of the [typeschema spec](https://typeschema.com/), it should allow any validation schema to be used without wrapping in a parser. This by itself, vastly improves the compatibility list, which adds @deepkit/type, @effect/schema, arktype, fastest-validator, io-ts and @sinclair/typebox to be used without the user wrapping the schema first.

Due to my unfamiliarity with trpc's codebase i'm sure this needs more effort and tests, so i'd like some guidance around what else should be done so this can be merged

Currently, typeschema has a bug with TS 5.5+ which breaks type inference. Due to this, most tests are failing, but applying patches from my [PR](https://github.com/decs/typeschema/pull/62) un-breaks them, so we should wait until that is merged.

To not introduce breaking changes, an additional try catch is added in ParseFn to try to wrap the input schema and extract its parse function

Note: I'm currently rebuilding [trpc-openapi](https://github.com/lobomfz/trpc-openapi-typeschema) based around this, so, with this addition, all [libraries](https://typeschema.com/) that support Serialization will be able to auto generate its spec, unlike alternatives which are based on zod.

## ✅ Checklist


packages/server/src/unstable-core-do-not-import/parser.ts
```diff
@@ -44,10 +51,10 @@ export type ParserWithInputOutput<TInput, TParsedInput> =
   | ParserZodEsque<TInput, TParsedInput>
   | ParserValibotEsque<TInput, TParsedInput>;
 
-export type Parser = ParserWithInputOutput<any, any> | ParserWithoutInput<any>;
+export type Parser = ParserWithInputOutput<any, any> | ParserWithoutInput<any> | TypeschemaSchema
```
KATT, line 54: This should be in `ParserWithInputOutput` instead

packages/server/src/unstable-core-do-not-import/parser.ts
```diff
@@ -57,6 +64,11 @@ export type inferParser<TParser extends Parser> =
         in: $InOut;
         out: $InOut;
       }
+    : TParser extends TypeschemaSchema
+    ? {
+        in: InferIn<TParser>;
+        out: Infer<TParser>;
+      }
```
KATT, line 71: This shouldn't need to be changed

KATT (changes requested):

KATT (changes requested):
Look at #5958 for a reference on how to do this

Nick-Lucas:
Hey @lobomfz  this has been idle for quite a long time so I'm going to close it out. The contribution would be very welcome though if you want to re-open

PR #5910, opened 2024-07-22
chore: bump a bunch of packages

juliusmarminge:
Bumps:

- tanstack
- typescript
- eslint
- typescript-eslint
- vite
- vitest
- next
- tailwind
- postcss
- ...

PR #5911, opened 2024-07-22
chore: bump `@playwright/test`

juliusmarminge:
leftover from #5910


juliusmarminge:
seems like it's cached the old versions too hard or something...

Issue #2330, opened 2022-07-27
`trpc.useQuery()` is not linted by `eslint-plugin-react-hooks`


itsMapleLeaf:
See the playground here: https://codesandbox.io/s/eslint-react-hooks-does-not-lint-on-lowercase-named-namespaces-ojbcw8?file=/src/App.js

The linter rule only recognizes un-namespaced hooks, like `useQuery()`, or if the namespace is capitalized, like `Trpc.useQuery()`

I don't think this requires a change in API, but it might be worth changing the convention in the docs with this in mind.

iway1:
I just do this:
```
const {useQuery, useMutation} = trpc;
```

Then use those in components directly. It's just more dev friendly IMO to import useQuery or useMutation instead of doing `trpc.useQuery`

KATT:
Interesting, I didn't know this! Do you know which lint rule it is? I'd like to open an issue

itsMapleLeaf:
> Interesting, I didn't know this! Do you know which lint rule it is? I'd like to open an issue

The lint rule is `react-hooks/rules-of-hooks`, and it lives on the react repo https://github.com/facebook/react/tree/main/packages/eslint-plugin-react-hooks

KATT:
Blocked: https://github.com/facebook/react/issues/25065

TkDodo:
This seems pretty important for v10. Anything we can do to speed-up things on react's side?

KATT:
> This seems pretty important for v10. Anything we can do to speed-up things on react's side?

If you have any strings to pull in the React team, pull 'em. ☺

PR #5913, opened 2024-07-22
feat: `queryOptions` API

juliusmarminge:
If we can focus our RQ wrapper on the [`queryOptions` API](https://tanstack.com/query/latest/docs/framework/react/reference/queryOptions#queryoptions), I think we could remove a whole lot of surface area and also provide functionality for all the built-in queryClient methods

```ts
import { useSuspenseQuery } from '@tanstack/react-query';
import { createFileRoute, Link } from '@tanstack/react-router';
import { useTRPC } from '~/trpc/react';
import { PostErrorComponent } from './posts.$postId';

export const Route = createFileRoute('/posts/$postId/deep')({
  loader: async ({ params: { postId }, context }) => {
    const data = await context.queryClient.ensureQueryData(
      //  ^?
      context.trpc.post.byId.queryOptions({ id: postId }),
    );

    return { title: data.title };
  },
  meta: ({ loaderData }) => [{ title: loaderData.title }],
  errorComponent: PostErrorComponent,
  component: PostDeepComponent,
});

function PostDeepComponent() {
  const { postId } = Route.useParams();
  const trpc = useTRPC();

  const postQuery = useSuspenseQuery(
    trpc.post.byId.queryOptions({ id: postId }),
  );
  postQuery.data;
  //        ^?

  return (
    <div className="space-y-2 p-2">
      <Link
        to="/posts"
        className="block py-1 text-blue-800 hover:text-blue-600"
      >
        ← All Posts
      </Link>
      <h4 className="text-xl font-bold underline">{postQuery.data.title}</h4>
      <div className="text-sm">{postQuery.data.body}</div>
    </div>
  );
}
```

If implemented correctly, it should be possible to remove all the currently wrapped functions in favor of using built-in's + queryOptions/useQueryOptions wrappers

```ts
queryClient.fetch(trpc.foo.queryOptions())
// instead of
trpc.foo.fetch()


const utils = trpc.useUtils();
const query = utils.foo.queryOptions();

useQuery(query)
//instead of
trpc.foo.useQuery()
```

Would even solve #2330


packages/react-query/src/shared/proxy/utilsProxy.ts
```diff
@@ -308,6 +335,7 @@ export const getQueryType = (
   utilName: keyof AnyDecoratedProcedure,
 ): QueryType => {
   switch (utilName) {
+    case 'queryOptions':
```
juliusmarminge, line 338: todo: make configurable, should be possible to use for infinite queries too

packages/tests/server/react/useQuery.test.tsx
```diff
@@ -311,6 +311,33 @@ describe('useQuery()', () => {
       ]
     `);
   });
+
+  test('using queryOptions', async () => {
+    const { client, App } = ctx;
+    function MyComponent() {
+      const queryOptions = client.post.byId.useQueryOptions({ id: '1' });
+      expect(queryOptions.trpc.path).toBe('post.byId');
+      const query1 = useQuery(queryOptions);
```
KATT, line 320: Niiiiiiceeeeeeeee

KATT: Will this work with the `trpc: {}` props as well?

juliusmarminge: Probably with some modifications.

This was very much a first draft. I think we neee to decide what to do with the tRPC client, if it should be injected from a factory or if react context is still required.

I think @Nick-Lucas doesnt like it requiring react context

KATT: I wonder if we should do a new `@trpc/tanstack-query` pkg (and maybe `@trpc/tanstack`) and deprecate `@trpc/react-query` if we go this direction 🤔 🤔 we could probably make a codemod that does the migration 

Would be so nice with a clean slate

patches/@tanstack__react-query@5.59.9.patch
juliusmarminge, line 1: 5.59.10 fails to release so pulled this in here for now xd

packages/react-query/src/utils/createUtilityFunctions.ts
```diff
@@ -30,6 +42,82 @@ export function createUtilityFunctions<TRouter extends AnyRouter>(
...
+          },
+        };
+
+        const result = await untypedClient.query(
+          ...getClientArgs(queryKey, actualOpts),
+        );
+
+        if (isAsyncIterable(result)) {
+          return buildQueryFromAsyncIterable(result, queryClient, queryKey);
+        }
```
juliusmarminge, line 105: Not sure if we should do this here actually since it not only affects hooks but also `queryClient.fetchQuery(options)` and in that situation you might want the iterable as-is. Thoughts?

KATT (approved):

KATT (dismissed):

KATT (approved):

PR #5930, opened 2024-08-02
fix(server): handle errors better in SSE

KATT:
Closes #5918

## 🎯 Changes

Handles errors as values in SSE

- Rather than `.error()`ing the stream, we emit a custom SSE-event caused `serialized-error` and then close the stream
- On the client, we can later on handle this serialized error in #5871


packages/server/src/unstable-core-do-not-import/stream/sse.ts
```diff
@@ -209,12 +244,12 @@ export function sseStreamConsumer<TData>(opts: {
...
       const reader = readable.getReader();
 
-      const iterator: AsyncIterator<inferTrackedOutput<TData>> = {
+      const iterator: AsyncIterator<ConsumerStreamResult<TData>> = {
         async next() {
           const value = await reader.read();
           if (value.done) {
             return {
-              value: undefined,
+              value: value.value,
```
KATT, line 252:
```suggestion
              value: undefined,
```

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -106,8 +106,14 @@ export function unstable_httpSubscriptionLink<
           });
 
           for await (const chunk of iterable) {
+            if (!chunk.ok) {
+              // TODO: handle in https://github.com/trpc/trpc/issues/5871
+              continue;
+            }
```
KATT, line 112: cc @BeBoRE 
we now serialize any error that happens in streams of SSE

PR #5958, opened 2024-08-21
ParserArkTypeEsque

mmkal:
Related to https://x.com/ssalbdivad/status/1826368190097895566

## 🎯 Changes

Adds a `ParserArkTypeEsque` as a supported input type.

~Note @ssalbdivad ark-type 2.0.0-beta.6 I think might have a bug in it - see the test using `arktype2` in this PR for a repro, the `expectTypeOf` assertion here fails, but I think it should pass (and does for arktype v1). Smaller repro: https://stackblitz.com/edit/rosmye?file=index.ts - I've set this PR as a draft for now in the meantime.~

Motivation: 

I wrote [trpc-cli](https://github.com/mmkal/trpc-cli) and it currently only supports zod. It's using zod-to-json-schema, and now that arktype 2 added the `.toJsonSchema()` method, it should be easy to add arktype support, once we can pass the arktype schema itself into the router, rather than the `assert` function.

## ✅ Checklist

- [ ] If necessary, I have added documentation related to the changes made. [not yet]

Note sure what happened with the commits, but the diff looks right now - I'm assuming this will be squashed so the individual commits won't be that important.


packages/tests/server/validators.test.ts
```diff
@@ -320,7 +321,57 @@ test('myzod', async () => {
...
+});
+
+test('arktype v2 schema', async () => {
+  const t = initTRPC.create();
+
+  const router = t.router({
+    num: t.procedure
+      .input(arktype2.type({ text: 'string' }))
+      .query(({ input }) => {
+        expectTypeOf(input).toEqualTypeOf<{ text: string }>();
```
mmkal, line 356: @ssalbdivad this is the assertion erroring because `text` seems to be any

ssalbdivad: Ahh, so there is a bug specifically in `2.0.0-beta.6` that prevents ArkType from resolving correctly with `moduleResolution: "node"`. It will be fixed in the next release.

That said, that setting should probably be updated to `NodeNext` this point anyways as the version of node associated with it is long since deprecated.

mmkal: Unfortunately changing it to NodeNext or Node16 gives lots of CJS/ESM errors. And updating the package.json to type `module` means adding file extensions to all relative imports.

I might give that a go (at least for packages/tests) and see if that's as much thread-pulling as is needed - but in a separate PR. This PR will have to wait either for that other one, or a new release of ArkType v2 beta with the bug fixed.

image

ssalbdivad: No issues with the new version, that will be out in the next couple days anyways.

That said, ArkType 2.0 only supports ESM which could be a separate issue depending on how this package consumes it.

mmkal: Bumped the arktype2 dependency to `2.0.0-rc.0` which fixed the type problem. Seemed to not have any ESM issues without touching tsconfig, so I'm satisfied!

KATT (approved):

mmkal:
@KATT thanks! FYI there was a websocket test error the first run of `main / test`, I think it's a flake: https://github.com/trpc/trpc/actions/runs/10610509957/attempts/1

KATT:
> @KATT thanks! FYI there was a websocket test error the first run of `main / test`, I think it's a flake: https://github.com/trpc/trpc/actions/runs/10610509957/attempts/1

Yeah, we might have a few flaky tests in the websocket tests, thanks for letting us know

PR #6047, opened 2024-09-26
feat(sse): ability to reconnect with new headers

### LINKED ISSUE #6046: feat: unstable_httpSubscriptionLink with headers polyfill does not update headers when reestablishing a lost connection

n/a

### Describe the bug

This is a known issue with event-source-polyfill https://github.com/Yaffle/EventSource/issues/137 but something we can probably fix in tRPC-land

Essentially EventSource is given headers and those headers are passed at instantiation. When a connection is lost EventSource will opaquely recover and reconnect, however it will use the original headers which may be outdated. This is a problem when the Authorization header contains a token which has expired since the EventSource was created

### Link to reproduction

https://github.com/trpc/trpc/pull/6047

### To reproduce

Pass a short lived header to unstable_httpSubscriptionLink.eventSourceOptions when using the polyfill, and then reboot the backend server. The header will be re-sent when connection is reestablished

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

## Funding

<a href="https://polar.sh/trpc/trpc/issues/6046">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/6046/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

Nick-Lucas:
Closes #6046 

## 🎯 Changes

## ✅ Checklist

### DIFF
--- a/packages/client/src/links/httpSubscriptionLink.ts
+++ b/packages/client/src/links/httpSubscriptionLink.ts
@@ -11,7 +11,6 @@ import {
 import { TRPCClientError } from '../TRPCClientError';
 import { getTransformer, type TransformerOptions } from '../unstable-internals';
 import { getUrl } from './internals/httpUtils';
-import type { CallbackOrValue } from './internals/urlWithConnectionParams';
 import {
   resultOf,
   type UrlOptionsWithConnectionParams,
@@ -37,7 +36,7 @@ type HTTPSubscriptionLinkOptions<TRoot extends AnyClientTypes> = {
   /**
    * EventSource options or a callback that returns them
    */
-  eventSourceOptions?: CallbackOrValue<EventSourceInit>;
+  eventSourceOptions?: () => EventSourceInit | Promise<EventSourceInit>;
 } & TransformerOptions<TRoot> &
   UrlOptionsWithConnectionParams;
 
--- a/packages/client/src/links/httpSubscriptionLink.ts
+++ b/packages/client/src/links/httpSubscriptionLink.ts
@@ -62,73 +62,95 @@ export function unstable_httpSubscriptionLink<
         let eventSource: EventSource | null = null;
         let unsubscribed = false;
 
-        run(async () => {
-          const url = getUrl({
-            transformer,
-            url: await urlWithConnectionParams(opts),
-            input,
-            path,
-            type,
-            signal: null,
-          });
-
-          const eventSourceOptions = await resultOf(opts.eventSourceOptions);
-          /* istanbul ignore if -- @preserve */
-          if (unsubscribed) {
-            // already unsubscribed - rare race condition
-            return;
-          }
-          eventSource = new EventSource(url, eventSourceOptions);
-          const onStarted = () => {
-            observer.next({
-              result: {
-                type: 'started',
-              },
-              context: {
-                eventSource,
-              },
+        function startSubscription() {
+          run(async () => {
+            const url = getUrl({
+              transformer,
+              url: await urlWithConnectionParams(opts),
+              input,
+              path,
+              type,
+              signal: null,
             });
 
-            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-            eventSource!.removeEventListener('open', onStarted);
-          };
-          // console.log('starting', new Date());
-          eventSource.addEventListener('open', onStarted);
-          const iterable = sseStreamConsumer<
-            Partial<{
-              id?: string;
-              data: unknown;
-            }>
-          >({
-            from: eventSource,
-            deserialize: transformer.output.deserialize,
-          });
-
-          for await (const chunk of iterable) {
-            if (!chunk.ok) {
-              // TODO: handle in https://github.com/trpc/trpc/issues/5871
-              continue;
+            const eventSourceOptions = await resultOf(opts.eventSourceOptions);
+            /* istanbul ignore if -- @preserve */
+            if (unsubscribed) {
+              // already unsubscribed - rare race condition
+              return;
             }
-            const chunkData = chunk.data;
 
-            // if the `tracked()`-helper is used, we always have an `id` field
-            const data = 'id' in chunkData ? chunkData : chunkData.data;
+            eventSource = new EventSource(url, eventSourceOptions);
+
+            const onStarted = () => {
+              observer.next({
+                result: {
+                  type: 'started',
+                },
+                context: {
+                  eventSource,
+                },
+              });
+
+              // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+              eventSource!.removeEventListener('open', onStarted);
+            };
+            // console.log('starting', new Date());
+            eventSource.addEventListener('open', onStarted);
+
+            eventSource.addEventListener('error', (ev) => {
+              if (
+                'status' in ev &&
+                typeof ev.status === 'number' &&
+                [401, 403].includes(ev.status)
+              ) {
+                console.log('Restarted EventSource due to 401/403 error');
+
+                const oldEventSource = eventSource;
+                oldEventSource?.close();
+
+                startSubscription();
+              }
+            });
+
+            const iterable = sseStreamConsumer<
+              Partial<{
+                id?: string;
+                data: unknown;
+              }>
+            >({
+              from: eventSource,
+              deserialize: transformer.output.deserialize,
+            });
+
+            for await (const chunk of iterable) {
+              if (!chunk.ok) {
+                // TODO: handle in https://github.com/trpc/trpc/issues/5871
+                continue;
+              }
+              const chunkData = chunk.data;
+
+              // if the `tracked()`-helper is used, we always have an `id` field
+              const data = 'id' in chunkData ? chunkData : chunkData.data;
+              observer.next({
+                result: {
+                  data,
+                },
+              });
+            }
+
             observer.next({
               result: {
-                data,
+                type: 'stopped',
               },
             });
-          }
-
-          observer.next({
-            result: {
-              type: 'stopped',
-            },
+            observer.complete();
+          }).catch((error) => {
+            observer.error(TRPCClientError.from(error));
           });
-          observer.complete();
-        }).catch((error) => {
-          observer.error(TRPCClientError.from(error));
-        });
+        }
+
+        startSubscription();
 
         return () => {
           observer.complete();
--- a/packages/client/src/links/httpSubscriptionLink.ts
+++ b/packages/client/src/links/httpSubscriptionLink.ts
@@ -94,21 +94,6 @@ export function unstable_httpSubscriptionLink<
           };
           eventSource.addEventListener('open', onStarted, { once: true });
 
-          eventSource.addEventListener('error', async (ev) => {
-            if (
-              'status' in ev &&
-              typeof ev.status === 'number' &&
-              [401, 403].includes(ev.status)
-            ) {
-              console.log('Restarted EventSource due to 401/403 error');
-
-              eventSource?.restart(
-                url,
-                await resultOf(opts.eventSourceOptions),
-              );
-            }
-          });
-
           const iterable = sseStreamConsumer<
             Partial<{
               id?: string;
@@ -117,6 +102,23 @@ export function unstable_httpSubscriptionLink<
           >({
             from: eventSource,
             deserialize: transformer.output.deserialize,
+            tryHandleError: async (ev) => {
+              if (
+                'status' in ev &&
+                typeof ev.status === 'number' &&
+                [401, 403].includes(ev.status) &&
+                eventSource
+              ) {
+                console.log('Restarted EventSource due to 401/403 error');
+
+                eventSource.restart(
+                  url,
+                  await resultOf(opts.eventSourceOptions),
+                );
+
+                return true;
+              }
+            },
           });
 
           for await (const chunk of iterable) {
--- a/packages/client/src/links/httpSubscriptionLink.ts
+++ b/packages/client/src/links/httpSubscriptionLink.ts
@@ -59,98 +59,91 @@ export function unstable_httpSubscriptionLink<
           throw new Error('httpSubscriptionLink only supports subscriptions');
         }
 
-        let eventSource: EventSource | null = null;
+        let eventSource: EventSourceWrapper | null = null;
         let unsubscribed = false;
 
-        function startSubscription() {
-          run(async () => {
-            const url = getUrl({
-              transformer,
-              url: await urlWithConnectionParams(opts),
-              input,
-              path,
-              type,
-              signal: null,
-            });
+        run(async () => {
+          const url = getUrl({
+            transformer,
+            url: await urlWithConnectionParams(opts),
+            input,
+            path,
+            type,
+            signal: null,
+          });
 
-            const eventSourceOptions = await resultOf(opts.eventSourceOptions);
-            /* istanbul ignore if -- @preserve */
-            if (unsubscribed) {
-              // already unsubscribed - rare race condition
-              return;
-            }
+          const eventSourceOptions = await resultOf(opts.eventSourceOptions);
+          /* istanbul ignore if -- @preserve */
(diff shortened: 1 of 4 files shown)


packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -140,3 +167,68 @@ export function unstable_httpSubscriptionLink<
...
+  private es: EventSource;
+
+  private listeners: Partial<Record<keyof EventSourceEventMap, any[][]>> = {};
+
+  constructor(url: string, options: EventSourceInit | undefined) {
+    this.es = new EventSource(url, options);
+  }
+
+  restart(url: string, options: EventSourceInit | undefined) {
+    this.es.close();
```
KATT, line 184: you probably wanna prevent this from emitting any events

Nick-Lucas: Agreed, haven't seen any negative effects, but no sense in risking it

Nick-Lucas: This one is done, removed all known listeners before closing the connection

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -79,22 +88,22 @@ export function unstable_httpSubscriptionLink<
...
+                eventSource: eventSource!.getEventSource(),
               },
             });
-
-            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-            eventSource!.removeEventListener('open', onStarted);
           };
-          // console.log('starting', new Date());
-          eventSource.addEventListener('open', onStarted);
+          eventSource.addEventListener('open', onStarted, { once: true });
```
Nick-Lucas, line 105: Important spec thing given the change in how this is run "once". When we do reinitialize, this callback will be called again with the new EventSource instance.

That felt right in-case anybody depends on holding that instance so they can update the reference and re-attach their own listeners

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -103,6 +112,24 @@ export function unstable_httpSubscriptionLink<
           >({
             from: eventSource,
             deserialize: transformer.output.deserialize,
+            tryHandleError: async (ev) => {
+              if (
```
Nick-Lucas, line 116: we don't offer the chance to reboot when the network error itself occurs and that event doesn't come with a status code, but when it recovers and there's an issue like 401 we do call this. So it can take an extra round trip to recover but feels easier to work with

www/docs/client/links/httpSubscriptionLink.md
```diff
@@ -178,8 +178,10 @@ If the client and server are not on the same domain, you can use `withCredential
 // [...]
 unstable_httpSubscriptionLink({
   url: 'https://example.com/api/trpc',
-  eventSourceOptions: {
-    withCredentials: true, // <---
+  eventSourceOptions() {
```
Nick-Lucas, line 181: I've removed the object variant of this because that way there is 0 nuance in when you can use it. Using a callback is a very small breaking change and gives tRPC more flexibility in calling it for newer values later

KATT: I don't see the motivation for it - what's the problem with `resultOf(opts.eventSourceOptions)`? 👀

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -37,7 +36,17 @@ type HTTPSubscriptionLinkOptions<TRoot extends AnyClientTypes> = {
...
+  /**
+   * For a given error, should we reinitialize the underlying EventSource?
+   *
+   * This is useful where a long running subscription might be interrupted by a recoverable network error,
+   * but the existing authorization header might have expired in the mean-time
+   */
+  shouldReinitialize?: (
+    status: number,
+    error: Event,
+  ) => boolean | Promise<boolean>;
```
KATT, line 49: What if there's a transient error in the `eventSourceOptions`-callback? Should that follow the same code path?

packages/client/src/links/httpSubscriptionLink.ts
KATT: 🤔 

This feels a bit off to me but I don't hate it either... 

I feel like `shouldReinitialize()` should be called whenever it's closed while there is an active subscription

KATT: Sketching

```ts
// closed unexpectedly 
type Closed = { type: "closed" }
// errors from the original EventSource
type Failure = { type: "event", event: Event; }

// if it's an error in i.e. the URL callback or the eventSourceOptions-callback 
type CallbackError = { type: "error", error: TRPCClientError }

type Failure = CallbackError | Closed | Failure
type ShouldReinitialize = (opts: Failure) => MaybePromise<boolean>;

```

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -37,7 +36,17 @@ type HTTPSubscriptionLinkOptions<TRoot extends AnyClientTypes> = {
...
    */
-  eventSourceOptions?: CallbackOrValue<EventSourceInit>;
+  eventSourceOptions?: () => EventSourceInit | Promise<EventSourceInit>;
+  /**
+   * For a given error, should we reinitialize the underlying EventSource?
+   *
+   * This is useful where a long running subscription might be interrupted by a recoverable network error,
+   * but the existing authorization header might have expired in the mean-time
+   */
+  shouldReinitialize?: (
```
KATT, line 46: naming is hard, just thinking of alternatives

```suggestion
  shouldRecreate?: (
```

```suggestion
  shouldRecreateOnError?: (
```

KATT (commented):
I like it but I think:

- I don't see why we need to remove the `eventSourceOptions`-object - I think the most common usecase is `{ withCredentials: true }`
- I think we might wanna make the `shouldReinitialize` more generic

let's discuss on Discord

packages/client/src/links/httpSubscriptionLink.ts
KATT: Maybe `shouldRecreateOn4xx` even if we want it to be specific

Currently naming suggests it's generic but it's very specific

packages/client/src/links/httpSubscriptionLink.ts
Nick-Lucas: I like the idea of that, I would also add a dedicated HttpError to that with status etc expanded because Event lacks the type safety here, but exposing the other states does make sense

www/docs/client/links/httpSubscriptionLink.md
Nick-Lucas: We're still using resultOf internally

If the goal of shouldReinitialize is to get the latest copy of eventSourceOptions, then passing an object implicitly makes that functionality useless. Only allowing a callback means we don't need a load of conditional docs "use x if you're doing y, unless z then use f" becomes "use x if you want it, y if you need it"

packages/client/src/links/httpSubscriptionLink.ts
Nick-Lucas: `shouldRecreateOnError` I definitely prefer, think we can commit to that

packages/client/src/links/httpSubscriptionLink.ts
Nick-Lucas: I'd see that as a userland problem and we just let it bubble

www/docs/client/links/httpSubscriptionLink.md
Nick-Lucas: Alternative would be a warning that if shouldReinitializeOnError is set, tRPC moans that eventSourceOptions needs to be a callback?

packages/client/src/links/httpSubscriptionLink.ts
Nick-Lucas: Right so I checked if there was any "Closed" type error via the test and doesn't seem to be, it may well fire on network disconnection but I've wrapped at "raw" and "http-error" types because I don't want to make too many assumptions about what non-http-error cases will look like but pulling out the status automatically makes a lot of sense as it's a common case

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -140,3 +203,74 @@ export function unstable_httpSubscriptionLink<
...
+
+  private listeners: Partial<Record<keyof EventSourceEventMap, any[][]>> = {};
+
+  constructor(url: string, options: EventSourceInit | undefined) {
+    this.es = new EventSource(url, options);
+  }
+
+  restart(url: string, options: EventSourceInit | undefined) {
+    for (const type in this.listeners) {
+      for (const [t, l, o] of this.listeners[
```
KATT, line 221: - nested loop is usually bad
- don't do one character variable

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -140,3 +203,74 @@ export function unstable_httpSubscriptionLink<
...
   };
 }
+
+/**
+ * We wrap EventSource so that is can be reinitialized with new options
+ */
+class EventSourceWrapper implements Partial<EventSource> {
+  private es: EventSource;
+
+  private listeners: Partial<Record<keyof EventSourceEventMap, any[][]>> = {};
```
KATT, line 213: looks like this could just be

```ts
  private listeners: Array<{
    type: keyof EventSourceEventMap;
    callback: (...args: any[]) => void
    options?: x
  }>
```

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -140,3 +203,74 @@ export function unstable_httpSubscriptionLink<
...
+    type: TEvent,
+    listener: (this: EventSource, ev: EventSourceEventMap[TEvent]) => any,
+    options?: boolean | EventListenerOptions,
+  ) {
+    this.listeners[type] ??= [];
+    for (const [t, l, o] of this.listeners[type] ?? []) {
+      if (t === type && l === listener) {
+        this.listeners[type].splice(this.listeners[type].indexOf([t, l, o]), 1);
+      }
+    }
```
KATT, line 272: this can be simplified to something like

```ts
const idx = this.listeners.findIndex(x => /* ... */)
const listener = this.listeners[idx]
if (!listener) {
  return
}
this.es.removeEventListener(listener.type, listener.callback, listener.options)

// ... splice
```

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -103,6 +123,49 @@ export function unstable_httpSubscriptionLink<
...
+                'status' in ev &&
+                typeof ev.status === 'number' &&
+                typeof opts.shouldRecreateOnError === 'function' &&
+                eventSource
+              ) {
+                let recreateOnErrorOpts: RecreateOnErrorOpt = {
+                  type: 'raw',
+                  event: ev,
+                };
+                if ('status' in ev && typeof ev.status === 'number') {
```
KATT, line 137: Avoid nested `if`s like this, this code is very hard to follow

do an early return or something instead

KATT (changes requested):
I'm fine with the functionality and the interface but the code needs to be simplified - it's very hard to follow:

- Nested ifs
- Nested for loops
- One-character variable names

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -103,6 +123,49 @@ export function unstable_httpSubscriptionLink<
...
+                  return false;
+                }
+
+                if (typeof opts.eventSourceOptions !== 'function') {
+                  // eslint-disable-next-line no-console
+                  console.warn(
+                    'shouldRecreateOnError has returned true but eventSourceOptions is not a function. This means eventSourceOptions will not be updated',
+                  );
+                }
+
```
KATT, line 158: Potential race condition here -

`shouldRestart` can be called twice before it gets a chance to restart

can you make the code in a way to avoid race conditions?

packages/client/src/links/httpSubscriptionLink.ts
Nick-Lucas: Type was slightly wrong with the `any` but best to use the Record as there are known keys and that makes adding and removing event listeners more efficient later

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -103,6 +123,49 @@ export function unstable_httpSubscriptionLink<
...
+                eventSource.restart(
+                  url,
```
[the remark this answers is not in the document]
Nick-Lucas: Didn't realise that, I'll also remove the warning then since there's more than 1 reason to use this method

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -140,3 +192,97 @@ export function unstable_httpSubscriptionLink<
...
+  private listeners: Partial<
+    Record<
+      keyof EventSourceEventMap,
+      Parameters<EventSource['addEventListener']>[]
+    >
+  > = {};
+  private *getAllEventListeners() {
+    for (const _type in this.listeners) {
+      const type = _type as keyof EventSourceEventMap;
+      for (const listener of this.listeners[type] ?? []) {
```
KATT, line 226: i'm guessing it could be
```suggestion
      const type = _type as keyof typeof this.listeners;
      for (const listener of this.listeners[type]) {
```

packages/server/src/unstable-core-do-not-import/stream/sse.ts
```diff
@@ -227,19 +228,38 @@ export function sseStreamConsumer<TData>(opts: {
...
-  eventSource.addEventListener('error', (cause) => {
+  eventSource.addEventListener('error', async (cause) => {
+    // We prevent more than 1 error handler from waiting at the same time
+    const result = await errorLock;
+    if (result === 'CANCEL_ALL') {
+      return;
+    }
+
+    const resolvable = createResolvable<void | 'CANCEL_ALL'>();
+    errorLock = resolvable.promise;
```
KATT, line 246: we have a utility for this already called `createDeferred`
```suggestion
    const deferred = createDeferred<void | 'CANCEL_ALL'>();
    errorLock = deferred.promise;
```

packages/server/src/unstable-core-do-not-import/stream/sse.ts
```diff
@@ -271,6 +291,20 @@ export function sseStreamConsumer<TData>(opts: {
...
+    resolve = _resolve;
+  });
+
+  if (resolve === NOOP) {
+    throw 'Unlock not assigned';
+  }
+
+  return { promise, resolve };
+}
+
```
KATT, line 307:
```suggestion
```

www/docs/client/links/httpSubscriptionLink.md
```diff
@@ -113,6 +115,12 @@ const trpc = createTRPCClient<AppRouter>({
             },
           }; // you either need to typecast to `EventSourceInit` or use `as any` or override the types by a `declare global` statement
         },
+
+        // Optional: EventSource will automatically recover from disconnections, but always uses the initial headers
+        // If your auth header expires, you can force a restart and fresh eventSourceOptions() call when shouldRecreateOnError() returns true
+        shouldRecreateOnError(status, _error) {
+          return opts.type === 'http-error' && [401, 403].includes(opts.status);
+        },
```
KATT, line 123: I think this is worth its own header in the docs as well

Nick-Lucas: Fair point, done and cleaned up this one since it becomes just noise

packages/client/src/links/httpSubscriptionLink.ts
Nick-Lucas: Above comment is outdated

Approach now just locks the error stream, and if an error gets handled will discard the waiting stream to avoid restarting again immediately. Feels like only compromises in this space so simple is good as a first approach I think

KATT (approved):

PR #6063, opened 2024-09-28
feat: implement output parsers for subscriptions

### LINKED ISSUE #6062: feat: subscriptions should work together with output parsers

Currently it is impossible to write something as:

```typescript
export const test = t.procedure
  .output(z.string())
  .subscription(async function* () {
    yield "hello";
    await sleep(1000);
    yield "world";
  });
```

It won't compile with a TypeScript error and also will fail at runtime because a `AsyncIterable<string>` is not a `string`.  
Client types happen to generate correctly.

It should be allowed, it should correctly type the function and it should validate the subscribed data, not the subscription itself.

The above code should be allowed and work as expected. Nothing new needs to be added for this to work.

The current solution is to skip using output parsers and instead provide the type manually, without any runtime safety:

```typescript
export const test = t.procedure
  .subscription<string>(async function* () {
    yield "hello";
    await sleep(1000);
    yield "world";
  });
```

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

## Funding

<a href="https://polar.sh/trpc/trpc/issues/6062">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/6062/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

lbguilherme:
Closes #6062

## 🎯 Changes

This PR allow us to specify output parsers on subscription procedures. It will work with legacy style observers and also with async generators. Types are correctly generated and values are checked (parsed) at runtime.

## ✅ Checklist

### DIFF
--- a/packages/server/src/observable/observable.ts
+++ b/packages/server/src/observable/observable.ts
@@ -1,3 +1,4 @@
+import type { MaybePromise } from '../unstable-core-do-not-import/types';
 import type {
   Observable,
   Observer,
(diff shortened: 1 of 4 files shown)


KATT (commented):
Let's discuss in the issue, see https://github.com/trpc/trpc/issues/6062#issuecomment-2380641605

PR #6064, opened 2024-09-28
fix(server) + feat(www): fix subscriptions output validation & refresh subscription docs 

### LINKED ISSUE #6062: feat: subscriptions should work together with output parsers

KATT:
- Closes #6062
- Closes #6063

## 🎯 Changes

- Moves subscription docs inside /server
- Moves WebSockets docs inside /server
- Moves a bunch of common code from SSE/WS to the /subscription docs
- Adds tests with validators of async iterables
- Fixes a type on subscriptions with output validators
- Fixes so that `onError`-callback is called on SSE output validation failures

## ✅ Checklist


packages/server/src/unstable-core-do-not-import/stream/sse.ts
```diff
@@ -92,6 +92,7 @@ export function sseStreamProducer(opts: SSEStreamProducerOptions) {
         closedPromise,
         maxDurationPromise.promise,
       ]);
+      console.log('next', next);
```
KATT, line 95:
```suggestion
```

PR #6070, opened 2024-09-30
ref

KATT:
Closes #2221

## 🎯 Changes

Reference code for https://github.com/trpc/trpc/discussions/2221


packages/tests/server/regression/issue-2221-router-factory-fn.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      )
+      .mutation(async (opts) => {
+        // [...db operations]
+
+        return {
+          ...opts.ctx.user,
+          ...opts.input,
+        };
+      }),
+  }));
```
KATT, line 66: cc @ahkhanjani @lanternlogic here's an example

PR #6073, opened 2024-10-01
feat: allow overriding of `mutationFn`

simonecervini:
## 🎯 Changes

This PR enables users to override the `mutationFn` value via mutation options and default mutation options. 

This is very useful for offline-first apps when you want to conditionally mutate something locally without a network request. For example:

```ts
const utils = createTRPCQueryUtils({
  queryClient: opts.queryClient,
  client: opts.trpcClient,
});

utils.user.updatePhoto.setMutationDefaults(({ canonicalMutationFn }) => ({
  mutationFn: async (...params: Parameters<typeof canonicalMutationFn>) => {
    // The user could also be a guest user (generated locally without a connection)
    // For guest users, it doesn't make sense to perform the mutation online
    // because they would probably receive an "UNAUTHORIZED" error
    const user = utils.user.get.getData()!;
    if (!user.isGuest) {
      return canonicalMutationFn(...params);
    }
  },
  onMutate: async (input) => {
    // classic optimistic update
  }
}));
```

## ✅ Checklist

### DIFF
--- a/packages/react-query/src/shared/hooks/createHooksInternal.tsx
+++ b/packages/react-query/src/shared/hooks/createHooksInternal.tsx
@@ -316,9 +316,12 @@ export function createRootHooks<
       {
         ...opts,
         mutationKey: mutationKey,
-        mutationFn: (input) => {
-          return client.mutation(...getClientArgs([path, { input }], opts));
-        },
+        mutationFn:
+          opts?.mutationFn ??
+          defaultOpts?.mutationFn ??
+          ((input) => {
+            return client.mutation(...getClientArgs([path, { input }], opts));
+          }),
         onSuccess(...args) {
           const originalFn = () =>
             opts?.onSuccess?.(...args) ?? defaultOpts?.onSuccess?.(...args);


simonecervini:
Hey @KATT, I opened this PR as a draft because I haven't added tests yet. Can you please let me know if you think this change makes sense for the library (because I can imagine the overriding of mutationFn was disabled on purpose)? 

If so, I can add tests and mark this as ready to review.

Thank you!

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -316,9 +316,12 @@ export function createRootHooks<
       {
         ...opts,
         mutationKey: mutationKey,
-        mutationFn: (input) => {
-          return client.mutation(...getClientArgs([path, { input }], opts));
-        },
+        mutationFn:
+          opts?.mutationFn ??
+          defaultOpts?.mutationFn ??
```
KATT, line 321: Probably not needed / good to allow overriding the mutationFn for alllllll mutations? 🤔 

If so, it should probably be more in the direction of [our `overrides:`-option](https://trpc.io/docs/client/react/useUtils#invalidate-full-cache-on-every-mutation)?

KATT (commented):
Sorry for the slow response, my github notifications are 💀 

This looks good to me! Just some tests

PR #6079, opened 2024-10-04
feat(server): add Standard Schema support

fabian-hiller:
## 🎯 Changes

An union of schema library authors (Zod, Valibot, ArkType, ...) collaborating on a standard interface for schema libraries called [Standard Schema](https://github.com/standard-schema/standard-schema). This simplifies implementation and prevents vendor lock-in. So far Valibot v1 supports this spec and Zod will follow soon. This PR adds support for all schema libraries that follow the v1 of this spec in the future.

> In the long run, the explicit implementation of Valibot and Zod specific code can be removed from tRPC.

## ✅ Checklist

## Summary by CodeRabbit

- **Dependencies**
	- Updated `valibot` package to version `1.0.0-beta.14` across multiple packages
	- Introduced support for parallel versioning of `valibot`

- **Validation**
	- Simplified input and output validation syntax
	- Removed `v.parser()` wrapper in validation methods
	- Added support for new schema validation approach

- **Error Handling**
	- Introduced `StandardSchemaV1Error` for more structured error reporting
	- Enhanced error formatting capabilities

- **Documentation**
	- Updated quickstart and server validation documentation to reflect new validation approach


packages/server/src/unstable-core-do-not-import/parser.ts
```diff
@@ -110,5 +118,17 @@ export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
     };
   }
 
+  if ('~standard' in parser) {
+    // StandardSchemaEsque
+    return async (value) => {
+      const result = await parser['~validate']({ value });
+      if (result.issues) {
+        console.error('Schema issues:', result.issues);
```
juliusmarminge, line 126:
```suggestion
```

packages/server/src/unstable-core-do-not-import/parser.ts
```diff
@@ -110,5 +118,17 @@ export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
...
+        throw new Error(result.issues[0].message);
```
juliusmarminge, line 127: Should there not be a StandardError or something? Having a generic error here makes it impossible to do errorFormatter like this one? https://trpc.io/docs/server/error-formatting#adding-custom-formatting

Can we attach all the issues on Error.cause or something?

fabian-hiller: I added our `SchemaError` utility class. Does this look okay to you? Or should we add `SchemaError` directly to the source code and not import it from `@standard-schema/utils`?

juliusmarminge: Yea no deps and mports. Just a way to revive the full error with all the issues instead of just the first one

packages/server/package.json
```diff
@@ -141,12 +142,15 @@
...
     "yup": "^1.0.0",
     "zod": "^3.0.0"
   },
   "funding": [
     "https://trpc.io/sponsor"
-  ]
+  ],
+  "dependencies": {
+    "@standard-schema/utils": "^0.1.0"
+  }
```
KATT, line 155: We don't want dependencies.

```suggestion
```

packages/server/src/unstable-core-do-not-import/parser.ts
```diff
@@ -110,5 +120,16 @@ export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
     };
   }
 
+  if ('~standard' in parser) {
+    // StandardSchemaEsque
+    return async (value) => {
+      const result = await parser['~validate']({ value });
+      if (result.issues) {
+        throw new SchemaError(result.issues);
+      }
```
KATT, line 129: If `result.reason` was an `Error`-type, we could've just rethrown here

But for now I think we gotta do something like

```ts
if (result.issues) {
  throw new Error("Validation error", { cause: result.issues })
}
```

For us, it would be a bit simpler if there was an actual `Error` here

packages/server/src/unstable-core-do-not-import/parser.ts
```diff
@@ -18,6 +20,13 @@ export type ParserArkTypeEsque<TInput, TParsedInput> = {
   infer: TParsedInput;
 };
 
+export type ParserStandardSchemaEsque<TInput, TParsedInput> = {
+  '~types'?: {
+    input: TInput;
+    output: TParsedInput;
+  };
```
KATT, line 27: We have some different wip-branches where we do infer errors too, does `~types` also have an `Error`-subclass?

In the case of zod, we could infer that pretty easily in #6027 where we played with this, see L12 here:

https://github.com/trpc/trpc/blob/de219a3c4aab74cd2682f408c88f884be9cef587/packages/server/src/%40trpc/parser.ts#L2-L15

KATT (commented):
I like that y'all are converging at a standard way of doing things

We don't want third party dependencies in trpc though

packages/server/src/unstable-core-do-not-import/parser.ts
KATT: As @juliusmarminge already pointed out, it'd be nice if it was something like

```ts
if (result.error) {
  throw result.error
}
```

tuples work well too

```ts
const [err, result] =  await parser['~validate'](value);

if (err) {
  throw err;
}
return result
```

fabian-hiller: I think forcing schema libraries to use a specific error class to be compatible with Standard Schema is not a good idea, as it would unnecessarily compromise bundle size, performance (creating `Error` instances is expensive), and flexibility.

packages/server/src/unstable-core-do-not-import/parser.ts
fabian-hiller: What's the benefit of inferring the error type?

KATT: Think e.g. a form submission - it'd be nice to infer the right path when rendering the errors

fabian-hiller: Why do you need the error class for this?

juliusmarminge: Cause there's no way to get the type of an error cause, so if you'd to `throw new Error("", { cause: issues })` you'd need a cast like `(err.cause as SchemaIssues<TSchema>).fieldErrors.email` for example. Or how would you infer here?

Would be nice if something strictly typed came out of the validate method directly as a built-in

fabian-hiller: Are Zod's errors strictly typed? I thought it is the same error type for every schema? Zod's issues are also not typed as far as I know. But maybe I'm wrong. Could the paths also be inferred from the input type?

juliusmarminge: they are: 
CleanShot 2024-10-08 at 00 24 15

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgLzgXzgMyhEcBEyEAJvgNwBQFAxhAHYDO8IAngMrUAWApiAIZwAvCgB0EAEYArbtRgAKJJggQAXKKZRgdAOZyAlABo44vlDXIRdAK4hx3KPvR6qtRvCjcGVgDbxhrDh5+EQY+TG4ABVMGbgU4OggYADFlNQByNKMTMzgANgBOJyoPL18RexwoCgB6arh6+oA9AH4KEp8YcqhKkUxvPhgYbjp9XuBub2IAUW7oBhEgA

fabian-hiller: Thank you for the example! I understand it much better now. I think this is also possible with Standard Schema. We know the input and output type, so we know all possible paths that can occur, and can transform them into an object of dot paths. I can provide example code.

fabian-hiller: The idea is basically to infer the input type from the schema. Standard Schema provides an `InferInput` util type for this. Then we could use a modified version of [ts-essentials'](https://github.com/ts-essentials/ts-essentials) `Paths` type, which converts the input type into a union of all possible paths. Finally, we pass this type to `Record` to construct the `FieldErrors` type.

```ts
type FieldErrors = Record<Paths<InferInput<TSchema>>, string[]>
```

fabian-hiller:
Does this PR look good to you, or should I change or modify anything?

packages/server/src/unstable-core-do-not-import/parser.ts
KATT: Using an object might not be the best idea as errors could occur on e.g both `foo` and `foo.bar` simultaneously - an array with issues and paths is probably best for a generic lib - you can supply functions to turn it into an object

fabian-hiller: > Using an object might not be the best idea as errors could occur on e.g both foo and foo.bar simultaneously

You don't think this is a problem, or are you aware of an edge case? That's why I wrote that the `Paths` type needs to be modified. I can provide a raw implementation if you are interested.

> an array with issues and paths is probably best for a generic lib

I completely agree. That's exactly what the Standard Schema does.

> you can supply functions to turn it into an object

I have thought about providing such a function with `@standard-schema/utils`, but such an implementation requires a lot of thought to make it work for everyone. Some libraries are probably only interested in nested messages, because they assume (or know) that the root type is always valid. Others are also interested in errors related to the root schema. There may also be edge cases where a path cannot be converted to a dot path. Then there are libraries that need the value to be a `string` and others that need it to be a `string[]`.

We are very interested in and grateful for feedback from libraries like tRPC to make the spec as perfect as possible.

juliusmarminge:
how done is the spec? are you not open to returning some tuple-like thing from `~validate()` so libs can rethrow?

fabian-hiller:
I have been working with Colin and others on the interface and we think the spec is already in good shape, but at the end of the day we are in beta and Standard Schema only works if we can get everyone on board. The idea of the beta is to get feedback like yours and we are more than happy to discuss things and improve the spec together. I am also available for a call to figure things out.

KATT (approved):

fabian-hiller:
We will probably release a new beta version of the spec this week. I keep you posted.

fabian-hiller:
I updated the Standard Schema version and improved the error handling.

packages/server/src/unstable-core-do-not-import/parser.ts
```diff
@@ -110,5 +118,25 @@ export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
...
 }
+
+class SchemaError extends Error {
+  public readonly issues: ReadonlyArray<v1.StandardIssue>;
+  constructor(issues: ReadonlyArray<v1.StandardIssue>) {
+    super(issues[0]?.message);
+    this.name = 'SchemaError';
+    this.issues = issues;
+  }
+}
```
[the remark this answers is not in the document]
fabian-hiller: Zod will probably still have inferrable errors via `.parse`, but throwing errors via the standard schema interface would force any schema library to add the spec as a runtime dependency to throw the same error class. This is probably not an option. If you prefer the previous implementation with `throw new Error('Schema Error:', { cause: result.issues });` I can revert it.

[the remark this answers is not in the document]
juliusmarminge: Even without the error class `result.issues` could be generic, no?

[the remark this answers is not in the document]
fabian-hiller: What would be the advantage of having `result.issues` as a generic for a library like tRPC?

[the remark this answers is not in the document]
juliusmarminge: Right now, none.

I'm just thinking about the spec in general and want it to be as good as possible. When doing form validation it'd be nice to have the possible fields that might error be autocompletable for example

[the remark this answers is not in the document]
fabian-hiller: I am not sure if `result.issues` as a generic will help with autocompletable fields, as this information can already be extracted from the input type of the schema using a [`Paths` utility type](https://github.com/ts-essentials/ts-essentials/tree/master/lib/paths). So far, the only schema library I know of that has almost type-safe issues is Valibot. Most other libraries just return a general issue interface.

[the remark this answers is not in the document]
juliusmarminge: CleanShot 2024-11-12 at 19 24 29@2x

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAKjgQwM5wF5wGZQiOAIgwgBNCAoCgYwgDtV5VqALAUxGTgF5MA6CACMAVm2owAFAG8KcOXEHIoALn6MowOgHMJASgoBffTXqM4UNqgCuAG3i9m7Tn1TJsbAApLUbaUarA2HASAIQW1nYuVtTUlqi6cDLy5pa2MHxsUHhQANxyAPT5cAA8ALSlcFpsdJnA1LLyhclyAHoA-FTJtAzw2MBsNqQAolnQ6LzhaRmjUG182DbIMDDVevP9gyPZqA1yfQPDM6h8cE3IVjAQcLTgNmwwyIJ3u6dF7YZAA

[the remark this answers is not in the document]
fabian-hiller: Thank you for this example. I understand you much better now. Since the Standard Schema spec is intentionally minimal, it most likely won't include a `flatten` method as part of its issues, but the spec already provides all the information to implement a `flatten` function yourself (or as part of our official [utils package](https://github.com/standard-schema/standard-schema/tree/main/packages/utils)).

A `flatten` function/method requires 3 things:
1. The **path information** to create a dot path
2. The **error messages** related to a specific path
3. The **input type** to make flatten type safe

Every issue of the Standard Schema spec includes a `.message` and optionally a `.path` property. This information can be used at runtime to create a flat error object. The input type can be used to make the output of `flatten` typesafe by extracting every possible nested path.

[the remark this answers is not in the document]
fabian-hiller: To summarise, if I understood you correctly, you were basically asking whether we should additionally include the type information in `result.issues` to make it easier to extract something like `fieldErrros` in a typesafe way.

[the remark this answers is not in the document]
juliusmarminge: alrght, sounds good.

then all that's left is to decide whether to use this custom `SchemaError` or `new Error("Schema validation failed", { cause: issues })`. What do you think? Will you try and recommend one path over the other for integrations?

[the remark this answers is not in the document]
fabian-hiller: I think something like a `SchemaError` might make sense. On the other hand, we are still figuring out if we should provide a `@standard-schema/utils` package with runtime code like a `SchemaError` class or not.

In the long run, a goal would be to have tRPC and other similar libraries mainly or only support Standard Schema as a common interface for schema libraries. If these libraries also commit to a unified implementation, e.g. by using `SchemaError`, the DX for both sides will be great. But we are still in the phase of figuring things out. Feel free to share your thoughts!

The current plan is to add more libraries to Standard Schema by creating more such pull requests. This way we will hopefully get more feedback. In 2 or 3 weeks we could imagine releasing a first v1 RC of the spec and making Standard Schema more public, for example by talking more about it on X.

packages/server/src/unstable-core-do-not-import/parser.ts
```diff
@@ -110,5 +118,25 @@ export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
     };
   }
 
+  if ('~standard' in parser) {
```
mmkal, line 121: It would probably be sensible to move this higher up. All of the existing validation libraries won't actually reach here. 

Now that standard-schema exists, the onus should be on validation libraries to implement it, and trpc's docs should slowly move towards "just use any standard-schema lib". At that point, most users will expect the `parser['~standard'].validate(...)` method to be used over `parser.parse(value)` or whatever. Hopefully there being any differences would be very rare, but still I think better to prioritise standard-schema.

juliusmarminge: that would be a breaking change though. since for example a validation failure on a zod schema would not throw ZodError when using `~standard.parse`

fabian-hiller: That's why I put it at the end. Once Standard Schema is more established, tRPC may move it to the top when they release a new version.

colinhacks:
Summarize some discussion points for clarity/posterity, and because we'll be pointing to this PR as a model for other integrators 👍

## 1 `Error` subclass 

We definitely don't want to require spec-compliant libraries to throw Errors for performance reasons. It's a huge hit to performance in the "validation failure" case and we don't want to impose a performance tax like that on everyone consuming spec-compliant validators.

> For us, it would be a bit simpler if there was an actual Error here

The hard part here is creating a spec that library authors will actually implement. It's expected that integrators like tRPC will need to do some massaging to fit the result of `~standard.validate()` into their own systems. No way to please everyone so the spec is just trying to be the lowest common denominator.

## 2 Error formatting / inferrable errors

Given a compliant schema, you have an inferred type and you have the error information (including paths for each issue). So you have all the ingredients you need to replicate Zod's current `ZodError#format()` and `ZodError#flatten()` logic and build a generic, structured result. As Fabian mentioned this is also Moreover, this logic would work for all spec-compliant schemas, not just Zod. Here's where that logic lives in Zod 3: https://github.com/colinhacks/zod/blob/f7ad26147ba291cb3fb257545972a8e00e767470/src/ZodError.ts#L221C3-L266C7 

As Fabian mentioned this could conceivably be implemented in the `@standard-schema/utils` package at some point. But this is entirely optional. I was mildly against the creation of a utils package to avoid any confusion, and to re-assure everyone that the spec is 100% static-only. It does seem to have caused some confusion here, so just to re-iterate: *there's absolutely no need for any schema library or framework/integrator to install any runtime dependency to benefit from Standard Schema*.

## 3 Zod 4 questions

> if they migrate to StandardSchema will they lose that? That'd be a big loss

> I am not sure what Zod's result.issues

There seems to be some confusion here. Zod 3.24 already implements the spec, and there were no changes made to `ZodError`. All of the functionality you're alluding to still exists. I can easily implement this spec without losing or changing of any functionality on Zod. The behavior & return type of Zod's first party `parse` methods can be completely independent of the `~standard.validate()` method. Is this making sense? 

You can see Zod's implementation of the spec here: https://github.com/colinhacks/zod/blob/f7ad26147ba291cb3fb257545972a8e00e767470/src/types.ts#L287-L294

To answer your other question @juliusmarminge, Zod 4 is pulling this logic out into separate `formatError` and `flattenError` functions for bundle size reasons but the functionality will still exist.

csvn:
Thanks for the update and great summary!

The PR looks pretty complete, what is there left to update here now? Is it to remove `@standard-schema/spec` as a `devDependency`?

Would be lovely to get this PR merged when it's ready, as working with `valibot@latest` (`1.0.0-beta.14`) has some issues with inferring Input/Output correctly due to the `schema._types` to `schema['~types']` changes.

Not trying to put pressure, just curious. The changes here seems to make it much nicer to handle validation/parsing as well (without wrapping in the `valibot.parse(schema)` method), so just looking forward to them being in tRPC v11!

fabian-hiller:
> .. as working with `valibot@latest` (`1.0.0-beta.14`) has some issues with inferring Input/Output correctly due to the `schema._types` to `schema['~types']` changes.

Valibot v1 supports Standard Schema. There should be no bugs. And tRPC should still support our old implementation. But I can look into it. Maybe I am wrong. Feel free to give me feedback. Maybe I missed something.

> Not trying to put pressure, just curious. The changes here seems to make it much nicer to handle validation/parsing as well (without wrapping in the `valibot.parse(schema)` method), so just looking forward to them being in tRPC v11!

Exactly 😎

csvn:
> Valibot v1 supports Standard Schema. There should be no bugs. And tRPC should still support our old implementation. But I can look into it. Maybe I am wrong. Feel free to give me feedback. Maybe I missed something.

@fabian-hiller I don't want to hijack this PR, my I described my issue in the section below.
It's a great idea for a shared API for validators, great initiative and thanks for all the hard work! 😍

<details>

<summary><strong>Click to show</strong></summary>

### Issue

The issue I get is with `valibot@1.0.0-beta.14` + `@trpc/server@11.0.0-rc.660`. It seems that the client is using the "Output" type on the client instead of the "Input" type for queries/mutations.

```ts
// On the server
const querySchema = v.object({
  q: v.optional(v.string(), '')
});

export const appRouter = router({
  list: publicProcedure.input(v.parser(v.object({
    q: v.optional(v.string(), '')
  })))
});

export type QueryInput = v.InferInput<typeof querySchema>;
             // ^ { q?: string | undefined; }
export type QueryOutput = v.InferOutput<typeof querySchema>;
             // ^ { q: string; }
```

```ts
// On the client
trpc.list.query({});
             // ^ Property 'q' is missing in type '{}' but required in type '{ q: string; }'.  ts(2345)

// The type here should be the same as `QueryInput`, but it is the same as `QueryOutput`
```

The only issue I've had is when using `optional()` with a default value, which makes the Input/Output types differ. `transform()` may also cause similar issues.

### Likely cause

I haven't looked too close at the issue (don't know how tRPC manages to infer the output type), but it looks to me like the `ParserValibotEsque` type is no longer matching the schema for the current Valibot `latest` version

image

For now, I got it working correctly by wrapping `v.parser` with a type adjustment, which makes tRPC able to correctly infer the Input type again:

```ts
publicProcedure
    .input(compat(v.parser(querySchema)))

function compat<T extends v.Parser<any, any>>(parser: T): T & {
  schema: {
    _types?: {
      input: v.InferInput<T['schema']>;
      output: v.InferOutput<T['schema']>;
    };
  }
} {
  return parser;
}
```

</details>

fabian-hiller:
Yes, what you say may be true. This should be fixed once Standard Schema is supported. It will also allow you to remove the `parser` function. In the long run, we should probably remove the Valibot specific code.

packages/server/src/unstable-core-do-not-import/parser.ts
```diff
@@ -1,3 +1,5 @@
+import type { v1 } from '@standard-schema/spec';
```
KATT, line 1: Please vendor this type instead of adding a library

```suggestion
```

packages/server/package.json
```diff
@@ -117,6 +117,7 @@
   },
   "devDependencies": {
     "@fastify/websocket": "^10.0.1",
+    "@standard-schema/spec": "^1.0.0-beta.3",
```
KATT, line 120:
```suggestion
```

KATT (commented):
I'm happy to see this PR get up to date and merged in general

If there's changes to the standard schema, I guess you can just add versioning too

The only tweak I'd want would be to vendor rather than `import`ing the standard schema

packages/server/src/unstable-core-do-not-import/parser.ts
fabian-hiller: Fixed. I created a `standardSchema.ts` file. Feel free to rename it or to move the code around.

KATT (approved):

PR #6082, opened 2024-10-06
chore(server): `EventSource` recreation improvements

KATT:
Continuation of #6046 / #6047

## 🎯 Changes

- Handle `EventSource`-connection fully inside of `sse.ts`
- Do not use non-custom `Event`-type - you gotta roll your own
- Pause all events while the connection logic is called
- prefix `shouldRecreate` with `experimental_` - I really wanna find a better solution than this where we don't need to sacrifice `tracked()`


packages/client/src/links/httpBatchStreamLink.ts
KATT, line 1: _(Changes here are unrelated, you can ignore them)_

packages/client/src/links/httpBatchLink.ts
KATT, line 1: (Changes here are unrelated, you can ignore them)

packages/client/src/links/httpSubscriptionLink.ts
```diff
@@ -75,102 +58,69 @@ export function unstable_httpSubscriptionLink<
...
+              input,
+              path,
+              type,
+              signal: null,
+            }),
+          init: () => resultOf(opts.eventSourceOptions),
+          signal,
+          deserialize: transformer.output.deserialize,
+          shouldRecreateOnError: opts.experimental_shouldRecreateOnError,
+        });
```
KATT, line 88: The stream is initialized with callbacks on how to get the URL and the options

PR #6088, opened 2024-10-07
fix(server): ensure order of observable subscription errors

oleasteo:
## 🎯 Changes

The `ReadableStreamDefaultController.error()` causes **any** future interactions with the associated stream to error.
Thus, if the underlying observable receives an error, it would have caused any `.next()` call on the iterator to throw instantly. In case of multiple emits, this would cause the error to propagate out of order.

In addition, this would have caused the `stopped` package in the ws adapter not to be sent. `await iterator.return?.()` would have thrown.

I'm not sure when this was introduced, but it seems to have been a somewhat recent regression.

## ✅ Checklist

- [x] ~If necessary, I have added documentation related to the changes made.~

### DIFF
--- a/packages/server/src/observable/observable.ts
+++ b/packages/server/src/observable/observable.ts
@@ -1,3 +1,4 @@
+import type { Result } from '../unstable-core-do-not-import';
 import type {
   Observable,
   Observer,
@@ -9,9 +10,9 @@ import type {
 
 /** @public */
 export type inferObservableValue<TObservable> = TObservable extends Observable<
-  infer TValue,
-  unknown
->
+    infer TValue,
+    unknown
+  >
   ? TValue
   : never;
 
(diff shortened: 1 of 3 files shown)


oleasteo:
Btw, I was unable to create a proper test case for the `"stopped"` message to be sent after an error as the `onStopped` callback did not seem to be called regardless of the respective `respond(...)` line to be reached.

Maybe, someone with a more profound insight should create such a test case (and / or fix a possible remaining bug).

KATT (approved):
thank you!

appreciate this a lot - the code around the websockets and all of this are an absolute mess right now; impressed you worked it out

KATT:
if you setup github sponsors I happily send you a thank you $ :)

KATT:
feel free to book me if you wanna get more involved, seems like trpc would be better off with it - https://cal.com/alexander

oleasteo:
Thank you for the blazing fast RTT and a great project altogether :tada:

PR #6108, opened 2024-10-12
fallback handler for DELETE requests

bompi88:
## 🎯 Changes

Fallback handler has been added for DELETE requests as already done for GET requests. A `Content-Type` header should not be required for `DELETE` requests.

## ✅ Checklist

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/http/contentType.ts
+++ b/packages/server/src/unstable-core-do-not-import/http/contentType.ts
@@ -262,7 +262,7 @@ function getContentTypeHandler(req: Request): ContentTypeHandler {
     return handler;
   }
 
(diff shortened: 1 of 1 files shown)


KATT (commented):
why would you call a trpc procedure with `DELETE`? we don't support it

bompi88:
@KATT I see now that I might need to open a PR in [trpc-to-openapi](https://github.com/mcampa/trpc-to-openapi) instead, as this one is using something imported from `unstable-core-do-not-import`. The name gives it away. Their DELETE endpoint does not work unless a `Content-Type`-header is provided. I traced the line down to this line of code, but it seems like it shouldn't have been imported at all.

KATT:
Let's do that instead!

Issue #3475, opened 2022-12-24
bug: Procedure return types result in `any` when using non-relative import paths


delyan-peev:
### Provide environment information

```
  System:
    OS: macOS 12.5
    CPU: (10) arm64 Apple M1 Pro
    Memory: 2.67 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh

  Binaries:
    Node: 16.18.1 - ~/.nvm/versions/node/v16.18.1/bin/node
    npm: 8.19.2 - ~/.nvm/versions/node/v16.18.1/bin/npm

  Browsers:
    Chrome: 108.0.5359.124
    Safari: 15.6

  npmPackages (Server):
    @trpc/server: ^10.4.3 => 10.4.3 
    typescript: ^4.9.4 => 4.9.4

  npmPackages (Client):
    @tanstack/react-query: ^4.19.1 => 4.19.1 
    @trpc/client: ^10.4.3 => 10.4.3 
    @trpc/react-query: ^10.4.3 => 10.4.3 
    @trpc/server: ^10.4.3 => 10.4.3 
    react: ^18.2.0 => 18.2.0 
    typescript: ^4.9.4 => 4.9.4
```

### Describe the bug

Hey guys,

First of all, thanks for the awesome package, it's been revolutionizing the way we write React apps. 

Not sure if this is a known issue, but return types of queries (and not only) in the client result in `any` when using non-relative (either absolute or TS aliased paths) imports on the server. I supposed it is a bug, since I did not find the docs mentioning relative paths are required.

Actually, if you change the import of the whole `entitiesRouter` to be non-relative (see example and description below), it will return an object of type `any` for `trpc.entities` in the client. So the issue is not only for queries but for anything that is not imported with a relative path in `trpc`.

This is kind of frustrating for us, since it requires us to use relative imports pretty much everywhere in the project and it can easily result in a relative imports hell.

E.g.
`import { A } from '../../../../../../wherever/whatever';`

Is there a workaround for this what would it be?

Apologies if this has been asked and answered, I searched through the issues and the docs, but did not find anything.

Regards,
Delyan

### Link to reproduction

https://github.com/delyan-peev/trpc-relative-paths-bug

### To reproduce

1. Run `npm install` in both the `api-server` and `client` directories.
2. Run `npx prisma generate client` in the `api-server` directory.
3. Observe the returned type of the data property in `client/src/Test.tsx`
[...]

Issue #2928, opened 2022-10-06
bug: no type safety when using nested routers?


jclaessens97:
### Provide environment information

```
❯ npx envinfo --system --binaries --browsers --npmPackages "typescript,next,react,@trpc/server,@trpc/client,@trpc/react,@trpc/client,@trpc/next,@tanstack/react-query"
Need to install the following packages:
  envinfo
Ok to proceed? (y) y

  System:
    OS: macOS 12.4
    CPU: (10) arm64 Apple M1 Max
    Memory: 16.16 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.asdf/installs/nodejs/16.14.0/bin/node
    npm: 8.3.1 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 105.0.5195.125
    Safari: 15.5
```

### Describe the bug

I followed a simple example on how to configure a multi router setup (using the child routers). However when I'm then trying to use it in my client, I only have limited type support. It only recognizes the first "layer" of the router (being the child router). On the childrouters itself I don't seem to have any typings.

Might be that I'm doing something wrong, but I followed the example very closely.

Another thing is that I'm working with a separate server, but I have to import the AppRouter in my web project directly from the API project. Is there another, cleaner way, of doing this? 

### Link to reproduction

https://github.com/jclaessens97/trcp-nested-route-repro

### To reproduce

Implement nested router, use it in client, see that type safety stops at the first layer.

### Additional information

versions:
```
    "@trpc/client": "10.0.0-proxy-beta.15",
    "@trpc/server": "10.0.0-proxy-beta.15",
```

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

KATT:
On my phone but this is not a bug in trpc. Might be a bad alias here? 
![Screenshot_2022-10-07-14-11-15-26_40deb401b9ffe8e1df2f1cc5ba480b12](https://user-images.githubusercontent.com/459267/194561929-e5cc86fa-ad12-4047-a31c-ebdd5ec47612.jpg)

PR #6112, opened 2024-10-12
docs(server/routers): add warning to use relative imports to trpc.ts

oleasteo:
## ✅ Checklist

### DIFF
--- a/www/docs/server/routers.md
+++ b/www/docs/server/routers.md
@@ -26,6 +26,10 @@ export const publicProcedure = t.procedure;
 
 You'll notice we are exporting certain methods of the `t` variable here rather than `t` itself. This is to establish a certain set of procedures that we will use idiomatically in our codebase.
 
+> [!WARNING]  
+> Due to typescript limitations, all imports toward the *trpc.ts* must be relative. You cannot use path aliases as defined within your *tsconfig.json*.
+> Otherwise, type inference in your client code won't work (the routers will resolve to `any`).
+
 ## Defining a router
 
 Next, let's define a router with a procedure to use in our application. We have now created an API "endpoint".


Nick-Lucas:
This really depends on your setup, typescript can absolutely bring in tRPC types by absolute import, so if you've had this experience there was likely a problem with your setup

Feel free to to come to our discord as someone may be able to help there

oleasteo:
@Nick-Lucas I've added an example with npm-workspaces, see #6114 

In there, you can reproduce the issue by replacing `from './trpc.js'` with `from '@/trpc.js'` in *server/user.router.ts*. This will result in `trpc.user: any` in the client.

So, if you don't happen to know a fix, we should definitely add such a warning (with the added context of npm workspaces) to the docs.

Nick-Lucas:
> @Nick-Lucas I've added an example with npm-workspaces, see #6114
> 
> In there, you can reproduce the issue by replacing `from './trpc.js'` with `from '@/trpc.js'` in _server/user.router.ts_. This will result in `trpc.user: any` in the client.
> 
> So, if you don't happen to know a fix, we should definitely add such a warning (with the added context of npm workspaces) to the docs.

It really is a typescript setup issue in the project, I pointed out one possible thing though I know you're also referring to a different file in this comment.

Absolute imports require proper configuration in both tsconfig and build tools. It's fiddly but there's no justification for this comment being added to the docs because it's at very least misleading. Some documentation on how to correctly set up typescript for various repo types would be far more helpful

oleasteo:
I don't accept "setup issue" until one shows me, how, exactly. As far as I'm concerned, this is reproducible simply once client and server ain't in the same package, with pretty straight-forward setup. Nothing complicated. Common use-case.

I've dug into the issues and found two other related ones: #3475 and #2928. Both, closed without any addressing the issue.
So, sorry if I have to insist on some kind of solution here ;-)

KATT:
@oleasteo 

- https://github.com/trpc/examples-next-prisma-starter/blob/97c3dfbf7d7902bc8394756a58ef916e7eda6d82/tsconfig.json#L22. 
- https://github.com/trpc/examples-next-prisma-starter/blob/97c3dfbf7d7902bc8394756a58ef916e7eda6d82/src/pages/post/%5Bid%5D.tsx#L7
- https://github.com/trpc/examples-next-prisma-starter/blob/97c3dfbf7d7902bc8394756a58ef916e7eda6d82/src/pages/post/%5Bid%5D.tsx#L35

oleasteo:
@KATT that's the client-side code.

https://github.com/trpc/examples-next-prisma-starter/blob/97c3dfbf7d7902bc8394756a58ef916e7eda6d82/src/server/routers/post.ts#L5
server-side there is a relative import. I'd expect that to cause the same issue if resolved via some path alias.

oleasteo:
Solved. I'll update #6114 shortly to provide a reference solution and will open a PR to update the FAQ/Troubleshooting.

PR #6114, opened 2024-10-13
chore(examples/minimal-react): type alias showcase

oleasteo:
Relates to #6112 

## 🎯 Changes

Added an example that uses [npm workspaces](https://docs.npmjs.com/cli/v10/using-npm/workspaces).

## ✅ Checklist


examples/npm-workspaces/client/src/index.ts
```diff
@@ -0,0 +1,57 @@
...
+import {
+  createTRPCClient,
+  splitLink,
+  unstable_httpBatchStreamLink,
+  unstable_httpSubscriptionLink,
+} from '@trpc/client';
+/**
+ * We only import the `AppRouter` type from the server - this is not available at runtime
+ */
+import type { AppRouter } from '@examples-npm-workspaces/server/src';
```
Nick-Lucas, line 13: If I understand right you're saying this is resolving 'any', but as best I can tell you haven't told typescript how to find this in the first place. You'll need a path to `@examples-npm-workspaces` in your tsconfig

oleasteo: It's resolved via the package.json `@examples-npm-workspaces/server` → npm workspaces.

examples/npm-workspaces/client/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "build": "tsc",
+    "dev": "wait-port 3000 && tsx watch src/index.ts",
+    "start": "node dist/index.js"
+  },
+  "dependencies": {
+    "@trpc/client": "npm:@trpc/client@next",
+    "@examples-npm-workspaces/server": "*"
```
juliusmarminge, line 12: For clarity this should be a dev dep cause we only need the type in dev. No runtime code is shared

examples/npm-workspaces/server/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "target": "esnext",
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
+    "esModuleInterop": true,
+    "strict": true,
+    "outDir": "dist",
+    "skipLibCheck": true,
+    "paths": {
+      "@/*": ["./src/*"]
+    }
```
juliusmarminge, line 12: Try adding a baseDir here and in client. Using path aliases without a base is a bit sketchy imo

oleasteo: Tried it, does not resolve the issue. I tried `baseUrl: "../"` as well (with adjusted path alias). Does not help either.

Also, https://www.typescriptlang.org/docs/handbook/modules/reference.html#baseurl

> baseUrl was designed for use with AMD module loaders. If you aren’t using an AMD module loader, you probably shouldn’t use baseUrl. Since TypeScript 4.1, baseUrl is no longer required to use [paths](https://www.typescriptlang.org/docs/handbook/modules/reference.html#paths) and should not be used just to set the directory paths values are resolved from.

So, it's actually recommended not so set baseUrl anymore.

juliusmarminge: TIL about that, but it doesn't change the fact that I've experienced weird behavior without it 😭

I'll take a look when I get home and see if I can get it working

oleasteo: Solved. See discord. I'll update the PR shortly.

juliusmarminge:
we already have an example that uses npm workspaces here: https://github.com/trpc/trpc/tree/next/examples/minimal-react

maybe making the necessary chagnes there instead?

oleasteo:
@juliusmarminge done.

juliusmarminge (changes requested):
Can you reset lockfile 

image

oleasteo:
@juliusmarminge I agree with using the `from 'examples-minimal-react-server'` instead of type alias. I've missed to port this when integrating into minimal-react.

I disagree with the other changes, though. They go against "Move Fast and Break Nothing."

Now, you don't get the updated `trpc` API on the client-side while using `npm run dev`, only after `npm run build`. I don't believe this reflects a common setup, honestly.

juliusmarminge:
Can fix that by running a watcher. Feel free to update to use live types if you rather have that though. Easy change to just point exports types condition directly at the source

oleasteo:
I am actually failing to "Easy change to just point exports types condition directly at the source".

If I try, I get 
```typescript
const trpc: "The property 'useContext' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'useUtils' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'Provider' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'createClient' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'useQueries' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'useSuspenseQueries' in your router collides with a built-in method, rename this router or procedure on your backend."
```
on the client. An error which seems misplaced and that does not help me in the slightest.

Sure, file watchers may be the best for big projects. But this is a **minimal** example, isn't it? How to set up file watchers would be different based on your build tooling. Where do we draw the line?

I would draw the line at a very *quickstart-esque* setup. It serves the needs to be able to start writing code that works and is type-safe. It should be able to provide a great DX in enabling this. But it does not need to serve the needs of a big project. You might want to add a FAQ entry for those performance issues instead?

juliusmarminge:
> I am actually failing to "Easy change to just point exports types condition directly at the source".
> 
> If I try, I get
> 
> ```ts
> const trpc: "The property 'useContext' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'useUtils' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'Provider' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'createClient' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'useQueries' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'useSuspenseQueries' in your router collides with a built-in method, rename this router or procedure on your backend."
> ```
> 
> on the client. An error which seems misplaced and that does not help me in the slightest.
> 
> Sure, file watchers may be the best for big projects. But this is a **minimal** example, isn't it? How to set up file watchers would be different based on your build tooling. Where do we draw the line?
> 
> I would draw the line at a very _quickstart-esque_ setup. It serves the needs to be able to start writing code that works and is type-safe. It should be able to provide a great DX in enabling this. But it does not need to serve the needs of a big project. You might want to add a FAQ entry for those performance issues instead?

Yea seems like the type got resolved to any 

CleanShot 2024-10-25 at 12 14 43@2x

I fixed it by using native node import aliases instead (supported by typescript ^5.4)

PR #6128, opened 2024-10-16
fix(sse): combine abort signal with internal one

### LINKED ISSUE #6126: bug: dangling async generator on internal cancel within `sseStreamProducer`

n/a

### Describe the bug

As discussed with @KATT :

sse.ts `sseStreamProducer(...)` has some cancel conditions on which to abort further handling (namely, `opts.emitAndEndImmediately` & `opts.maxDurationMs`).
When those trigger, the emit of further data is stopped, but the userland abort signal is not coupled with this. So, any potential userland event listener does not stop immediately, but only after the next `yield` (as we at least call `iterator.return()`).

### Link to reproduction

n/a

### To reproduce

n/a

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

## Funding

<a href="https://polar.sh/trpc/trpc/issues/6126">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/6126/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

oleasteo:
Closes #6126

## 🎯 Changes

Added `AbortController` to internal `subscription` processing to avoid dangling generators in userland.

Consider userland like this:

```ts 
procedure.subscription(async function* (opts) {
    // listen for new events
    for await (const [data] of on(ee, 'add', { signal: opts.signal })) {
      const post = data as Post;
      yield post;
    }
  })
```

Before the changes, this would have had a dangling event listener (for just the very next event) **after** `opts.emitAndEndImmediately` or `opts.maxDurationMs` triggered. With the changes, the `opts.signal` triggers with these cancel conditions, so the event listener will be cleaned up properly.

## ✅ Checklist

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/stream/utils/asyncIterable.ts
+++ b/packages/server/src/unstable-core-do-not-import/stream/utils/asyncIterable.ts
@@ -27,6 +27,7 @@ export async function* withCancel<T>(
 interface TakeWithGraceOptions {
   count: number;
   gracePeriodMs: number;
+  onCancel?: () => void;
 }
 
 /**
@@ -36,7 +37,7 @@ interface TakeWithGraceOptions {
  */
 export async function* takeWithGrace<T>(
   iterable: AsyncIterable<T>,
-  { count, gracePeriodMs }: TakeWithGraceOptions,
+  { count, gracePeriodMs, onCancel }: TakeWithGraceOptions,
 ): AsyncGenerator<T> {
   const iterator = iterable[Symbol.asyncIterator]();
   const timer = createPromiseTimer(gracePeriodMs);
@@ -53,7 +54,7 @@ export async function* takeWithGrace<T>(
       }
       yield result.value;
       if (--count === 0) {
-        timer.start();
+        timer.start().promise.then(onCancel, noop);
       }
     }
   } finally {
--- a/packages/server/src/unstable-core-do-not-import/utils.ts
+++ b/packages/server/src/unstable-core-do-not-import/utils.ts
@@ -87,3 +87,32 @@ export function assert(
 export function sleep(ms = 0): Promise<void> {
   return new Promise<void>((res) => setTimeout(res, ms));
 }
+
+/**
+ * Ponyfill for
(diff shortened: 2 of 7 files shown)


Thompson1985 (approved):

packages/server/src/unstable-core-do-not-import/utils.ts
```diff
@@ -68,3 +68,51 @@ export function noop(): void {}
...
+  }
+
+  const ac = new AbortController();
+
+  for (const signal of signals) {
+    if (signal.aborted) {
+      trigger();
+    } else if (!ac.signal.aborted) {
+      signal.addEventListener('abort', trigger, { once: true });
+    }
```
KATT, line 107: Simplified

```suggestion
    if (signal.aborted) {
      trigger();
      break;
    }
    signal.addEventListener('abort', trigger, { once: true });
```

packages/server/src/unstable-core-do-not-import/stream/utils/asyncIterable.ts
```diff
@@ -27,6 +27,7 @@ export async function* withCancel<T>(
 interface TakeWithGraceOptions {
   count: number;
   gracePeriodMs: number;
+  onCancel?: () => void;
```
KATT, line 30: Can't this be an abort signal instead of a callback?

Also, I think it's only used in one place, no need to make it optional

KATT (approved):

packages/server/src/unstable-core-do-not-import/stream/utils/asyncIterable.ts
oleasteo: It could be an abort ctrl passed in. But I like to not make util functions too specific. If there's a simpler or more generic way, without adding pretty much any effort, it's best to keep util functions generic or simple. In this case, I'd argue it's both more simple and generic. Lean function headers are great.

packages/server/src/unstable-core-do-not-import/utils.ts
oleasteo: oops. I misinterpreted the github mobile view and thought this was already edited into the PR :see_no_evil:

PR #6131, opened 2024-10-17
fix: update fastify dependency

jonapgar-groupby:
## 🎯 Changes

Update Fastify dependency to support passing Response object directly to Reply.send.
Earlier versions of fastify do not support passing a Response object directly. The capability was introduced here and is first available on version 4.26.0

https://github.com/fastify/fastify/commit/101ba57f894281421d902bf509adb4270473799d

Without this change, users who upgrade to v11 of tRPC who are using a prior version of fastify will receive an empty object `{}` for every response. No errors are emitted from the server, but tRPC clients will reject the response as invalid

 
## ✅ Checklist

### DIFF
--- a/examples/fastify-server/package.json
+++ b/examples/fastify-server/package.json
@@ -19,7 +19,7 @@
     "@fastify/websocket": "^10.0.1",
     "@trpc/client": "npm:@trpc/client@next",
     "@trpc/server": "npm:@trpc/server@next",
-    "fastify": "^4.16.0",
+    "fastify": "^4.26.0",
     "superjson": "^1.12.4",
     "tslib": "^2.5.0",
     "ws": "^8.0.0",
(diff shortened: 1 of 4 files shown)


KATT (changes requested):
Please update the test/support for the new fastify version

jonapgar-groupby:
Ah thanks I missed that 4.26.2 was coming from in the lockfile for examples/fastify-server. This caused the discrepancy with the referenced types. 

Should be all good now

KATT (commented):
I don't have any problem updating this, but it doesn't matter / "do" anything afaict?

We just have fastify as a devDep so you can install whatever fastify version you want in your apps

KATT:
> Probably the best solution would be to declare fastify as a peer dep and mark it optional via[ peerDependenciesMeta](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#peerdependenciesmeta). The same could be done for `express` and several other dependencies currently declared as dev dependencies.

This would be great!

PR #6144, opened 2024-10-20
fix(server/util): dangling event listener in `AbortSignal.any` ponyfill

oleasteo:
Simple improvement as proposed by @KATT https://github.com/trpc/trpc/pull/6128#pullrequestreview-2379887271

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/utils.ts
+++ b/packages/server/src/unstable-core-do-not-import/utils.ts
@@ -102,9 +102,9 @@ export function abortSignalsAnyPonyfill(signals: AbortSignal[]): AbortSignal {
   for (const signal of signals) {
     if (signal.aborted) {
       trigger();
-    } else if (!ac.signal.aborted) {
-      signal.addEventListener('abort', trigger, { once: true });
+      break;
     }
+    signal.addEventListener('abort', trigger, { once: true });
   }
 
   return ac.signal;


packages/server/src/unstable-core-do-not-import/utils.ts
```diff
@@ -102,9 +102,9 @@ export function abortSignalsAnyPonyfill(signals: AbortSignal[]): AbortSignal {
   for (const signal of signals) {
     if (signal.aborted) {
       trigger();
-    } else if (!ac.signal.aborted) {
-      signal.addEventListener('abort', trigger, { once: true });
+      break;
```
juliusmarminge, line 105:
```suggestion
      continue;
```
?

oleasteo: No, `break` is correct. The ac cannot be aborted more than once. No need for further listeners.

juliusmarminge: But they're different signals?

Break here will exit the loop so the behavior won't be the same as before

oleasteo: Yeah, in theory, this would've had dangling event listeners on the trailing signals beforehand.
In practice, there is only one place this util function is in use. At that usage, there are only two signals passed.

Yes, for a very rare situation, a very minor bug has been fixed.

oleasteo: Changed the commit msg in squash & PR title

juliusmarminge: Ok 👍

I didn't review the original PR so just saw the title and this caught my eyes. I trust you have more context on what's happening here

KATT (approved):

PR #6164, opened 2024-10-29
chore(deps): modernize deps & tooling

juliusmarminge:
Got a bit bigger than i first intended

- pnpm 9
- eslint 9
- eslint flat config
- prettier v3
- next 15


examples/.experimental/next-app-dir/test/server-http.test.ts
juliusmarminge, line 1: honestly think we might just wanna deprecate all of this stuff now that next is revamping all their caching but for now the `server-invoker` still works in 15 so we can keep it in for now

juliusmarminge:
Not sure if the railway failures are real or not.. never understood when those fail/pass

KATT (approved):
huge-upgrade-trent-arant

PR #6177, opened 2024-11-01
chore: isolate db for each pr of sse example

KATT:
This was all stupid, Railway creates a new db for every deploy, I'm silly


examples/next-sse-chat/src/server/railwayPrNumber.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * @example 'trpc-pr-5821'
+ */
+const railwayEnv = process.env.RAILWAY_ENVIRONMENT_NAME;
```
KATT, line 4: RAILWAY_ENVIRONMENT_NAME doesn't seem available here

cc @juliusmarminge can you spot what I'm doing wrong in the Dockerfile or whatever?  

dockerfile logs it but it doesn't reach the build smh

examples/next-sse-chat/src/server/db/schema.ts
```diff
@@ -1,8 +1,15 @@
...
+import { railwayPrNumber } from '../railwayPrNumber';
 
-const pgTable = pgTableCreator((name) => `sse-chat_${name}`);
+const prPrefix = railwayPrNumber ? `__pr-${railwayPrNumber}_` : '';
+
+if (prPrefix) {
+  console.log('pull request prefix', prPrefix);
+}
+
+const pgTable = pgTableCreator((name) => `${prPrefix}sse-chat_${name}`);
```
KATT, line 12: idea is to add a prefix for each PR environment so they're isolated

examples/next-sse-chat/src/server/railwayPrNumber.ts
juliusmarminge: You gotta add it to turbo.json to inject it into the script

KATT: awesomeee thanks

examples/.railway/next-sse-chat/Dockerfile
```diff
@@ -5,6 +5,8 @@ RUN corepack enable
 
 ARG DATABASE_URL
 ENV DATABASE_URL=$DATABASE_URL
+ARG RAILWAY_ENVIRONMENT_NAME
+ENV RAILWAY_ENVIRONMENT_NAME=$RAILWAY_ENVIRONMENT_NAME
```
juliusmarminge, line 9: is this needed? we need the database_url here cause of it running migrations but i thought railway injected their stuff automatically?

examples/next-sse-chat/src/server/db/schema.ts
juliusmarminge: note that ayou need to add this to drizzle config too under `schemaFilter` to make sure migration script also filters to use the prefix

KATT:
This was all stupid, Railway creates a new db for every deploy

PR #6206, opened 2024-11-08
chore(server): refactor some stream-stuff

KATT:
- `withPing()` made next event get swallowed
- Refactor jsonl so we use the term "encode"/"decode" rather than "hydrate"/"dehydrate"
- ~~Also refactored stuff to use explicit resource management~~ I tried, but build doesn't let it pass currently
- Create a `writeResponseToNodeHTTPResponse` helper 
- Refactor JSONL so we don't need deferreds

PR #6222, opened 2024-11-11
chore: add `using`

KATT:
Fork off #6206 that adds `using`-keyword


scripts/tsconfigs.ts
KATT, line 1: Basically this scripts creates a standard tsconfig in each package, that extends its `./tsconfig.build.json` where you can e.g. enable react etc

scripts/tsconfigs.ts
```diff
@@ -0,0 +1,36 @@
...
+  absolute: true,
+});
+
+// Process each package
+for (const packagePath of packagePaths) {
+  const packageDir = path.dirname(packagePath);
+  // Skip the tests package
+  if (path.basename(packageDir) === 'tests') {
+    continue;
+  }
```
oleasteo, line 27: Maybe also skip if `tsconfig.json` already exists?

scripts/tsconfigs.ts
oleasteo: Why is this script added to the repo in the first place? Looks like more of a one-shot to me.

oleasteo (approved):

Issue #6193, opened 2024-11-06
bug: SSE Subscription Never Aborts


yspreen:
### Provide environment information

  System:
    OS: Linux 5.0 undefined
    CPU: (6) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.20.3 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @tanstack/react-query: ^5.59.17 => 5.59.20 
    @trpc/client: ^11.0.0-rc.608 => 11.0.0-rc.608+f75de97b3 
    @trpc/next: ^11.0.0-rc.608 => 11.0.0-rc.608+f75de97b3 
    @trpc/react-query: ^11.0.0-rc.608 => 11.0.0-rc.608+f75de97b3 
    @trpc/server: ^11.0.0-rc.608 => 11.0.0-rc.608+f75de97b3 
    next: 15.0.2 => 15.0.2 
    react: ^19.0.0-rc-7c8e5e7a-20241101 => 19.0.0-rc-fb9a90fa48-20240614 
    typescript: ^5 => 5.6.3 

### Describe the bug

Subscription never gets aborted

### Link to reproduction

https://stackblitz.com/~/github.com/yspreen/trpc-bug

### To reproduce

https://stackblitz.com/~/github.com/yspreen/trpc-bug

### 👨‍👧‍👦 Contributing

- [X] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

## Funding
* You can sponsor this specific effort via a [Polar.sh](https://polar.sh) pledge below
* We receive the pledge once the issue is completed & verified

<a href="https://polar.sh/trpc/trpc/issues/6193">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/6193/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/trpc/trpc/issues/6193/pledge.svg">
</picture>
</a>

yspreen:
here's the code:
```

export const appRouter = router({
  loopBased: procedure.subscription(async function* ({ signal }) {
    try {
[...]

PR #6214, opened 2024-11-10
feat(client): handle timeout in SSE link

KATT:
Closes #

## 🎯 Changes

Adds ability to timeout an SSE connection without losing tracked events.

From testing, browsers seem really bad at keeping track of a server that e.g. gets killed.

Should I add a default maybe? 

Thinking something like default timeout 30s, default ping every 15s

PR #6228, opened 2024-11-13
Abort SSE when ReadableStream is cancelled

tarekwiz:
## 🎯 Changes
Added AbortController to the createReadableStream function, allowing it to abort once the request is cancelled.
This is to fix an issue with SSE where the server would never send an abort signal when the client closes the connection.

https://github.com/trpc/trpc/issues/6193

## ✅ Checklist

- [*] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [*] If necessary, I have added documentation related to the changes made.
- [*] I have added or updated the tests related to the changes made.


paradoxloop:
@tarekwiz is this meant to solve the problem of a client not being able to reconnect to a stream if it (the client) cancelled the stream (e.g reloading page)? An issue that resulting on this kind error on front end:

`Uncaught TypeError: ReadableStreamDefaultController.close: Cannot close a stream that has errored.`

We applied your code changes as a patch and it didn't solve this issue.

tarekwiz:
> @tarekwiz is this meant to solve the problem of a client not being able to reconnect to a stream if it (the client) cancelled the stream (e.g reloading page)? An issue that resulting on this kind error on front end:
> 
> `Uncaught TypeError: ReadableStreamDefaultController.close: Cannot close a stream that has errored.`
> 
> We applied your code changes as a patch and it didn't solve this issue.

No this is not meant to fix that issue. This is meant to fix an issue on the server where the server would not call abort on the abortcontroller when the connection is dropped

KATT:
> @tarekwiz is this meant to solve the problem of a client not being able to reconnect to a stream if it (the client) cancelled the stream (e.g reloading page)? An issue that resulting on this kind error on front end:
> 
> `Uncaught TypeError: ReadableStreamDefaultController.close: Cannot close a stream that has errored.`
> 
> We applied your code changes as a patch and it didn't solve this issue.

@paradoxloop I've not experienced this - could you please try the latest release and see? We made some changes to client's stream as part of #6214. If not, please create a new issue

paradoxloop:
thanks @KATT just tried to latest release "633" and this does not make the issue go away.

Will try to create a MVR early next week however this work is attempt to move away from websockets for a particular tool and there seems to be lots of footguns and the tooling/documentation around SSE is limited (more generally not specifically TRPC) so might have to pause it.

tarekwiz:
@KATT #6193 Was closed but this PR that fixes it was never merged.

KATT:
> @KATT #6193 Was closed but this PR that fixes it was never merged.

@tarekwiz see https://github.com/trpc/trpc/issues/6193#issuecomment-2475035156

> > I don't think #6228 changes anything - what it aims to fix should have been a symptom of us not triggering the abort signal correctly, but the stream itself should be canceled as long as you listen to the abort signals [as defined in the docs](https://trpc.io/docs/server/subscriptions#basic-example).

KATT:
> Nope the abort controller is never called in the latest next release. This fix is needed. You can reproduce it by using Astro’s fetch implementation

Please give us better steps to reproduce it or supply a reproduction test - we can't merge this as-is without tests or anything as we then can't guarantee we won't break it in the future

KATT:
In the last PR we made, we deleted the signal on the `sseStreamProducer` as we abort the stream when we write the response:

https://github.com/trpc/trpc/blob/09a76869e7b1c0b56db4b5bc20ef7cde3015cbdb/packages/server/src/adapters/node-http/writeResponse.ts#L48

So there's no signal needed on the readable stream itself.

We trigger the abort signal like this:

https://github.com/trpc/trpc/blob/09a76869e7b1c0b56db4b5bc20ef7cde3015cbdb/packages/server/src/adapters/node-http/incomingMessageToRequest.ts#L139-L140

This should be sufficient. If it isn't in your environment, please add logs to all possible close-events akin to the above and let us know which trigger

Docs in Node.js:

- https://nodejs.org/docs/latest-v22.x/api/http.html#event-close
- https://nodejs.org/docs/latest-v22.x/api/net.html#event-end

Again, we need easier steps to reproduce than _"You can reproduce it by using Astro’s fetch implementation"_ - make a reproduction repro and tell us how to reproduce it in your repo 🙏

tarekwiz:
There you go https://github.com/tarekwiz/trpc-sse-bug
If you open the page, then close the browser, the loop never finishes and abort controller is never aborted.

KATT:
Abort signal is not triggered, no, but the stream ends as expected:

https://github.com/tarekwiz/trpc-sse-bug/pull/1

tarekwiz:
Ok let's fix the abort controller not being aborted

KATT:
> Ok let's fix the abort controller not being aborted

I'm honestly unsure how to do this nicely - the signal is there to handle _aborts_ and it does that reliably now.

We'd have to add a new abort controller inside of `resolveResponse` that was able to sniff when any `new Response()`'s body stream eventually ends.

Your PR added this for one of our possibly `Response`s but we'd need to do it for all of them

We could do some sort of wrapper of response in `resolveResponse` like

```ts
function createResponse(
  body: BodyInit | null,
  init: ResponseInit,
  onEndController: AbortController,
) {
  const response = new Response(body, init);

  // Somehow sniff when `response.body`'s stream has ended - unsure how?

  return response;
}
```


Unsure how to do it. Ideas?

Also unsure what you need this for?

tarekwiz:
So basically the current situation is that the function inside the subscription would only reach the "finally" or error state when the server tries to send a message but the connection is closed. We need the abort controller to be able to save up on server resources so that once the request is closed, we can cleanup the resources used by that user. Otherwise we would have to wait for the server to send information to the client. Maybe adding ping would solve that but I haven't tested it.

tarekwiz:
By the way, when sse.ping is enabled, the finally block is never reached. This took me 2 hours of debugging to find @KATT

tarekwiz:
Also, how does this look to you

```js
function createResponse(
  body: BodyInit | null,
  init: ResponseInit,
  onEndController: AbortController,
) {
  // check if body is a stream
  if (body instanceof ReadableStream) {
   // override the cancel function of the ReadableStream to cancel the abort controller.
    body.cancel = async () => {
      onEndController.abort();
    };
  }

  return new Response(body, init);
}```

KATT:
> Also, how does this look to you
> 
> ```js
> function createResponse(
>   body: BodyInit | null,
>   init: ResponseInit,
>   onEndController: AbortController,
> ) {
>   // check if body is a stream
>   if (body instanceof ReadableStream) {
>    // override the cancel function of the ReadableStream to cancel the abort controller.
>     body.cancel = async () => {
>       onEndController.abort();
>     };
>   }
> 
>   return new Response(body, init);
> }```
> ```

Feels wrong but maybe it could work... the signal would still wouldn't be triggered if the body just ended "normally"

tarekwiz:
What is a "normal" way of closing a readable stream?

KATT:
> What is a "normal" way of closing a readable stream?

When the stream is done but wasn't actually cancelled - 040e633810a982f79d89217ca413af2e3ef3d79b

tarekwiz:
Hmmm not sure tbh..

KATT:
This could maybe work: https://github.com/trpc/trpc/blob/16f40e59ab90536e986115b7620755aa1c163555/packages/server/src/unstable-core-do-not-import/stream/utils/readableStreamFrom.test.ts#L52-L72

tarekwiz:
Yeah that looks solid.

KATT:
There's more to it most likely, but feel to do a PR using this code, I don't know when I'll have more time for this

PR #6234, opened 2024-11-14
Polymorphism types and tests for Query/Mutation Options Client, more Utility Factory methods

Nick-Lucas:
* Adds polymorphism tests where a shared router type can be used to create shared components
* Adds queryKey and queryFilter factories with tests
* Adds InferInput and InferOutput helpers which can be used against Query & Mutation Options factories to get their types

image

## 🎯 Changes

## ✅ Checklist


packages/tests/server/react/polymorphism.legacy.factory.tsx
```diff
@@ -0,0 +1,106 @@
+//
```
Nick-Lucas, line 1: Ignore these `.legacy` files, they're just straight renames of the old tests so we can review the diff of the new ones

KATT (approved):

packages/tests/server/react/polymorphism.test.tsx
```diff
@@ -263,85 +319,84 @@ describe('polymorphism', () => {
...
-      await userEvent.click($.getByTestId('refreshBtn'));
-
-      await waitFor(() => {
-        expect($.container).toHaveTextContent(
-          'Last Export: `Search for Polymorphism React` (Ready!)',
-        );
-      });
-    });
-  });
+  // describe('sub-typed factory', () => {
```
Nick-Lucas, line 322: I... don't think subtyping is going to be possible without essentially rewriting the types deeply to remove all callbacks which reference the input or output types

Woulds love to be proven wrong, but structural typing is known not to function when there's an inner callback and the strucutral-ness ends up going the wrong direction and breaking 😔

image

Nick-Lucas: I'm not sure the exact cause in this case but I've seen this plenty of times and it sucks

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
```diff
@@ -33,10 +33,30 @@ import {
   trpcSubscriptionOptions,
   type TRPCSubscriptionOptions,
 } from './subscriptionOptions';
-import type { QueryType, ResolverDef } from './types';
+import type { QueryType, ResolverDef, TRPCQueryKey } from './types';
 import { getQueryKeyInternal } from './utils';
 
-export interface DecorateQueryProcedure<TDef extends ResolverDef> {
+interface QueryKeyStuff {
```
juliusmarminge, line 39: `DecorateRouter` ?

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
```diff
@@ -33,10 +33,30 @@ import {
...
+  /**
+   * Simple utility to invalidate a query or router
+   *
+   * @see https://tanstack.com/query/latest/docs/framework/react/guides/query-invalidation
+   */
+  invalidate: () => Promise<void>;
+}
+
+export interface DecorateQueryProcedure<TDef extends ResolverDef>
+  extends Omit<QueryKeyStuff, 'getQueryKey'> {
```
juliusmarminge, line 56: `invalidate` here will need to accept input right? so this extend isn't totally accurate.

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
```diff
@@ -46,9 +66,19 @@ export interface DecorateQueryProcedure<TDef extends ResolverDef> {
...
+   * Useful for query invalidation
+   *
+   * @see https://tanstack.com/query/latest/docs/framework/react/guides/query-invalidation
+   */
+  getQueryKey: (input?: TDef['input']) => TRPCQueryKey;
 }
 
 export interface DecorateMutationProcedure<TDef extends ResolverDef> {
+  _input: TDef['input'];
+  _output: TDef['output'];
```
juliusmarminge, line 80: how are these shown in IDE? are they shown before or after the "real" stuff?

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
```diff
@@ -172,32 +205,40 @@ export function createTRPCOptionsProxy<TRouter extends AnyRouter>(
...
-    const queryKey = getQueryKeyInternal(path, input, queryType);
+    const queryKey = getQueryKeyInternal(path, arg1, queryType ?? 'any');
 
     const contextMap: Record<UtilsMethods, () => unknown> = {
+      _input: null as any,
+      _output: null as any,
+      getQueryKey: () => queryKey,
+      invalidate: async () => {
+        await opts.queryClient.invalidateQueries({
+          queryKey: queryKey,
```
juliusmarminge, line 219: again, queryKey here contains input but that's not what you have in the type above

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
```diff
@@ -172,32 +205,40 @@ export function createTRPCOptionsProxy<TRouter extends AnyRouter>(
...
+        });
+      },
       infiniteQueryOptions: () =>
         trpcInfiniteQueryOptions({
-          opts: userOptions,
+          opts: arg2,
           path,
           queryClient: opts.queryClient,
-          queryKey,
+          queryKey: queryKey as any as TRPCQueryKey,
```
juliusmarminge, line 227: why are these type casts needed?

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
```diff
@@ -46,9 +66,19 @@ export interface DecorateQueryProcedure<TDef extends ResolverDef> {
    * @see https://tanstack.com/query/latest/docs/framework/react/reference/infiniteQueryOptions#infinitequeryoptions
    */
   infiniteQueryOptions: TRPCInfiniteQueryOptions<TDef>;
+
+  /**
+   * Useful for query invalidation
+   *
+   * @see https://tanstack.com/query/latest/docs/framework/react/guides/query-invalidation
+   */
+  getQueryKey: (input?: TDef['input']) => TRPCQueryKey;
```
Nick-Lucas, line 75: ^ @juliusmarminge

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
Nick-Lucas: Before which isn't preferable. We probably don't want these here at all though, I just added them for easy access while experimenting. There will be a better solution I'm sure

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
Nick-Lucas: No longer needed, good point.

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
Nick-Lucas: I'm not that happy with this solution by the way, it's just we need the methods in 3 places so I was trying to dry it out

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
juliusmarminge: yea but the Omit just leaves `invalite` in there whose type is wrong. 

Is this a future-proofing or else you could just inline both since they'll both be separate from the base type you're extending

Nick-Lucas: Oh sorry you're right, yeah I need to do the same thing here as I did for getQueryKey

Probably just need inlining, but would like to figure out the right API before going too hard here

packages/tests/server/react/polymorphism.test.tsx
```diff
@@ -202,21 +259,33 @@ describe('polymorphism', () => {
...
+          //   trpc.queryClient.invalidateQueries({
+          //     queryKey: trpc.github.getQueryKey(),
+          //   }),
+
+          // Option B: QueryFilter factory, re-usable in multiple places
+          // mutationFn: () =>
+          //   trpc.queryClient.invalidateQueries(trpc.github.getQueryFilter()),
+
+          // // Option C: invalidate method, simple, but cancellation, refetches etc would need their own
+          mutationFn: () => trpc.github.invalidate(),
```
Nick-Lucas, line 282: A few examples here of possible APIs

```tsx
const invalidate = useMutation({
  // Option A: fully explicit
  mutationFn: () =>
    trpc.queryClient.invalidateQueries({
      queryKey: trpc.github.getQueryKey(),
    }),

  // Option B: QueryFilter factory, re-usable in multiple places
  mutationFn: () =>
    trpc.queryClient.invalidateQueries(trpc.github.getQueryFilter()),

  // // Option C: invalidate method, simple, but cancellation, refetches etc would need their own
  mutationFn: () => trpc.github.invalidate(),
});
```

Option B is probably the right one, it's actually quite re-usable which is interesting, as invalidate isn't trhe only thing to take QueryFilters, so keeps our API surface down

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
```diff
@@ -46,9 +85,31 @@ export interface DecorateQueryProcedure<TDef extends ResolverDef> {
...
+  /**
+   * @internal prefer using InferInput to access input type
+   */
+  '~input': TDef['input'];
+
+  /**
+   * @internal prefer using InferOutput to access input type
+   */
+  '~output': TDef['output'];
+
```
Nick-Lucas, line 112: Putting the type on the object makes for a very cheap way of accessing type later, but providing some Infer helpers instead suggesting users utilise this directly. Prefixing with ~ forces autocomplete to sort it to the end

packages/tests/server/react/polymorphism.test.tsx
```diff
@@ -262,86 +245,6 @@ describe('polymorphism', () => {
       });
     });
   });
-
```
Nick-Lucas, line 265: It does not look like this kind of subtyping (the input/output being subtyped) can be possible in the new world, but that's probably okay because it's a much more complex style of polymorphism that probably warrants WET over DRY to reduce userland complexity

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
```diff
@@ -33,10 +33,48 @@ import {
...
+   * @see https://tanstack.com/query/latest/docs/framework/react/guides/query-keys
+   */
+  queryKey: () => TRPCQueryKey;
+
+  /**
+   * Calculate a Tanstack Query Filter for a Route
+   *
+   * @see https://tanstack.com/query/latest/docs/framework/react/guides/filters
+   */
+  queryFilter: () => QueryFilters;
```
juliusmarminge, line 52: I'm still not convinced this is a very useful method, especially since it doesn't take any of the other query filter args (predicate functions etc). why not just use `{ queryKey: trpc.foo.queryKey() }`

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
```diff
@@ -46,9 +84,33 @@ export interface DecorateQueryProcedure<TDef extends ResolverDef> {
...
+
+  /**
+   * Calculate a Tanstack Query Filter for a Query Procedure
+   *
+   * @see https://tanstack.com/query/latest/docs/framework/react/guides/filters
+   */
+  queryFilter: (input?: TDef['input']) => QueryFilters;
 }
 
 export interface DecorateMutationProcedure<TDef extends ResolverDef> {
```
juliusmarminge, line 103: should this have `.mutationKey()` ?

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
```diff
@@ -172,23 +237,31 @@ export function createTRPCOptionsProxy<TRouter extends AnyRouter>(
...
     const contextMap: Record<UtilsMethods, () => unknown> = {
+      '~input': undefined as any,
+      '~output': undefined as any,
+      queryKey: () => queryKey,
+      queryFilter: (): QueryFilters => {
+        return {
+          queryKey: queryKey,
+        };
+      },
       infiniteQueryOptions: () =>
```
juliusmarminge, line 254: for consistency with others :)

```suggestion
      infiniteQueryOptions: () => {
        return trpcInfiniteQueryOptions({
          opts: arg2,
          path,
          queryClient: opts.queryClient,
          queryKey: queryKey,
          query: callIt('query'),
        });
      },
```

packages/tests/server/react/polymorphism.test.tsx
juliusmarminge, line 1: should this be local to the new package? i can see the argument for having it here (keeping it here for colocatino with old) but i also prefer when the test is close to the source

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
Nick-Lucas: I think the idea would be that we have full control, so if we do want to set some of the other filters internally we can without breaking anybody's apps

I'm open to not having it, just felt like it's appropriate for a React Query factory to provide

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
Nick-Lucas: I think that's a different user story to queryKey really, I'm not very familiar with mutationKey so wouldn't want to rush into adding that without knowing the story, but it seems to be more about remotely executing a mutation which wouldn't make a lot of sense with tRPC

juliusmarminge: It's needed for `queryClient.isMutating()` for example, a utility we currently have

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
Nick-Lucas: Could also add options to override some bits, but I didn't want to dive straight into that world, worth analysing some use cases

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
Nick-Lucas: Got it, are you okay if we implement that as additional work into your branch? The scope of this PR already expanded a lot and it was all purely in service of polymorphism support. mutationKey isn't needed for that case

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
juliusmarminge: I'd say until we have identified these use cases and added support for the other filter options just the queryKey suffices? Not a hill I'd die on so if you still see value in it go ahead and merge 👍

juliusmarminge (approved):

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
Nick-Lucas: Wait, what's the correct branch? I thought I had it from your PR?

juliusmarminge: This is Alex's R19 branch which I think isn't working in some areas 🤔

Nick-Lucas: Crap. Well that threw me for a loop and into git hell: https://github.com/trpc/trpc/pull/6244

Nick-Lucas: Closing this in favour of the other MR

PR #6239, opened 2024-11-15
fix(server): `unstable_httpBatchStreamLink` hiccups

KATT:
Closes #

## 🎯 Changes

Introduced bugs in some edge cases in #6206 that were show-cased when reducing timeout for iterations in the tests

## Summary

- Reintroduce deferreds in jsonl consumer
- Add `using` a bit more in tests, for fun

## Notes

Hide whitespace when reviewing

CleanShot 2024-11-15 at 12 54 12@2x

## ✅ Checklist

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/stream/jsonl.ts
+++ b/packages/server/src/unstable-core-do-not-import/stream/jsonl.ts
@@ -425,6 +425,10 @@ export async function jsonlStreamConsumer<THead>(opts: {
 
   type ControllerChunk = ChunkData | StreamInterruptedError;
   type ChunkController = ReadableStreamDefaultController<ControllerChunk>;
+  /**
+   * This is needed as new values can come in before the controller has read the chunk
+   * Not pretty, could likely be refactored and omitted somehow
+   */
   const chunkDeferred = new Map<ChunkIndex, Deferred<ChunkController>>();
 
   const controllers = new Map<ChunkIndex, ChunkController>();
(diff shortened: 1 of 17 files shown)


packages/server/src/unstable-core-do-not-import/stream/jsonl.ts
KATT, line 1: cc @oleasteo 

I think we could refactor this whole file to leverage async iterables for the production of the stream.... and some in the consumer now

i've made a bit of spaghetti right now 😳

packages/server/src/unstable-core-do-not-import/stream/sse.ts
```diff
@@ -32,7 +32,7 @@ export interface PingOptions {
 export interface SSEStreamProducerOptions<TValue = unknown> {
   serialize?: Serialize;
   data: AsyncIterable<TValue>;
-  abortCtrl: AbortController;
```
KATT, line 35: This shouldn't be needed anymore

We have this: https://github.com/trpc/trpc/blob/8b37fed52bc53acea6df9bb0896a8cdc89f8e39c/packages/server/src/adapters/node-http/writeResponse.ts#L47-L49

packages/server/src/unstable-core-do-not-import/stream/utils/asyncIterable.ts
```diff
@@ -81,8 +77,6 @@ export async function* takeWithGrace<T>(
       yield result.value;
       if (--count === 0) {
         timerPromise = timer.start();
-        // eslint-disable-next-line @typescript-eslint/no-floating-promises
-        timerPromise.then(() => opts.abortCtrl.abort());
```
KATT, line 85: Shouldn't be needed as ending the stream will cause the abort controller to be triggered

PR #6259, opened 2024-11-22
chore(server): stream improvements

KATT:
## 🎯 Changes

- Adds abortsignal to `observableToAsyncIterable`
- Gets rid of `createReadableStream` from `sse.ts`
- Use async iterators when producing jsonl stream
- Refactor jsonl consumption to have a distinct manager for streams for different parts of the streams

## ✅ Checklist


packages/server/src/unstable-core-do-not-import/stream/jsonl.ts
```diff
@@ -266,20 +277,62 @@ function createBatchStreamProducer(opts: ProducerOptions) {
...
+    >();
+
+    // Continue processing while there are still active iterators
+    while (iterators.size > 0) {
+      // Ensure we have a pending promise for each iterator's next value
+      for (const [idx, it] of iterators.entries()) {
+        if (!nextPromises.has(idx)) {
+          nextPromises.set(idx, it.next());
+        }
+      }
```
KATT, line 298: this could probably be done a bit more optimally

it's a bit of an unecessary loop in a loop

feel free to have a play @oleasteo :)

packages/server/src/unstable-core-do-not-import/stream/jsonl.ts
```diff
@@ -117,101 +112,87 @@ class MaxDepthError extends Error {
...
   let counter = 0 as ChunkIndex;
   const placeholder = 0 as PlaceholderValue;
 
-  const stream = createReadableStream<ChunkData>();
-  const pending = withRefCount(new Set<ChunkIndex>(), () => {
-    if (!stream.cancelled()) {
-      stream.controller.close();
-    }
-  });
+  const set = new Set<{
```
oleasteo, line 122: Rename `const set` to e.g. `const queue`, as we do not use any features that `Set` is known for ⇒ confusing to read. Also, add a comment on why we use `Set` (simple and fast `delete` API)

packages/server/src/unstable-core-do-not-import/stream/jsonl.ts
```diff
@@ -266,20 +247,42 @@ function createBatchStreamProducer(opts: ProducerOptions) {
...
+        entry.nextPromise = entry.iterator.next();
+        set.add(entry);
+      }
+    }
+  } finally {
+    // Properly clean up any remaining iterators by calling return()
+    // Ensures resources are released if the loop exits early (e.g. due to error)
+    await Promise.all(Array.from(set).map((it) => it.iterator.return?.()));
+    set.clear();
+  }
```
oleasteo, line 275: Iterating over sets is a slow operation. We use it frequently here. Since we don't add new entries after the initial `encode`s, a RingBuffer structure would be optimal. This should improve readability as we get rid of the false pretense of using sets while also improving performance.

This is too much to implement as a review feedback, but keep in mind as a possible refactor.

packages/server/src/unstable-core-do-not-import/stream/jsonl.ts
```diff
@@ -382,18 +396,129 @@ function createConsumerStream<THead>(
...
+  /**
+   * Gets a reader for consuming the stream's data
+   */
+  getReader: () => ReadableStreamDefaultReader<ControllerChunk>;
+}
+
+/**
+ * Creates a handler for managing stream controllers and their lifecycle
+ */
+function createStreamManager(abortController: AbortController) {
```
oleasteo, line 453: Plural "Streams" `createStreamsManager`

oleasteo (changes requested):
Overall, I like the changes. But it's difficult to review such big PRs. Consider using smaller PRs next time.

One change that I'm not particularly fond of is the "Fix so that the jsonl output is actual JSON". Why is this done in the first place? It makes the `jsonl.ts` not produce `jsonl` anymore xD

packages/server/src/unstable-core-do-not-import/stream/jsonl.ts
KATT: Never used a ring buffer - wanna show how in a fork of this fork?

packages/server/src/unstable-core-do-not-import/stream/jsonl.ts
```diff
@@ -25,17 +25,16 @@ type CHUNK_VALUE_TYPE_PROMISE = typeof CHUNK_VALUE_TYPE_PROMISE;
...
+const PROMISE_STATUS_FULFILLED = 1;
 type PROMISE_STATUS_FULFILLED = typeof PROMISE_STATUS_FULFILLED;
-const PROMISE_STATUS_REJECTED = 1;
-type PROMISE_STATUS_REJECTED = typeof PROMISE_STATUS_REJECTED;
 
-const ASYNC_ITERABLE_STATUS_RETURN = 0;
+const ASYNC_ITERABLE_STATUS_YIELD = 1;
+type ASYNC_ITERABLE_STATUS_YIELD = typeof ASYNC_ITERABLE_STATUS_YIELD;
+const ASYNC_ITERABLE_STATUS_RETURN = 2;
 type ASYNC_ITERABLE_STATUS_RETURN = typeof ASYNC_ITERABLE_STATUS_RETURN;
```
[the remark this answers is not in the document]
KATT: I don't like enums in TypeScript

https://www.totaltypescript.com/why-i-dont-like-typescript-enums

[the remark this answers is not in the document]
oleasteo: I agree that numeric enums behave silly for the runtime. That's why I usually only use string enums.

But having multiple `const`s like you do here makes things unnecessarily bad to read. So long as you don't mess with `Object.entries(numericEnum)` and similar, you are perfectly fine using them.

If you don't want to use string enums in this case, we could compromise on `as const` objects instead?

```ts
const VALUE_STATE_ENUM = {
  ERROR: 0,
  RETURN: 1,
  YIELD: 2,
} as const;
```

should do the trick.

PR #6267, opened 2024-11-25
fix(server): `withPing` never calls `.return()` on iterator

KATT:
Fixes https://github.com/trpc/trpc/pull/6228#issuecomment-2496254179

> By the way, when sse.ping is enabled, the finally block is never reached. This took me 2 hours of debugging to find @KATT
> 
> _Originally posted by @tarekwiz in https://github.com/trpc/trpc/issues/6228#issuecomment-2496254179_
> 

            

## 🎯 Changes

Leans heavily into explicit resource management to ensure we don't miss calling `.return()` when mucking about with iterators

### Hide whitespace when reviewing

CleanShot 2024-11-26 at 18 27 44


scripts/getRollupConfig.ts
KATT, line 1: cc @juliusmarminge this looks ok?

packages/server/src/unstable-core-do-not-import/stream/utils/withPing.test.ts
```diff
@@ -104,3 +105,30 @@ test('yield values from source iterable', async () => {
...
+
+  const iterable = withPing(gen(), 50);
+
+  for await (const i of iterable) {
+    if (i === 10) break;
+  }
+
+  await vi.waitFor(() => {
+    expect(stopped).toBe(true);
+  });
```
KATT, line 133: cc @oleasteo this failed before

packages/server/package.json
```diff
@@ -116,6 +116,9 @@
   "publishConfig": {
     "access": "public"
   },
+  "dependencies": {
+    "@swc/helpers": "^0.5.15"
```
juliusmarminge, line 120: Why we need this?

KATT: because explicit resource managament adds a bunch of code everywhere it's used

with this it'll reference a tree-shakable library instead

i don't mind either tbh but this is how i got it to work

we could also use `tslib` which has the same purpose

KATT: i think if we set .....

https://github.com/trpc/trpc/blob/55fae79efdc0f2e911829baea2ce7228a9b457e5/scripts/getRollupConfig.ts#L100

... to `false`, we can delete this but then we'll add a bunch of noise to the bundle (which is fine) - adding `tslib`/`@swc/helpers` seems to be the recommended approach b/c then all libs "share" these additions

KATT: #6270

PR #6301, opened 2024-12-07
Bugfix: Adjust invalidate typings of utils proxy of react-query

NiklasPor:
Sorry, the last PR was from the wrong base branch and targeted at the wrong upstream branch.

## 🎯 Changes

Currently the query parameter of the predicate function inside the invalidate function of the react-query utils is typed in the following way:
```ts
Query<
  inferProcedureInput<TProcedure>,
  TRPCClientError<TRoot>,
  inferProcedureInput<TProcedure>,
  ...
>
```

That's incorrect, because the Query type requires the result data type in the third parameter:
```ts
declare class Query<TQueryFnData = unknown, TError = DefaultError, TData = TQueryFnData, TQueryKey extends QueryKey = QueryKey> extends
```

That's why we have to update the typing to use the output as a third parameter:
```ts
Query<
  inferProcedureInput<TProcedure>,
  TRPCClientError<TRoot>,
  inferTransformedProcedureOutput<TRoot, TProcedure>
  ...
>
```

## ✅ Checklist

- [ ] I have added or updated the tests related to the changes made. ➡️ Is there a useful way to test this?

### DIFF
--- a/packages/react-query/src/shared/proxy/utilsProxy.ts
+++ b/packages/react-query/src/shared/proxy/utilsProxy.ts
@@ -22,6 +22,7 @@ import type {
   AnyRouter,
   DeepPartial,
   inferProcedureInput,
+  inferProcedureOutput,
   inferTransformedProcedureOutput,
   ProtectedIntersection,
   RouterRecord,
@@ -242,7 +243,7 @@ export type DecorateQueryProcedure<
     filters?: Omit<InvalidateQueryFilters, 'predicate'> & {
       predicate?: (
         query: Query<
-          inferProcedureInput<TProcedure>,
+          inferProcedureOutput<TProcedure>,
           TRPCClientError<TRoot>,
           inferTransformedProcedureOutput<TRoot, TProcedure>,
           QueryKeyKnown<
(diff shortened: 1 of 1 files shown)


packages/react-query/src/shared/proxy/utilsProxy.ts
```diff
@@ -244,7 +244,7 @@ export type DecorateQueryProcedure<
         query: Query<
           inferProcedureInput<TProcedure>,
```
Nick-Lucas, line 245: the first parameter is the Output and the 3rd parameter is the Output after select(), so if the 3rd is wrong we'll also need to change the 1st

I'm scratching my head though because I'm surprised we got so far with it being wrong, but you do seem to be correct

NiklasPor: I think it's just a feature which isn't used very often.
I don't even actually use it, I was just playing around with the query invalidator and tried to invalidate queries based on the current query data (`{ predicate => query.state.data.includes(...) }`) and then found out that the query data was typed to the input of the query, not the output.

Nick-Lucas: If you can get the first parameter changed too then we'll get this in :)

Nick-Lucas (approved):

Nick-Lucas:
Thanks for this! lgtm

NiklasPor:
> Thanks for this! lgtm

My pleasure, one of my favorite projects ❤️

KATT (approved):

PR #6311, opened 2024-12-11
Alexdb/better structural sharing

### LINKED ISSUE #6306: feat: Better structural sharing in tanstack-query

TRPC uses `superjson` by default to provide types beyond json-serializable ones, however this breaks default structural sharing in tanstack-query.
The most obvious impact is that any native `Date` field in a response object will cause the data object returned by a `useQuery` hook to have a different identity.

Discussion for context: https://github.com/TanStack/query/pull/6481

Given that trpc uses `superjson` by default, I think I'd like to see a default `structuralSharing` function that handles equality for the types handled by superjson.

* Userland functions
* An external package that provides the functionality and can be easily dropped in. This might already exist but I was not able to find a suitable lib.
* Lobby popular parsing packages (superjson, I use devalue) to add `replaceEqualDeep` equivalent functions, since the logic of the two operations are tightly related. Using two different algos could introduce very hard-to-detect bugs.

### Additional information

Adding a default function would be a breaking change :)

I don't know what the performance impact of everything would be.

### 👨‍👧‍👦 Contributing

aldebout:
Closes #6306

## 🎯 Changes

This PR introduces a new function to handle structural sharing of non-JSON-serializable results in tanstack-query.
With this feature, API endpoints that returns `Date`s, `Set`s and `Map`s should now have more stable data object identity between renders.

## ✅ Checklist

### DIFF
--- a/packages/react-query/package.json
+++ b/packages/react-query/package.json
@@ -94,5 +94,8 @@
   },
   "funding": [
     "https://trpc.io/sponsor"
-  ]
+  ],
+  "dependencies": {
+    "fast-deep-equal": "^3.1.3"
+  }
 }
--- a/packages/react-query/src/shared/hooks/createHooksInternal.tsx
+++ b/packages/react-query/src/shared/hooks/createHooksInternal.tsx
@@ -13,6 +13,7 @@ import {
 } from '@tanstack/react-query';
 import type { TRPCClientErrorLike } from '@trpc/client';
 import { createTRPCUntypedClient } from '@trpc/client';
+import { replaceEqualDeep } from '@trpc/react-query/utils/structuralSharing';
 import type { AnyRouter } from '@trpc/server/unstable-core-do-not-import';
 import { isAsyncIterable } from '@trpc/server/unstable-core-do-not-import';
 import * as React from 'react';
@@ -182,6 +183,7 @@ export function createRootHooks<
       void prefetchQuery(queryKey, opts as any);
     }
     const ssrOpts = useSSRQueryOptionsIfNeeded(queryKey, {
+      structuralSharing: replaceEqualDeep,
       ...defaultOpts,
       ...opts,
     });
--- a/packages/react-query/src/utils/structuralSharing.ts
+++ b/packages/react-query/src/utils/structuralSharing.ts
@@ -1,0 +1,83 @@
+import equal from 'fast-deep-equal/es6';
+
+/**
+ * Taken from @tanstack/query-core utils.ts
+ * Modified to support Date, Map, Set and NaN comparisons
+ *
+ * This function returns `a` if `b` is deeply equal.
+ * If not, it will replace any deeply equal children of `b` with those of `a`.
+ */
+export function replaceEqualDeep(a: any, b: any): any {
+  if (equal(a, b)) {
+    return a;
+  }
+
+  const array = isPlainArray(a) && isPlainArray(b);
+
+  if (array || (isPlainObject(a) && isPlainObject(b))) {
+    const aItems = array ? a : Object.keys(a);
+    const aSize = aItems.length;
+    const bItems = array ? b : Object.keys(b);
+    const bSize = bItems.length;
+    const copy: any = array ? [] : {};
+
+    let equalItems = 0;
+
+    for (let i = 0; i < bSize; i++) {
+      const key = array ? i : bItems[i];
+      if (
+        !array &&
+        a[key] === undefined &&
+        b[key] === undefined &&
+        aItems.includes(key)
+      ) {
+        copy[key] = undefined;
+        equalItems++;
+      } else {
+        copy[key] = replaceEqualDeep(a[key], b[key]);
+        if (copy[key] === a[key] && a[key] !== undefined) {
+          equalItems++;
+        }
+      }
+    }
+
+    return aSize === bSize && equalItems === aSize ? a : copy;
+  }
+
+  return b;
+}
+
+function isPlainArray(value: unknown) {
+  return Array.isArray(value) && value.length === Object.keys(value).length;
+}
+
+// Copied from: https://github.com/jonschlinkert/is-plain-object
+function isPlainObject(o: any): o is object {
+  if (!hasObjectPrototype(o)) {
+    return false;
+  }
+
+  // If has no constructor
+  const ctor = o.constructor;
+  if (ctor === undefined) {
+    return true;
+  }
+
+  // If has modified prototype
+  const prot = ctor.prototype;
+  if (!hasObjectPrototype(prot)) {
+    return false;
+  }
+
+  // If constructor does not have an Object-specific method
+  if (!prot.hasOwnProperty('isPrototypeOf')) {
+    return false;
+  }
+
+  // Most likely a plain Object
+  return true;
+}
+
+function hasObjectPrototype(o: any): boolean {
+  return Object.prototype.toString.call(o) === '[object Object]';
+}
--- a/packages/tests/server/react/structuralSharing.test.ts
+++ b/packages/tests/server/react/structuralSharing.test.ts
@@ -1,0 +1,136 @@
+import { replaceEqualDeep } from '@trpc/react-query';
+import { parse, stringify } from 'superjson';
+
+test('identical plain object', async () => {
+  const data = { a: 1, b: 2 };
+  const oldResult = parse(stringify(data));
+  const newResult = parse(stringify(data));
+
+  expect(replaceEqualDeep(oldResult, newResult)).toBe(oldResult);
+});
+
+test('different plain objects', async () => {
+  const oldResult = parse(stringify({ a: 1, b: 2 }));
+  const newResult = parse(stringify({ a: 1, b: 'a' }));
+
+  expect(replaceEqualDeep(oldResult, newResult)).not.toBe(oldResult);
+});
+
+test('nested plain objects - fully identical', async () => {
+  const oldResult = parse(stringify({ a: 1, b: { c: 2 } }));
+  const newResult = parse(stringify({ a: 1, b: { c: 2 } }));
+
+  expect(replaceEqualDeep(oldResult, newResult)).toBe(oldResult);
+});
+
+test('nested plain objects - same nested object', async () => {
+  const oldResult = parse(stringify({ a: 1, b: { c: 3 } }));
+  const newResult = parse(stringify({ a: 2, b: { c: 3 } }));
+
+  expect(replaceEqualDeep(oldResult, newResult)).not.toBe(oldResult);
+  expect(replaceEqualDeep(oldResult, newResult).b).toBe(oldResult.b);
+});
+
+test('stable for dates - identical', async () => {
+  const date = new Date('2024-04-01');
+  const oldResult = parse(stringify({ a: date }));
+  const newResult = parse(stringify({ a: date }));
+
+  expect(replaceEqualDeep(oldResult, newResult)).toBe(oldResult);
+});
+
+test('stable for dates - different', async () => {
+  const unchangedObj = { nested: { value: 42 } };
+  const oldResult = parse(
+    stringify({ a: new Date('2024-04-01'), b: unchangedObj }),
+  );
+  const newResult = parse(
+    stringify({ a: new Date('2024-04-02'), b: unchangedObj }),
+  );
+
+  expect(replaceEqualDeep(oldResult, newResult)).not.toBe(oldResult);
+  expect(replaceEqualDeep(oldResult, newResult).b).toBe(oldResult.b);
+});
+
+test('stable for maps - identical', async () => {
+  const map = new Map<unknown, unknown>([
+    ['a', 1],
+    [3, 'foo'],
+  ]);
+  const oldResult = parse(stringify({ a: map }));
+  const newResult = parse(stringify({ a: map }));
+
+  expect(replaceEqualDeep(oldResult, newResult)).toBe(oldResult);
+});
+
+test('stable for maps - different order is still equal', async () => {
+  const map = new Map<unknown, unknown>([
+    ['a', 1],
+    [3, 'foo'],
+  ]);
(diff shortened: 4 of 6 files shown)


aldebout:
Current state:

* The [first commit](https://github.com/trpc/trpc/pull/6311/commits/b59bf5b7b47cda96dda506ea6e8d0d23c85b17ca) introduces the function and exports it, the [second commit](https://github.com/trpc/trpc/pull/6311/commits/71e835dc57110ab3e10d694def8c9725c5a03e9e) uses it by default making this a breaking change.
* I would appreciate some help writing tests for the second part, I'm not familiar with testing hooks in components, we need to get the first result, invalidate, get the second result and compare both.
* I'm not sure where we should reflect this in the docs?

packages/react-query/src/utils/structuralSharing.ts
```diff
@@ -0,0 +1,83 @@
+import equal from 'fast-deep-equal/es6';
```
KATT, line 1: We don't like to have dependencies, so please remove this

packages/react-query/package.json
```diff
@@ -94,5 +94,8 @@
   },
   "funding": [
     "https://trpc.io/sponsor"
-  ]
+  ],
+  "dependencies": {
+    "fast-deep-equal": "^3.1.3"
```
KATT, line 99: We don't want dependencies

```suggestion
```

packages/tests/server/react/structuralSharing.test.ts
KATT, line 1: Not a blocker, but if you're doing unit tests, I think you can have this right next to the file you're testing rather than in the testing package

KATT: Would also like to see some actual integration tests that leverages this - see `useQuery.test.ts` etc

packages/react-query/src/shared/hooks/createHooksInternal.tsx
```diff
@@ -182,6 +183,7 @@ export function createRootHooks<
       void prefetchQuery(queryKey, opts as any);
     }
     const ssrOpts = useSSRQueryOptionsIfNeeded(queryKey, {
+      structuralSharing: replaceEqualDeep,
```
KATT, line 186: This should be opt-in based on some setting or grabbed from context or whatever

KATT (dismissed):
Needs

- A test that leverages the structural sharing in a test that renders a react app
- Removal of dependency
- Making it opt-in somehow (I'm guessing this is something you pass to the react query proivder?)

packages/react-query/src/utils/structuralSharing.ts
aldebout: Done, I copy-pasted the code.

www/docs/client/react/setup.mdx
```diff
@@ -97,6 +97,25 @@ export function App() {
 The reason for using `useState` in the creation of the `queryClient` and the `TRPCClient`, as opposed to declaring them outside of the component, is to ensure that each request gets a unique client when using SSR. If you use client side rendering then you can move them if you wish.
 :::
 
+:::tip
+If you're using a transformer like `superjson`, you probably want React Query to be able to use [structural sharing](https://tanstack.com/query/latest/docs/framework/react/guides/render-optimizations#structural-sharing) for non-JSON types like Date, Set and Map. You can do this by setting the `structuralSharing` option to `replaceEqualDeep` in the `queryClient` options.
```
KATT, line 101: Add a blurb _why_ you'd want structural sharing 🙏

packages/react-query/src/utils/structuralSharing.ts
```diff
@@ -0,0 +1,154 @@
+/**
+ * Taken from @tanstack/query-core utils.ts
+ * Modified to support Date, Map, Set and NaN comparisons
+ *
+ * This function returns `a` if `b` is deeply equal.
+ * If not, it will replace any deeply equal children of `b` with those of `a`.
+ */
+export function replaceEqualDeep(a: any, b: any): any {
```
KATT, line 8: Since we have this as the only dependency instead, can we just import it instead?

aldebout: As discussed, I updated the wording.

packages/react-query/src/utils/structuralSharing.ts
```diff
@@ -0,0 +1,154 @@
+/**
+ * Inspired by @tanstack/query-core utils.ts
+ * Modified to support Date, Map, Set and NaN comparisons
+ *
+ * This function returns `a` if `b` is deeply equal.
+ * If not, it will replace any deeply equal children of `b` with those of `a`.
+ */
+export function replaceEqualDeep(a: any, b: any): any {
```
KATT, line 8: I think we should allow custom comparators here too - for instance if you use any custom data types like `Temporal.PlainDate` etc?

Could be a factory function like


Also, I'm not sold on the naming 🤔 🤔

packages/react-query/src/utils/structuralSharing.test.ts
```diff
@@ -0,0 +1,154 @@
+import { replaceEqualDeep } from '@trpc/react-query';
```
KATT, line 1:
```suggestion
import { replaceEqualDeep } from './structuralSharing';
```

packages/react-query/src/utils/structuralSharing.ts
aldebout: I changed to a factory function + a default comparator, let me know what you think.

I  also renamed but I'm still not sure it's great.

packages/react-query/src/utils/structuralSharing.ts
```diff
@@ -0,0 +1,132 @@
+/**
+ * Inspired by @tanstack/query-core utils.ts
+ * Modified to support Date, Map, Set and NaN comparisons
+ *
+ * This function returns `a` if `b` is deeply equal.
+ * If not, it will replace any deeply equal children of `b` with those of `a`.
+ */
+export function createStructuralSharingFunction(
+  equalityFn: (a: any, b: any) => boolean,
```
KATT, line 9: nit:
```suggestion
  equalityFn: (a: unknown, b: unknown) => boolean,
```

KATT:
Thanks for this, will do proper review later in the week!

packages/react-query/src/utils/structuralSharing.ts
```diff
@@ -0,0 +1,132 @@
...
+
+function isPlainArray(value: unknown) {
+  return Array.isArray(value) && value.length === Object.keys(value).length;
+}
+
+// Copied from: https://github.com/jonschlinkert/is-plain-object
+function isPlainObject(o: any): o is object {
+  if (!hasObjectPrototype(o)) {
+    return false;
+  }
```
KATT, line 107: We have a much simplified version here:

https://github.com/trpc/trpc/blob/3957a7dc564b14f52ce6ba5078490e73b6b6be98/packages/server/src/unstable-core-do-not-import/utils.ts#L26-L32

Which one is wrong? 😅

packages/react-query/src/utils/structuralSharing.ts
```diff
@@ -0,0 +1,132 @@
...
+  if (a.valueOf !== Object.prototype.valueOf)
+    return a.valueOf() === b.valueOf();
+  if (a.toString !== Object.prototype.toString)
+    return a.toString() === b.toString();
+  return false;
+}
+
+function isPlainArray(value: unknown) {
+  return Array.isArray(value) && value.length === Object.keys(value).length;
+}
```
KATT, line 101: How come `isPlainArray` rather than `Array.isArray()`? What's the difference?

packages/react-query/src/utils/structuralSharing.ts
```diff
@@ -0,0 +1,132 @@
...
+    return b;
+  };
+
+  return structuralSharingFunction;
+}
+
+export const defaultStructuralSharingFunction =
+  createStructuralSharingFunction(isEqual);
+
+function isEqual(a: any, b: any) {
```
KATT, line 86:
```suggestion
export function defaultIsEqual(a: unknown, b: unknown) {
```

export it so it can be used when making a custom `createStructuralSharingFunction`

www/docs/client/react/setup.mdx
```diff
@@ -97,6 +97,27 @@ export function App() {
 The reason for using `useState` in the creation of the `queryClient` and the `TRPCClient`, as opposed to declaring them outside of the component, is to ensure that each request gets a unique client when using SSR. If you use client side rendering then you can move them if you wish.
 :::
 
+:::tip
+React Query uses a technique called [structural sharing](https://tanstack.com/query/latest/docs/framework/react/guides/render-optimizations#structural-sharing) to ensure that as many references as possible will be kept intact between re-renders.
```
KATT, line 101: I still want an explanation of why it's important to keep values intact between re-renders

I've personally haven't ever felt the need of this optimization and I refetch aggressively [on any mutation](https://trpc.io/docs/client/react/useUtils#invalidate-full-cache-on-every-mutation)

KATT (changes requested):
- Fails CI, please fix
- Also add a motivation on why I'd want to use this

packages/react-query/src/utils/structuralSharing.ts
aldebout: This comes for RQ and I believe most of it aims to avoid edge cases like this one https://github.com/TanStack/query/pull/7357.
Because we do `copy = {}` if there's a custom constructor we lose something.

packages/react-query/src/utils/structuralSharing.ts
aldebout: Similar to the object, I believe this is to avoid edge cases https://github.com/TanStack/query/pull/3669.

www/docs/client/react/setup.mdx
aldebout: I think there's two ways to look at it:

Say you have a page with an "edit" form that uses RHF `useForm` and fetches the current data client-side. There's multiple ways to do this, but there's a chance you have something like

```tsx
useEffect(() => {reset(data.values)}, [data.values])
```

somewhere.

With structural sharing, even if the query is refetched for some reason (for example, you switch tabs with the default behavior), the identity of `data.values` is stable as long at the data stays the same. Without structural sharing, the identity is not stable, the form is reset and you lose any changes you were in the process of making.
I think there are other ways to tackle this specific use-case, but overall I think it's a nice thing that helps react-query work in a more intuitive way.

The second argument is that structural sharing is already present in RQ by default, for json-compatible structures. Since trpc has a RQ integration, I think it makes sense to also provide this feature for those that want it, keeping behavior similar to vanilla RQ while benefiting from the type safety of trpc.
FWIW, stability for query results that contain `Date` objects has been requested / tried [multiple](https://github.com/TanStack/query/pull/6481) [times](https://github.com/TanStack/query/pull/7151) at RQ.

packages/react-query/src/utils/structuralSharing.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+test('stable for sets - different order is still equal', async () => {
+  const set = new Set([1, 2, 3]);
+  const oldResult = parse(stringify({ a: set }));
+  const newResult = parse(stringify({ a: new Set([3, 2, 1]) }));
+
+  expect(defaultStructuralSharingFunction(oldResult, newResult)).toBe(
+    oldResult,
+  );
+});
```
KATT, line 149: I am wondering if this is a bug or a feature? Is this how RQ handles arrays?

packages/react-query/src/utils/structuralSharing.test.ts
```diff
@@ -0,0 +1,208 @@
...
+      ]),
+    }),
+  );
+
+  expect(defaultStructuralSharingFunction(oldResult, newResult)).toBe(
+    oldResult,
+  );
+});
+
+test('stable for maps - different keys are not equal', async () => {
```
KATT, line 112: Not sure if order should matter or not here?

packages/react-query/src/utils/structuralSharing.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    oldResult.b,
+  );
+});
+
+test('sets - different sizes are not equal', async () => {
+  const prev = new Set([1]);
+  const next = new Set([1, 2]);
+
+  expect(defaultStructuralSharingFunction(prev, next)).toEqual(next);
+});
```
KATT, line 173: added this test and fixed the bug

packages/react-query/src/utils/structuralSharing.ts
```diff
@@ -0,0 +1,148 @@
...
+ *   return false
+ * });
+ */
+export function createStructuralSharingFunction(
+  customEqualityFunction: (a: unknown, b: unknown) => boolean,
+) {
+  const equalCheck = (a: unknown, b: unknown) => {
+    return customEqualityFunction(a, b) || defaultIsEqual(a, b);
+  };
+  const structuralSharingFunction = (prev: any, next: any): any => {
```
KATT, line 22: renamed these to `prev` and `next` fyi

packages/react-query/src/utils/structuralSharing.ts
```diff
@@ -0,0 +1,148 @@
...
+    if (equalCheck(prev, next)) {
+      return prev;
+    }
+
+    const array = isPlainArray(prev) && isPlainArray(next);
+    const plainObject = isPlainObject(prev) && isPlainObject(next);
+    const map = prev instanceof Map && next instanceof Map;
+    const set = prev instanceof Set && next instanceof Set;
+
+    if (array || plainObject || map) {
```
KATT, line 32: The code within here is.... horribly hard to follow

should we split it up into less DRY operations instead?

KATT (commented):
changed some variable names, fixed a bug, and left some comments

KATT:
please re-request review or ping me on discord when you have responded etc!

PodkopovP:
Anything happen with this one?

PR #6329, opened 2024-12-21
Always show an error message even if a non-error object with no message is thrown

juliesaia:
## 🎯 Changes

Currently, there's an edge case where if a non-error object with no `message` property is thrown, no message is shown at all. 

In my case this came up with this line in the package `@instantdb/admin`: `Promise.reject({ status: res.status, body: json });`

In an effort to show *something* to the user to help with debugging, this PR adds a last resort of stringifying the entire error object as a message.

I restricted this to dev because of similar logic for hiding `err.stack`, it's not feasible to access `config.isDev` from an error constructor though so hoping that process.env.NODE_ENV is good enough.

## ✅ Checklist

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/error/TRPCError.ts
+++ b/packages/server/src/unstable-core-do-not-import/error/TRPCError.ts
@@ -25,6 +25,10 @@ export function getCauseFromUnknown(cause: unknown): Error | undefined {
     for (const key in cause) {
       err[key] = cause[key];
     }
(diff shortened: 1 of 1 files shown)


KATT (changes requested):
Please add a test that shows how the behaviour was wrong before

juliesaia:
Dug into this a little deeper, ended up being a bug in next's edge runtime specifically 😅 sorry about that!

https://github.com/vercel/next.js/issues/74218

PR #6355, opened 2025-01-02
chore: improve jsonl async code

KATT:
Closes #

## 🎯 Changes

Adds a `raceAsyncIterables`-function that can race async iterables

No reason, was just fun

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/stream/utils/raceAsyncIterables.ts
+++ b/packages/server/src/unstable-core-do-not-import/stream/utils/raceAsyncIterables.ts
@@ -9,7 +9,58 @@ type ActiveStateIdle = Brand<1, 'Idle'>;
 const ActiveStateIdle = 1 as ActiveStateIdle;
 type State = ActiveStatePending | ActiveStateIdle;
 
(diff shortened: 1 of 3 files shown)


packages/server/src/unstable-core-do-not-import/stream/utils/raceAsyncIterables.ts
```diff
@@ -0,0 +1,183 @@
...
+>;
+
+interface RaceAsyncIterables<TYield>
+  extends AsyncIterable<TYield, void, unknown> {
+  add(iterable: AsyncIterable<TYield>): void;
+}
+
+/**
+ * Creates a new async iterable that merges multiple async iterables into a single stream.
+ * Values from the input iterables are yielded in the order they resolve, similar to Promise.race().
```
oleasteo, line 66: Since we only complete once all iterables have completed, it's more comparable to a `merge` than a `race`. I don't see the similarity to `Promise.race()` that strongly.
The completion condition and the fact that iterable return values are ignored should be added to the documentation.

oleasteo (approved):
Looks good to me (aside from documentation and naming, see comment).

PR #6377, opened 2025-01-10
feat(react-query): Support for the useQueries combine option

### LINKED ISSUE #6375: feat: Support `combine` option on `useQueries`

Tanstack Query has a `combine` option on `useQueries` that lets you process multiple results before passing them along.

This nice because the array that `useQueries` does not have a stable identity.

```ts
const flatten = <T>(arr: T[][]): T[] => arr.flat()

function Component() {
  const { data: books } = trpc.books.useQueries((t) =>
    decades.map((decade) => t.book.findInDecade({ decade })),
  )

  // Memo is invalidated every render
  const allBooks = useMemo(flatten, [books])
}
```

The `combine` solves this with some special handling of [memoization](https://tanstack.com/query/v5/docs/framework/react/reference/useQueries#memoization):

> The combine function will only re-run if:
>
> - the combine function itself changed referentially
> - any of the query results changed

Right now, though, tRPC cannot send a `combine` function to `useQueries`.

Ideally I could just pass the `combine` option through:

```ts
const flatten = <T>(arr: T[][]): T[] => arr.flat()

function Component() {
  const { data: books } = trpc.books.useQueries(
    (t) => decades.map((decade) => t.book.findInDecade({ decade })),
    { combine: flatten },
  )

  // `books` has a stable identity, unless the combine function changes, decades changes,
  // or something is invalidated.
}
```

The `callback, { combine }` api is maybe a little odd? Then maybe we support an overload signature that takes one object.

```ts
const { data: books } = trpc.books.useQueries({
  queries: (t) => decades.map((decade) => t.book.findInDecade({ decade })),
  combine: flatten,
})
```

### Prior Issues

This comment https://github.com/trpc/trpc/pull/4662#pullrequestreview-1559140858
Leads to this closed issue https://github.com/trpc/trpc/issues/4678
Which was closed as `not planned`, but with no explanation.

### Relevant Code

The handoff from tRPC to tanstack query happens here:
https://github.com/trpc/trpc/blob/next/packages/react-query/src/shared/hooks/createHooksInternal.tsx#L681-L709

And only takes one parameter: the queries.

### Existing Docs

But the [documentation site](https://trpc.io/docs/client/react/useQueries#context) says that you can pass a context as the second parameter, but I think this actually doesn't work at all, and is a type error to do so. If that is to be supported, that would need to be worked into the api as well. Perhaps by sticking close to how tanstack does it:

```ts
const { data: books } = trpc.books.useQueries({
  queries: (t) => decades.map((decade) => t.book.findInDecade({ decade })),
  combine: flatten,
}, customCtx)
```

### 👨‍👧‍👦 Contributing

AlexJWayne:
Closes #6375 

## 🎯 Changes

Adds support for the `combine` option of `useQueries`.

## Notes

This API probably needs to be thought about, but currently looks like this:

```ts
    const results = client.useQueries(
      (t) => [t.post.byId({ id: '1' }), t.post.byId({ id: '2' })],
      {
        combine: (results) =>
          results.map((result) => result.data).join(' and '),
      },
    );
```

I wasn't able to solve the typing issue here https://github.com/trpc/trpc/compare/next...AlexJWayne:trpc:use-queries-combine?expand=1#diff-9ddf87d87d9e564be08667cf18fc907796457d5d18386c4ff1ffb7438b42a7a5R706 but I'm not sure it is solvable and this isn't the only `as any` in this file.

## ✅ Checklist

### DIFF
--- a/packages/react-query/src/internals/useQueries.ts
+++ b/packages/react-query/src/internals/useQueries.ts
@@ -189,6 +189,7 @@ export type TRPCUseQueries<TRouter extends AnyRouter> = <
       TRouter['_def']['record']
     >,
   ) => readonly [...QueriesOptions<TQueryOptions>],
+  options?: { combine: any },
 ) => QueriesResults<TQueryOptions>;
 
 /**
--- a/packages/react-query/src/internals/useQueries.ts
+++ b/packages/react-query/src/internals/useQueries.ts
@@ -182,6 +182,7 @@ export type SuspenseQueriesOptions<
  */
 export type TRPCUseQueries<TRouter extends AnyRouter> = <
   TQueryOptions extends UseQueryOptionsForUseQueries<any, any, any, any>[],
+  TCombinedResult = QueriesResults<TQueryOptions>,
 >(
   queriesCallback: (
     t: UseQueriesProcedureRecord<
@@ -189,8 +190,10 @@ export type TRPCUseQueries<TRouter extends AnyRouter> = <
       TRouter['_def']['record']
     >,
   ) => readonly [...QueriesOptions<TQueryOptions>],
-  options?: { combine: any },
-) => QueriesResults<TQueryOptions>;
+  options?: {
+    combine?: (results: QueriesResults<TQueryOptions>) => TCombinedResult;
+  },
+) => TCombinedResult;
 
 /**
  * @internal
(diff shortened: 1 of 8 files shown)


AlexJWayne:
If the overall approach and API is approved then I will add docs to this PR as well

KATT (approved):

Issue #4876, opened 2023-10-04
feat: TRPC Proxy Client's context type


yikkwongwong:
### Describe the feature you'd like to request

Currently TRPC Proxy Client context is of type `ClientContext` and is not an extendable interface like [`OperationContext`](https://github.com/trpc/trpc/issues/4518).

### Describe the solution you'd like to see

Can TRPC Proxy client's context follow `OperationContext`? Or make it an `interface` so we can extend on it.

### Describe alternate solutions

I am not sure why Proxy Client did not use OperationContext but the terminating link is using it. Would be good if both types are synced.

### 👨‍👧‍👦 Contributing

- [ ] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

juliusmarminge:
I'm not sure I understand what you want to achieve. You can put whatever you want in the request context, do you want to restrict that on some global type? Can you provide some code snippet of what you want to achieve?

yikkwongwong:
Heyy @juliusmarminge, yup we can put whatever we want in the request context, but the types cannot be inferred. TRPCProxyClient is using ClientContext and not an extendable interface. OperationContext is fixed in https://github.com/trpc/trpc/issues/4518.

Example
```
interface CustomContext { }

declare module '@trpc/client' {
    interface OperationContext extends CustomContext {} //works
    interface ClientContext extends CustomContext {} //nope
}
```

Not-Jayden:
+1 on this. I want to add a way to configure retry options per procedure call in certain cases, but there's no type safe way of being able to do this via context.

Are there any good workarounds?

PR #6383, opened 2025-01-10
chore: move test helpers to corresponding package

KATT:
Closes #

## 🎯 Summary

I'm having a stab of doing some housekeeping of breaking apart `packages/tests`.

I'd want tests to do with `react-query` to live in that package, etc and I think this would move us to that direction.

## How

The idea is that each package can export test helpers that we can have internally — e.g. `@trpc/server` can have a `@trpc/server/__tests__` (which is ignored in `package.json`) that we can use internally.

This allows us to have

- Generic test helpers in `@trpc/server`
- Test helpers that only rely on `@trpc/server` in `@trpc/server`
- Test helpers that rely on client + server in `@trpc/client` (e.g. our `routerToServerAndClientNew`-monster)

## Challenges with the current approach that I'd want help with 🙋‍♂️

I think the tsconfig/rollup setup needs to be tweaked to fix the below problems. Could be a follow-up?

- Had to do a hack (see `__getSpy()`)
- Auto-import of the utils doesn't seem to work that reliably with how I set it up
- "Find references" of anything within `__tests__` doesn't seem to work across projects
- Rollup gives warnings about our test files - unsure why they aren't excluded
- I'd like to nuke `/packages/tests` completely. It's a lot of grunt work to move them to the right packages

## Summary by CodeRabbit

- **Chores**
  - Updated version control settings to include test directories.
  - Added a development dependency for enhanced WebSocket support.
- **Tests**
  - Improved test resource management, error handling, and timer utilities for more reliable testing.
- **Refactor**
  - Consolidated and standardized test infrastructure for better maintainability and stability.
- **Style**
  - Refined formatting configurations for a cleaner and more consistent import order.

These improvements reinforce overall application reliability while keeping runtime behavior for end‑users unchanged.


packages/client/src/__tests__/testClientResource.ts
```diff
@@ -0,0 +1,128 @@
...
+    lazy: {
+      enabled: true,
+      closeMs: 0,
+    },
+    ...opts?.wsClient,
+  });
+
+  const transformer = router._def._config.transformer as any;
+  const trpcClientOptions = {
+    links: [
```
KATT, line 65: hello #4876 and #6370

juliusmarminge (approved):

PR #6435, opened 2025-01-30
fix(server): do not abort when reading http2 req.body

### LINKED ISSUE #6432: bug: Response not written to POST requests on HTTP2 connection

```
  System:
    OS: Linux 6.12 Arch Linux
    CPU: (16) x64 AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics
    Memory: 7.71 GB / 30.06 GB
    Container: Yes
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.2.0 - ~/.nvm/versions/node/v22.2.0/bin/node **(also tried 23.6.1)**
    Yarn: 1.22.22 - ~/.nvm/versions/node/v22.2.0/bin/yarn
    npm: 10.7.0 - ~/.nvm/versions/node/v22.2.0/bin/npm
    bun: 1.0.3 - ~/.bun/bin/bun
  npmPackages:
    @tanstack/react-query: ^5.64.2 => 5.64.2 
    @trpc/client: 11.0.0-rc.718 => 11.0.0-rc.718+2018e3914 
    @trpc/react-query: 11.0.0-rc.718 => 11.0.0-rc.718+2018e3914 
    @trpc/server: 11.0.0-rc.718 => 11.0.0-rc.718+2018e3914 
    react: ^18.3.1 => 18.3.1 
    typescript: 5.7.2 => 5.7.2
```

### Describe the bug

When using http2, the response body is always empty when the method is POST. I was able to trace the problem and discovered that the reason for this is that the request is being aborted in response to this event:

https://github.com/trpc/trpc/blob/776d07336bddd01ac2ad02da0360fb71aa734471/packages/server/src/adapters/node-http/incomingMessageToRequest.ts#L147

If I don't pass the abort signal in the stream pipe, it is able to write to the stream without any issue and everything works. Here is where I made that change:

https://github.com/trpc/trpc/blob/776d07336bddd01ac2ad02da0360fb71aa734471/packages/server/src/adapters/node-http/writeResponse.ts#L50-L52

I'll be honest, it doesn't make sense to me why the response close event fires and yet I can still write it to it just fine. I'm also not sure how to deal with this situation without breaking proper abort handling elsewhere.

### Link to reproduction

https://github.com/cseickel/trpc-http2-post-example

### To reproduce

Unfortunately it does not run in stakblitz for me, maybe because it requires an ssl connection. You can clone the linked repo and follow the readme.

https://github.com/cseickel/trpc-http2-post-example

### Additional information

I understand that no where does the project explicitly state that HTTP2 is supported, but I'm hoping it is within scope. The reason I want to use HTTP2 is that it is really important for using the SSE subscription feature. The reason for this is that each subscription opens a new connection, and with HTTP/1 the browser will not open more than 6 connections of any type to one server. This limit applies across multiple tabs as well, so it is really hard to stay under that limit. 

HTTP/2 removes this limit and makes SSE feasible.

Thank you for the awesome project!

### 👨‍👧‍👦 Contributing

cseickel:
Closes #6432

## 🎯 Changes

Listen to the request `close` event instead of the `end` event to handle aborted requests.

This is actually how it was prior to PR #6206. There was no specific explanation as to why the change was made there so I'm not sure if there is a reason it was needed. I'm working under the assumption that this change was a mistake because it seems like `close` was the correct event to listen for.

The specific problem I am fixing here is that when the request is a POST, it reads the body stream. Once the stream has been fully read, the `end` event will be fired which then triggers the abort signal. I ran into this problem with HTTP2 requests specifically, but I could see how this could be problematic elsewhere as well.

## ✅ Checklist

## Summary by CodeRabbit

- **Bug Fixes**
	- Updated request abort mechanism to improve socket handling and request termination.

### DIFF
--- a/packages/server/src/adapters/node-http/incomingMessageToRequest.ts
+++ b/packages/server/src/adapters/node-http/incomingMessageToRequest.ts
@@ -145,7 +145,7 @@ export function incomingMessageToRequest(
   };
 
   res.once('close', onAbort);
-  req.socket?.once?.('end', onAbort);
+  req.socket?.once?.('close', onAbort);
 
   // Get host from either regular header or HTTP/2 pseudo-header
   const url = createURL(req);


cseickel:
CodeRabbit is pretty awesome. Also, I'm embarrassed to say that I did not run the tests before submitting but now that I did I see the link to the issue (#6193) that explains why you are listing to 'end'.

I will fix this up and add proper tests. I'm thinking the fix is to just unsubscribe from 'end' just before we read the body, and then resubscribe afterwards so that reading the body does not abort but you still have your bug fix for the other situation.

cseickel:
This is a little bigger than I initially thought when I opened this. Here's what I think needs to be done:

1. Repeat all express/standalone adapter tests against an http1 and an http2 server.
  - I know at least the "read POST body" one will fail with http2 
2. Change the test added for handling aborted sse connections, which just fires the 'end' event and expects the connection to close. I really think this is wrong and only seemed to fix the problem because it aggressively closed too many connections without cause. The test should be more precise:
  - Create a subscription endpoint with cleanup code
  - Setup a subscription
  - Ensure the event source is active
  - Have the client close the connection
  - Assert that the event source is ended an cleanup code has run
3. Fix failing tests.

I'll join the discord this weekend to try and discuss this.

KATT:
The failing test you have here is safe to ignore as it's a mocked request.

If we could refactor the `incomingMessageToRequest.test.ts` to use real http + http2 servers rather than mocking, that'd be a great test... 🤔

cseickel:
> If we could refactor the `incomingMessageToRequest.test.ts` to use real http + http2 servers rather than mocking, that'd be a great test... 🤔

Absolutley, I don't see why we couldn't do that. It's certainly worth try at least. I'll give it a go over the weekend.

KATT:
Amazing!! 🤩

KATT:
> I'm not positive when I will get that done, so if someone else wants to work on it immediately then just let me know.

Thanks, no stress.

If you could just make a test with a real HTTP2 server that fails, that'd be amazing. I can handle the rest.

KATT:
Superseded by #6452 that adds more "official" support

@cseickel, could you please review it? 🙏

PR #6223, opened 2024-11-12
chore(client): refactor & undeprecate `wsLink`

hmatthieu:
Closes #6109 

# WebSocket Implementation Refactoring

This PR addresses #6109 by completely refactoring the WebSocket implementation to improve maintainability. The new implementation maintains all existing functionality while providing a cleaner and more modular architecture.

## 🔄 Architecture Improvements
Split the monolithic WebSocket client into focused, single-responsibility modules:
- `wsLink`: Streamlined to handle only connection state observability
- `WsClient`: Acts as the high-level coordinator between modules
- `WsConnection`: Provides low-level WebSocket control
- `RequestManager`: Dedicated to managing tRPC request lifecycles
- `ReconnectManager`: Handles reconnection logic and backoff strategies
- `PingPong`: Implements connection health monitoring

## ✅ Checklist

- [x] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [x] If necessary, I have added documentation related to the changes made.
- [x] I have added or updated the tests related to the changes made.

## Summary by CodeRabbit

- **New Features**
  - Improved real-time communication with advanced connection management, including lazy initialization, keep-alive support, and robust error handling.
  - Introduced a new WebSocket client class with enhanced features for managing TRPC operations.
  - Added a new function for creating a WebSocket client with customizable options.
  - Implemented a comprehensive WebSocket connection management system with ping-pong mechanisms.
  - Enhanced the `RequestManager` class for better lifecycle management of WebSocket requests.
  - Added new utility functions for error handling and connection message construction.
  - Updated the WebSocket link functionality for improved communication between client and server.

- **Bug Fixes**
  - Ensured that WebSocket connections now close gracefully by handling closures asynchronously, reducing potential disruptions.

- **Refactor**
  - Streamlined the internal structure of WebSocket modules to enhance maintainability and simplify configuration.
  - Modified the management of WebSocket server handlers for improved flexibility.
  - Updated export paths and organization for better module structure.

- **Tests**
  - Expanded coverage with new scenarios that simulate connection failures and verify active subscriptions for improved stability.

PR #6453, opened 2025-02-05
fix(client): trigger ws reconnect on keepalive timeout

TomKaltz:
## 🎯 Changes

This fixes a bug where a pong timeout (when keepalive is true) destroys the connection but does not properly trigger the reconnection logic.

## ✅ Checklist

## Summary by CodeRabbit

- **Refactor**
  - Improved error messaging for connection interruptions. Users will now see a clearer, more descriptive notification when a connection fails to receive the expected acknowledgment, making it easier to understand and diagnose connectivity issues.


packages/client/src/links/wsLink.ts
```diff
@@ -402,6 +402,11 @@ export function createWSClient(opts: WebSocketClientOptions) {
               pongTimeout = setTimeout(() => {
                 const wasOpen = self.state === 'open';
                 destroy();
```
KATT, line 404: I'm not sure `destroy()` is needed if we do `onCloseOrError`?

KATT (commented):
This code is all being rewritten by @hmatthieu in #6223 so can't guarantee any fixes you do here will be permanent without a regression test

packages/client/src/links/wsLink.ts
TomKaltz: You're right the reconnection logic handles closing the original connection.  I've remove the call to destroy()

KATT (approved):

TomKaltz:
This won't automatically merge but I look forward to the work @hmatthieu is doing and if I get some free time I will try to see if I can contribute in some way to that effort.

PR #6488, opened 2025-02-12
custom content types in responses

KATT:
I think it would be fairly easy to allow custom content types in responses

[**Read the RFC**](https://github.com/trpc/trpc/blob/02-12-response-types-idea/.rfcs/001-custom-content-types.mdx)

### DIFF
--- a/.rfcs/001-custom-content-types.mdx
+++ b/.rfcs/001-custom-content-types.mdx
@@ -1,0 +1,80 @@
+## Custom response types (RFC)
+
+You can return a `Response` object from your procedure by specifying `'Response'` as the output type. This is useful when you need full control over the response, such as returning files, streams, or custom headers.
+
+This feature allows you to have full control over the HTTP response, enabling use cases like:
+
+- File downloads and other binary data
+- Custom content types
+- Custom headers and status codes
+
+Note that when using `'Response'`:
+
+- You're responsible for properly formatting the response
+- You're responsible for setting appropriate headers
+- You're responsible for doing manual deserialization and type safety on the client when needed
+- The responses can not be batched and will be rejected by the server if attempted (needs to be called with `httpLink()`)
+
+```ts twoslash
+import { initTRPC } from '@trpc/server';
+
+export const t = initTRPC.create();
+const publicProcedure = t.procedure;
+
+export const appRouter = t.router({
+  // Simple file download example
+  downloadFile: publicProcedure
+    .output('Response')
+    .query(() => {
+      return new Response('Hello World', {
+        headers: {
+          'Content-Type': 'text/plain',
+          'Content-Disposition': 'attachment; filename="hello.txt"',
+        },
+      });
+    }),
+
+  // Stream example with custom headers
+  streamData: publicProcedure
+    .input(z.object({ count: z.number().min(1) }))
+    .output('Response')
+    .mutation(async ({ input }) => {
+      const stream = new ReadableStream({
+        async start(controller) {
+          for (let i = 0; i < input.count; i++) {
+            controller.enqueue(`data: ${i}\n\n`);
+            await new Promise(r => setTimeout(r, 1000));
+          }
+          controller.close();
+        },
+      });
+
+      return new Response(stream, {
+        headers: {
+          'Content-Type': 'text/event-stream',
+          'Cache-Control': 'no-cache',
+          'Connection': 'keep-alive',
+        },
+      });
+    }),
+
+  // Binary file example
+  getImage: publicProcedure
+    .input(z.object({ id: z.string() }))
+    .output('Response')
+    .query(async ({ input }) => {
+      const imageBuffer = await getImageFromDatabase(input.id); // hypothetical function
+
+      return new Response(imageBuffer, {
+        headers: {
+          'Content-Type': 'image/jpeg',
+          'Cache-Control': 'public, max-age=3600',
+        },
+      });
+    }),
+});
+```
+
+### Why a static `.output('experimental_Response')`?
+
+The `.output('experimental_Response')` is static because it changes the type of the procedure in a way that is not compatible with the usual tRPC type system and allows us to reject a request if it is tried to be called with e.g. batching.


.rfcs/001-custom-content-types.mdx
juliusmarminge, line 1: Putting my snippet in here for the case of public transparancy, but I think this can be more of a user-land thing that would be more generalizable with procedure builders:

```
import { TRPCError, initTRPC } from "@trpc/server";
import { fetchRequestHandler } from "@trpc/server/adapters/fetch";
import { z } from "zod";

const t = initTRPC
  .meta<{
    allowBatch?: boolean;
  }>()
  .context<{
    isBatch: boolean;
  }>()
  .create();

const conditionalBatchProcedure = t.procedure.use(async (opts) => {
  if (opts.ctx.isBatch && opts.meta?.allowBatch === false) {
    throw new TRPCError({
      code: "BAD_REQUEST",
      message: "Batch requests are not allowed",
    });
  }

  return opts.next();
});

// Procedure that doesn't allow batching
const noBatchProcedure = conditionalBatchProcedure.meta({
  allowBatch: false,
});

// Procedure that only allows returning responses
const responseProcedure = noBatchProcedure.output(z.instanceof(Response));

const myProcedure = responseProcedure.mutation(async (opts) => {
  // ...
  return new Response("Hello, world!");
});

const router = t.router({
  myProcedure,
});

const handler = (req: Request) => {
  return fetchRequestHandler({
    endpoint: "/api/trpc",
    req,
    router,
    createContext: (opts) => {
      return {
        isBatch: opts.info.isBatchCall,
      };
    },
  });
};

export { handler as GET, handler as POST };
```

.rfcs/001-custom-content-types.mdx
```diff
@@ -0,0 +1,161 @@
...
+
+This has been a long-requested feature in [#1937](https://github.com/trpc/trpc/issues/1937) which has been partially solved already by allowing the use of handling e.g. `FormData` in mutation inputs.
+
+## tl;dr
+
+```ts
+// Backend:
+const appRouter = router({
+  hello: publicProcedure
+    .output('Response') // <-- define that we're returning a Response
```
KATT, line 13: We will probably drop this part

```suggestion
```

packages/client/src/internals/types.ts
```diff
@@ -67,6 +67,10 @@ export type NodeJSReadableStreamEsque = {
   ): NodeJSReadableStreamEsque;
 };
 
+interface HeadersEsque {
+  get(key: string): string | null;
+}
```
juliusmarminge, line 72: i thought we had one of these already 🤔

packages/client/src/links/internals/httpUtils.ts
```diff
@@ -223,12 +228,18 @@ export async function fetchHTTPResponse(opts: HTTPRequestOptions) {
   });
 }
 
-export async function httpRequest(
+export async function jsonHttpRequest(
```
juliusmarminge, line 231: confusing naming when this retuns both types?

packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
```diff
@@ -351,6 +351,18 @@ export async function resolveResponse<TRouter extends AnyRouter>(
           type: proc._def.type,
           signal: opts.req.signal,
         });
+
+        if (
+          config.experimental?.outputResponse &&
+          info.isBatchCall &&
+          data instanceof Response
+        ) {
```
juliusmarminge, line 359: pretty sure it can't be batched whether you have the flag enabled or not?

```suggestion
        if (info.isBatchCall && data instanceof Response) {
```

packages/tests/server/___testHelpers.ts
```diff
@@ -117,9 +119,21 @@ export function routerToServerAndClientNew<TRouter extends AnyRouter>(
   });
   const trpcClientOptions = {
     links: [
-      httpBatchLink({
-        url: httpUrl,
-        transformer: opts?.transformer as any,
+      splitLink({
```
juliusmarminge, line 122: we should also be able to use this with `unstable_httpBatchStreamLink` right? as the responses there aren't batched, just the requests?

juliusmarminge: nvm im dumb it's 1 response just the json chunks being streamed down...

packages/client/src/links/internals/httpUtils.ts
KATT: lol yeah 🤣
All the utility functions here are a mess

packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
KATT: yeah just wanted to avoid a breaking change

packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
```diff
@@ -371,9 +379,27 @@ export async function resolveResponse<TRouter extends AnyRouter>(
 
     // ----------- response handlers -----------
     if (!info.isBatchCall) {
-      const [call] = info.calls;
+      const call = info.calls[0]!;
       const [error, result] = await rpcCalls[0]!;
 
+      if (!error && result?.data instanceof Response) {
```
Nick-Lucas, line 385: So my sense is that we should follow the patterns established by content types for inputs. It would be more consistent and more http-native

1. Have multiple response strategies, JsonResponseType, BlobResponseType, etc
2. Detect the strategy for a given return type
3. Use the strategy to apply headers and such, native http style, check if it supports batching etc
4. On the client read the headers native http and deserialise appropriately
5. Profit

packages/tests/server/___testHelpers.ts
```diff
@@ -117,9 +119,21 @@ export function routerToServerAndClientNew<TRouter extends AnyRouter>(
...
+        condition(op) {
+          // check for context property `skipBatch`
+          return Boolean(op.context['skipBatch']);
```
Nick-Lucas, line 125: This is a neat solution, but I'd love to see another. Maybe we generate new proxy methods when a procedure is a non-json type?

Ie for the new client: `path.to.foo.blobQueryOptions({ })`

packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
```diff
@@ -371,9 +379,27 @@ export async function resolveResponse<TRouter extends AnyRouter>(
...
+        if (!config.experimental?.outputResponse) {
+          throw new Error(
+            'You need to activate outputResponse in the root config to use this feature - e.g. `initTRPC.create({ experimental: { outputResponse: true } })`',
+          );
+        }
+
+        const headers = new Headers(response.headers);
+        headers.set('trpc-response-output', '1');
+
+        return new Response(response.body, {
```
Nick-Lucas, line 396: This seems more inspired by jstack than tRPC

tRPC would say this is a `.output()` call. I don't know if we want to be exposing http inside procedures really? One of the cool things is we/others could build adapters and links which run over entirely different protocols and this would get in the way of that.

Nick-Lucas: Advantage of an output() call is in theory that could be checked for some metadata before the procedure ever gets executed to complain about being part of a batched call

```ts
const blobOutput = {
  parse(),
  trpc: { supportsBatching: false }
}

procedure.output(blobOutput).query()
```

packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
KATT: yes i agree

how do we get there

packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
KATT: i'm unsure what you're suggesting here and how it would look like - what are you saying a procedure should return in this case instead?

packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
KATT: like, i'd want to have e.g. logic for all diffrent type of content-type (inputs) somewhere and all accept (outputs) somewhere where we have vertical slices of both client + server logic (similar to how jsonl/sse is, but more generic)

the `resolveResponse()`-fn is a mess but unsure exactly how to structure this idea

packages/client/src/links/httpBatchLink.ts
```diff
@@ -70,6 +70,9 @@ export function httpBatchLink<TRouter extends AnyRouter>(
             },
             signal,
           });
+          if (res.type === 'response') {
```
Nick-Lucas, line 73: Maybe should be a symbol? Feels like a breaking change otherwise

packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
Nick-Lucas: the way we do request content types is pretty clean no?

packages/server/src/unstable-core-do-not-import/http/contentType.ts
packages/client/src/links/internals/contentTypes.ts

Having an array of supported content types with a known interface (strategy pattern) and matching an appropriate strategy is clean and modular, and could be exposed on the public API for extension one day

packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
Nick-Lucas: Updated my code example above

AhmedBaset:
Any timeline for this?

PR #6262, opened 2024-11-22
chore: `@trpc/upgrade` with queryOptions API codemod

juliusmarminge:
Closes #

## 🎯 Changes

- Blog: https://www-git-query-optoins-codemod-trpc.vercel.app/blog/introducing-tanstack-react-query-client
- Docs: https://www-git-query-optoins-codemod-trpc.vercel.app/docs/client/tanstack-react-query/setup

## ✅ Checklist

- [ ] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [ ] If necessary, I have added documentation related to the changes made.
- [ ] I have added or updated the tests related to the changes made.

## Summary by CodeRabbit

- **New Features**
  - Enhanced integration with TanStack React Query for a more streamlined and type-safe data fetching/mutation experience.
  - Introduced a new CLI upgrade tool to simplify project migrations.

- **Documentation**
  - Launched updated guides covering migration, setup, and usage of the new integration.
  - Refreshed website content and announcements, including new blog posts and author profiles.

- **Chores**
  - Updated dependencies and refactored core code to improve maintainability and overall performance.

PR #6489, opened 2025-02-12
fix: grab call expression deeper than 2 calls

barelyhuman:
moves up the tree to find the actuall call expression and use the correct replace helper

Closes #

## 🎯 Changes

Backward compatible change to make sure the correct call expression is picked in a nested expression. 

Old code picked `utils.x.helper()`  but wouldn't do it for `utils.x.y.z.helper()`, this fixes that

## ✅ Checklist


barelyhuman:
> not able to test now, but does this make it pass on the next-prisma-starter example?? 👀

That's the only one I tested it on, I'm going to mark it as draft so I can test the other examples, will do it post day job today so give me a bit. 

I'm guessing there's a few similar issues so should be ablel to wrap it up in one PR, if they are different however, I'll raise a different PR

packages/upgrade/test/__snapshots__/transforms.test.ts.snap
```diff
@@ -118,7 +118,10 @@ export function Component() {
 
   const mutation = useMutation(
     trpc.post.create.mutationOptions({
-      onSettled: () => queryClient.invalidateQueries(trpc.post.queryFilter()),
+      onSettled: () => {
+        queryClient.invalidateQueries(trpc.post.queryFilter());
+        queryClient.invalidateQueries(trpc.post.list.queryFilter());
+      },
```
juliusmarminge, line 124: pushed up this test which is currently failing:
CleanShot 2025-02-12 at 12 36 11

this is the output when running it on the example:

CleanShot 2025-02-12 at 12 34 22

it doesn't preserve the full path

juliusmarminge (changes requested):

packages/upgrade/test/__snapshots__/transforms.test.ts.snap
barelyhuman: understood, will be handled and i’ll add subsequent tests if needed as well

juliusmarminge: yea this is very testable stuff, so easiest way I found is just to find failing cases by running the codemod on a real codebase (our examples serve as small codebases in this sense 😁) and then just copy the failing cases over to a unit test and iterate until it passes :)

I just don't have a lot of time to do it myself anymore hence the request for community help. Appreciate you taking your time to help out a ton!

juliusmarminge:
> @juliusmarminge Before I add more utils for the transformer, any standard / doc on the folder structure or do I just make a `lib`/`utils` folder?

do what you feel is right!

barelyhuman:
Confirmed, working with all examples, no crashes or warnings 

If there's any additional behavior that needs to be added, do let me know

juliusmarminge (approved):
Awesome! Thank you!

You wanna keep working on the provider transform bugs I mentioned here? https://github.com/trpc/trpc/pull/6262#issuecomment-2651959435

juliusmarminge:
> query though, wouldn't it be nice to have it as a separate PR ?

Yea I'll merge this one and you can open a new one :)

juliusmarminge:
> Awesome! Thank you!
> 
> You wanna keep working on the provider transform bugs I mentioned here? [#6262 (comment)](https://github.com/trpc/trpc/pull/6262#issuecomment-2651959435)

Re this: The `next-see-chat` example is a fairly basic one that I think we should be able to fully migrate by running the CLI, so that's a good test to run and make sure that everything in there transforms correctly and the packages are installed etc etc

PR #6502, opened 2025-02-17
chore: easier call signature

juliusmarminge:
Always annoyed myself that you need to specify the type - also `path` should probably be optional as we want procedure to be more standalone and callable outside of routers?

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/procedureBuilder.test.ts
+++ b/packages/server/src/unstable-core-do-not-import/procedureBuilder.test.ts
@@ -230,6 +230,44 @@ test('inferProcedureBuilderResolverOptions', async () => {
   }
 });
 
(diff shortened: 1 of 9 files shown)


packages/server/src/unstable-core-do-not-import/procedureBuilder.test.ts
```diff
@@ -230,6 +230,44 @@ test('inferProcedureBuilderResolverOptions', async () => {
...
+
+        return {
+          result: '__result' as const,
+          path: opts.ctx.path,
+        };
+      });
+
+    const result = await proc({
+      ctx: {},
+      getRawInput: async () => ({
```
juliusmarminge, line 259: still a bit weird with the async function, but other than that procedures are easily callable as-is now?

PR #6537, opened 2025-02-20
feat: offer an alternative lambda streaming adapter

olup:
Closes #

## 🎯 Changes

Its a feature. It adds a new adapter that work with lambda streaming.

## ✅ Checklist

### DIFF
--- a/packages/server/src/adapters/aws-lambda/global.d.ts
+++ b/packages/server/src/adapters/aws-lambda/global.d.ts
@@ -1,0 +1,24 @@
+import type { Writable } from 'node:stream';
+
+// Those types are made available globally by lambda's runtime
+declare global {
+  namespace awslambda {
+    export namespace HttpResponseStream {
+      function from(writable: Writable, metadata: ANY): Writable;
+    }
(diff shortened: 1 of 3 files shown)


packages/server/src/adapters/aws-lambda/global.d.ts
KATT, line 1: We should avoid adding `.d.ts`-files I think as it pollutes the global scope

I think you can do something like this locally to the file where it's needed

packages/server/src/adapters/aws-lambda/index.ts
```diff
@@ -82,3 +83,58 @@ export function awsLambdaRequestHandler<
     return await planner.toResult(response);
   };
 }
+
+export function awsLambdaStreamingRequestHandler<
```
KATT, line 87: Nice

zirkelc:
Hey @olup I implemented the same adapter in PR #6680 before I noticed that you already had a draft :-D

Please have a look there, maybe we can merge parts of it together.

PR #6547, opened 2025-02-22
fix(wsLink): wsLink should await for connectionParams before sending …

hmatthieu:
## 🎯 Changes

Added a test case to ensure wsLink awaits urlConnectionParams before sending any messages. urlConnectionParams resolution is now handled at the wsConnection level. The wsClient can use the WebSocket once the connection parameters are sent.

## ✅ Checklist

## Summary by CodeRabbit

- **Refactor**
	- Streamlined the WebSocket connection initialization process by simplifying handshake operations and improving asynchronous handling for more reliable connection setups.
	- Removed unused properties and methods related to connection message building in the WebSocket client.

- **Tests**
	- Introduced two new tests for asynchronous authentication to ensure that delayed credential retrieval integrates seamlessly with connection establishment and reconnections.


TomKaltz:
@hmatthieu this may have introduced a regression, ping timeout/reconnect seems to have stopped working with this PR.

hmatthieu:
> @hmatthieu this may have introduced a regression, ping timeout/reconnect seems to have stopped working with this PR.

While the connectionParams async function resolves, reconnection will indeed not be active. But IMO that's fine — we don't want to retry if we haven't even finished our previous attempt, right?

Or are you talking about something else ?

TomKaltz:
> @hmatthieu this may have introduced a regression, ping timeout/reconnect seems to have stopped working with this PR.

Yes, upon a loss in network the ping never times out to close the current connection.  When the network comes back up a new connection is forced but connectionParams are not sent on the new connection, resulting in authentication errors.

TomKaltz:
My other concern is that the connection is not being forcefully closed upon ping timeout like before.  Therefor client code is not notified of the loss in connection until the network comes back which is unintuitive.

hmatthieu:
@TomKaltz After further investigation:

Preliminary notes:

Previous wsLink behavior regarding keep-alive:
When wsLink didn't receive a pong from the server, it closed the connection and dispatched an onClose event. The app could react to this onClose event, but the connection was not automatically reopened.

After refactoring:
The behavior remains the same, but the connection now automatically reopens (which is IMO a bugfix). On my side, onClose followed by onOpen events are dispatched, allowing the app to react accordingly.

I couldn't reproduce your latest concern. I'll push my latest commit to this PR and would appreciate your feedback.

TomKaltz:
I can confirm that proper reconnect upon network restore is now working.  What is not working optimally IMHO is that the original connection is not forcefully closed when a keepaline ping times out.  Due to this, the connection just hangs open until the network connection returns, then the original connection is finally closed and a new one is opened.  I'm using chrome devtools network tab "offline" throttling to simulate a dropped connection.

KATT (approved):

PR #6563, opened 2025-02-24
DRAFT/STRAWMAN/not currently intended for merging: add trpc.foo.useQuery back?

mmkal:
Related to https://github.com/trpc/trpc/discussions/6240 - CC @juliusmarminge 

This PR isn't really a "pull" request - just as a place to allow discussion with commenting on code inline etc. if it's helpful.

Since @KATT suggested something to fill the gap between the old react-query API and the new one _might_ be welcome, I thought I'd just poke around and see what the dumbest possible implementation of it might look like.

I want to be clear though that I'm aware I _wasn't_ involved in the discussions around the new API and what the specific pitfalls of the old one were, so I am very sorry if this is just prompting people to repeat themselves!

## 🎯 Changes

For now, this just adds a `useQuery(...)` method to the `DecorateQueryProcedure` interface. Something similar would be needed for `useMutation` and `useSubscription` and `useInfiniteQuery`, but I didn't bother with those because I wanted to sanity check/find out pitfalls first.

DX for consumers wise, I think it brings back pretty much the usage pattern as before, along with all the shortcomings about new react-query features and so on.

Maintenance wise:

Code: it doesn't seem bad? I haven't been in the weeds in trpc for a while so I don't remember which helpers came into existence and when, but all I really did was copy-paste the `TRPCQueryOptions` interface into a `TRPCUseQueryResult` interface, with identical generics, but made the return types of each overload `UseQueryResult<TData, TRPCClientErrorLike...>`. The runtime implementation was trivial - just do what we're already doing for `queryOptions` and pass it to a `useQuery` imported from `@tanstack/react-query`

Of course even if this approach, there's more to consider than just the code. Test and docs would of course be needed, and decisions needed about the docs. trpc would now need to decide when it promotes which variant. My intuition says to promote `trpc.greeting.useQuery()` by default, and suggest `useQuery(trpc.greeting.queryOptions())` as an escape hatch, for when you need to override one of the options, use a new react-query feature, or use a different react-query version.

Also, it's hard to predict what'll happen in future. But, if react-query came out with a new version that renamed `queryFn` to `queryFunction` or something, I think it'd be simple for users to do something like `"react-query-v123": "npm:@tanstack/react-query@^123.0.0"` in their package.json, then something like this in their code:

```tsx
import {trpc} from '~/wherever'
import {useQuery} from 'react-query-v123'

export default function Foo() {
  const hello = trpc.hello.useQuery()
  
  const {queryFn, ...goodbyeOptions} = trpc.goodbye.queryOptions()
  const goodbye = useQuery({...goodbyeOptions, queryFunction: queryFn})

  return <>{hello.data} {goodbye.data}</>
}
```

### DIFF
--- a/packages/tanstack-react-query/src/internals/createOptionsProxy.ts
+++ b/packages/tanstack-react-query/src/internals/createOptionsProxy.ts
@@ -1,4 +1,4 @@
-import type { DataTag, QueryClient, QueryFilters } from '@tanstack/react-query';
+import { useQuery, type DataTag, type QueryClient, type QueryFilters } from '@tanstack/react-query';
 import type { TRPCClient, TRPCRequestOptions } from '@trpc/client';
 import { getUntypedClient, TRPCUntypedClient } from '@trpc/client';
 import type {
@@ -22,6 +22,7 @@ import {
   trpcMutationOptions,
   type TRPCMutationOptions,
 } from './mutationOptions';
+import type { TRPCUseQueryResult} from './queryOptions';
 import { trpcQueryOptions, type TRPCQueryOptions } from './queryOptions';
 import {
   trpcSubscriptionOptions,
@@ -86,6 +87,8 @@ export interface DecorateQueryProcedure<TDef extends ResolverDef> {
    * @see https://trpc.io/docs/client/tanstack-react-query/usage#queryOptions
    */
   queryOptions: TRPCQueryOptions<TDef>;
+
+  useQuery: TRPCUseQueryResult<TDef>;
 
   /**
    * Create a set of type-safe infinite query options that can be passed to `useInfiniteQuery`, `prefetchInfiniteQuery` etc.
@@ -318,6 +321,15 @@ export function createTRPCOptionsProxy<TRouter extends AnyTRPCRouter>(
           queryKey: getQueryKey(),
           query: callIt('query'),
         });
+      },
+      useQuery: () => {
+        return useQuery(trpcQueryOptions({
+          opts: arg2,
+          path,
+          queryClient: opts.queryClient,
+          queryKey: getQueryKey(),
+          query: callIt('query'),
+        }));
       },
       mutationOptions: () => {
         return trpcMutationOptions({
--- a/examples/minimal-react/client/src/Greeting.tsx
+++ b/examples/minimal-react/client/src/Greeting.tsx
@@ -3,6 +3,15 @@ import { trpc } from './utils/trpc';
 
 export function Greeting() {
   const greeting = useQuery(trpc.greeting.queryOptions({ name: 'tRPC user' }));
+  const greeting2 = trpc.greeting.useQuery({name: 'tRPC user'});
+  const greeting3 = trpc.greeting.useQuery();
+  const greeting4 = trpc.greeting.useQuery({ name: 'tRPC user' }, {enabled: true});
+  const greeting5 = trpc.greeting.useQuery(undefined, {enabled: true});
 
-  return <div>{greeting.data?.text}</div>;
+  greeting2.data satisfies typeof greeting.data;
+  greeting3.data satisfies typeof greeting.data;
+  greeting4.data satisfies typeof greeting.data;
+  greeting5.data satisfies typeof greeting.data;
+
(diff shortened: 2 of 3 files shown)


packages/tanstack-react-query/src/internals/createOptionsProxy.ts
```diff
@@ -87,6 +93,8 @@ export interface DecorateQueryProcedure<TDef extends ResolverDef> {
    */
   queryOptions: TRPCQueryOptions<TDef>;
 
+  useQuery: TRPCUseQueryResult<TDef>;
```
juliusmarminge, line 96: We could flag this so it's not exposed by default, and people coming from the old API could enable to have a more familiar API.

Something like:

```ts
createTRPCOptionsProxy<AppRouter, {
  enableOldHooks: true
}>()
```

juliusmarminge (commented):
Thanks for the PR, good way to start a more formal discussion! 

The new structure of the new package allows for introducing this without much overhead so maintenance wise this would be fine.

One problem is we'd need to reintroduce multiple init methods in order to not expose react hooks for server callers for example?

> My intuition says to promote trpc.greeting.useQuery() by default, and suggest useQuery(trpc.greeting.queryOptions()) as an escape hatch

My main problem with this is you're leading developers down a "bad path" since they would not benefit from eslint rules nor the compiler, two aspects that this new API that isnt talked about

We'll definitely discuss this, both internally and on this PR though.

Nick-Lucas:
Just to echo what Julius is saying, the old integration doesn't look like it's going to age well when React Compiler comes out, and there's a long-standing issue where eslint's hooks rules don't lint it because the rules don't consider our proxy to be valid. The classic client permits (and in the case of our polymorphism work actively encourages) patterns which break the rules of hooks, so will not compile in the future, and we should have emphasised that in the announcement post.

I would love to find a way to support custom helpers where teams and codebases want these sorts of wrappers (at their own peril) but I personally think the core surface area of the TRQ client should avoid anything that could become a foot gun in the wrong hands

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
```diff
@@ -1,4 +1,9 @@
-import type { DataTag, QueryClient, QueryFilters } from '@tanstack/react-query';
+import {
+  useQuery,
```
juliusmarminge, line 2: This might be rejected by Next.js if using server components like this: https://trpc.io/docs/client/tanstack-react-query/server-components

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
mmkal: Yeah I guess, but as you pointed out in https://github.com/trpc/trpc/discussions/6508#discussioncomment-12280845 there is already away to get the old API, if it is indeed considered "old". My hope with this is to figure out if we can keep around the old API for most use cases indefinitely, because it's better DX. But if that were the direction making it opt in and using the label "old" wouldn't make sense.

If we'd just be adding it to help people migrate extra slowly, then yeah that could be sensible.

mmkal:
Re compiler and eslint: this is scope creep, but I just pushed a change which wraps `rules-of-hooks` in the eslint config to enable it for `trpc.foo.useQuery()` expressions, along with an example of a violation in Greeting.tsx:

image

It's a hack, which depends on the implementation of [RuleOfHooks.ts](https://github.com/facebook/react/blob/ebc22ef7e15bf38dc91b7033782cedc2f43f7d6e/packages/eslint-plugin-react-hooks/src/RulesOfHooks.ts), clearly a better solution would be an actual resolution to https://github.com/facebook/react/issues/25065. But, I also was able to make a very small modification to `eslint-plugin-react-hooks` which would allow for resolving that bug, either in a configurable way or not, so maybe the react team would accept a pull request.

I don't know about React Compiler, but maybe something similar would be possible?

Nick-Lucas:
> I don't know about React Compiler, but maybe something similar would be possible?

So you can get an early view on what's acceptable with https://www.npmjs.com/package/eslint-plugin-react-compiler - it's been a while since I tried it but my work codebase uses tRPC's polymorphism and because it involves passing hooks through props it unfortunately rejects that code. Most tRPC code might go unjected though I'm not sure if anything would de-opt under the hood because of the chained hooks

PR #6587, opened 2025-03-07
feat(server): simpler `next()` calls in middlewares

KATT:
Closes #

## 🎯 Changes

No more `ctx: {}` on `.next()` calls in middlewares.

**Before**

```ts
return opts.next({
  ctx: {
    x:  'y'
  }
})
```
**After**

```ts
return opts.next({
  x: 'y'
})
```

### Why

Cleaner.

We don't have any other things that are ever passed on `.next({ })`. I expected us to have more properties that weren't only internal, but we never did.

## ✅ Checklist

- [x] It is backwards compatible

## Summary by CodeRabbit

- **Refactor**
	- Streamlined response structures across authentication, middleware, and procedure calls by returning key data (e.g., session and user) directly instead of nested within additional objects.
	- Modified input handling in middleware to simplify context overrides and enhance clarity.

- **Documentation**
	- Updated migration guides and usage documentation to reflect the new, simplified syntax for middleware calls and context handling.

- **Tests**
	- Revised test cases to align with the new procedure signatures and input definitions, eliminating legacy middleware patterns.
	- Added new test cases to validate the removal of unnecessary context wrappers and ensure proper error handling.


packages/tests/server/regression/issue-4527-nested-middleware-root-context.test.ts
KATT, line 1: I don't see why we'd want to override all of these as part of the middlewares

packages/tests/server/getRawInput.test.ts
```diff
@@ -1,43 +1,12 @@
...
 import type { TRPCError } from '@trpc/server';
 import { initTRPC } from '@trpc/server';
-
-test('middleware swap', async () => {
-  const t = initTRPC.create();
-  const router = t.router({
-    test: t.procedure
-      .use((opts) =>
-        opts.next({
-          getRawInput: () => Promise.resolve('foo'),
```
KATT, line 11: I don't see why one would need to do this

packages/server/src/unstable-core-do-not-import/middleware.ts
```diff
@@ -102,14 +118,57 @@ export type MiddlewareFunction<
...
+       */
+      <$ContextOverride>(
+        opts:
+          | {
+              ctx: $ContextOverride;
+            }
+          | {
+              /**
+               * @deprecated
+               * The ability to override the `input`-key will be removed in v12 - please open an issue if you need this feature
```
Nick-Lucas, line 155: will it be removed? isn't input overriding a valid tool in tRPC?

packages/server/src/unstable-core-do-not-import/middleware.ts
```diff
@@ -203,7 +262,7 @@ export function createInputMiddleware<TInput>(parse: ParseFn<TInput>) {
             }
           : parsedInput;
 
-      return opts.next({ input: combinedInput });
+      return opts.next({ [nextInputSymbol]: combinedInput });
```
Nick-Lucas, line 265: So that was my question as I thought createInputMiddleware used this.

I'm not entirely convinced there's a solid reason to take away input overriding from userland, though I'm also not sure if it's used by anybody for anything useful

Nick-Lucas:
So this is fine, I am a sceptic about removing input overriding from the public API, and I do wonder if (while cleaner) this is really a step further toward an "obvious" API for users, but there implementation looks decent

packages/server/src/unstable-core-do-not-import/middleware.ts
KATT: I can't think of a usecase where it's needed

Less is more, reducing api surface area feels like a good thing

PR #6634, opened 2025-03-25
fix: recognize `effect` standard-schema parsers 3.14.2

mmkal:
As of a few minutes ago, the [docs for using `effect`](https://trpc.io/docs/server/validators#with-effect) on trpc.io are broken!

This is because of https://github.com/Effect-TS/effect/pull/4648:

the `Schema.standardSchemaV1(...)` function now returns a class with a static `~standard` property, which means `getParserFn` treats it as a function, rather than a standard-schema compliant struct.

Closes # (no issue yet, too new and I thought I'd just submit a fix)

This is all kind of my fault, the change to effect went in because I'm trying to add effect support to trpc-cli: https://github.com/mmkal/trpc-cli/pull/63

## 🎯 Changes

- First commit fdacff8fdd864f2afd27295f1c79c7ce74cbc629: add failing test case with `effect@3.14.2`. See failure here: https://github.com/trpc/trpc/pull/6634#issuecomment-2752094491
- Second commit 789c99948726786c33e8191c51cd020119d3bc90: fix by checking if the parser looks like a standard-schema struct, even if it's a function

## ⏭️ Next

For the next major change, I think `getParseFn` should be simplified significantly, maybe to allow only two modes: standard-schema and raw function. Something like this:

```ts
export function getParseFn<TType>(procedureParser: Parser): ParseFn<TType> {
  const parser = procedureParser as any;

  if ('~standard' in parser) {
    // StandardSchemaEsque
    return async (value) => {
      const result = await parser['~standard'].validate(value);
      if (result.issues) {
        throw new StandardSchemaV1Error(result.issues);
      }
      return result.value;
    };
  }

  if (typeof parser === 'function') {
    // ParserCustomValidatorEsque
    return parser;
  }

  throw new Error('Could not find a validator fn');
}
```

This would cover all popular validation libraries, and put the onus on new ones to conform to standard-schema if they want to work well with trpc. Or, a less aggressively "breaking" version where we keep all the existing checks for common validation libraries (including old versions etc.) and just put in the _preference_ for standard-schema. Even that shouldn't happen without a major bump though because it'll change error formatting for anyone who uses zod, which is just too disruptive.

Note that this isn't the first time a problem like this has arisen, arktype also has parsers that are functions with properties attached: https://github.com/trpc/trpc/pull/5958

## ✅ Checklist

## Summary by CodeRabbit

- **Chores**
  - Introduced a new dependency to enhance underlying functionality.
  
- **Tests**
  - Added a new test to validate input processing and error handling with the new dependency.


mmkal:
Failed successfully:

image

Run: https://github.com/trpc/trpc/actions/runs/14066659207/job/39390772887

Will push the fix now.

juliusmarminge (approved):

PR #6640, opened 2025-03-27
fix(client): Websocket not reconnecting when lazy is enabled 

bnussman:
## 🎯 Changes

- Fixes issue where websocket won't reconnect when it is closed unexpectedly when lazy mode is enabled

## Reasoning 💭 

If your tRPC server goes down and you have a client that uses lazy mode, it won't attempt to reconnect. Clients that use lazy mode should be able to gracefully reconnect.

## ✅ Checklist

### DIFF
--- a/packages/client/src/links/wsLink/wsClient/wsClient.ts
+++ b/packages/client/src/links/wsLink/wsClient/wsClient.ts
@@ -330,14 +330,12 @@ export class WsClient {
       handleCloseOrError(event);
       this.callbacks.onClose?.(event);
 
-      if (!this.lazyMode) {
-        this.reconnect(
-          new TRPCWebSocketClosedError({
-            message: 'WebSocket closed',
-            cause: event,
-          }),
-        );
-      }
+      this.reconnect(
+        new TRPCWebSocketClosedError({
+          message: 'WebSocket closed',
+          cause: event,
+        }),
+      );
     });
 
     ws.addEventListener('error', (event) => {


packages/client/src/links/wsLink/wsClient/wsClient.ts
```diff
@@ -330,14 +330,12 @@ export class WsClient {
       handleCloseOrError(event);
       this.callbacks.onClose?.(event);
 
-      if (!this.lazyMode) {
```
bnussman, line 333: @hmatthieu Do you think this will have any unintended side-effects?

packages/client/src/links/wsLink/wsClient/wsClient.ts
```diff
@@ -330,7 +330,10 @@ export class WsClient {
...
+      if (
+        !this.lazyMode ||
+        (this.lazyMode && this.requestManager.hasPendingRequests())
+      ) {
         this.reconnect(
```
hmatthieu, line 337: Isn't this equivalent to 

```suggestion
      if (
        !this.lazyMode || this.requestManager.hasPendingRequests())
      ) {
        this.reconnect(
``` ?

File: packages/client/src/links/wsLink/wsClient/wsClient.ts [hunk 3]
```diff
@@ -330,7 +330,10 @@ export class WsClient {
       handleCloseOrError(event);
       this.callbacks.onClose?.(event);
 
-      if (!this.lazyMode) {
+      if (
+        !this.lazyMode ||
+        (this.lazyMode && this.requestManager.hasPendingRequests())
```
hmatthieu, line 335: In lazy mode, I believe we wait for the next request from the client before reconnecting. You're right, though—active subscriptions are a different case. That said, I don’t think we want this same behavior for other pending requests, like queries or mutations.

```suggestion
        !this.lazyMode || /* something like: this.requestManager.hasPendingSubscriptions() */
```

At first, I thought we might want to generalize this behavior to *any* pending request, but now I’m not so sure. If that were the case, we’d probably want to write a test like this for a simple query:

**Test: A pending query triggers a reconnect**

1. The client opens a connection to the server  
2. The client sends a query, but the server doesn't respond  
3. The connection is destroyed  
4. The client reconnects and resends the same query?

But while writing this, I realized we probably *don’t* want that behavior. For queries (or mutations), if the connection drops, we generally treat the request as failed and expect the client to initiate a new request to trigger a lazy reconnect.


WDYT ?

packages/client/src/links/wsLink/wsClient/wsClient.ts
bnussman: Yes, it is. My bad! 

Updated the PR to reflect this

packages/client/src/links/wsLink/wsClient/wsClient.ts
bnussman: Agreed. I don't think we want this behavior for queries/mutations. 

Added `hasPendingSubscriptions` in 366a5e3

KATT (approved):

PR #6661, opened 2025-04-02
Make opts optional in trpcSubscriptionOptions.

### LINKED ISSUE #6660: bug: subscriptionOptions causes a crash without second parameter

```
  System:
    OS: macOS 15.3.2
    CPU: (14) arm64 Apple M4 Pro
    Memory: 2.83 GB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.14.0 - ~/.local/state/fnm_multishells/57632_1743508339299/bin/node
    npm: 10.9.2 - ~/.local/state/fnm_multishells/57632_1743508339299/bin/npm
    pnpm: 10.6.2 - ~/.local/state/fnm_multishells/57632_1743508339299/bin/pnpm
    bun: 1.2.7 - /run/current-system/sw/bin/bun
  Browsers:
    Chrome: 134.0.6998.166
    Safari: 18.3.1
  npmPackages:
    @tanstack/react-query: 5.71.3 => 5.71.3
    @trpc/client: ^11.0.1 => 11.0.1
    @trpc/server: ^11.0.1 => 11.0.1
    next: ^15.2.4 => 15.2.4
    react: ^19.0.0 => 19.1.0
    typescript: ^5.8.2 => 5.8.2
```

### Describe the bug

this code

```mjs
/**
 * @internal
 */ const trpcSubscriptionOptions = (args)=>{
    const { subscribe, path, queryKey, opts } = args;
    const input = queryKey[1]?.input;
    const enabled = 'enabled' in opts ? !!opts.enabled : input !== skipToken;
```

triggers 

`Error: opts is not an Object. (evaluating ''enabled' in opts')`

when subscriptionOptions is called with only one parameter, like

```ts
trpc.sseSetupEvents.subscriptionOptions({ appId })
```

changing this line to

```ts

trpc.sseSetupEvents.subscriptionOptions({ appId }, {})
```

fixes the crash, but the types definitely indicate that this second param should be optional. so this looks like a bug to me.

### Link to reproduction

the minimal starter doesn't use app router so I don't think I can easily set up repro with that link

### To reproduce

```ts
trpc.sseSetupEvents.subscriptionOptions({ appId })
```

### Additional information

—

### 👨‍👧‍👦 Contributing

yspreen:
Closes #6660

## 🎯 Changes

Make opts optional in trpcSubscriptionOptions.

## ✅ Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced subscription flexibility by making options parameter optional, preventing potential runtime errors when options are absent.

### DIFF
--- a/packages/tanstack-react-query/src/internals/subscriptionOptions.ts
+++ b/packages/tanstack-react-query/src/internals/subscriptionOptions.ts
@@ -125,7 +125,7 @@ export const trpcSubscriptionOptions = (args: {
   subscribe: typeof TRPCUntypedClient.prototype.subscription;
   path: readonly string[];
   queryKey: TRPCQueryKey;
-  opts: AnyTRPCSubscriptionOptionsIn;
+  opts?: AnyTRPCSubscriptionOptionsIn;
 }): AnyTRPCSubscriptionOptionsOut => {
   const { subscribe, path, queryKey, opts = {} } = args;
   const input = queryKey[1]?.input;
--- a/packages/tanstack-react-query/src/internals/subscriptionOptions.ts
+++ b/packages/tanstack-react-query/src/internals/subscriptionOptions.ts
@@ -127,7 +127,7 @@ export const trpcSubscriptionOptions = (args: {
   queryKey: TRPCQueryKey;
   opts: AnyTRPCSubscriptionOptionsIn;
 }): AnyTRPCSubscriptionOptionsOut => {
-  const { subscribe, path, queryKey, opts } = args;
+  const { subscribe, path, queryKey, opts = {} } = args;
   const input = queryKey[1]?.input;
   const enabled = 'enabled' in opts ? !!opts.enabled : input !== skipToken;
 


juliusmarminge (commented):
We could do this but I'm not quite sure when you wouldn't want to pass the second arg?

yspreen:
@juliusmarminge currently the types are broken because the second parameter is market as optional, so it shouldn't cause a crash to leave it out.

I also believe it's in the spirit of this library to have empty parameter objects be empty. 

```ts
trpc.someSubscription.subscriptionOptions({ myParam: "some value" }, {})
```

adds nothing over

```ts
trpc.someSubscription.subscriptionOptions({ myParam: "some value" })
```

juliusmarminge:
> currently the types are broken because the second parameter is market as optional, so it shouldn't cause a crash to leave it out.

oh I misunderstood then, sorry.

Still curious when you wouldn't want to pass any of the callback arguments, so open to discuss whether the bug is actually that the type allows for optional, and not that we don't handle it

yspreen:
@juliusmarminge 

the callbacks are one way to handle subscription data. you can also use
```ts
const { data, error } = trpc.someSubscription.subscriptionOptions({ myParam: "some value" })
```

so this is a perfectly valid use case. I don't see what needs to be passed here.

juliusmarminge (approved):

PR #6689, opened 2025-04-09
fix(client): sseStreamConsumer to use instance's properties instead of global

goknsh:
## 🎯 Changes

Fixes `sseStreamConsumer` to use the `eventSource` instance's properties created from the passed in `opts.EventSource` instead of the global `EventSource`, which is not available in environments like React Native.

## ✅ Checklist

## Summary by CodeRabbit

- **Chores**
	- Updated internal error handling for streaming events to refine the management of error scenarios. This update is purely under-the-hood and does not alter any public-facing functionality.


packages/server/src/unstable-core-do-not-import/stream/sse.ts
```diff
@@ -346,7 +346,7 @@ export function sseStreamConsumer<TConfig extends ConsumerConfig>(
           _es = null;
         });
         eventSource.addEventListener('error', (event) => {
-          if (eventSource.readyState === EventSource.CLOSED) {
+          if (eventSource.readyState === eventSource.CLOSED) {
```
KATT, line 349:
```suggestion
          if (eventSource.readyState === opts.EventSource.CLOSED) {
```

KATT (dismissed):
tweaked it - we should be using the passed in ponyfill instead of the global indeed

packages/server/src/unstable-core-do-not-import/stream/sse.ts
goknsh: Can't make this change, results in a type error. The types have `CLOSED` existing on the instance only. https://github.com/trpc/trpc/blob/9f1e419518a121a77c2437a986d75aea265e467a/packages/server/src/unstable-core-do-not-import/stream/sse.types.ts#L24-L27

KATT: Test it with `@canary`, I think this should work

KATT (approved):

PR #6694, opened 2025-04-10
feat(client): add `retryDelayMs`-option to `retryLink`

rynobax:
Addresses https://github.com/trpc/trpc/discussions/6425

## 🎯 Changes

Add a `retryDelay` option to RetryLink.  Right now it defaults to immediately retrying on failure, this allows introducing a delay.

IMO this should have a reasonable default (what i provided in the docs is the tanstack query default).  Since that would be a breaking change, I think it makes sense to do it in a follow up pr.

## ✅ Checklist

## Summary by CodeRabbit

- **New Features**
  - Enabled customizable retry delays between attempts, allowing users to define specific wait times based on the retry iteration.
  - Introduced a new function for simulating operation links that can fail a specified number of times before succeeding.

- **Bug Fixes**
  - Enhanced testing capabilities for the retry mechanism with additional test cases covering various scenarios.

- **Documentation**
  - Updated help materials to guide users on configuring the new customizable retry delay option.

### DIFF
--- a/packages/client/src/links/retryLink.ts
+++ b/packages/client/src/links/retryLink.ts
@@ -15,7 +15,7 @@ interface RetryLinkOptions<TInferrable extends InferrableClientTypes> {
   /**
    * The delay between retries in ms (defaults to 0)
    */
-  retryDelay?: (attempt: number) => number;
+  retryDelayMs?: (attempt: number) => number;
 }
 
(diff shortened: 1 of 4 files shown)


KATT (approved):
Added tests and fixed an edge-case when unsubscribing during the delay

PR #6806, opened 2025-06-04
feat(client): `localLink`

### LINKED ISSUE #3768: feat: add localLink that runs the router in the same process as the client

Hi there!

It would be great to have a `LocalLink` that allows me to do:
This would directly run the router in the same process as the client instead of transferring requests to a server.

### But why??

This is motivated mainly due to testing and fast prototyping:
1. This allows to test the client with the server code (with a small caveat, see below)
2. I can start writing a fully functional client that saves data in local storage instead of a database, and when I want to add a database, I don't need to rewrite the procedures, giving me significantly faster prototyping speed.

```ts
import { createRouter } from '../path/to/server/trpc';
const client = createTRPCProxyClient<ReturnType<typeof createRouter>>({
  links: [
    localLink({ router: createRouter({ database }) }),
  ],
});```

For the feature itself it doesn't matter that the router is created by a function call, but this solves a problem I mentioned above: What if part of the code only runs on the server, e.g. database access? Well, just inject that dependency, and, in case you are running on the client, use a browser-based database.

Not sure, to be honest.

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

KATT:
Closes #3768

## 🎯 Changes

Adds a `localLink`

## ✅ Checklist

### DIFF
--- a/packages/client/src/internals/signals.ts
+++ b/packages/client/src/internals/signals.ts
@@ -52,3 +52,19 @@ export function raceAbortSignals(
 
   return ac.signal;
 }
+
+export function abortSignalToPromise(signal: AbortSignal): Promise<never> {
+  return new Promise((_, reject) => {
+    if (signal.aborted) {
+      reject(signal.reason);
+      return;
+    }
+    signal.addEventListener(
+      'abort',
+      () => {
+        reject(signal.reason);
+      },
+      { once: true },
+    );
+  });
+}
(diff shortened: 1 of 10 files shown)


KATT:
@coderabbitai full review

CarlosZiegler:
Looks interesting, some real example how/ why use that ?

PR #6827, opened 2025-06-15
Update awesome-trpc.mdx to include Deep Fish

noahgsolomon:
including my own project deep fish as a project which is powered by tRPC


noahgsolomon (commented):
it's not open sourced right now we plan on open sourcing in august once v1 is out

PR #6828, opened 2025-06-16
fix(`server`): content-type on streamed responses should be `application/jsonl`

### LINKED ISSUE #6819: bug: `httpBatchStreamLink` responses use wrong `content-type` header (`application/json`)

```
  System:
    OS: macOS 15.5
    CPU: (12) arm64 Apple M4 Pro
    Memory: 4.10 GB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
  Browsers:
    Chrome: 137.0.7151.69
    Safari: 18.5
```

### Describe the bug

Image

### Link to reproduction

### To reproduce

### Additional information

Image

### 👨‍👧‍👦 Contributing

KATT:
Closes #6819

## 🎯 Changes

content-type on streamed responses should be `application/jsonl`

## ✅ Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected the response header to dynamically match the requested content-type for batch streaming responses.
- **New Features**
  - Enabled specifying the accept header via URL query parameters for HTTP requests.
- **Tests**
  - Added a test to verify that the content-type header is accurately set to 'application/jsonl' for applicable batch responses.


packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
```diff
@@ -504,7 +504,7 @@ export async function resolveResponse<TRouter extends AnyRouter>(
     // batch response handlers
     if (info.accept === 'application/jsonl') {
       // httpBatchStreamLink
-      headers.set('content-type', 'application/json');
+      headers.set('content-type', 'text/event-stream+json');
```
Nick-Lucas: Should it be `application/stream+json`?

Was just trying to look it up out of interest and only found this:

https://stackoverflow.com/questions/52098863/whats-the-difference-between-text-event-stream-and-application-streamjson

BeBoRE: https://en.wikipedia.org/wiki/JSON_streaming
Seems like the only formalized format is 'application/json-seq', but that uses record seperator-delimited JSON instead of newline seperated like jsonl.

That Spring issue seems to have settled on using 'application/x-ndjson'.

https://github.com/spring-projects/spring-framework/issues/21283#issuecomment-665114578

KATT: Nice. Going to try these and see if the network tab in chrome displays them nicely

KATT: None of `application/*` looks nice in the output in chrome inspector

Might look at refactoring so it actually gives back an `text/event-stream` because it looks good 😅

KATT: `text/event-stream` seems to be the one that gives the nicest output (tested in #6846)


CleanShot 2025-06-28 at 16 16 46@2x

KATT: but i won't care about that for now

PR #6853, opened 2025-07-02
add new extension row to extensions table

dugajean:
## 🎯 Changes

Add my library to the extensions docs under Awesome tRPC Collection

## ✅ Checklist

### DIFF
--- a/www/docs/community/awesome-trpc.mdx
+++ b/www/docs/community/awesome-trpc.mdx
@@ -28,6 +28,7 @@ A collection of resources on tRPC.
 | msw-trpc - tRPC support for MSW                                               | https://github.com/maloguertin/msw-trpc              |
 | trpc-cli - Turn a tRPC router into a type-safe, fully documented CLI          | https://github.com/mmkal/trpc-cli                    |
 | trpc-live - Live query solution for tRPC                                      | https://github.com/strblr/trpc-live                  |
+| trpc-formdata - Easily create FormData procedures that support file uploads   | https://github.com/dugajean/trpc-formdata            |
 
 ### Frontend frameworks
 


KATT:
We support formdata with the httpLink (https://trpc.io/docs/server/non-json-content-types / https://github.com/trpc/examples-next-formdata)

How is this different? I know our documentation is a bit bad here 😬

flythenimbus:
@KATT I'm just wrapping that so the user doesn't have to do so much plumbing themselves. I essentially add a fetch wrapper to remove the JSON content type header so that the multipart formdata header is set by the browser with the correct boundary. I do this all within a new link called `formDataLink` which is a wrapper of `httpLink` with the custom fetch.

flythenimbus:
@KATT Can we please get this merged? I believe it's quite cumbersome to get FormData acceptance implemented by hand each time. The library is well implemented and documented. Feel free to give it a review, it's pretty small.

juliusmarminge:
I'm with @KATT on this one. We already support this OOTB and leading users down a 3rd party library might be confusing. We should instead improve the docs and/or make adjustments to core to make this easier.

Can you elaborate on what's cumbersome with this examples referenced from this doc page: https://trpc.io/docs/server/non-json-content-types

- client: https://github.com/juliusmarminge/trpc-interop/blob/66aa760141030ffc421cae1a3bda9b5f1ab340b6/src/client.tsx#L10-L18
- server: https://github.com/juliusmarminge/trpc-interop/blob/66aa760141030ffc421cae1a3bda9b5f1ab340b6/src/server.ts#L28-L41

_(it might be even easier than that with zod 4 having some natively supported file schemas now)_

Issue #6870, opened 2025-07-22
feat:  native support for AsyncLocalStorage NodeJS within trpc context


ARAldhafeeri:
### Describe the feature you'd like to request

AsyncLocalStorage provides stores that remain coherent through asynchronous operations, perfectly complementing tRPC's focus on developer experience. It's a powerful tool that can enhance context management in tRPC applications. Below are different approaches to integrating it.

### **Native Implementation**  
`initTRPC.context.create` allows  **optionally** to enable async storage on the context. Which will enable the asyncStorage implicity and the user can interact with that storage anywhere  down the async request chain.
 
```typescript
const t = initTRPC.context<Context>().create({
  asyncStorage: new AsyncLocalStorage<Context>(),
});

// Context automatically propagates everywhere
```

**Pros:**  
✅ Cleaner, more intuitive API  
✅ No boilerplate in middleware  
✅ Harder to misuse
✅ Official support means better docs & type safety  

**Cons:**  
❌ Requires changes to tRPC core  
❌ Increased maintenance burden  
❌ Less flexibility for custom implementations  

###  Manual Implementation
user simply create async storage within the context and create middleware to run the storage explicitly

```typescript
const asyncStorage = new AsyncLocalStorage<Context>();
const t = initTRPC.context<{ 
  original: Context;  // Original context 
  storage: typeof asyncStorage 
}>().create();

// Middleware must explicitly run storage:
t.procedure.use(async (opts) => {
  return asyncStorage.run(opts.ctx.original, () => opts.next());
});
```

**Pros:**  
✅ Works today (no core changes needed)  
✅ Full control over storage behavior  
✅ Can be incrementally adopted  

**Cons:**  
⚠️ Boilerplate in every middleware  
⚠️ Less discoverable (users must know to set it up)  
⚠️ Easy to misuse (e.g., forgetting `run()`)  
⚠️ Pollutes context shape with `original` and `storage`  

Where this can be implemented ? 
- probably within `create` method in initTRPCBuilder 

### Additional information

I have also developed a library a while ago that provides excellent async context ulitlities maybe I can implement something like that in trpc : 
https://www.npmjs.com/package/node-requext

### 👨‍👧‍👦 Contributing

- [x] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

KATT:
I'm a bit confused on what you're suggesting or asking for here exactly

ARAldhafeeri:
Hello Katt,

[...]

PR #6881, opened 2025-08-07
 new way to integrate AsyncLocalStorage 

ARAldhafeeri:
This PR addresses the feature request here https://github.com/trpc/trpc/issues/6870

## Problem Statement

Currently, integrating AsyncLocalStorage with tRPC requires:

1. Manual context passing
2. Global procedure middleware wrapping

This creates unnecessary boilerplate for a common need.

## Solution

Add optional `asyncStorage` to runtime config that automatically wraps all procedures to improve developer experience.

### Why Runtime Configuration?

- **Architectural Fit**: Async context is a runtime concern, not request context
- **Backward Compatible**: Optional parameter maintains existing behavior
- **Clean DX**: Single configuration point vs middleware wiring

### Propagation Path

1. Run time config → Root config → Procedure builder → Resolver → Procedure caller
2. Preserves full type safety throughout
3. inject context into the storage for type safety

### Performance

- Minimal overhead (AsyncLocalStorage is optimized for this)
- Only active when configured
- short-circuit pattern  run the async storage on the same index as middleware chain

### Advanced Use Cases

- **Nested Procedures**: Automatically inherit storage
- **Overrides**: Still possible via manual `.use()` wrapping with not passing asyncStorage as runtime configuration
- **Middleware**: Runs within storage context

## Usage Example

```tsx
const storage = new AsyncLocalStorage();
const t = initTRPC.create({
  asyncStorage: storage 
});

// Any procedure automatically has context storage
const router = t.router({
  getUser: t.procedure.query(() => {
    deepFunction(); // Can access context via storage.getStore()
  })
});

```

## Alternatives Considered

1. **Context-Based**: Rejected - wrong abstraction l[a](http://async-context-management.md/)yer
2. **Plugin System**: Overkill -  focused use case
3. **Global Middleware**: existing solution before this with .use on a procedure  

## Final note :

- I have also updated the documentation and added example for this, as new concept in docs docs/server/[a](http://a[sync-context-management.md](http://sync-context-management.md/)/)sync-context-management.md

## ✅ Checklist

- [ x] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [x ] If necessary, I have added documentation related to the changes made.
- [x ] I have added or updated the tests related to the changes made.

## Summary by CodeRabbit

* **New Features**
  * Added support for integrating AsyncLocalStorage for automatic context propagation across all procedures.
* **Documentation**
  * Introduced new documentation explaining async context management and usage of AsyncLocalStorage with procedures.
* **Tests**
  * Added tests demonstrating AsyncLocalStorage integration with minimal configuration.
* **Chores**
  * Improved test setup for generating self-signed TLS certificates for better cross-platform compatibility.

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/initTRPC.ts
+++ b/packages/server/src/unstable-core-do-not-import/initTRPC.ts
@@ -1,4 +1,4 @@
-import { AsyncLocalStorage } from 'async_hooks';
+import type { AsyncLocalStorage } from 'async_hooks';
 import {
   defaultFormatter,
   type DefaultErrorShape,
--- a/packages/server/src/unstable-core-do-not-import/initTRPC.ts
+++ b/packages/server/src/unstable-core-do-not-import/initTRPC.ts
@@ -1,3 +1,4 @@
+import { AsyncLocalStorage } from 'async_hooks';
 import {
   defaultFormatter,
   type DefaultErrorShape,
@@ -47,6 +48,11 @@ export interface RuntimeConfigOptions<
    * @see https://trpc.io/docs/v11/data-transformers
    */
   transformer?: DataTransformerOptions;
+
+  /**
+   * option to enable async storage for all procedures
+   */
+  asyncStorage?: AsyncLocalStorage<any>;
 }
 
 type ContextCallback = (...args: any[]) => object | Promise<object>;
@@ -165,6 +171,10 @@ class TRPCBuilder<TContext extends object, TMeta extends object> {
        * @internal
        */
       $types: null as any,
+      /**
+       * pass asyncStorage as runtime configuration
+       */
+      asyncStorage: opts?.asyncStorage,
     };
 
     {
@@ -177,6 +187,7 @@ class TRPCBuilder<TContext extends object, TMeta extends object> {
         );
       }
     }
+
     return {
       /**
        * Your router config
@@ -189,6 +200,7 @@ class TRPCBuilder<TContext extends object, TMeta extends object> {
        */
       procedure: createBuilder<$Root['ctx'], $Root['meta']>({
         meta: opts?.defaultMeta,
+        asyncStorage: opts?.asyncStorage,
       }),
       /**
        * Create reusable middlewares
--- a/packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
+++ b/packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
@@ -1,3 +1,4 @@
+import { AsyncLocalStorage } from 'async_hooks';
 import type { inferObservableValue, Observable } from '../observable';
 import { getTRPCErrorFromUnknown, TRPCError } from './error/TRPCError';
 import type {
@@ -87,6 +88,10 @@ type ProcedureBuilderDef<TMeta> = {
   subscription?: boolean;
   type?: ProcedureType;
   caller?: CallerOverride<unknown>;
+  /**
+   * option to enable async storage for all procedures
+   */
+  asyncStorage?: AsyncLocalStorage<any>;
 };
 
 type AnyProcedureBuilderDef = ProcedureBuilderDef<any>;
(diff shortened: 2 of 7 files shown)


ARAldhafeeri:
Hello, any updates on this ?

packages/server/src/unstable-core-do-not-import/initTRPC.ts
```diff
@@ -1,3 +1,4 @@
+import type { AsyncLocalStorage } from 'async_hooks';
```
KATT, line 1: We shouldn't import from Node, instead we should use a duck type that looks like `AsyncLocalStorage`

packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
```diff
@@ -624,6 +629,24 @@ async function callRecursive(
...
+        result = await callRecursive(
+          0,
+          { ..._def, asyncStorage: undefined },
+          opts,
+        );
+      });
+
+      // Return the result from the async context
+      return result!;
+    }
```
KATT, line 648:
```suggestion
    // If this is the first call (index 0) and asyncStorage is configured, wrap the entire execution
    if (index === 0 && _def.asyncStorage) {
      return await _def.asyncStorage.run(opts.ctx, async () => {
        // Process the middleware chain without asyncStorage to prevent infinite recursion
        return await callRecursive(
          0,
          { ..._def, asyncStorage: undefined },
          opts,
        );
      });
    }
```

packages/tests/server/validators.test.ts
```diff
@@ -705,6 +705,76 @@ test('async validator fn', async () => {
...
+test('Summon async local storage across all procedures with minimal configuration and boilerplate.', async () => {
+  type Context = {
+    foo: string;
+  };
+
+  const contextStorage = new AsyncLocalStorage<Context>();
+
+  const t = initTRPC.context<Context>().create({
+    asyncStorage: contextStorage,
+  });
```
KATT, line 717: Hmmm, with the types you've put in, it always has to be a context storage - I think it'd be nice to separate the context from the data in the store so your async storage can have a type that doesn't match the context..

```ts
type Context = {
  foo: string;
};

const contextStorage = new AsyncLocalStorage<{
  something: string;
  ctx: Context;
}>();

const t = initTRPC.context<Context>().create({
  asyncStorage: {
    storage: contextStorage,
    // called whenever a procedure is called
    getStore: async (opts) => {
      return {
        ctx: opts.ctx,
        something: 'else',
      };
    },
  },
});
```

packages/server/src/unstable-core-do-not-import/initTRPC.ts
KATT: Replace with an interface like

```ts
interface AsyncStorageLike<T> {
  run<TResult>(store: T, fn: (store: T) => Promise<TResult>): Promise<TResult>;
}
```

packages/tests/server/validators.test.ts
ARAldhafeeri: Hello 👋,

What do you think about simply using:

```ts
AsyncLocalStorage<unknown>
```

That way developers can cast to their own type when needed. Since ALS is strictly server-side, type safety across the client–server boundary doesn’t add much practical value, it just introduces extra boilerplate.

On the breaking `getStore` change: I see the biggest win in keeping `AsyncLocalStorage` integration a **one-line config**. Developers can define their async local storage (with any helper methods they want) in one place and export only the storage. Inside procedures, they can call `getStore` when needed.

Not every procedure needs async local storage, but to make one-line config possible, we currently run storage on all procedures. The overhead of running storage each time is minimal, and developers can still wrap their own helpers around it if desired.

Your call though,  if you’d prefer me to implement this differently as you instructed, let me know. You’re the creator and know best.

ARAldhafeeri: Hello Katt, 
Any update on this ?

ARAldhafeeri:
Hello , any updates on this ?

Nick-Lucas:
I think this could instead be a community middleware. Put it first in the chain and wrap `next()` and it would work the same

PR #6888, opened 2025-08-11
fix: prefer Standard Schema for input/output type inference

### LINKED ISSUE #6887: feat: migrate to Zod v4 ways of retrieving input/output types

The `_input` and `_output` attributes that are being used to retrieve the types of Zod schemas are [now deprecated](https://github.com/colinhacks/zod/blob/v4.0.15/packages/zod/src/v4/classic/schemas.ts#L26). There is at least [one bug](https://github.com/colinhacks/zod/issues/5073) where `_output` is incorrect (while the non-deprecated replacement (`z.output<...>` or `...['_zod']['output']`) is correct), causing the input in the resolver to have an incorrect type.

As of v3.24.0, Zod supports Standard Schema, which could also be used to extract the types.

Type inference should use one of the newer and more correct ways of retrieving input/output types, where available.

Just wait for the bug to be fixed. That will probably work well enough in practice, but involves depending on something deprecated and potentially incorrect in other ways.

### Additional information

I'm willing to look into implementing a fix, though I would need guidance on what the desired solution should be, especially regarding compatibility with different versions of Zod.

### 👨‍👧‍👦 Contributing

dzhu:
## 🎯 Changes

Since it is the new and standard thing, we should use Standard Schema to extract input and output types whenever the schema implements it.

Fixes #6887

## ✅ Checklist

- [x] I have added or updated the tests related to the changes made. (note: The test uses `toEqualTypeOf` in a resolver definition like the other Zod branded type test, which doesn't seem to actually cause `pnpm test` to signal failure if the equality is not satisfied, though it does show up as an error in a TypeScript-aware editor. I'm not sure if a failure there is actually being caught in any automatically visible way.)

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/parser.ts
+++ b/packages/server/src/unstable-core-do-not-import/parser.ts
@@ -62,17 +62,22 @@ export type ParserWithInputOutput<TInput, TParsedInput> =
 export type Parser = ParserWithInputOutput<any, any> | ParserWithoutInput<any>;
 
 export type inferParser<TParser extends Parser> =
-  TParser extends ParserWithInputOutput<infer $TIn, infer $TOut>
+  TParser extends ParserStandardSchemaEsque<infer $TIn, infer $TOut>
     ? {
         in: $TIn;
         out: $TOut;
       }
-    : TParser extends ParserWithoutInput<infer $InOut>
+    : TParser extends ParserWithInputOutput<infer $TIn, infer $TOut>
       ? {
-          in: $InOut;
-          out: $InOut;
+          in: $TIn;
+          out: $TOut;
         }
-      : never;
+      : TParser extends ParserWithoutInput<infer $InOut>
+        ? {
+            in: $InOut;
+            out: $InOut;
+          }
+        : never;
 
 export type ParseFn<TType> = (value: unknown) => Promise<TType> | TType;
 
--- a/packages/tests/server/validators.test.ts
+++ b/packages/tests/server/validators.test.ts
@@ -935,6 +935,11 @@ test('zod4 branded types', () => {
         expectTypeOf(opts.input.accountId).toEqualTypeOf<Types['output']>();
         return opts.input;
       }),
+
+    top: t.procedure.input(AccountId).query((opts) => {
+      expectTypeOf(opts.input).toEqualTypeOf<Types['output']>();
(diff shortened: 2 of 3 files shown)


packages/server/src/unstable-core-do-not-import/parser.ts
```diff
@@ -62,17 +62,22 @@ export type ParserWithInputOutput<TInput, TParsedInput> =
 export type Parser = ParserWithInputOutput<any, any> | ParserWithoutInput<any>;
 
 export type inferParser<TParser extends Parser> =
-  TParser extends ParserWithInputOutput<infer $TIn, infer $TOut>
+  TParser extends ParserStandardSchemaEsque<infer $TIn, infer $TOut>
```
Nick-Lucas, line 65: Thanks for the contribution! This is a great idea and the docs changes look good

In this case couldn't we change the precedence of `ParserWithInputOutput` to keep the conditional type simple?

```ts
export type ParserWithInputOutput<TInput, TParsedInput> =
  | ParserStandardSchemaEsque<TInput, TParsedInput>
  | ParserZodEsque<TInput, TParsedInput>
  | ParserValibotEsque<TInput, TParsedInput>
  | ParserArkTypeEsque<TInput, TParsedInput>;
```

dzhu: I tried that, but it didn't work. My understanding, as a relative newcomer to TypeScript, is that there isn't actually precedence there; rather, the final type is the union of the outcomes for each individual branch.

For example, if you try to compile the following code, you should see an error that indicates that `Get<{ a: number; b: string }>` evaluates to `string | number`.
```ts
interface A<T> {
  a: T;
}
type B<T> = {
  b: T;
};

type U<T> = A<T> | B<T>;
type Get<X> = X extends U<infer $T> ? $T : never;

const x: Get<{ a: number; b: string }> = {};
```

I think the same thing applies in the Zod case, but it's hidden because the branded type is a subtype of the base type, so the union just equals the base type. (If you brand `string` with `"brand"`, the branded type is `string & z.BRAND<"brand">`.)

Nick-Lucas: Yes I believe you're correct, thanks for clarifying!

Nick-Lucas (approved):

PR #6976, opened 2025-10-17
feat(tanstack-react-query): Add QueryKey and MutationKey Prefix option

### LINKED ISSUE #4989: feat: overwrite query key on trpc.useQuery options

I have query procedure `listPosts` which accepts some generic `parameters` and and optional `lastUpdatedAt` timestamp. 
The query key looks something like this:

```
[["posts","list"],{"input":{ "parameters": {} },"type":"query"}]
```

This query is expensive, so I cache it also on the server with the given `parameters` as caching key in case multiple clients request the same data. When I receive the same `parameters` on the server, I return the data from the cache.

Now I need some way for clients to force a refresh on the server returning new data instead of cached data for the same `parameters`. So in this case, I query the `listPosts` with an optional `lastUpdatedAt` timestamp. The query key is now: 

```
[["posts","list"],{"input":{ "parameters": {}, "lastUpdatedAt": 123456789 },"type":"query"}]
```

The servers receives the same `parameters` as before and finds the data in the cache, but the `lastUpdatedAt` timestamp is greater than the cached timestamp. So it runs the expensive operation on the server and returns the new data (and caches it).

Now I have two query keys on the client which are logically the same. I must update the query data of key 1 and key 2 with new data, because query key 2 is only temporary to trigger a server cache-invalidation. The `lastUpdatedAt` that I sent in query key 2 is a local state that will get lost after page navigation. When I later get back to the `ListPosts` page, it will use query key 1 again with data which was maybe set by query key 2.

I'm currently using `onSuccess` of `useQuery` to set the query data of all query keys matching the input without `lastUpdatedAt`:

```ts
type Input = {
  parameters: Record<string, any>;
  lastUpdatedAt?: number;
};

const query = trpc.posts.list.useQuery(input, {
  onSuccess(data) {
    const { lastUpdatedAt, ...restInput } = input;
    // update the query cache for all entries with the same parameters, but ignoring the lastUpdatedAt date
    queryClient.setQueriesData(getQueryKey(trpc.posts.list, restInput), data);
  },
});
```

That means it sets the data for query key 1 (without `lastUpdatedAt`) and 2 (with `lastUpdatedAt`). 

---

However, `onSuccess` was deprecated in React Query v5, so I'm looking for an alternative option. 
Allowing to overwrite the generated tRPC query key would make it possible to send different inputs to the server but with the same query key. 

I would like to overwrite the query key via the tRPC query options:

```ts
const query = trpc.posts.list.useQuery(input, {
  trpc: {
    // static query key overwrite
    queryKey: [["posts","list"],{"input":{ "parameters": {} },"type":"query"}],
    // OR
    // dynamic query key overwrite
    queryKey: () => {
      // remove field that should not be passed to server
      const { lastUpdatedAt, ...restInput } = input;
      return getQueryKey(trpc.posts.list, restInput)
    },
  },
});
```

That means the following two inputs produce the same query key:

```ts
// = query key: [["posts","list"],{"input":{ "parameters": {} },"type":"query"}]
const input1 = {
  parameters: {},
}

// = query key: [["posts","list"],{"input":{ "parameters": {} },"type":"query"}]
const input2 = {
  parameters: {},
  lastUpdatedAt: 123456789
}
```

By default, React Query would return the data from the query cache for both inputs. So we would need to use the `refetch()` function to manually trigger a refetch to the server with new input. 

1. `useQuery.onSuccess()` was deprecated in v5, but could be implemented again with a `useEffect()` with dependency to `useQuery.data`

```ts
  const { data } = trpc.posts.list.useQuery(input);

  // update all queries every time the data changes
  React.useEffect(() => {
    const { lastUpdatedAt, ...restInput } = input;
    queryClient.setQueriesData(getQueryKey(trpc.posts.list, restInput), data);
  }, [data, queryClient, input]);
```

2. use `useQuery` directly with the proxy client and change the query key

```ts
const { lastUpdatedAt, ...restInput } = input;
const queryKey = getQueryKey(trpc.posts.list, restInput);

const query = useQuery({
  queryKey,
  queryFn: (context) => {
    return utils.client.posts.list.query(input)
  },
});
```

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

Nick-Lucas:
Closes #4989

## 🎯 Changes

Supports queryKeyPrefix on TRPCProvider with the Tanstack Query Client, allowing all cached queries in a react tree to be prefixed, for instance with a user ID. This allows easier implementation of user/session switching while keeping data cached for switching back and forth

```tsx
import { createTRPCContextWithFeatureFlags } from '@trpc/tanstack-react-query';
import type { AppRouter } from '../server/router';

// First enable the feature, we need a new function for this for typescript reasons, but it's a drop-in change
export const { TRPCProvider, useTRPC, useTRPCClient } =
  createTRPCContextWithFeatureFlags({
    enableKeyPrefix: true,
  })<AppRouter>();

export function App() {
  return (
    <QueryClientProvider client={queryClient}>
      <TRPCProvider
        trpcClient={trpcClient}
        queryClient={queryClient}

        // Optional: provide a prefix for all query/mutation keys
        queryKeyPrefix={userId}

        // Also can be an array
        queryKeyPrefix={[userId, orgId]}
      >
        {/* Your app here */}
      </TRPCProvider>
    </QueryClientProvider>
  );
}
```

Nuances:

* In order to use a prefix, the feature flag much be enabled when creating the context
* If a runtime prefix is not provided then the element won't be in the key - I _believe_ this is correct behaviour else `invalidateQueries([[]])` would invalidate `[["somePrefix"]]` as well
* If a runtime prefix is provided then query keys will include the prefix element

## ✅ Checklist

## Summary by CodeRabbit

* **New Features**
  * Optional keyPrefix support for TRPC providers to namespace query/mutation keys and avoid cache collisions across instances

* **Improvements**
  * Prefix-aware key handling throughout client APIs for consistent key generation and usage
  * TypeScript typings extended to propagate keyPrefix feature flags across public APIs

* **Documentation**
  * Added "Query Key Prefixing" setup, examples, and guidance for multi-provider isolation

* **Tests**
  * Added tests validating multiple TRPC providers and prefixed/unprefixed key behavior

### DIFF
--- a/www/docs/client/tanstack-react-query/usage.mdx
+++ b/www/docs/client/tanstack-react-query/usage.mdx
@@ -331,10 +331,12 @@ export function MyComponent() {
   const billingList = useQuery(billing.list.queryOptions());
   const accountList = useQuery(account.list.queryOptions());
 
-  return <div>
-    <div>Billing: {JSON.stringify(billingList.data ?? null)}</div>
-    <div>Account: {JSON.stringify(accountList.data ?? null)}</div>
-  </div>;
+  return (
+    <div>
+      <div>Billing: {JSON.stringify(billingList.data ?? null)}</div>
+      <div>Account: {JSON.stringify(accountList.data ?? null)}</div>
+    </div>
+  );
 }
 ```
 


www/docs/client/tanstack-react-query/server-components.mdx
```diff
@@ -336,7 +336,7 @@ export function prefetch<T extends ReturnType<TRPCQueryOptions<any>>>(
   queryOptions: T,
 ) {
   const queryClient = getQueryClient();
-  if (queryOptions.queryKey[1]?.type === 'infinite') {
+  if (queryOptions.queryKey[2]?.type === 'infinite') {
```
KATT: This will cause a breaking change for some - can we do it so we only prepend a prefix if there is one?

KATT: `TRPCQueryKey` should probably become `TRPCQueryKey<TIsPrefixed extends boolean = false /* for backwards compat*/>`?

Nick-Lucas: It's a consideration I want to be sure about definitely, here are my current thoughts:

* we take data from keys by index in multiple places so having an optional prefix risks weird bugs, and I opted for simplicity on first pass
* we'd want to use a prefix because then myrouter.pathKey() captures/invalidates the whole prefix but nothing else, which I think is desirable behaviour 
* do we want to worry about folk who are trying to directly manipulate query keys? Anybody using the client as designed won't notice the change when they upgrade, keys will just change format

KATT: Or rather...


```ts
/**
 * @internal
 */
export type TRPCKeyPrefix = string | undefined;

type TRPCQueryKeyBase = [
  path: readonly string[],
  opts?: { input?: unknown; type?: Exclude<QueryType, 'any'> },
];

type TRPCQueryKeyWithPrefix<TPrefix extends string> = [
  prefix: TPrefix,
  path: readonly string[],
  opts?: { input?: unknown; type?: Exclude<QueryType, 'any'> },
];

/**
 * @public
 */
export type TRPCQueryKey<TPrefix extends TRPCKeyPrefix = undefined> =
  TPrefix extends string ? TRPCQueryKeyWithPrefix<TPrefix> : TRPCQueryKeyBase;

type TRPCMutationKeyBase = [path: readonly string[]];

type TRPCMutationKeyWithPrefix<TPrefix extends string> = [
  prefix: TPrefix,
  ...TRPCMutationKeyBase,
];

/**
 * @public
 */
export type TRPCMutationKey<TPrefix extends TRPCKeyPrefix = undefined> =
  TPrefix extends string
    ? TRPCMutationKeyWithPrefix<TPrefix>
    : TRPCMutationKeyBase;

```

Nick-Lucas: I don't think we'll be able to propagate types based on whether a prefix is set and retain the ideal API. The prefix would need to be known at the point all the types are created rather than served to the Provider dynamically

So we can either:

a. make the query key a union type between prefixed and non-prefixed. Projects indexing it for whatever reason might then get type errors and need to check if length is 2 or 3 before indexing it
b. add the prefix into the path segment of the key, though I believe that could be more breaking if any code depends on reading that back
c. change the query key type as proposed. Which also breaks existing indexing similar to above options
d. we use your proposed approach but make the TPrefix argument a feature flag which enables the functionality, and force people to enable the feature from the proxy factory so that TRPCProvider comes with the feature enabled and query keys are triples thereafter
e. not do this at all until v12

I think D is the only option we have to entirely avoid breaking changes. What do you think?

juliusmarminge: Yes and then with d you can also make it so the provider props is

`TFeatureFlag extends { prefix: true } ? BaseProps & { queryKeyPrefix: unknown } : BaseProps`

or whatnot

Nick-Lucas: Having a play and urgh D touches _everything_

We'll have to propagate this flag through every single type in the tanstack client

This is going to be a big PR 😭

Floriferous:
This is super useful, we've setup a hack that invalidates everything when orgId/userId changes, because it's being sent as part of our Clerk token to the backend, instead of being included in the query key. 

Awesome, looking forward to getting this merged 👌

examples/minimal-react/playwright.config.ts
```diff
@@ -14,8 +15,10 @@ const config: PlaywrightTestConfig = {
     ...devices['Desktop Chrome'],
     baseURL: baseUrl,
     headless: opts.headless,
+    video: process.env['CI'] ? 'retain-on-failure' : 'off',
+    screenshot: process.env['CI'] ? 'only-on-failure' : 'off',
```
Nick-Lucas, line 19: there seems to be a race condition in the e2e testing, and enabling recording slows down playwright enough to fix it... Obviously not the right fix, but I can't reproduce it locally or with recording enabled to see what's happening 🙈

PR #6986, opened 2025-10-24
simplify

KATT:
simplify a little bit

### DIFF
--- a/www/docs/client/tanstack-react-query/setup.mdx
+++ b/www/docs/client/tanstack-react-query/setup.mdx
@@ -153,14 +153,14 @@ export function App() {
 If you want to prefix all queries and mutations with a specific key (for example, to namespace them by user ID), you can enable the `enablePrefix` feature flag when creating the tRPC context.
 
 ```tsx title='components/App.tsx'
-import { createTRPCContextWithFeatureFlags } from '@trpc/tanstack-react-query';
+import { createTRPCContext } from '@trpc/tanstack-react-query';
 import type { AppRouter } from '../server/router';
 
-// First enable the feature
-export const { TRPCProvider, useTRPC, useTRPCClient } =
-  createTRPCContextWithFeatureFlags({
-    enableKeyPrefix: true,
-  })<AppRouter>();
+// Enable the prefix feature flag
+export const { TRPCProvider, useTRPC, useTRPCClient } = createTRPCContext<
+  AppRouter,
+  { enablePrefix: true }
+>();
 
 // Then use it in your app
 export function App() {
--- a/packages/tanstack-react-query/src/index.ts
+++ b/packages/tanstack-react-query/src/index.ts
@@ -1,7 +1,4 @@
-export {
-  createTRPCContext,
-  createTRPCContextWithFeatureFlags,
-} from './internals/Context';
+export { createTRPCContext } from './internals/Context';
 export type {
   TRPCOptionsProxy,
   inferInput,
--- a/packages/tanstack-react-query/src/internals/Context.tsx
+++ b/packages/tanstack-react-query/src/internals/Context.tsx
@@ -6,25 +6,6 @@ import type { TRPCOptionsProxy } from './createOptionsProxy';
 import { createTRPCOptionsProxy } from './createOptionsProxy';
 import type { DefaultFeatureFlags, FeatureFlags } from './types';
 
-export interface CreateTRPCContextOpts<TEnablePrefix extends boolean> {
-  /**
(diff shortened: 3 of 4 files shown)


www/docs/client/tanstack-react-query/setup.mdx
```diff
@@ -153,14 +153,14 @@ export function App() {
...
 
-// First enable the feature
-export const { TRPCProvider, useTRPC, useTRPCClient } =
-  createTRPCContextWithFeatureFlags({
-    enableKeyPrefix: true,
-  })<AppRouter>();
+// Enable the prefix feature flag
+export const { TRPCProvider, useTRPC, useTRPCClient } = createTRPCContext<
+  AppRouter,
+  { enablePrefix: true }
```
Nick-Lucas, line 162: This is a breaking change I believe, you can't ignore this type  param if you want to provide the router type. Or am I missing something?

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
```diff
@@ -154,18 +152,16 @@ export interface DecorateInfiniteQueryProcedure<
...
+  extends TypeHelper<TDef>,
+    DecorateRouterKeyable<TDef['featureFlags']> {
   /**
    * Create a set of type-safe query options that can be passed to `useQuery`, `prefetchQuery` etc.
    *
    * @see https://tanstack.com/query/latest/docs/framework/react/reference/queryOptions#queryoptions
    * @see https://trpc.io/docs/client/tanstack-react-query/usage#queryOptions
    */
-  queryOptions: TRPCQueryOptions<TDef, TFeatureFlags>;
+  queryOptions: TRPCQueryOptions<TDef>;
```
Nick-Lucas, line 164: I was wondering if TDef would be the better way to do this, just wasn't sure since it didn't go to 100% of branches, thanks for implementing!

packages/tanstack-react-query/src/internals/utils.ts
```diff
@@ -178,28 +160,21 @@ export function getQueryKeyInternal<TPrefix extends readonly string[]>(
...
-        ...(opts.type && opts.type !== 'any' && { type: opts.type }),
+        ...(typeof input !== 'undefined' &&
+          input !== skipToken && { input: input }),
+        ...(type && type !== 'any' && { type: type }),
       },
     ];
+  });
+
+  if (opts.prefix) {
+    base.unshift(opts.prefix);
```
Nick-Lucas, line 175: Yeah this is better

www/docs/client/tanstack-react-query/setup.mdx
KATT: Ay. Did i forget a default value?

KATT: The way I work with new params is usually:

1. Add it as required
2. Implement it everywhere
3. Then, add default value for backwards compat

That way, I don't miss adding it somewhere internally

KATT: It's implemented with a default a value already https://github.com/trpc/trpc/blob/4abdf7be265a520463639da83b768c1de9a56783/packages/tanstack-react-query/src/internals/Context.tsx#L30-L33

Did I miss anything?

Nick-Lucas: Hm okay, weirdly I thought I'd ruled out an optional type param as possible, and Julius concurred it was impossible, but trying in tsplayground now is works. So it's fine I guess!

https://www.typescriptlang.org/play/?ts=5.8.3#code/GYVwdgxgLglg9mABFApgZygHgIIBpEBCiAvIsAIYA2aKAfABQCUAXIgNp6EC6iA3gFCIhiAE4ooIEUjY9yaROTABPfgF9+-CAgyj0ISlBLJ0WAETlTDRkA

I do though think that inferring from a runtime variable is better forward-thinking, in case we want it in the future, and because it's a little friendlier for DX to have a physical input for configuring this, and that's where it does break:

https://www.typescriptlang.org/play/?ts=5.8.3#code/GYVwdgxgLglg9mABFApgZygHgIIBpEBCiAvIsAIYA2aKAfABQBGAXIQJSsDaehAuogG8AUIlGIATiighxSTv3JpE5MAE8hAXyFCICDBPQhKUEsnRYAROQsMLjC2yFA

Nick-Lucas: Perhaps the compromise would be something like:

`const { useTRPC } = createTRPCContext<AppRouter>().setFlags({ enablePrefix: true })`

PR #6987, opened 2025-10-25
Prefix query key 2

KATT:
- prefix always a string (but added as an array, we can make it more flexible at a later point)
- force setting a query key on provider
- get rid of that `readonly string[]`

### DIFF
--- a/www/docs/client/tanstack-react-query/setup.mdx
+++ b/www/docs/client/tanstack-react-query/setup.mdx
@@ -180,7 +180,7 @@ export function App() {
       <TRPCProvider
         trpcClient={trpcClient}
         queryClient={queryClient}
-        keyPrefix={'service-1'}
+        keyPrefix="service-1"
       >
         {/* Your app here */}
       </TRPCProvider>


packages/tanstack-react-query/src/internals/Context.tsx
```diff
@@ -10,14 +10,19 @@ export interface CreateTRPCContextResult<
...
-      : never;
-  }>;
+  TRPCProvider: React.FC<
+    {
+      children: React.ReactNode;
+      queryClient: QueryClient;
+      trpcClient: TRPCClient<TRouter>;
+    } & (TFeatureFlags['enablePrefix'] extends true
+      ? {
+          queryKeyPrefix: string;
```
Nick-Lucas: One thing I had at the back of my mind is this isn't strictly a correct name, it also prefixes the mutation key. But equally I wasn't sure about confusing folk with `keyPrefix` since that's not really TRQ terminology

packages/tanstack-react-query/test/__helpers.tsx
```diff
@@ -41,7 +41,9 @@ export function testReactResource<TRouter extends AnyTRPCRouter>(
 
   const { TRPCProvider, useTRPC, useTRPCClient } = createTRPCContext<
     TRouter,
-    { enablePrefix: true }
+    {
+      enablePrefix: true;
+    }
   >();
```
Nick-Lucas, line 49: Did you catch my last comments here?

https://github.com/trpc/trpc/pull/6986#discussion_r2462699138

Sent after the PR was merged but I do think that having a physical parameter to pass in at runtime would be a little better for DX, and so we can utilise that state for other flags we might invent that actually need the variable at runtime

packages/tanstack-react-query/src/internals/createOptionsProxy.ts
```diff
@@ -289,17 +289,16 @@ export type TRPCOptionsProxy<
 > &
   DecorateRouterKeyable<TFeatureFlags>;
 
-export interface TRPCOptionsProxyOptionsBase<
+export type TRPCOptionsProxyOptionsBase<
```
KATT, line 293: this is technically a breaking change but one i can live with

packages/tanstack-react-query/test/__helpers.tsx
KATT: I added type tests now

packages/tanstack-react-query/src/internals/Context.tsx
KATT: I just renamed it to `keyPrefix` anyway

PR #7000, opened 2025-11-06
fix: add minimal url validation to httpLink options

bwalendz:
## Problem

You can provide an invalid url to `httpLink` options, either a string or null/undefined. If this is provided by an untyped env variable it can easily go unnoticed. In addition, the `httpLink` options typing is not guaranteed during runtime and/or can be bypassed by `any`.

Currently with the invalid url scenario, the error is somewhat opaque:
```
TypeError: Cannot read properties of undefined (reading 'toString')
    at resolveHTTPLinkOptions
  cause: TypeError: Cannot read properties of undefined (reading 'toString')
      at resolveHTTPLinkOptions
      at httpBatchLink
```

## Solution
Add minimal relative/absolute url validation inside `resolveHTTPLinkOptions`, which currently accepts `url: string | URL`, that will produce a user friendly error if the url is invalid.

We could only check for null/undefined but I think light validation of the url value and failing fast is more appropriate because a valid url is necessary downstream and can make it more obscure to debug the source of the issue.

The new error will display:
```
Error: A valid relative or absolute url is required for `httpLink` options, provided: undefined
    at resolveHTTPLinkOptions
    at httpBatchLink
```

## Supported HTTP URL Examples
This should be feature parity with the no validation version.

String Relative: `/api/trpc`
String Absolute: `http://localhost:3000/api/trpc`
URL Object: `new URL(...)`

## Error Example
```typescript
const url = import.meta.env.VITE_FORGOT_TO_DEFINE; // undefined, can be typed as any

const trpc = createTRPCClient<AppRouter>({
  links: [
    httpBatchLink({
      url,
    }),
  ],
});
```

## Tests
- Updated existing tests that previously used a non-valid URL
- Added new tests for relative and absolute urls and malformed

## Summary by CodeRabbit

* **New Features**
  * Added URL validation and normalization for HTTP links with descriptive error messages for invalid inputs

* **Tests**
  * Updated tests to use concrete local endpoints for reliability
  * Added coverage for relative, absolute, and malformed URL scenarios in HTTP link handling

### DIFF
--- a/packages/tests/server/links.test.ts
+++ b/packages/tests/server/links.test.ts
@@ -6,10 +6,8 @@ import {
   httpBatchLink,
   httpBatchStreamLink,
   httpLink,
-  httpSubscriptionLink,
   loggerLink,
   retryLink,
-  splitLink,
   TRPCClientError,
 } from '@trpc/client';
 import { createChain } from '@trpc/client/links/internals/createChain';


packages/client/src/links/internals/httpUtils.ts
```diff
@@ -16,6 +16,20 @@ import type { TransformerOptions } from '../../unstable-internals';
 import { getTransformer } from '../../unstable-internals';
 import type { HTTPHeaders } from '../types';
 
+function parseUrlOption(url: string | URL) {
```
Nick-Lucas, line 19: I really like the idea of validating this, it makes sense and seems the various cases have been well considered. 

I would rather throw more meaningful errors up front than return null and ask the consumer to throw a general error though

PR #7037, opened 2025-12-11
fix(server): pass maxDurationMs timeout signal to subscription handlers

### LINKED ISSUE #6991: bug: maxDurationMs does not interrupt the subscription when not yielding

```
  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @tanstack/react-query: ^5.80.3 => 5.90.5 
    @trpc/client: canary => 11.7.1-canary.1+3ae1c3651 
    @trpc/next: canary => 11.7.1-canary.1+3ae1c3651 
    @trpc/react-query: canary => 11.7.1-canary.1+3ae1c3651 
    @trpc/server: canary => 11.7.1-canary.1+3ae1c3651 
    next: ^15.3.1 => 15.5.6 
    react: ^19.1.0 => 19.2.0 
    typescript: ^5.9.2 => 5.9.3 
```

### Describe the bug

When testing subscriptions on Vercel, I was surprised to experience timeouts of my lambdas, even though I've set  `sse.maxDurationMs` to a duration shorter than my lambda's max duration.

After running more tests, I figured the subscription was correctly terminated but only when it yields a value:
https://stackblitz.com/edit/github-rmw4rvvn-lqfxkrsa?file=src%2Fpages%2Fapi%2Ftrpc%2F%5Btrpc%5D.ts

If I do not yield anything from the subscription handler, then it's never terminated and the signal is never aborted:
https://stackblitz.com/edit/github-rmw4rvvn-ertp6cer?file=src%2Fpages%2Fapi%2Ftrpc%2F%5Btrpc%5D.ts

Btw, I also noticed that when the subscription is terminated by `maxDurationMs`, it takes some time for the client to notice and reconnect. It would be great if there was a built-in way to proactively reconnect from the client in order to minimise the latency during reconnection. Maybe that could be a default behaviour when `maxDurationMs` is set? Please let me know if I'm missing something here.

### Link to reproduction

https://stackblitz.com/edit/github-rmw4rvvn-ertp6cer?file=src%2Fpages%2Fapi%2Ftrpc%2F%5Btrpc%5D.ts

### To reproduce

Just look at the console log. Notice that we never log "🚫 subscription ended" but continue loggign `🔄 loop ${count}` even though the 2s `maxDurationMs` is passed.

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

tt-a1i:
## Summary

When using `httpSubscriptionLink` with `maxDurationMs`, the timeout was not being propagated to the subscription handler's `opts.signal`. This caused subscriptions that wait for the signal to abort (without yielding values) to hang indefinitely, as described in #6991.

**The fix:**
- Creates a combined `AbortController` that aborts when either:
  1. The original request signal aborts (client disconnect)
  2. The `maxDurationMs` timeout fires
- Passes this combined signal to the subscription handler via `opts.signal`
- Uses `AbortController.abort()` with `AbortError` (not `AbortSignal.timeout()` which throws `TimeoutError`)

Now subscription handlers can properly respond to the `maxDurationMs` timeout by:
- Checking `opts.signal.aborted`
- Listening for the `'abort'` event on `opts.signal`

## Test plan

- [x] Added regression test that verifies `maxDurationMs` aborts subscriptions that don't yield
- [x] All existing `httpSubscriptionLink.test.ts` tests pass (16 tests)
- [x] Full server test suite passes (484 tests)

Fixes #6991

## Summary by CodeRabbit

* **Bug Fixes**
  * Subscription timeouts (maxDurationMs) are now enforced via a shared request-level cancellation signal, improving propagation and cleanup across calls; stream-level timeout wrapping was removed.

* **Tests**
  * Added regression tests ensuring subscriptions are aborted when maxDurationMs is set (including non-yielding cases) and that timers/resources are cleared on unsubscribe.


tt-a1i:
Good catch, thanks for digging into this.

I agree `AbortSignal.timeout()` doesn’t really solve the leak here since there’s no way to cancel it early, so the timer can hang around until it fires.

I pushed an update that:
- switches back to a manual `setTimeout` + `clearTimeout`
- removes the request signal listener when either the timeout or the request abort wins
- wraps the returned observable/async-iterable so cleanup also runs on completion/unsubscribe (and if the procedure throws), not only on request abort
- keeps this scoped to `subscription` procedures only

Commit: cbaa9390d

tt-a1i:
Quick update: the remaining `Lint and auto-fix` failure was just an eslint rule (`@typescript-eslint/no-redundant-type-constituents`) complaining about `unknown | null` in the async-iterable wrapper.

I pushed a tiny follow-up that switches that temp variable to plain `unknown` and clears it with `undefined` after each `yield` (same GC intent, but satisfies the rule).

Commit: dad98d002

PR #7073, opened 2025-12-23
feat(server): bind procedure return type to output parser input

### LINKED ISSUE #6978: feat: Bind return type of a procedure to the input type of the `output` schema

With Zod V4.1 introducing [Codecs](https://zod.dev/codecs), we get a more "lightweight" option for encoding and decoding data (e.g., instead of using `Transformers`). However, the current implementation of the type inference binds the returned type of a procedure to the output type of the parse function, rather than the input.

This restricts the output schema in that it can't apply transformations and is only capable of validations 1:1.

For example, if I have a schema with a codec `stringToBigInt` (as per the docs) and would look to do something like the code below, I'll face type inference issues stating that `Type string is not assignable to bigint`, even though the underlying schema and runtime types would align.

```ts
function encode<TSchema extends z.ZodType>(schema: TSchema) {
  return (value: unknown) => schema.encode(value as z.output<TSchema>) // as-coercion needed due to the input type of the `ParseFn` being `unknown`
}

//...
procedure.output(encode(z.object({x: stringToBigInt}))).query(() => ({x: 0n}))
```

To get this working, it seems like what's needed is to enable chained middlewares to align output types of prior middlewares as input types to the successors.

The [`MiddlewareOKResult`](https://github.com/trpc/trpc/blob/cd64ab3b9684b24ceec8bc8d5b247db3db23cd06/packages/server/src/unstable-core-do-not-import/middleware.ts#L21) could have a type parameter which defaults to `unknown` which then is set in [createOutputMiddleware](https://github.com/trpc/trpc/blob/cd64ab3b9684b24ceec8bc8d5b247db3db23cd06/packages/server/src/unstable-core-do-not-import/middleware.ts#L216) instead of leaving the `outputMiddleware` as an `AnyMiddlewareFunction`, which is used to narrow the type on the middleware output type.

From what I can see, the `ParseFn` type would also need an argument type parameter `TInput` (default to `unknown`), e.g.,
```ts
type ParseFn<TOutput, TInput = unknown> = (value: TInput) => Promise<TOutput> | TOutput
```
of which there are then some areas where types need to be updated to ensure it aligns well.

Right now, I have to do type coercions that I don't necessarily like to make this work. The solution is to do the following (using the initial example):

```ts
function encode<TSchema extends z.ZodType>(schema: TSchema) {
  return (value: unknown) => schema.encode(value as z.output<TSchema>) // as-coercion needed due to the input type of the `ParseFn` being `unknown`
}

//...
const schema = z.object({x: stringToBigInt})
procedure.output(encode(schema)).query(() => ({x: 0n}) as unknown as z.input<typeof schema>) // coerce to input type although it's the output shape to be encoded through the codec to be the input shape
```

### Additional information

If I have missed any requirements and this is a useful feature for others, please let me know how I can assist. 
If there are other ways of going about this in a more typesafe way, do let me know as well! 

### 👨‍👧‍👦 Contributing

bas0N:
Closes #6978

## 🎯 Changes

This PR adds support for binding the return type of a procedure to the **input type** of the output schema, enabling proper type inference for Zod v4.1 Codecs and other bidirectional transform scenarios.

### Problem

With Zod v4.1 Codecs (bidirectional transforms), schemas can have different input and output types. For example, a `stringToBigInt` codec would:
- **Input type** (what goes into the encoder): `bigint`
- **Output type** (what comes out after encoding): `string`

Previously, `ParseFn<TType>` took `unknown` as input, losing type information about what the parse function expected. This caused type errors when using output transforms/codecs where the resolver should return the schema's input type

### Solution

1. **Updated `ParseFn` type** to support an input type parameter:
   
   type ParseFn<TOutput, TInput = unknown> = (value: TInput) => Promise<TOutput> | TOutput;
   2. **Added `TData` type parameter** to `MiddlewareOKResult`, `MiddlewareErrorResult`, and `MiddlewareResult` to properly track data types through the middleware chain.

3. **Updated `MiddlewareFunction`** with `TOutputIn` and `TOutputOut` type parameters for proper output type flow.

4. **Updated `createOutputMiddleware`** to accept `ParseFn<TOutputOut, TOutputIn>` for correct type inference.

### Result

Now output transforms work correctly with proper type inference:

```
// Schema that transforms bigint → string
const schema = z.object({ x: z.bigint().transform(x => x.toString()) });

procedure
  .output(schema)
  .query(() => {
    // Resolver returns bigint (schema input type)
    return { x: 123n };
    // Client receives { x: string } (schema output type)
  });
```

## ✅ Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved type safety for middleware input/output and result payloads, making middleware pipelines' data shapes clearer.
  * Output handling now uses explicit encoding paths so server-to-client transforms are consistently applied.

* **Tests**
  * Added tests verifying output encoding/serialization (including date, bigint, and nested transforms) to ensure resolver→client values are correctly encoded.


lukeramsden:
I don't think this will properly support codecs at runtime - https://zod.dev/codecs - it requires using .encode on the 'backward' pass.  I had a fairly minimal example here https://github.com/lukeramsden/trpc/pull/1/changes#diff-dd323865a9c57111229561a1ffe12931bb744ad7956ca5b596173a488d1d517c

KATT (commented):
Hey! Appreciate the work you did on this PR!

I need to look at the original issue to understand in what way we don't support this - I'll upgrade zod on another branch.

However, I don't see what this PR does, seeing that the tests seemingly pass even if I revert the files changed?

Even after doing the below, I don't see any type errors in the `outputParser.test.ts`
```shell
git checkout origin/main packages/server/src/unstable-core-do-not-import/middleware.ts packages/server/src/unstable-core-do-not-import/parser.ts
```

KATT:
I've bumped zod to v4 on `main` now so that should make it easier to make a real repro

bas0N:
@KATT @lukeramsden  Thanks for the feedback! Updated the PR to add runtime .encode() support for Zod v4 codecs - added getEncodeFn() that uses .encode() when available and falls back to .parse() for unidirectional transforms. Also replaced the tests with actual z.codec() tests that fail without these changes

Ari4ka (approved):

packages/tests/server/outputParser.test.ts
```diff
@@ -369,3 +369,120 @@ test('async validator fn', async () => {
...
+    encode: (big) => big.toString(),
+  });
+
+  const schema = z.object({ value: stringToBigint });
+
+  const router = trpc.router({
+    q: trpc.procedure.output(schema).query(() => {
+      // For codec output encoding, we return what encode() expects (bigint)
+      // and cast to satisfy the type system which expects the input type
+      return { value: 123n } as unknown as { value: string };
```
Nick-Lucas, line 397: wouldn't want to merge this without seeing that userland typesafety is better than this

Nick-Lucas:
given the age of this one (sorry we missed it) and the comment I've left I'm going to close, but happy to take an updated PR

PR #7100, opened 2026-01-13
feat(client,server): add experimental_encoder option for WebSocket connections

SirajChokshi:
Related to discussions in #1937

## 🎯 Changes
Adds Encoders. They sit below transformers in the message pipeline, converting tRPC messages to their wire format (e.g. JSON text frames or binary formats like CBOR, protobufs, msgpack, etc) and converting back to POJOs on the server.

- Added option `experimental_encoder` for WebSocket clients
    - `jsonEncoder` is used if unspecified (backwards compatible format)
    - `encoder` runs after `transformer` for websocket clients
- Adds NextJS websockets encoder example with msgpack

#### Message lifecycle
```mermaid
flowchart LR
    subgraph Client
        direction TB
        A[User Data] --> B[Transformer.serialize]
        B --> C[Encoder.encode]
    end
    
    C <--> D((WebSocket))
    
    subgraph Server
        direction TB
        D --> E[Encoder.decode]
        E --> F[Transformer.deserialize]
        F --> G[User Data]
    end
```

### Encoder Implementation Notes
Using this feature adds means encoder authors have to understand the points below. I lean towards allowing encoder (and transformer) authors to do whatever they want, so I did not abstract much away. I also figure most users will use off-the-shelf encoders for popular codecs as opposed to rolling their own.

Any suggestions on flattening/rolling these decisions into tRPC core without causing unexpected behavior are appreciated!

#### 1. `nullToUndefined` for Binary Encoders
E.g. MessagePack (and other binary formats) converts all undefined to null, but tRPC expects undefined for missing optional fields.

We could update [`parseTRPCMessage.ts`](https://github.com/trpc/trpc/blob/10b4bde0cb497c29d00f3c78c09fe8b64391f93c/packages/server/src/unstable-core-do-not-import/rpc/parseTRPCMessage.ts) to coerce `lastEventId` of `null` to `undefined`. I avoided touching these internals as I did not fully understand the implications / nuance between null and undefined.

#### 2. ArrayBuffer / Uint8Array Normalization
Different environments (NodeJS, browser, etc) use different formats for binary messages. Encoder implementors should include something like:
```typescript
const bytes = data instanceof ArrayBuffer ? new Uint8Array(data) : data;
```

#### Example

```typescript
import { decode, encode } from '@msgpack/msgpack';
import type { Encoder } from '@trpc/server/adapters/ws';

import { appRouter } from './routers/_app';

export const msgpackEncoder: Encoder = {
  encode: (data) => encode(stripUndefined(data)),
  //                       ^^^^^^^^ deeply replaces `null` with `undefined`
  decode: (data) => decode(data instanceof ArrayBuffer ? new Uint8Array(data) : data)
  },
};

const wss = new WebSocketServer({
  port: 3001,
});


const handler = applyWSSHandler({
  wss,
  router: appRouter,
  experimental_encoder: msgpackEncoder,
});
```

## ✅ Checklist

## Summary by CodeRabbit

* **New Features**
  * Experimental custom wire encoder support for WebSocket connections, enabling binary encoders (e.g., MessagePack) and improved real-time streaming.
  * Client and server updated to use selectable encoders for sending/receiving WebSocket frames.

* **Documentation**
  * Added a Next.js example with README showing setup, usage, and how to observe MessagePack binary frames.

* **Tests**
  * New comprehensive test suite validating encoder interoperability, subscriptions, and error paths.

### DIFF
--- a/www/docs/client/links/httpBatchLink.md
+++ b/www/docs/client/links/httpBatchLink.md
@@ -19,8 +19,8 @@ const client = createTRPCClient<AppRouter>({
   links: [
     httpBatchLink({
       url: 'http://localhost:3000',
+      // transformer,
     }),
-    // transformer,
   ],
 });
 ```


SirajChokshi:
Hey @hmatthieu,

Alex (KATT) said you should take a look at this PR ([Context on Discord](https://discord.com/channels/867764511159091230/980118117676118067/1460983659342594252)). LMK what you think / what to iterate on :)

packages/client/src/links/wsLink/wsClient/wsClient.ts
```diff
@@ -315,9 +319,12 @@ export class WsClient {
     ws.addEventListener('message', ({ data }) => {
       this.inactivityTimeout.reset();
 
-      if (typeof data !== 'string' || ['PING', 'PONG'].includes(data)) return;
+      // Handle PING/PONG as text regardless of encoder
+      if (typeof data === 'string' && ['PING', 'PONG'].includes(data)) return;
```
Nick-Lucas: I wonder if this might actually be a change of behaviour which is undesirable for the JSON encoder? 

I guess you've changed it because the data might actually not be string now? But that implies this logic is tied into the encoder itself and we need a 3rd method like "isValid(data): boolean"

SirajChokshi: > I wonder if this might actually be a change of behaviour which is undesirable for the JSON encoder?

We can't really abort early if the request isn't a string anymore since we are choosing to interpret more than just JSON text frames.

I believe the old `typeof data !== 'string'` guard was just there for type-safety (to avoid compilation errors) as opposed to a real runtime concern since the previous behavior was that tRPC server sends text frames no matter what via JSON.stringify(). WebSocket event data is typed as `any` in TypeScript 5.9, so the guard allows for JSON.parse() to compile a few lines down in the old code.

So, previously this case wasn't possible and now it is. I decided to bubble it up into an error when it gets to the default encoder. This primarily tries to flag the developer that their client and server are not agreeing on the encoder.
```
// guard from default JSONEncoder.decode()
if (typeof data !== 'string') {
      throw new Error(
        'jsonEncoder received binary data. JSON uses text frames. ' +
          'Use a binary encoder for binary data.',
      );
    }
```

**To make this clearer I removed the first part of the condition, so we're just checking for PING/PONG messages since that covers the string check anyway.**

Nick-Lucas (dismissed):

PR #7108, opened 2026-01-20
fix(tanstack-react-query): mark react-dom as optional peer dependency

### LINKED ISSUE #7081: bug: @trpc/tanstack-react-query should mark react-dom as optional peer dependency

see below

## Problem

  `@trpc/tanstack-react-query` lists `react-dom` as a required peer dependency, but React Native apps don't use `react-dom`. This causes npm install failures when using strict peer dep resolution.

  ## Current behavior

  ```json
  // packages/tanstack-react-query/package.json
  {
    "peerDependencies": {
      "react-dom": ">=18.2.0"
    }
    // no peerDependenciesMeta
  }
  ```

  ## Expected behavior

  ```json
  {
    "peerDependencies": {
      "react-dom": ">=18.2.0"
    },
    "peerDependenciesMeta": {
      "react-dom": { "optional": true }
    }
  }
  ```

  ## Context

  - `@tanstack/react-query` itself only requires `react`, not `react-dom`
  - React Native is a supported use case for tRPC
  - Other packages like `react-redux` use this pattern correctly ([reference](https://github.com/reduxjs/react-redux/issues/1559))

  ## Workaround

  Using `legacy-peer-deps=true` in `.npmrc`

### Link to reproduction

n/a

### To reproduce

see above

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

bbinya1224:
Closes #7081

## 🎯 Changes

This PR marks `react-dom` as an optional peer dependency in the `@trpc/tanstack-react-query package`.

### Why?
Currently, @trpc/tanstack-react-query lists react-dom as a required peer dependency.  
However, React Native and Expo applications do not use react-dom.  
This causes installation failures when using strict peer dependency resolution.

### What?
Added `peerDependenciesMeta` to `packages/tanstack-react-query/package.json`.
Set `react-dom` as `optional: true.`

This change allows mobile developers to use tRPC with TanStack Query without being forced to install react-dom, while still maintaining the correct dependency relationship for web developers.

## ✅ Checklist

- [x] If necessary, I have added documentation related to the changes made. (N/A - this is a configuration fix)
- [x] I have added or updated the tests related to the changes made. (N/A - package.json dependency change verified by build)

## Summary by CodeRabbit

* **Chores**
  * Removed react-dom from peer dependencies in the affected React query packages while leaving React as a required peer dependency. This simplifies peer dependency requirements, reduces strictness during installation, and improves compatibility with consumer projects that manage their own DOM renderer versions.


packages/tanstack-react-query/package.json
```diff
@@ -59,6 +59,11 @@
     "react-dom": ">=18.2.0",
     "typescript": ">=5.7.2"
   },
+  "peerDependenciesMeta": {
```
Nick-Lucas: I think it would actually be completely safe to remove the peer dependency entirely rather than just making it optional. I'm not entirely sure why it's there because we never import it directly so is probably a mistake!

bbinya1224: Thanks for the feedback! I've updated the PR to completely remove `react-dom` from the peer dependencies instead of just making it optional.

I noticed that `@trpc/react-query` also has the same issue. Would you like me to fix that in this PR as well, or should I create a separate PR for it?

Nick-Lucas: Go for it! So long as CI passes we should be golden

bbinya1224: Done! I've applied the changes to both packages. Thanks for the quick feedback!

First OSS contribution—really excited about this! 🫣

Nick-Lucas: Nice one!

If you want to do more then pop up in discord, we're always looking for more contributors :)

https://discord.com/channels/867764511159091230/980118117676118067

PR #7182, opened 2026-02-21
feat: support passing FormData/Octet types through serializer/deserializer

### LINKED ISSUE #7180: bug: when `serializer` in `httpLink` can not return `FormData`, even though tRPC can serialize it when using in `input` directly

```
  System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @tanstack/react-query: ^5.80.3 => 5.90.21 
    @trpc/client: canary => 11.9.1-canary.9+d92cc45b3 
    @trpc/next: canary => 11.9.1-canary.9+d92cc45b3 
    @trpc/react-query: canary => 11.9.1-canary.9+d92cc45b3 
    @trpc/server: canary => 11.9.1-canary.9+d92cc45b3 
    next: ^15.3.8 => 15.5.12 
    react: ^19.1.0 => 19.2.4 
    typescript: ^5.9.2 => 5.9.3 
```

### Describe the bug

I have written the following serializer/deserializer pair for use with `httpLink`. It allows serializing `Blob` objects contained inside of a plain object to be serialized as `FormData` fields. When you pass it an object `{ file: new Blob(["foo"]) }`, it returns a formdata with field `data` set to `{"file": <pointer to file 0>}` and field `blob` set to `new Blob(["foo"])`.

If I pass this FormData object directly to a tRPC mutation, it gets correctly serialized. In this case I have to manually deserialize it on the server though. If I instead pass this function as a `serializer` to `httpLink`, the `FormData` returned from `serializer` is encoded as JSON (resulting in `{}` being the value sent). This obviously then breaks things.

<details>
<summary>Serializer/deserializer for Blob</summary>

```
import { parse, registerCustom, stringify } from "superjson";

let BLOBS: Blob[] = [];

registerCustom({
  isApplicable: (v) => v instanceof Blob,
  serialize: (blob) => {
    BLOBS.push(blob);
    return { index: BLOBS.length - 1 };
  },
  deserialize: (obj) => BLOBS[obj.index],
}, "Blob");

export function serialize(data: unknown): FormData | string {
  BLOBS = [];
  try {
    const serialized = stringify(data);
    if (BLOBS.length === 0) return serialized;
    const formData = new FormData();
    formData.append("data", serialized);
    for (const blob of BLOBS) formData.append("file", blob);
    console.log(formData);
    return formData;
  } finally {
    BLOBS = [];
  }
}

export function deserialize(val: FormData | string): unknown {
  let serialized: string;
  if (typeof val === "string") {
    serialized = val;
    BLOBS = [];
  } else {
    console.log(val);
    const data = val.get("data");
    if (typeof data !== "string") throw new Error("Expected 'data' field to be a string");
    const files = val.getAll("file");
    for (const file of files) {
      if (!(file instanceof Blob)) throw new Error("Expected 'file' field to be a Blob");
    }
    serialized = data;
    BLOBS = files as Blob[];
  }
  try {
    return parse(serialized);
  } finally {
    BLOBS = [];
  }
}
```

</details>

### Link to reproduction

https://stackblitz.com/edit/trpc-serializing-formdata?file=src%2Fpages%2Findex.tsx

### To reproduce

Press button in the demo above.

### Additional information

My proposal is to allow `serializer` to return `FormData`, and to call `deserializer` even with `FormData` values from the wire. The latter is a breaking change, and would require a major version bump.

For now, I suggest a new option in the `transformer` called `deserializeNonJsonTypes` that defaults to `false`,  but can be set to `true` to enable this behaviour.

For consistency, `serializer` should probable be called also for `FormData` inputs (right now it is not). This is again a breaking change, so should be flagged by a `serializeNonJsonTypes` option.

And furthermore I am saying `FormData`, but it should probably apply equally to all values that `isNonJsonSerializable` returns `true` for.

### 👨‍👧‍👦 Contributing

lucacasonato:
Closes #7180

## 🎯 Changes

This commit adds support for:

- passing `FormData`, `Blob` and `ReadableStream` input values through client
  serializer (when `transformer.unstable_serializeNonJsonTypes` is enabled) and
  deserializing them on the server (when
  `transformer.unstable_deserializeNonJsonTypes` is enabled)
- returning `FormData`, `Blob` and `ReadableStream` values from serializer, and
  serializing them correctly on the client (ie not through JSON)

## ✅ Checklist

- [x] I have followed the steps listed in the
      [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).

## Summary by CodeRabbit

* **New Features**
  * Added support for sending FormData and binary (octet-stream) data in HTTP requests.
  * Added configuration flags to enable or disable serialization and deserialization of non-JSON types.

* **Tests**
  * Updated test expectations for FormData handling in GET requests.

### DIFF
--- a/www/docs/server/non-json-content-types.md
+++ b/www/docs/server/non-json-content-types.md
@@ -137,6 +137,104 @@ export const appRouter = t.router({
 
 For a more advanced code sample you can see our [example project here](https://github.com/juliusmarminge/trpc-interop/blob/66aa760141030ffc421cae1a3bda9b5f1ab340b6/src/server.ts#L28-L43)
 
+### Using Data Transformers with Non-JSON Inputs
+
+By default, non-JSON inputs like `FormData` and binary types bypass the data transformer — they are sent as-is from the client and arrive as-is on the server. This means that even if you have a transformer like `superjson` configured, it will not run on these input types.
+
+If you want your transformer to process these types (for example, to convert a `FormData` into a custom serialization format), you can opt in with two flags on the input transformer:
+
+- **`unstable_serializeNonJsonTypes`** — when `true`, the client-side `serialize()` function will be called on `FormData`, `Blob`, and `Uint8Array` inputs before sending them.
+- **`unstable_deserializeNonJsonTypes`** — when `true`, the server-side `deserialize()` function will be called on `FormData` and `ReadableStream` inputs before passing them to the resolver.
+
+These flags are independent — you can use one without the other.
+
+#### Example: Serializing FormData to JSON on the client
+
+If `serialize()` converts a `FormData` into a plain object, the request will be sent as `application/json` instead of `multipart/form-data`:
+
+```ts
+const transformer = {
+  input: {
+    serialize(obj) {
+      if (obj instanceof FormData) {
+        const data = {};
+        obj.forEach((v, k) => {
+          data[k] = String(v);
+        });
+        return data;
+      }
+      return obj;
+    },
+    deserialize: (obj) => obj,
+    unstable_serializeNonJsonTypes: true,
+  },
+  output: {
+    serialize: (obj) => obj,
+    deserialize: (obj) => obj,
+  },
+};
+```
+
+#### Example: Deserializing FormData on the server
+
+If you want the server-side transformer to convert `FormData` before it reaches the resolver:
+
+```ts
+const transformer = {
+  input: {
+    serialize: (obj) => obj,
+    deserialize(obj) {
+      if (obj instanceof FormData) {
+        const data = {};
+        obj.forEach((v, k) => {
+          data[k] = String(v);
+        });
+        return data;
+      }
+      return obj;
+    },
+    unstable_deserializeNonJsonTypes: true,
+  },
+  output: {
+    serialize: (obj) => obj,
+    deserialize: (obj) => obj,
+  },
+};
+```
+
+#### Example: Converting plain objects to FormData in the serializer
+
+The serializer can also _return_ a `FormData` or binary type from a plain object input. tRPC will detect the output type and set the correct `Content-Type` header automatically:
+
+```ts
+const transformer = {
+  input: {
+    serialize(obj) {
+      // Convert plain object → FormData on the client
+      const fd = new FormData();
+      fd.set('payload', JSON.stringify(obj));
+      return fd;
+    },
+    deserialize(obj) {
+      // Convert FormData → plain object on the server
+      if (obj instanceof FormData) {
+        return JSON.parse(obj.get('payload'));
+      }
+      return obj;
+    },
+    unstable_deserializeNonJsonTypes: true,
+  },
+  output: {
+    serialize: (obj) => obj,
+    deserialize: (obj) => obj,
+  },
+};
+```
+
+:::note
+Batch links (`httpBatchLink` and `httpBatchStreamLink`) do not support `FormData` or binary type inputs. If the transformer returns a `FormData` or binary type, batch links will throw an error. Use `httpLink` or a `splitLink` for these cases.
+:::
+
 ### `File` and other Binary Type Inputs
 
 tRPC converts many octet content types to a `ReadableStream` which can be consumed in a procedure. Currently these are `Blob` `Uint8Array` and `File`.


packages/server/src/unstable-core-do-not-import/http/contentType.ts
```diff
@@ -240,7 +244,12 @@ const octetStreamContentTypeHandler: ContentTypeHandler = {
       });
     }
     const getInputs = memo(async () => {
-      return req.body;
+      const stream = req.body;
+      const transformer = opts.router._def._config.transformer.input;
+      if (transformer.unstable_deserializeNonJsonTypes) {
```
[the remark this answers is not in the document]
lucacasonato: This is incorrect. An empty POST body will have a zero length readable stream, not `null` here.

Nick-Lucas:
Thanks for this, if you could get it rebased to resolve the conflicts I can have a proper look over, there's quite a bit to get my head around here. Not promising it will get merged though as I want to be really sure we're not taking on debt/complexity which can be avoided before committing to that.

packages/client/src/links/internals/httpUtils.ts
```diff
@@ -82,11 +82,34 @@ type GetInputOptions = {
...
+        !opts.transformer.input.unstable_serializeNonJsonTypes
+      ) {
+        throw new Error(
+          'Batch link does not support FormData or Octet type inputs, use httpLink',
+        );
+      }
+      const serialized = opts.transformer.input.serialize(_input);
+      if (isFormData(serialized) || isOctetType(serialized)) {
+        throw new Error(
+          'Batch link does not support FormData or Octet type inputs, use httpLink',
```
Nick-Lucas, line 106: I understand why centralising this logic is attractive, but I would rather accept some duplication in exchange for not mixing a bunch of different links into centralised utilities, makes it hard to unpick later if everything is coupled like that. Could we split this back out and just add the required functionality to each Link? Happy with adding more utilities at a similar level to isFormData etc if useful!

packages/client/src/links/internals/httpUtils.ts
```diff
@@ -119,9 +144,22 @@ export const getUrl: GetUrl = (opts) => {
...
   }
-  if (opts.type === 'query' || opts.type === 'subscription') {
+  if (
+    (opts.type === 'query' || opts.type === 'subscription') &&
+    opts.methodOverride !== 'POST'
+  ) {
     const input = getInput(opts);
-    if (input !== undefined && opts.methodOverride !== 'POST') {
+    if (input !== undefined) {
+      if (isFormData(input)) {
```
Nick-Lucas, line 153: Probably `isNonJsonSerializable` would be better suited here, we can get the class-name in a more general way and just have the one conditional. Easier to add more supported content types in the fduture

packages/client/src/links/internals/httpUtils.ts
```diff
@@ -131,12 +169,29 @@ export const getUrl: GetUrl = (opts) => {
   return url;
 };
 
-export const getBody: GetBody = (opts) => {
+export const getBodyAndContentType: GetBodyAndContentType = (opts) => {
```
Nick-Lucas, line 172: I am wondering if there was a simpler way to go about all of this. Isn't the root problem here that tRPC makes decisions on whether something is FormData(etc) before it decides that something is JSON and runs the transformer?

* The transformer flag you proposed is reasonable, and I wonder if we could simply hoist the transform call to earlier in the process and pass around `{ input, transformedInput }` for the remainder? Not sure how this would work in terms of performance cost though
* or maybe instead of trying to leverage the transformer our links need a `prepareInput(opts: { value, type, path, methodOverride, etc }): unknown` callback which gives you an opportunity to enrich or transform an input on the fly?

I think I lean toward the latter, because it might suit more use cases that just yours, and doesn't require twisting the transformers away from being a standardised interface which SuperJSON+others support, used only for JSON transport

packages/server/src/unstable-core-do-not-import/http/contentType.ts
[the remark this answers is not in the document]
Nick-Lucas: The type of stream disagrees here so I would go with that, it's just transformer.serialize takes any as the input so typescript isn't complaining

Nick-Lucas (commented):
Hey thanks for this. I've left a few comments but one of note that I think could take this in a much simpler direction, interested in your thoughts. 

Transformers weren't really intended for anything other than JSON type-safety and this feels like a pretty risky implementation to try and merge, so I'd love to find another way

PR #7199, opened 2026-03-02
test(tests): add superjson error deserialization regression tests

### LINKED ISSUE #7083: v11: tRPC server returns non-superjson error responses causing client 'Unable to transform response from server'
## Summary

Using tRPC v11 with `transformer: superjson` configured on the server, error responses (for example input validation errors) are emitted as plain JSON instead of the expected superjson wire format. Clients configured with `superjson` fail with:

> TRPCClientError: Unable to transform response from server

### Minimal reproduction

Repo: https://github.com/gerome650/trpc-superjson-repro

Steps:

1. `pnpm install`
2. `node server.mjs`
3. `node client.mjs`

The client intentionally calls the mutation without input to trigger validation.
**Expected**: server's error responses use the router's transformer (superjson) so the client can decode them.
**Actual**: client fails with "Unable to transform response from server"; server returns error body as plain JSON.

### Evidence

- tRPC versions: @trpc/server: 11.7.1, @trpc/client: 11.7.1, superjson: 1.13.3
- Error response body from server (when calling mutation without required input):
```json
[{"error":{"json":{"message":"Invalid input: expected object, received undefined","code":-32600,"data":{"code":"BAD_REQUEST","httpStatus":400}}}}]
```

- Client error:
```
TRPCClientError: Unable to transform response from server
    at TRPCClientError.from (.../@trpc/client/dist/TRPCClientError-CjKyS10w.mjs:41:10)
    at .../@trpc/client/dist/httpBatchLink-BOe5aCcR.mjs:237:37
```

### Observation

The error response is wrapped in `{"json": {...}}` structure (indicating some serialization attempt) but lacks the `"meta"` key that superjson format requires. The response appears to be a hybrid format that neither plain JSON parsers nor superjson deserializers can handle correctly.

### Suggestion

The HTTP adapter's error response path should apply the router's transformer (`superjson`) consistently so the client can decode error responses. Currently, the error path emits a partial serialization that doesn't match the expected superjson wire format.

Please let me know if you need additional logs or a different reproduction scenario.

eren-karakus0:
Closes #7083

## Changes

Add regression tests for #7083 to verify that error responses are properly serialized and deserialized when the server is configured with `transformer: superjson`.

### Background

The existing test suite covered:
- superjson for successful responses (data round-trips)
- Error responses with mixed transformers (superjson input + devalue output)

But there were **no tests** for the combination of **pure superjson + error responses**, which is exactly the scenario reported in #7083.

### Investigation

I was unable to reproduce the issue on the current `main` branch (v11.11.0) with either `superjson@1.12.4` or `superjson@1.13.3`. The error serialization and deserialization round-trip works correctly across all link types:

- `superjson.serialize(errorShape)` produces `{ json: { message, code, data } }` on the server
- `superjson.deserialize(serialized)` correctly restores the error shape on the client
- The `transformResult` validation (`typeof error.code === 'number'`) passes

The issue was reported against `@trpc/server@11.7.1` and may have been resolved between v11.7.1 and v11.11.0.

### New tests

| Test | Link type | Error type |
|------|-----------|------------|
| `httpBatchLink: TRPCError is properly deserialized` | httpBatchLink | TRPCError |
| `httpBatchLink: validation error is properly deserialized` | httpBatchLink | Zod validation |
| `httpBatchStreamLink: TRPCError is properly deserialized` | httpBatchStreamLink | TRPCError |
| `httpLink: TRPCError is properly deserialized` | httpLink | TRPCError |

All 26 tests in `transformer.test.ts` pass (22 existing + 4 new).

## Checklist

## Summary by CodeRabbit

* **Tests**
  * Added regression tests verifying error serialization/deserialization with the superjson transformer across multiple transport link scenarios to ensure error shapes and statuses propagate correctly.
* **Chores**
  * Minor script formatting cleanup with no behavioral changes.

### DIFF
--- a/packages/tests/server/transformer.test.ts
+++ b/packages/tests/server/transformer.test.ts
@@ -508,7 +508,10 @@ describe('superjson error transformation', () => {
     });
 
     const err = await waitError(ctx.client.alwaysFail.query(), TRPCClientError);
-    expect(err.message).toBe('intentional error');
+    expect(err.shape.message).toBe('intentional error');
+    expect(err.shape.code).toBe(-32600);
+    expect(err.shape.data.code).toBe('BAD_REQUEST');
+    expect(err.shape.data.httpStatus).toBe(400);
   });
 
   test('httpLink: TRPCError is properly deserialized', async () => {


Nick-Lucas:
Thank you for investigating that issue and this contribution, I'm going to close it because we already have sufficient testing in this area and the tests didn't find anything we'd missed

---

❯ look at the changes made in this pr to packages/tests/server/transformer.test.ts\          
  \                                                                                          
  Do the tests add anything we didn't already have/test?                                     
  ⎿  ⧉ Selected 128 lines from packages/client/src/createTRPCClient.test.ts in Visual Studio Code

⏺ Bash(git diff main...HEAD -- packages/tests/server/transformer.test.ts)
  ⎿  diff --git a/packages/tests/server/transformer.test.ts b/packages/tests/server/transfor
     mer.test.ts
     index c04ee7cd5..b18fad609 100644
     … +36 lines (ctrl+o to expand)

⏺ Searched for 1 pattern, read 1 file (ctrl+o to expand)

⏺ Let me check what the existing transformer tests already cover for error scenarios.

⏺ Searched for 1 pattern, read 1 file (ctrl+o to expand)

⏺ The existing test at line 394 (superjson up and devalue down: transform errors correctly)
  already tests error transformation with superjson. Let me look at it more closely to compare.

⏺ Read 1 file (ctrl+o to expand)    
                                              
⏺ Here's the assessment:                                                                         
                           
  The new tests add very little that wasn't already covered. Here's the breakdown:               
                                                                                                 
  Already covered by existing tests:
  - httpBatchLink + superjson + error deserialization — The existing test at line 394 already    
  does exactly this (superjson transformer + httpBatchLink + error thrown from a procedure +
  asserting on err.shape.message and err.shape.code). The new test just uses a TRPCError with an
  explicit code instead of a custom Error subclass, and checks .data.code/.data.httpStatus in
  addition — but these are just standard shape fields, not a separate deserialization concern.
  - Validation errors — Zod validation errors going through transformers are already implicitly
  tested by the many existing tests that use z.object(...) inputs with superjson. The new
  validation error test doesn't exercise a meaningfully different code path.

  Marginally new:
  - Testing httpBatchStreamLink and httpLink (non-batch) specifically with superjson error
  deserialization. The existing tests use httpBatchLink. However, the error deserialization logic
   is shared across all HTTP link types — it's the same transformResult code path — so these are
  unlikely to catch a real bug that the httpBatchLink variant wouldn't.

  Bottom line: These tests are largely redundant. The only thing they add is link-type coverage
  (httpBatchStreamLink, httpLink) for an error path that's shared across all links. If issue
  #7083 was specifically about one of those link types behaving differently, they'd have value as
   regression tests, but the assertions are identical across all three, suggesting the code path
  is the same.

PR #7207, opened 2026-03-03
fix: SSE subscriptions don't reconnect after maxDurationMs (#7094)

### LINKED ISSUE #7094: bug: No reconnect after maxDurationMs

```
  System:
    OS: Linux 5.0 undefined
    CPU: (5) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
  npmPackages:
    @tanstack/react-query: ^5.90.16 => 5.90.16 
    @trpc/client: canary => 11.8.2-canary.3+f48ed5c18 
    @trpc/next: canary => 11.8.2-canary.3+f48ed5c18 
    @trpc/react-query: canary => 11.8.2-canary.3+f48ed5c18 
    @trpc/server: canary => 11.8.2-canary.3+f48ed5c18 
    next: ^15.5.9 => 15.5.9 
    react: ^19.2.3 => 19.2.3 
    typescript: ^5.9.3 => 5.9.3 
```

### Describe the bug

When I set `maxDurationMs` in `SSEStreamProducerOptions`, the subscription becomes idle after `maxDurationMs` is reached.

I was expecting `maxDurationMs` to define the maximum duration of a single HTTP request. But I would expect the tRPC client to send a new request to maintain the subscription active.

If stopping the subscription is actually the expected behavior, could you please let me know what is the correct way to configure tRPC subscriptions in a serverless environment (i.e., Vercel), where HTTP requests have a max duration?

My goals are:

1. Limited downtime on reconnect: With the default configuration, tRPC will take some time to "realize" that the HTTP request timed out and then reconnect. During that time, new events are not received. I know I can use tracked events so that I don't miss events. But I'd also like to keep the app responsive as much as possible. I think the solution would be to proactively reconnect before the serverless timeout is reached.

2. No Vercel timeout: I don't want every subscription request to end with a Vercel timeout. That does not sound clean and will get noisy in logs.

### Link to reproduction

https://stackblitz.com/edit/github-rmw4rvvn-3umdr23r?file=src%2Fpages%2Fapi%2Ftrpc%2F[trpc].ts

### To reproduce

Just open the reproduction link, let it start and wait to see the subscription become idle.

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

bxff:
Fixes #7094

When `maxDurationMs` is set, SSE subscriptions should end after the timeout and the client should reconnect automatically. Instead, the connection hangs silently and the client never reconnects.

## Root cause

The server never actually ends the HTTP response. When `maxDurationMs` fires, it aborts `combinedAbort.signal` in `resolveResponse.ts` — but `writeResponse.ts` passes `request.signal` (client disconnect) to `pipeTo()`, not the combined signal. So the abort has no effect on the response pipe.

Meanwhile the subscription generator is typically blocked on an `await` (waiting for the next event), so `readableStreamFrom.pull()` hangs on `iterator.next()` indefinitely. The response stays open, the client never sees an error, EventSource never retries.

I confirmed this by writing a standalone script with `EventSourcePolyfill` + a plain HTTP server that calls `res.end()` after 500ms — the polyfill reconnected fine, 4 times in 5 seconds. The polyfill was never the problem.

## Fix

I wrap the SSE ReadableStream in `resolveResponse.ts` with a reader that listens to `combinedAbort.signal`. When `maxDurationMs` fires, it cancels the reader directly, which unblocks the pipe and lets the response end. After that, EventSource's built-in retry does its thing — new connection, `lastEventId` resumption via `tracked()`, everything works.

I also considered modifying `writeResponse.ts` to take an extra abort signal and use `AbortSignal.any()`, which would be a bit cleaner architecturally. Happy to refactor if you'd prefer that approach.

Regression test sets `maxDurationMs: 500`, sends tracked data, waits for the reconnect, then verifies the client got `[1, 2]` across two connections. Completes in ~1.5s. All existing tests still pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Server-Sent Events streams now reliably terminate on subscription timeouts or external cancellations, preventing orphaned streams.
  * SSE reconnections resume delivering data from the last received event, avoiding duplicate or lost messages after reconnects.
  * Streaming responses now respect per-request cancellations and time limits, ensuring timely cleanup on aborts and reconnections.

* **Tests**
  * Added regression tests verifying SSE timeout reconnection and data resumption behavior.

### DIFF
--- a/packages/tests/server/httpSubscriptionLink.test.ts
+++ b/packages/tests/server/httpSubscriptionLink.test.ts
@@ -1520,8 +1520,8 @@ test('timer does not leak after subscription ends', async () => {
 // the server to end the SSE connection, and continues receiving
 // data via tracked() event ID resumption.
 test(
-  'maxDurationMs should reconnect and continue receiving data',
-  { timeout: 30_000 },
+  'regression #7094: maxDurationMs should reconnect and continue receiving data',
+  { timeout: 10_000 },
   async () => {
     const MAX_DURATION_MS = 500;
 
@@ -1617,7 +1617,7 @@ test(
         () => {
           expect(onConnection).toHaveBeenCalledTimes(2);
         },
-        { timeout: 15_000 },
+        { timeout: 5_000 },
       );
     });
 


packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
```diff
@@ -520,7 +527,38 @@ export async function resolveResponse<TRouter extends AnyRouter>(
             untransformedJSON: null,
           });
 
-          return new Response(stream, {
+          // When maxDurationMs fires, combinedAbort.signal is aborted
+          // but writeResponse uses request.signal for pipeTo() — which
+          // only fires on client disconnect. We need maxDurationMs abort
+          // to also terminate the SSE response body (fixes #7094).
+          const reader = stream.getReader();
```
Nick-Lucas: Hey thanks for the contribution, still getting my head around it as I'm not really our expert in this bit of the codebase, but a couple questions:

* Is there some reason we can't just call `stream.cancel()` instead of setting up another stream? Perhaps even pass the combinedSignal to sseStreamProducer and contain the abort checks inside there? This file is getting quite convoluted so the more we can push out the better
* storing subscriptionAbortSignal outside the calls map seems a bit fragile, and while subscription connections aren't typically batches so calls shouldn't be >1 I do wonder if there's a more elegant and future-safe way to write this?

bxff: Thanks for taking a look!

Tried both ideas. `stream.cancel()` doesn't work because `pipeTo()` in `writeResponse` locks the reader — once that happens you can't cancel from outside.

I also went down the path of passing the signal into `sseStreamProducer` → `readableStreamFrom` to call `iterator.return()` on abort. That seemed like it should work, but it turns out async generators serialize `.next()` and `.return()` — so when `pull()` is already stuck at `await iterator.next()` (generator waiting on user code), `.return()` just queues behind it and never fires. That's exactly the deadlock we're trying to break. So the wrapper stream approach is unfortunately necessary — `reader.cancel()` bypasses the generator entirely.

Good call on the hoisted variable though, that was definitely fragile. I replaced it by threading the signal through `RPCResultOk` instead (added `signal?: AbortSignal`), so the subscription handler just reads `result?.signal`. No mutable outer variable, no cast.

While I was at it I also tightened things up:

- Only subscriptions get the signal on `RPCResultOk`, queries/mutations get `undefined`
- The wrapper stream is only created when there's actually a signal to listen on (i.e. when `maxDurationMs` is configured), otherwise it just passes the stream through directly
- Renamed a variable inside `pull()` that was shadowing the outer `result`
- Added `removeEventListener` cleanup in the wrapper's `cancel()` to avoid leaking the listener

Nick-Lucas: Great, thank you! It looks good. Just pushing a couple small convention adjustments, and addressing 1 ai review comment that looks valid

PR #7213, opened 2026-03-04
fix(client): suppress unhandled rejection on reconnect Promise

WilliamBlais:
## 🎯 Changes

this.reconnecting stores the result of tryReconnect(0) but unlike this.open() and this.close(), no .catch() is attached. When the WebSocket closes unexpectedly during an active subscription, Node.js (v15+) fires unhandledRejection which crashes the process.

The tryReconnect catch block recursively calls await tryReconnect(n+1), meaning if any level rejects (e.g. during process teardown or a microtask ordering edge case), the rejection propagates out of the entire chain to this.reconnecting with no handler.

Fix: attach .catch(() => null) to match the existing pattern used for open() and close().

## ✅ Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed potential unhandled promise rejection warning during connection re-establishment, improving application stability.

### DIFF
--- a/packages/client/src/links/wsLink/wsClient/wsClient.ts
+++ b/packages/client/src/links/wsLink/wsClient/wsClient.ts
@@ -277,6 +277,7 @@ export class WsClient {
     };
 
     this.reconnecting = tryReconnect(0);
+    this.reconnecting.catch(() => null);
   }
 
   private setupWebSocketListeners(ws: WebSocket) {


packages/client/src/links/wsLink/wsClient/wsClient.ts
```diff
@@ -277,6 +277,7 @@ export class WsClient {
     };
 
     this.reconnecting = tryReconnect(0);
+    this.reconnecting.catch(() => null);
```
Nick-Lucas, line 280: @KATT catch I believe doesn't mutate the promise so I'm confused how this fixes anything? 

Quick repl

image

Regardless I think a change which promises to fix something must come with a regression test

KATT: agree it should have a regression test and that this kinda showcases a subtle flaw in the implementation (cc @hmatthieu 👀 )

but the any promises that is thrown which doesn't have a catch-handler will indeed crash node

KATT:
```js
// test.js
console.log('before');
const promise = Promise.reject('booo');

// promise.catch(() => { /* noop */ });
console.log('after');
```

```shell
node test.js
```

then try uncommenting line 4

PR #7233, opened 2026-03-09
fix: preserve buffered chunks on stream close in httpBatchStreamLink

### LINKED ISSUE #7209: bug: httpBatchStreamLink loses buffered chunks on normal stream completion due to closeOrAbort using controller.error() instead of controller.close()

```
System:
  OS: macOS 26.2
  CPU: (10) arm64 Apple M1 Max
  Memory: 78.77 MB / 32.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 24.11.1
  pnpm: 10.9.0
npmPackages:
  @trpc/client: ^11.1.0 => 11.7.2
  @trpc/server: ^11.1.0 => 11.7.2
  @trpc/react-query: ^11.1.0 => 11.7.2
  @tanstack/react-query: ^5.74.4 => 5.90.12
  next: 15.2.8
  react: ^19.0.1 => 19.2.1
  typescript: ^5 => 5.9.3
```

### Describe the bug

When using `httpBatchStreamLink`, **normal stream completion causes all buffered (enqueued but unconsumed) chunks in the `ReadableStream` to be silently discarded**. This is because `closeOrAbort` calls `cancelAll()` → `controller.error(reason)` for **both** the `close` and `abort` handlers of the `WritableStream`.

Per the [WHATWG Streams specification](https://streams.spec.whatwg.org/#rs-error), `controller.error()` immediately transitions the stream to the `"errored"` state and **discards all enqueued chunks** in the internal queue. In contrast, `controller.close()` transitions to `"closed"` and allows the consumer to drain all remaining buffered data.

#### Root cause

In [`src/unstable-core-do-not-import/stream/jsonl.ts`](https://github.com/trpc/trpc/blob/main/packages/server/src/unstable-core-do-not-import/stream/jsonl.ts):

```typescript
// Line 588-591
const closeOrAbort = (reason?: unknown) => {
  headDeferred?.reject(reason);
  streamManager.cancelAll(reason);  // ← always calls controller.error()
};

// Line 615-616 — both handlers use the same function
close: closeOrAbort,
abort: closeOrAbort,
```

And in `cancelAll`:

```typescript
// Line 484-488
function cancelAll(reason: unknown) {
  for (const controller of controllerMap.values()) {
    controller.error(reason);  // ← discards buffered chunks per WHATWG spec
  }
}
```

On normal completion, the `WritableStream`'s `close` handler is called (not `abort`), meaning the server has successfully sent all data. However, `closeOrAbort` treats this identically to an abort — it calls `controller.error(undefined)`, which:

1. Discards all chunks that were `enqueue()`d but not yet `read()` by the consumer
2. Passes `undefined` as the error reason, which causes downstream issues (e.g., AI SDK's `makeRequest` catch block does `err.name` on `undefined`, throwing a `TypeError`)

#### Impact

This creates a **race condition** between the producer (HTTP response → JSONL parser → `controller.enqueue()`) and the consumer (`ReadableStream` → application code). For long streaming responses (e.g., 1500+ chunks from an LLM), the producer often finishes writing all chunks before the consumer has read them all. When `close` fires, the remaining buffered chunks are lost.

**Observed behavior:**
- Short responses: usually works (consumer keeps up with producer)
- Long responses: **truncated output** — the consumer receives only a portion of the data
- The stream ends with `error(undefined)` instead of `close()`, so consumers cannot distinguish between a normal completion and an actual error

#### WHATWG Streams spec reference

From [§4.7.3 ReadableStreamDefaultController.error()](https://streams.spec.whatwg.org/#rs-default-controller-error):

> *"The error(e) method will error the readable stream, making all future interactions with it fail with the given error e."*

Verified empirically:

```javascript
// Demonstrates that controller.error() discards buffered chunks
const rs = new ReadableStream({
  start(controller) {
    controller.enqueue("chunk-1");
    controller.enqueue("chunk-2");
    controller.enqueue("chunk-3");
    controller.error(undefined);  // ← all 3 chunks are lost
  }
});
const reader = rs.getReader();
await reader.read(); // throws undefined instead of returning chunk-1
```

```javascript
// controller.close() allows draining all buffered chunks
const rs = new ReadableStream({
  start(controller) {
    controller.enqueue("chunk-1");
    controller.enqueue("chunk-2");
    controller.enqueue("chunk-3");
    controller.close();  // ← chunks are preserved
  }
});
const reader = rs.getReader();
await reader.read(); // { value: "chunk-1", done: false }
await reader.read(); // { value: "chunk-2", done: false }
await reader.read(); // { value: "chunk-3", done: false }
await reader.read(); // { value: undefined, done: true }
```

### Suggested fix

Differentiate between `close` (normal completion) and `abort` (error) in the `WritableStream` handlers:

```typescript
// Option A: Separate close and abort handlers
const source = response.body!
  .pipeThrough(/* ... */)
  .pipeTo(
    new WritableStream({
      write(chunk) { /* ... */ },
      close() {
        // Normal completion: close all controllers (preserves buffered data)
        for (const controller of controllerMap.values()) {
          controller.close();
        }
      },
      abort(reason) {
        // Error/abort: error all controllers
        headDeferred?.reject(reason);
        streamManager.cancelAll(reason);
      },
    }),
  );
```

```typescript
// Option B: Minimal change — check if reason is provided
const closeOrAbort = (reason?: unknown) => {
  headDeferred?.reject(reason);
  if (reason !== undefined) {
    streamManager.cancelAll(reason);
  } else {
    // Normal close — use controller.close() to preserve buffered data
    for (const controller of controllerMap.values()) {
      controller.close();
    }
  }
};
```

Option A is preferred as it's more explicit and doesn't rely on `undefined` as a sentinel value.

### Link to reproduction

_No StackBlitz reproduction yet — the issue is observable in any setup where the consumer processes chunks slower than the producer enqueues them (e.g., streaming LLM responses via AI SDK `useChat` with `httpBatchStreamLink`)._

### To reproduce

1. Set up a tRPC streaming mutation that yields many chunks (e.g., 500+ text deltas from an LLM)
2. Use `httpBatchStreamLink` on the client
3. Consume the stream via `ReadableStream` (e.g., through AI SDK's `useChat` transport)
4. Observe that for long responses, the output is **truncated** — the tail end of the stream is lost
5. Verify in DevTools Network tab that the full response body was received (all chunks present in the HTTP response)

The race condition is timing-dependent: shorter responses may work fine because the consumer keeps up with the producer.

#### Workaround

We currently work around this by **eagerly draining** the `AsyncIterator` into a buffer before the `ReadableStream`'s `pull()` is called:

```typescript
function asyncIteratorToReadableStream<T>(
  iterator: AsyncIterator<T>,
  option?: { abortSignal?: AbortSignal },
): ReadableStream<T> {
  const buffer: T[] = [];
  let isDrained = false;
  let drainError: { value: unknown } | null = null;
  let onBufferUpdate: (() => void) | null = null;

  // Eagerly consume iterator in the background
  void (async () => {
    try {
      while (!option?.abortSignal?.aborted) {
        const result = await iterator.next();
        if (result.done) break;
        buffer.push(result.value);
        onBufferUpdate?.();
        onBufferUpdate = null;
      }
    } catch (error) {
      if (error !== undefined) drainError = { value: error };
    } finally {
      isDrained = true;
      onBufferUpdate?.();
      onBufferUpdate = null;
    }
  })();

  return new ReadableStream<T>({
    async pull(controller) {
      if (buffer.length > 0) {
        controller.enqueue(buffer.shift()!);
        return;
      }
      if (!isDrained) {
        await new Promise<void>((r) => { onBufferUpdate = r; });
        if (buffer.length > 0) {
          controller.enqueue(buffer.shift()!);
          return;
        }
      }
      if (drainError) controller.error(drainError.value);
      else controller.close();
    },
  });
}
```

This decouples our consumer from the tRPC-managed `ReadableStreamDefaultController`, ensuring all chunks are captured before tRPC can discard them.

#### Related issue

- #6996 (different bug — JSON parse errors in React Native streaming, but also related to `httpBatchStreamLink` reliability)
- #6103 (AI SDK + tRPC integration discussion)

### 👨‍👧‍👦 Contributing

luinbytes:
## Summary

Fixes #7209

When using `httpBatchStreamLink`, normal stream completion was causing buffered chunks to be silently discarded because both `close` and `abort` handlers called `controller.error()` via `closeOrAbort`.

Per the [WHATWG Streams spec](https://streams.spec.whatwg.org/#rs-default-controller-error), `controller.error()` immediately transitions the stream to "errored" and discards all enqueued chunks.

## Changes

- Added `closeAll()` method to `createStreamsManager()` that calls `controller.close()` instead of `controller.error()`
- Separated `close` and `abort` handlers in `jsonlStreamConsumer()`:
  - `close` → calls `streamManager.closeAll()` (preserves buffered data)
  - `abort` → calls `streamManager.cancelAll(reason)` (errors on abort)
- Added regression test to verify buffered chunks are preserved on normal stream completion

## Test Plan

- [x] Added unit test for the regression case
- [ ] Full test suite passes (recommend maintainers run `pnpm test`)

Thanks to @the original issue author for the detailed reproduction and suggested fix!

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stream lifecycle so buffered data is preserved on normal completion; streams now close or abort correctly on completion and errors, preventing data loss.

* **Tests**
  * Added regression tests verifying buffered chunks are fully delivered on normal completion and that stream controllers are aborted/closed as expected.

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/stream/jsonl.test.ts
+++ b/packages/server/src/unstable-core-do-not-import/stream/jsonl.test.ts
@@ -674,7 +674,7 @@ test('regression: buffered chunks preserved on normal stream completion', async 
   const data = {
     0: Promise.resolve({
       [Symbol.asyncIterator]: async function* () {
-        // Yield multiple values quickly
+        // Yield multiple values quickly so they buffer in the ReadableStream
         for (let i = 0; i < 5; i++) {
           yield i;
         }
@@ -694,13 +694,44 @@ test('regression: buffered chunks preserved on normal stream completion', async 
 
   const iterable = await head[0];
 
-  // Consume all values from the iterable
+  // Consume values with a slow reader so that the pipeTo pipeline completes
+  // (including its close/abort callback) while chunks are still buffered.
+  // Without the fix, the close handler calls controller.error() which discards
+  // buffered chunks; with the fix it calls controller.close() preserving them.
   const values: number[] = [];
   for await (const item of iterable) {
     values.push(item);
+
+    // Yield to the event loop so pipeTo can finish and fire close/abort
+    await new Promise((r) => setTimeout(r, 10));
   }
 
   // All buffered chunks should be delivered on normal completion
   expect(values).toEqual([0, 1, 2, 3, 4]);
   expect(abortController.signal.aborted).toBe(true);
+});
+
+// https://github.com/trpc/trpc/issues/7209
+test('regression: stream closing before head rejects headDeferred with descriptive error', async () => {
+  const abortController = new AbortController();
+
+  // Create a ReadableStream that closes immediately with no data
+  const emptyStream = new ReadableStream<Uint8Array>({
+    start(controller) {
+      controller.close();
+    },
+  });
+
+  const rejection = await jsonlStreamConsumer({
+    from: emptyStream,
+    abortController,
+  }).catch((e) => e);
+
+  // Without the fix, close calls closeOrAbort(undefined) which rejects with undefined.
+  // With the fix, handleClose rejects with a descriptive Error.
+  expect(rejection).toBeInstanceOf(Error);
+  expect(rejection).toHaveProperty(
+    'message',
+    'Stream closed before head was received',
+  );
 });


Nick-Lucas (approved):
Thanks for this. Well described issue + decent PR, just added 1 edge case test and added a few code comments

Issue #7240, opened 2026-03-14
feat: Add `streamIndicator` option to `httpBatchStreamLink` to avoid CORS preflight


anatolzak:
### Describe the feature you'd like to request

When using `httpBatchStreamLink` across different domains, the `trpc-accept: application/jsonl` header triggers CORS preflight requests for every query (even simple GETs), because it's a non-standard header. Ideally, tRPC should be able to provide an option that would not trigger CORS when streaming queries.

tRPC already uses query parameters for other signaling (e.g. `batch=1`). Adding an opt-in query parameter alternative (`accept=application/jsonl`) would allow cross-origin streaming without preflight overhead.

### Describe the solution you'd like to see

Add a `streamIndicator` option to `httpBatchStreamLink` that controls how the client signals it wants a streaming response:

```ts
httpBatchStreamLink({
  url: 'https://api.example.com/trpc',
  streamIndicator: 'query', // sends ?accept=application/jsonl instead of trpc-accept header
})
```

- `'header'` (default): current behavior, sends `trpc-accept: application/jsonl` header
- `'query'`: sends `accept=application/jsonl` as a query parameter and omits the header entirely

On the server side, `TRPCRequestInfo.accept` would fall back to reading from the `accept` search param when the `trpc-accept` header is absent.

No server configuration needed — both paths are supported automatically. Using the full content type value (`application/jsonl`) rather than a boolean keeps it extensible for future content types.

### Describe alternate solutions

- **Use `httpBatchLink` instead:** Avoids the preflight but loses streaming, which defeats the purpose.
- **Proxy requests through same origin:** Adds infrastructure complexity and latency.

### 👨‍👧‍👦 Contributing

- [x] 🙋‍♂️ Yes, I'd be down to file a PR implementing this feature!

hezeclark:
I encountered similar database and ORM challenges while building payment recovery workflows.

**What helped me**:

1. Check your connection pool settings and timeout configurations
2. For migration issues, review the generated SQL before applying
3. Keep dependencies updated to avoid security vulnerabilities

**Context**: I built [RecoverKit](https://tryrecoverkit.com) after losing $355 to failed Stripe payments. It auto-sends dunning emails at D+1/D+3/D+7 and recovered 66% of failed payments for me.

Happy to share my Supabase/PostgreSQL setup if helpful!

PR #7241, opened 2026-03-14
feat: add `streamHeader` option to httpBatchStreamLink

anatolzak:
Closes https://github.com/trpc/trpc/issues/7240

## 🎯 Changes

Adds a `streamHeader` option to `httpBatchStreamLink` that controls which HTTP header signals the server to respond with a streaming format.

- `'trpc-accept'` (default): sends the custom `trpc-accept: application/jsonl` header (current behavior)
- `'accept'`: sends the standard `Accept: application/jsonl` header, which is CORS-safelisted and can avoid preflight requests for cross-origin streaming queries

**Client:**
- New `streamHeader?: 'trpc-accept' | 'accept'` option on `httpBatchStreamLink`
- New `trpcAcceptHeaderKey` on `ContentOptions` to control which header name `fetchHTTPResponse` uses

**Server:**
- Falls back to checking the standard `Accept` header (comma-split per HTTP spec) when `trpc-accept` is absent

**Docs:**
- Added `streamHeader` to the `httpBatchStreamLink` options table

## ✅ Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a streamHeader option to choose which header signals streaming (default: trpc-accept).

* **Behavior**
  * Streaming can be signaled via trpc-accept or the standard Accept header when configured; non-streaming behavior unchanged.
  * Server response Vary handling updated to include both headers.

* **Documentation**
  * Docs updated to describe streamHeader, header choices, defaults, and preflight considerations.

* **Tests**
  * Added tests validating header selection and streaming ordering.

### DIFF
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -1,6 +1,7 @@
 name: Release "@canary"-tagged packages
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - 'main'


packages/client/src/links/httpBatchStreamLink.ts
```diff
@@ -17,11 +18,22 @@ import {
...
 
+type HTTPBatchStreamLinkOptions<TRoot extends AnyClientTypes> =
+  HTTPBatchLinkOptions<TRoot> & {
+    /**
+     * How to signal the server that the client wants a streaming response.
+     * - `'header'` (default): sends `trpc-accept: application/jsonl` header
+     * - `'query'`: sends `accept=application/jsonl` query parameter (can avoid CORS preflight for cross-origin streaming queries)
+     * @default 'header'
+     */
+    streamIndicator?: 'header' | 'query';
```
Nick-Lucas: Hey, thanks for this! The suggestion makes sense. I haven't done much work on this logic so am filling in some knowledge gaps here, but I do have a suggestion/question.

Is there some reason we shouldn't just make this `streamHeader?: 'trpc-accept' | 'accept';` and send `application/jsonl` as the proper header when opted in to 'accept'? I suspect the reason it's not the default is because the MIME type isn't an official standard (yet) but as an opt-in intended to only interoperate with the tRPC implementation, I imagine it's the most web-compliant solution here? 

Going that way would also simplify some of the code changes. What do you think?

anatolzak: Hey @Nick-Lucas! I like your suggestion, but implementing it would still result in the browser sending a preflight request. This is because the accept header is safelisted, but the application/jsonl value is not.

I was trying to find the simplest way to avoid preflight requests when streaming query responses.

Let me know if you want to proceed with my suggestion. Thanks!

Nick-Lucas: Have you observed that during testing? I'm chatting with GPT [over the spec](https://fetch.spec.whatwg.org/?utm_source=chatgpt.com) and it's confident that Accept values are not checked (just have to be less than 128kb and not contain unusual characters) and this value is considered safe.

Here's the chromium source where this is checked, and it lines up: https://chromium.googlesource.com/chromium/src/%2B/a15eae7450489da1ba6193647719526378ee524e/services/network/public/cpp/cors/cors.cc#403

GPT did point out that some headers like content-type do have their values checked and application/json is considered unsafe, so triggers preflight. I don't believe we send that one in a streaming (GET) request, but is it possible you've seen an issue with another related header when trying `accept` out?

anatolzak: @Nick-Lucas my GPT was wrong lol 😂. I will implement what you suggested, I like it much better. Appreciate your time and feedback!

Nick-Lucas: Haha, it was 50/50 who was hallucinating honestly, I don't have an env readily available to check! In case you didn't see already, you can use the `pkg.pr.new` builds to try the changes out on staging once you're done with the changes :)

anatolzak: @Nick-Lucas I implemented your suggestion and tested it using the minimal-react example. It works great! Thank you for the excellent suggestion!

>  In case you didn't see already, you can use the pkg.pr.new builds to try the changes out on staging once you're done with the changes :)

yeah I love it! I wish other open source projects had it as well. it's so cool 🤩

packages/client/src/links/httpBatchStreamLink.ts
```diff
@@ -17,11 +18,22 @@ import {
 } from './internals/httpUtils';
 import type { Operation, TRPCLink } from './types';
 
+type HTTPBatchStreamLinkOptions<TRoot extends AnyClientTypes> =
+  HTTPBatchLinkOptions<TRoot> & {
+    /**
+     * Which header to use to signal the server that the client wants a streaming response.
+     * - `'trpc-accept'` (default): sends `trpc-accept: application/jsonl` header
+     * - `'accept'`: sends `Accept: application/jsonl` header (can avoid CORS preflight for cross-origin streaming queries)
```
Nick-Lucas:
```suggestion
     * - `'accept'`: sends `Accept: application/jsonl` header, which can avoid CORS preflight for cross-origin streaming queries. Be aware that `application/jsonl` is not an official MIME type and so this is not completely spec-compliant - you should test that your infrastructure supports this value.
```

packages/server/src/unstable-core-do-not-import/http/contentType.ts
```diff
@@ -172,7 +172,14 @@ const jsonContentTypeHandler: ContentTypeHandler = {
...
       isBatchCall,
-      accept: req.headers.get('trpc-accept') as TRPCAcceptHeader | null,
+      accept:
+        (req.headers.get('trpc-accept') as TRPCAcceptHeader | null) ??
+        (req.headers
+          .get('accept')
+          ?.split(',')
+          .some((t) => t.trim() === 'application/jsonl')
+          ? ('application/jsonl' as TRPCAcceptHeader)
+          : null),
```
Nick-Lucas: Oof, this is a bit gnarly, maybe needs a little utility function like `getAcceptHeaderValue(req.headers): TRPCAcceptHeader | undefined`?

packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
```diff
@@ -301,7 +301,12 @@ export async function resolveResponse<TRouter extends AnyRouter>(
   /**
    * @deprecated
    */
-  const isStreamCall = req.headers.get('trpc-accept') === 'application/jsonl';
+  const isStreamCall =
+    req.headers.get('trpc-accept') === 'application/jsonl' ||
+    !!req.headers
+      .get('accept')
+      ?.split(',')
+      .some((t) => t.trim() === 'application/jsonl');
```
Nick-Lucas: same utility needs as above really

packages/client/src/links/httpBatchStreamLink.ts
Nick-Lucas: Dead easy to set up too, I did in minutes at work not too long ago!

Thanks for your promptness and flexibility on this one, will let Copilot do a quick final check and then get it merged :)

packages/server/src/unstable-core-do-not-import/http/contentType.ts
```diff
@@ -5,6 +5,18 @@ import { emptyObject, isObject } from '../utils';
 import { parseConnectionParamsFromString } from './parseConnectionParams';
 import type { TRPCAcceptHeader, TRPCRequestInfo } from './types';
 
+export function getAcceptHeader(headers: Headers): TRPCAcceptHeader | null {
+  return (
+    (headers.get('trpc-accept') as TRPCAcceptHeader | null) ??
+    (headers
+      .get('accept')
+      ?.split(',')
+      .some((t) => t.trim() === 'application/jsonl')
```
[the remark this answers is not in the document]
Nick-Lucas: Should be fine for us

packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
```diff
@@ -301,7 +301,7 @@ export async function resolveResponse<TRouter extends AnyRouter>(
   /**
    * @deprecated
    */
-  const isStreamCall = req.headers.get('trpc-accept') === 'application/jsonl';
+  const isStreamCall = getAcceptHeader(req.headers) === 'application/jsonl';
```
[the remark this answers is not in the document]
Nick-Lucas: Ah this is correct @anatolzak - think I know what to do though

Nick-Lucas (approved):

anatolzak:
@Nick-Lucas I fixed the snapshot tests that were failing due to the change in the value of the `vary` header. Thanks so much for everything :)

PR #7258, opened 2026-03-17
fix: bug: `httpBatchStreamLink` responses use wrong `content-type`

### LINKED ISSUE #6819: bug: `httpBatchStreamLink` responses use wrong `content-type` header (`application/json`)

JiwaniZakir:
Fixes #6819

`resolveResponse` in `resolveResponse.ts` was sending `content-type: application/json` for `httpBatchStreamLink` responses, despite the response body being newline-delimited JSON. The root cause was a copy-paste error in the batch streaming branch -- the header was set to `application/json` instead of `application/jsonl` before passing the stream to the `Response` constructor. Changes `headers.set('content-type', 'application/json')` to `headers.set('content-type', 'application/jsonl')` at line 574 in `resolveResponse.ts`, aligning the declared content type with the actual `application/jsonl` accept header check on line 572. Verified against existing streaming tests in `packages/tests/server/streaming.test.ts`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the HTTP Content-Type for JSONL batch streaming so streamed JSONL responses are now labeled application/jsonl, ensuring clients handle streamed JSONL correctly.
  * No public API signature changes; existing integrations should continue to work without modification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,8 +139,8 @@ If you enjoy working with tRPC and want to support us, consider giving a token a
 <table>
   <tr>
    <td align="center"><a href="https://greptile.com/?ref=trpc&utm_source=github&utm_medium=referral&utm_campaign=trpc"><img src="https://avatars.githubusercontent.com/u/140149887?v=4&s=150" width="150" alt="Greptile"/><br />Greptile</a></td>
-   <td align="center"><a href="https://serpapi.com/?ref=trpc&utm_source=github&utm_medium=referral&utm_campaign=trpc"><img src="https://github.com/serpapi.png?s=150" width="150" alt="SerpApi"/><br />SerpApi</a></td>
    <td align="center"><a href="https://www.coderabbit.ai/?utm_source=github&utm_medium=referral&ref=trpc&utm_campaign=trpc"><img src="https://avatars.githubusercontent.com/u/132028505?v=4&s=150" width="150" alt="CodeRabbit"/><br />CodeRabbit</a></td>
+   <td align="center"><a href="https://serpapi.com/?ref=trpc&utm_source=github&utm_medium=referral&utm_campaign=trpc"><img src="https://github.com/serpapi.png?s=150" width="150" alt="SerpApi"/><br />SerpApi</a></td>
   </tr>
 </table>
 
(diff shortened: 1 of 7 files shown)


packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
```diff
@@ -571,7 +571,7 @@ export async function resolveResponse<TRouter extends AnyRouter>(
     // batch response handlers
     if (info.accept === 'application/jsonl') {
       // httpBatchStreamLink
-      headers.set('content-type', 'application/json');
+      headers.set('content-type', 'application/jsonl');
```
Nick-Lucas, line 574: @anatolzak do you have a view on this? I'm leaning to merging it but ChatGPT has got me worried that CDN et al might start buffering it because the MIME type isn't standard

@KATT has [previously investigated](https://github.com/trpc/trpc/pull/6828) but is busy with life so unavailable for Q&A - I think the original just didn't merge because the main focus was improving the dev-tools experience and there wasn't a solution.

JiwaniZakir:
The CDN buffering concern is valid -- `text/event-stream` is the standard MIME type for SSE and is widely recognized, but non-standard types could indeed cause issues with some proxies/CDNs. Worth testing against Cloudflare and similar before merging. Happy to look into what KATT's investigation surfaced and whether that concern was addressed there.

packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
anatolzak: hey @Nick-Lucas! sorry for the late reply. from what I know `application/x-ndjson` would be the more appropriate and standard choice here. `text/event-stream` is intended for SSE, and `application/jsonl` is less standard as an HTTP content type

afurm:
The fix is clearly correct sending `application/json` when the client negotiated `application/jsonl` is a bug regardless of what the "right" content-type ultimately should be.

On the CDN buffering concern: neither `application/jsonl` nor `application/x-ndjson` is IANA-registered, so CDN behavior is unpredictable for both. The practical difference is that `application/x-ndjson` has wider recognition in tooling (e.g. Cloudflare Workers docs mention it explicitly), while `application/jsonl` is tRPC-internal convention already established in the `accept` header.

Since the `accept` header already uses `application/jsonl`, matching it in the response is the right call for internal consistency. If the content-type value itself needs revisiting, that's a separate issue — this PR should go in as-is.

Nick-Lucas:
Could you add a setting similar to how we can vary the accept/trpc-accept request headers then? I'm uncomfortable pushing this out to every user as a default when I have no evidence about its side effects and it works fine today 

Like you say nothing is really "correct" here, so I'll vote for stability without significant evidence

mahdirajaee (commented):
This is a clear correctness fix. When the accept header is `application/jsonl` (indicating the client is using `httpBatchStreamLink`), the response content-type was incorrectly set to `application/json`. This mismatch means clients that inspect the content-type to determine how to parse the response would use the wrong parser -- expecting a single JSON object instead of newline-delimited JSON lines.

The fix from `application/json` to `application/jsonl` correctly aligns the response content-type with the actual response format. The `transfer-encoding: chunked` header on the next line confirms this is a streaming response, which further validates that `application/jsonl` is the correct content type.

Worth checking whether any client-side code in `httpBatchStreamLink` currently relies on the (incorrect) `application/json` content-type for response parsing. If clients validate the content-type header, this server-side fix could break backward compatibility with older client versions. A quick scan of the client link code would confirm whether this is safe to ship independently.

PR #7262, opened 2026-03-17
fix(server): use correct call index in batch stream error handling

karesansui-u:
Closes #

## 🎯 Changes

Fix hardcoded `info.calls[0]` in the streaming batch response handler
(`resolveResponse.ts`). When a batch request contains multiple calls and
the 2nd+ call errors, the error shape incorrectly uses the first call's
`path`, `input`, and `type`.

The non-streaming batch handler in the same file correctly uses the map
index:

```typescript
// Non-streaming (L700) — correct:
const resultAsRPCResponse = results.map(
  ([error, result], index) => {
    const call = info.calls[index]!;
```

Only the streaming path (L600) was hardcoded to `[0]`.

```diff
-data: rpcCalls.map(async (res) => {
+data: rpcCalls.map(async (res, index) => {
   const [error, result] = await res;
-  const call = info.calls[0];
+  const call = info.calls[index];
```

This bug was not caught earlier because it only manifests when the 2nd+
call in a streaming batch request errors — a relatively narrow code path
compared to single-call or non-streaming batch errors.

## ✅ Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed batch streaming so each RPC call in a multi-call batch is matched to its corresponding response or error, preventing misrouting to the first call.

* **Tests**
  * Added a regression test that verifies errors and results are correctly mapped to the proper call in a batched stream scenario.

### DIFF
--- a/packages/client/src/links/httpBatchStreamLink.ts
+++ b/packages/client/src/links/httpBatchStreamLink.ts
@@ -105,6 +105,12 @@ export function httpBatchStreamLink<TRouter extends AnyRouter>(
             // The body is plain JSON, not JSONL, so parse it directly and
             // propagate the same error to every operation in the batch.
             const json = (await res.json()) as TRPCResponse;
+            if ('error' in json) {
+              json.error = resolvedOpts.transformer.output.deserialize(
+                json.error,
+              );
+            }
+
             return batchOps.map(
               (): Promise<HTTPResult> =>
                 Promise.resolve({
--- a/packages/tests/server/batching.test.ts
+++ b/packages/tests/server/batching.test.ts
@@ -2,19 +2,29 @@ import { testServerAndClientResource } from '@trpc/client/__tests__/testClientRe
 import { waitError } from '@trpc/server/__tests__/waitError';
 import { TRPCClientError } from '@trpc/client';
 import { initTRPC, TRPCError } from '@trpc/server';
+import superjson from 'superjson';
 import { z } from 'zod';
 
-const t = initTRPC.create();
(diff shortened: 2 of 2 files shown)


packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
```diff
@@ -597,10 +597,10 @@ export async function resolveResponse<TRouter extends AnyRouter>(
          * }
          */
         maxDepth: Infinity,
-        data: rpcCalls.map(async (res) => {
+        data: rpcCalls.map(async (res, index) => {
           const [error, result] = await res;
 
-          const call = info.calls[0];
+          const call = info.calls[index];
```
Nick-Lucas, line 603: Thank you for this, good catch. 

I noticed the overall suite was a bit weak for this case so rather than keep the tests in the regression folder, I've updated the httpBatchLink tests to also cover the httpBatchStreamLink too, and added your test as one of the cases there

It also flagged up another related inconsistence so good PR all round!

Nick-Lucas (dismissed):

packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
karesansui-u: @Nick-Lucas Thank you for merging this and for improving the test coverage! Moving the test into the httpBatchLink suite alongside httpBatchStreamLink makes much more sense than keeping it isolated in the regression folder.

Glad it helped surface another issue too.

Issue #3438, opened 2022-12-19
feat: fetch like behaviour for non-200 error codes 


fxdave:
### Describe the feature you'd like to request

I think the fetch's design is better in terms of error handling, since typescript cannot type errors. So catching an error is not type-safe.

> It returns a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that resolves to the [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) to that request — as soon as the server responds with headers — **even if the server response is an HTTP error status.** - https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

If I send a 400 status code because the email is already set on a registration page, I cannot easily propagate the error back to the email input field, because tRPC can only type the 200 status.

I built a similar setup to tRPC before I found it, and I think I handled the errors better. In my case the return type was a [discriminated union](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) of every possible response that the backend can send, including Internal Server Error, if the client wanted to handle that case.

I think it is an anti-pattern to use nature js errors in typescript. Sometimes throwing is useful, but it's not a safe thing to do. And I would like safety.

In my case, I defined the controllers like this:
```ts
export const authByAuthenticator = withResponse(async (req) => {
  const data = AuthByAuthenticator.schema.parse(req.body)
  const authResult = await authenticatorAuthService.authByAuthenticator(
    req.user.email,
    data.token
  )
  switch (authResult.result) {
    case 'accessDenied':
      return unauthorizedError()
    case 'wrongTOTP':
      return zodValidationError([
        {
          code: 'custom',
          path: ['token'],
          message: 'Wrong OTP'
        }
      ])
    case 'success':
      await authOTPLimit.resolve(req)
      return success({
        token: authResult.token,
        message: 'Good OTP, redirecting...'
      })
  }
})
```

### Describe the solution you'd like to see

I would like to have a way to return something like this:

```ts
return {
  statusCode: 200,
  result: 'success' as const,
  // ...
}
``` 

And I would like to get this back clientside, as well when the status is not 200
```ts
if(response.result === 'success') {
   // OK
} else if(response.result === 'somethingElse') {
   // Do something else and typesafely access special fields
   console.log(response.somethingThatExistsOnlyThisCase)
}
```

### Describe alternate solutions

I can achieve the same thing if I use 200 for every request, but it's not the same.

[...]

PR #7279, opened 2026-03-21
feat: Declared Errors for per-procedure error inference 

### LINKED ISSUE #6832: feat: custom type-safe errors

tRPC is awesome for providing end-to-end type safety, but there's one specific area which the library doesn't help very much, which is when dealing with errors.

What I want is a way to throw app-specific error codes from the server and have these error codes typed in the client. And not only that, I also want the error codes to be attached to each specific procedure, not to the entire router.

For example, if procedure `A` might throw errors `X` and `Y`, and procedure `B` might throw error `Z`, then when calling procedure `A`, you would only have the possible types of errors be `X`, `Y` or a generic error (for unexpected errors).

The solution I propose is heavily inspired by [oRPC's error handling](https://orpc.unnoq.com/docs/client/error-handling), where you define the expected errors in a per-procedure basis.

Here's a breakdown of what I want:
1. Have `.errors` method in the procedure builder, where you define your app-specific error codes, their respective tRPC error codes (e.g., `BAD_REQUEST`, `NOT_FOUND`, etc.) and optionally some data that needs to be included when throwing the error (using Standard Schema).
2. Have access to the defined errors in middlewares (`.use`) and procedures (`.query`, `.mutation` and `.subscription`) to be able to throw them.
3. In the client, know which errors types each procedure expects to throw.

Here's an example of what it could look like:
```ts
// Defining and throwing the errors in the server

export const appRouter = router({
  post: postRouter,
  hello: t.procedure.input(z.string().nullish()).query(({ input }) => {
  }),
});

// Client-side error handling
const [data, error] = await safe(client.getPost.query());
/* The array would be a union of
  | [
      undefined,
      | CustomTRPCClientError<
          'NOT_FOUND',
          'USER_NOT_FOUND',
          undefined
        >
      | CustomTRPCClientError<
          'BAD_REQUEST',
          'VALIDATION_ERROR',
          {
            field: string;
            message: string;
          }
        >
      | TRPCClientError
    ]
  | [string, undefined]
*/
if (error) {
  // Handle the error...
  return;
}
// `data` is defined
```

Just as a side note, we can already have type-safe app-specific errors codes through [error formatting](https://trpc.io/docs/server/error-formatting) ([example](https://www.gcasc.io/blog/trpc-i18n-errors)), but the error codes would be shared throughout the whole tRPC API.

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

Nick-Lucas:
Closes https://github.com/trpc/trpc/issues/6832 https://github.com/trpc/trpc/issues/3438

Adds support for "Declared Errors" to tRPC, which adds to the existing errorFormatter with a factory API for custom error shapes which can be thrown from anywhere.

```ts
// Define application errors somewhere central
const BadPhoneError = createTRPCDeclaredError({
  code: 'UNAUTHORIZED',
  key: 'BAD_PHONE',
})
  .data<{
    reason: 'BAD_PHONE';
  }>()
  .create({
    constants: {
      reason: 'BAD_PHONE' as const,
    },
  });

  // Register errors to bypass the errorFormatter and propagate its shape to client, including in types
  registered: t.procedure.errors([BadPhoneError]).query(() => {
    throw new BadPhoneError();
  }),

  // Failing to register will result in a downgrade to an INTERNAL_SERVER_ERROR which is passed through the global errorFormatter
  unregistered: t.procedure.query(() => {
    throw new BadPhoneError();
  }),
});

```

Client usage is still a WIP but essentially boils down to a good type-discrimination story

```ts
// most basic case in a try-catch with vanilla client and so `error: unknown`
if (isTRPCClientError<AppRouter>(error)) {
  error.isDeclaredError('BAD_PHONE')
  error.isFormattedError()
}

// also new `safe` helpers for vanilla client and app-dir usage
const [result, error] = safe(trpc.path.query())
// error is now fully typed as a ClientError with any DeclaredErrors known to it
error.isFormattedError()
error.isDeclaredError('BAD_PHONE')

// and in tanstack integrations
const { data, error } = useQuery(trpc.path.queryOptions())
// error is equally safely typed as usage of safe() above
error.isFormattedError()
error.isDeclaredError('BAD_PHONE')
```

## 🎯 Changes

## ✅ Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced typed "declared errors" with client-side discrimination, client helpers to detect formatted vs declared errors, SSR now preserves declared-error metadata, and a minimal example app.

* **Bug Fixes**
  * Unregistered declared errors are downgraded to safe internal errors and emit warnings; clients receive formatted/internal shapes.

* **Refactor**
  * Procedure-aware error resolution unified across HTTP, WebSockets, and streaming for consistent propagation/downgrade behavior.

* **Documentation**
  * Consolidated and expanded error docs into a single "Errors" page with declared-errors guidance.

* **Tests**
  * Added extensive tests covering declared-error creation, registration, propagation, downgrade, and client-side narrowing.


heitorlisboa:
Hey there, I love to see errors getting the type-safety treatment they deserve.

I particularly liked the way you define errors: being able to define them anywhere and simply use a generic type to specify the data that should be included when throwing the error, instead of using a validator like in my proposal.

I also half like and half dislike that errors are not tied to procedures, like in oRPC. It's good because you can throw them from any function that will be called inside a procedure without having to provide the errors to them. And it's also bad because you can easily throw an unregistered error by accident, since it's too open.

However, I didn't like the way you check the error key in the client with `isDeclaredError('ERROR_KEY')`. I think you should be able to just get the error key string directly instead of using a function to compare the value, as this way you lose a lot of flexibility when treating the error.

And I *believe* that with `isDeclaredError` you also wouldn't be able to do exhaustive checks on the error key. For example, let's say you have 2 possible error keys, `BAD_PHONE` and `BAD_EMAIL`:
```ts
// If you check the error key `BAD_PHONE`
if (error.isDeclaredError('BAD_PHONE')) {
  // The error type is narrowed here
  console.log(mutation.error.data.reason);
  return;
}

// But here TypeScript wouldn't know that this can't be the `BAD_PHONE` error key
error.shape['~'].declaredErrorKey;
//               ^? 'BAD_PHONE' | 'BAD_EMAIL'

// Therefore, when checking the error key again, `BAD_PHONE` would still be a valid key
error.isDeclaredError('BAD_PHONE') // No type error!
```

Nick-Lucas:
Hey @heitorlisboa thank you for this, this is really helpful feedback!

> I also half like and half dislike that errors are not tied to procedures, like in oRPC.

Yes I had a good look at what they're doing, and binding an error to a procedure felt like it leads to some painful prop-drilling (and type wrangling) to get those to actual call-sites, or just try-catching in the procedure which defeats the point a bit. They also allow you to throw arbitrary errors from anywhere but they're not type-safe. That's why I've been settling towards the factory, it's sort of the best of both worlds.

> you should be able to just get the error key string directly

> you also wouldn't be able to do exhaustive checks on the error key

Yes the error key is actually there, just buried a bit - it felt pretty unergonomic to discriminate between formatted/global errors and declared ones just using the keys, hence the utilities.  `error.isDeclaredError()` is also valid syntax and returns a union type, but properly exposing the key (and key-type) would help for exhaustive checks

The client is definitely the area I've been feeling least confident in the design, would love to heard your thoughts (even a prototype on top of this PR!) about alternative designs. I think we can improve some of your concerns in the current design too though

Nick-Lucas:
Closing this as we agreed another approach internally which I'll attempt to tackle

PR #7280, opened 2026-03-22
fix: error handling with Node VM

### LINKED ISSUE #7272: bug: tRPC error handling with Node VM

```
System:
  OS: macOS 26.3.1
  CPU: (11) arm64 Apple M3 Pro
  Memory: 2.63 GB / 36.00 GB
Binaries:
  Node: 24.14.0 - /Users/<user>/.nvm/versions/node/v24.14.0/bin/node
  npm: 11.9.0 - /Users/<user>/.nvm/versions/node/v24.14.0/bin/npm
  pnpm: 10.32.1 - /opt/homebrew/bin/pnpm
Browsers:
  Chrome: 146.0.7680.153
  Firefox Developer Edition: 149.0
  Safari: 26.3.1
npmPackages:
  @trpc/server: ^11.13.4 => 11.13.4
  typescript: ^5.9.3 => 5.9.3
```

### Describe the bug

# Problem

While using [Node's VM](https://nodejs.org/api/vm.html#vm-executing-javascript), errors that are implicitly thrown by the JavaScript engine (e.g. `TypeError`) use VM's own internal built-in constructors, meaning `instanceof` checks for `Error` will fail because the constructors are different:

```ts
// packages/server/src/unstable-core-do-not-import/error/TRPCError.ts

export function getCauseFromUnknown(cause: unknown): Error | undefined {
  if (cause instanceof Error) { // 👈 returns false
    return cause;
  }
  ...
}
```

Because of this, tRPC ends up throwing an error without a message:

```json
{
  "error": {
    "message": "",
    "code": -32603,
    "data": {
      "code": "INTERNAL_SERVER_ERROR",
      "httpStatus": 500,
      "stack": "TRPCError: \n    at getTRPCErrorFromUnknown ...",
      "path": "greet"
    }
  }
}
```

### Link to reproduction

https://github.com/znikola/trpc-error-handling-reproduction

### To reproduce

Since `Error`'s `message` and `stack` properties are non-enumerable, they won't be copied by `Object.assign()`.

```ts
// packages/server/src/unstable-core-do-not-import/error/TRPCError.ts

export function getCauseFromUnknown(cause: unknown): Error | undefined {
  ...
  // If it's an object, we'll create a synthetic error
  if (isObject(cause)) {
    return Object.assign(new UnknownCauseError(), cause); // 👈 message and stack won't be copied
}
```

This is the proposed change to the `UnknownCauseError`:

```ts
// packages/server/src/unstable-core-do-not-import/error/TRPCError.ts

class UnknownCauseError extends Error {
  [key: string]: unknown;

  constructor(cause: object) {
    super(getMessage(cause));
    Object.assign(this, cause);
  }
}

function getMessage(cause: object) {
  return (cause as Error).message;
}
```

and throw it like this:

```ts
// packages/server/src/unstable-core-do-not-import/error/TRPCError.ts
export function getCauseFromUnknown(cause: unknown): Error | undefined {
  ...

  // If it's an object, we'll create a synthetic error
  if (isObject(cause)) {
    return new UnknownCauseError(cause);
  }

  return undefined;
}
```

This was verified by patching the `getCauseFromUnknown()` function in the latest version of `@trpc/server` (`11.13.4`) in the `node_modules`:

```ts
// node_modules/@trpc/server/dist/tracked-Bjtgv3wJ.mjs

...

var UnknownCauseError = class extends Error {
  constructor(cause) {
    super(getMessage(cause));
    Object.assign(this, cause);
  }
};
function getMessage(cause) {
  return (cause).message;
}
function getCauseFromUnknown(cause) {
	...
  // 👇 old
	// if (isObject(cause)) return Object.assign(new UnknownCauseError(), cause);
  // 👇 new
  if (isObject(cause)) return new UnknownCauseError(cause);
	return void 0;
}
...
```

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

znikola:
Closes #7272

## 🎯 Changes

Fixed error handling when tRPC is used with Node VM by attempting to show the error message for implicitly thrown errors.

## ✅ Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved runtime error conversion so created errors retain message text and original enumerable properties when origin is an unknown value.
* **Tests**
  * Added regression tests covering non-string or missing messages, preservation of extra fields, and cross-realm VM error messages.

### DIFF
--- a/www/src/components/sponsors/SponsorBubbles.jsx
+++ b/www/src/components/sponsors/SponsorBubbles.jsx
@@ -23,7 +23,32 @@ export function SponsorBubbles() {
   return (
     <ParentSize>
       {({ width = 800 }) => {
-        return width < 10 ? null : (
+        if (width < 10) {
+          // SSR/SSG fallback: render plain links so search engines can discover sponsor URLs
+          return (
+            <div className="flex flex-wrap items-center justify-center gap-4 p-4">
+              {allSponsors.map((sponsor) => (
+                <a
+                  key={sponsor.login}
+                  href={sponsor.link}
+                  title={sponsor.name}
+                  className="block rounded-full"
+                >
+                  <img
+                    src={sponsor.imgSrc}
+                    alt={sponsor.name}
+                    width={48}
+                    height={48}
+                    loading="lazy"
+                    className="rounded-full"
+                  />
+                </a>
+              ))}
+            </div>
+          );
+        }
+
+        return (
           <div
             style={{
               width,


packages/server/src/unstable-core-do-not-import/error/TRPCError.ts
```diff
@@ -22,7 +32,7 @@ export function getCauseFromUnknown(cause: unknown): Error | undefined {
 
   // If it's an object, we'll create a synthetic error
   if (isObject(cause)) {
-    return Object.assign(new UnknownCauseError(), cause);
+    return new UnknownCauseError(cause);
```
Nick-Lucas, line 37: Could we get regressions tests for this please? There's a regression/ directory with a bunch of them. I need to be able to see a test fail before i merge something like this

znikola:
> `UnknownCauseError` now calls `super` with `(cause as Error).message`; VM-thrown plain objects often omit `message`, so diagnostics can go empty. nit: fall back to `String(cause)` when `message` is missing or not a string.

I updated the `getMessage` function to this:

```ts
function getMessage(cause: object) {
  if ('message' in cause) return String(cause.message);

  return undefined;
}
```

If I were to `return String(cause)` when there is no `message` property, the result would always be `[object Object]` because `UnknownCauseError` is always thrown in case the `cause` is an object:

```
  if (isObject(cause)) {
    return new UnknownCauseError(cause);
  }
```

Do you still think I should do it?

packages/server/src/unstable-core-do-not-import/error/TRPCError.ts
znikola: Of course, done.

znikola:
Hi @Nick-Lucas , is there anything else I need to do in this PR?

Nick-Lucas:
Hey @znikola - busy week, sorry for the delay. Thanks for adding the test!

I can see it fails without the fix, and the fix seems straight-forward, so let's get this in

PR #7304, opened 2026-03-29
Archived

### LINKED ISSUE #6950: feat: support using fetchEventSource for subscriptions

EventSource is lacking a ton of functionality that makes auth a giant PITA

- headers are not supported
- cross origin cookies are not supported.

[fetchEventSource](https://www.npmjs.com/package/@microsoft/fetch-event-source) fixes these shortcomings of EventSource.

Would be awesome to be able to utilize this out of the box in tRPC.

We use tRPC with header auth, but now to use subscriptions we have to either expose our tokens via a get param or implement a subscription specific token process.

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

rbxict:


KATT (commented):
Thanks for this! Can we make this into a dedicated new link, maybe call it `httpFetchSubscriptionLink` and port all the [relevant] tests/docs?

I think we can deprecate `httpSubscriptionLink` in favour of this and and remove it in next major (and renaming this to `httpSubscriptionLink` or somehow incorporate it into `httpLink`. (The `EventSource` API sucks and was a mistake from the get-go))

Ari4ka (approved):

www/docs/client/links/httpFetchSubscriptionLink.md
```diff
@@ -0,0 +1,72 @@
+---
+id: httpFetchSubscriptionLink
+title: HTTP Fetch Subscription Link
+sidebar_label: HTTP Fetch Subscription Link
+slug: /client/links/httpFetchSubscriptionLink
+---
+
+`httpFetchSubscriptionLink` is a [**terminating link**](./overview.md#the-terminating-link) that uses a fetch-driven [Server-sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) (SSE) transport for subscriptions.
```
Nick-Lucas, line 8: Is the Mozilla link accurate? I need to have a deeper look at this PR in general, but from my understanding (and that link) SSE uses EventSource. The alternative to SSE and WebSockets over fetch is [Long Polling](https://javascript.info/long-polling) - is that what we're actually using here?

We're definitely motivated to add this as an option and thank you for the contribution! I'll give it a good read over later in the week when I have time :)

KATT: `fetch` can be used on the client in place of EventSource. As previously noted in my review, EventSource sucks and is a pain to work with, the only benefit is the eventsource tab inside of Chrome devtools but I think that has since been adapter do be pretty if the server just sends the right eventsource headers too

Nick-Lucas: Right okay! ChatGPT agrees

https://chatgpt.com/share/e/69cb9046-36d8-8012-9b1a-7cf4785912bb

So we're basically implementing the underlying protocol from scratch without the abstraction, neat

KATT: Yeah. It's a bad abstraction though so we shouldn't rely on it for api as we're doing here though

On my phone so hard to pseudo code, but it could likely just be an async generator function for the connection and the data

KATT: Can't view that link btw

PR #7345, opened 2026-04-20
fix(server): make batch stream content-type configurable

monsonego:
Addresses #6819

## 🎯 Changes

- make the streamed batch response `content-type` configurable via `initTRPC.create({ jsonl: { contentType } })`
- keep `application/json` as the default for backward compatibility, matching the stability concerns raised in `#7258`
- support `application/jsonl`, `application/x-ndjson`, or another content type string when a server wants to opt in
- document how to configure the response header and add tests for the default plus opt-in content types

## Why

`httpBatchStreamLink` advertises JSON Lines, but changing the default response header for all existing users could be a breaking change. This keeps the current default stable while adding the setting requested in `#7258`, so servers can choose the streamed `content-type` that works best for their tooling or infrastructure.

## ✅ Checklist

- [x] I have added documentation related to the changes made.

## Validation

- `corepack pnpm --dir packages/server build`
- `corepack pnpm --dir packages/server typecheck`
- `corepack pnpm --dir packages/tests typecheck`

## Summary by CodeRabbit

* **New Features**
  * Servers can configure the streamed response content-type (e.g., application/jsonl, application/x-ndjson, etc.); default is now application/json.

* **Documentation**
  * Streaming docs updated to reflect the new default and show how to configure JSONL streaming options.

* **Tests**
  * Streaming tests added/updated to verify configurable content-type values and streamed response behavior.

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
+++ b/packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
@@ -572,7 +572,10 @@ export async function resolveResponse<TRouter extends AnyRouter>(
     // batch response handlers
     if (info.accept === 'application/jsonl') {
       // httpBatchStreamLink
-      headers.set('content-type', 'application/jsonl');
+      headers.set(
+        'content-type',
+        config.jsonl?.contentType ?? 'application/json',
+      );
       headers.set('transfer-encoding', 'chunked');
       const headResponse = initResponse({
         ctx: ctxManager.valueOrUndefined(),
--- a/packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
+++ b/packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
@@ -572,7 +572,7 @@ export async function resolveResponse<TRouter extends AnyRouter>(
     // batch response handlers
     if (info.accept === 'application/jsonl') {
       // httpBatchStreamLink
-      headers.set('content-type', 'application/json');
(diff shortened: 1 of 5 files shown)


packages/server/src/unstable-core-do-not-import/http/resolveResponse.ts
```diff
@@ -572,7 +572,7 @@ export async function resolveResponse<TRouter extends AnyRouter>(
     // batch response handlers
     if (info.accept === 'application/jsonl') {
       // httpBatchStreamLink
-      headers.set('content-type', 'application/json');
+      headers.set('content-type', 'application/jsonl');
```
Nick-Lucas: Could you action what was discussed in the other PR? https://github.com/trpc/trpc/pull/7258#issuecomment-4111668657

That PR has gone stale but we can't merge this as a potential breaking change

monsonego: @Nick-Lucas 
Thanks for pointing me back to #7258. I updated this PR to follow that approach: the streamed response `content-type` is now configurable via `initTRPC.create({ jsonl: { contentType } })`, while `application/json` remains the default for backwards compatibility.

I also updated the docs and tests to cover both the default behavior and the opt-in cases. All checks are green now.

Nick-Lucas:
Thanks for this! Will look it over when I have a bit of time, often quite busy mid-week :)

Nick-Lucas (approved):

PR #7346, opened 2026-04-20
fix(server): stop leaking hashed dts chunks from public types

### LINKED ISSUE #6998: bug: Error when removing baseUrl - unstable-core-do-not-import type issue

src/middleware/jwt/index.ts:11:14 - error TS2742: The inferred type of 'jwtMiddleware' cannot be named without a reference to '../../../node_modules/@trpc/server/dist/unstable-core-do-not-import.d-1RewV6pM.mjs'. This is likely not portable. A type annotation is necessary.

### Describe the bug

Hi there,

I encountered an issue when I removed the baseUrl configuration. After removing baseUrl, I get the following TypeScript error

It seems like the issue is related to unstable-core-do-not-import. I am currently using tsgolint for linting, and tsgo no longer supports baseUrl, which might be the root cause of this issue.

Steps to reproduce:

Remove the baseUrl configuration from tsconfig.json.

### Link to reproduction

none

### To reproduce

remove tsconfig **baseUrl**

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

monsonego:
Closes #6998

## Summary
- normalize public `.d.mts` / `.d.cts` entrypoints after the server build so they reference stable `unstable-core-do-not-import` declaration paths instead of hashed chunk filenames
- keep the shared internal declaration chunks intact for the build itself
- add regression coverage for the normalization helper and remove `baseUrl` from the server-side monotest fixtures that exercise the no-`baseUrl` consumer path from #6998

## Why
Consumers emitting declarations without `baseUrl` can surface `TS2742` because inferred exported types point at `dist/unstable-core-do-not-import.d-<hash>.mjs`, which is not a portable public path. The public entrypoints should only expose stable declaration paths that are part of the published package surface.

## Validation
- `corepack pnpm --dir packages/server build`
- `node node_modules/typescript/lib/tsc.js -p packages/server/tsconfig.json --pretty false`
- `corepack pnpm --filter @monotest/api... build`

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for TypeScript declaration file normalization

* **Chores**
  * Enhanced build process to properly normalize generated TypeScript declaration files
  * Adjusted TypeScript configurations for test environments


KATT (commented):
Would it possible to do add any [regression] tests for this?

monsonego:
@KATT , Added regression coverage for this.

The PR now includes tests for the declaration import normalization itself, and CI also covers the no-`baseUrl` monotest path from `#6998` to make sure the generated public types stay portable. 
All checks are green now.

PR #7359, opened 2026-04-24
fix: clean up procedure builder typecast using Object.assign

gugli4ifenix-design:
## Problem

`createProcedureCaller` manually assigns `_def`, `procedure`, and `meta` properties to the procedure function, then uses a double typecast (`as unknown as AnyProcedure`). There is a `FIXME` comment noting this should be improved.

## Solution

Replace the manual property assignments with `Object.assign`, which:
- Produces the same runtime result
- Makes the intent clearer ("create a function with these additional properties")
- Removes the need for intermediate mutation
- Resolves the FIXME

The `as unknown as AnyProcedure` cast remains because `AnyProcedureBuilderDef` does not carry the full `AnyProcedure` shape (missing `$types`, `experimental_caller`). Fixing the underlying types would be a separate, larger change.

## Changes

1 file, net -1 line. No behavioral changes.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal code construction by streamlining the procedure builder implementation. No user-facing changes.

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
+++ b/packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
@@ -695,10 +695,9 @@ function createProcedureCaller(_def: AnyProcedureBuilderDef): AnyProcedure {
     return result.data;
   }
 
-  procedure._def = _def;
-  procedure.procedure = true;
-  procedure.meta = _def.meta;
-
-  // FIXME typecast shouldn't be needed - fixittt
-  return procedure as unknown as AnyProcedure;
+  return Object.assign(procedure, {
+    _def,
+    procedure: true as const,
+    meta: _def.meta,
+  }) as unknown as AnyProcedure;
 }


packages/server/src/unstable-core-do-not-import/procedureBuilder.ts
```diff
@@ -695,10 +695,9 @@ function createProcedureCaller(_def: AnyProcedureBuilderDef): AnyProcedure {
...
-  procedure.procedure = true;
-  procedure.meta = _def.meta;
-
-  // FIXME typecast shouldn't be needed - fixittt
-  return procedure as unknown as AnyProcedure;
+  return Object.assign(procedure, {
+    _def,
+    procedure: true as const,
+    meta: _def.meta,
+  }) as unknown as AnyProcedure;
```
Nick-Lucas, line 702: This hasn't fixed anything though, right? It's now allocating a new object but still has the type assertion 

Very open to this being improved!

Nick-Lucas (changes requested):

gugli4ifenix-design:
You're right — the type assertion stays because `AnyProcedureBuilderDef` doesn't carry `$types`, `experimental_caller`, etc. Fixing that would mean either widening the def type or restructuring how procedures are built, which felt out of scope for a cleanup.

The win here is small: Object.assign makes the intent explicit ("create a function-with-properties") vs mutating then casting, and removes the FIXME that's been sitting there. Happy to close this if the tradeoff isn't worth it — or if you'd prefer, I can take a look at the underlying type mismatch in a follow-up.

PR #7377, opened 2026-05-14
fix(server): reject routers with incompatible procedure context

### LINKED ISSUE #4306: bug: conflicting `ctx` types allowed at compile time

```
  System:
    OS: macOS 13.2.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 57.83 GB / 96.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.15.0 - ~/.nvm/versions/node/v18.15.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.15.0/bin/yarn
    npm: 9.5.0 - ~/.nvm/versions/node/v18.15.0/bin/npm
  Browsers:
    Chrome: 112.0.5615.137
    Safari: 16.3
```

```
"@trpc/server": "10.23.0",
```

### Describe the bug

routers don't seem to check the types of their procedures or sub-routers. self contained example:

```ts
import {initTRPC} from '@trpc/server'
import {createNextApiHandler} from '@trpc/server/adapters/next'

const trpca = initTRPC.context<{a: string}>().create()
const trpcb = initTRPC.context<{b: string}>().create()

const parent = trpca.router({
  a: trpca.procedure.query(({ctx}) => ctx.a.slice()),
  b: trpcb.procedure.query(({ctx}) => ctx.b.slice()),
})

export default createNextApiHandler({
  router: parent,
  createContext: () => ({a: 'aaa'}),
})
```

this is fine at compile time, and hitting `/api/trpc.a` works, but hitting `/api/trpc.b` gets `TypeError: Cannot read properties of undefined (reading 'slice')` (since `ctx.b` is never defined anywhere) - I shouldn't have been allowed to use `trpcb.procedure.query(...)` on a router for `trpca`, since they have different contexts.

### Link to reproduction

https://www.typescriptlang.org/play?ts=5.0.3#code/JYWwDg9gTgLgBAb2AO2DAKgJQAoGEC+cAZlBCHAOQACMUYAxgPQDOAplAG7sUBQoksRPSisAhjFYA5VgA8YAQTDAAEqOQATADbtCJMpRp0mbTu0aj1osBKjNGyWTF496EZM3i0GouAF44KGhYeAB0rsgScgA8CKIAXHAeUCgA5vgAfAAUAJRhIuKsOS5uHnBe9ABGfgGoGDi4YW6RMDEVCUmpGTl5YhJFxe7wYKIiEdXloiGkAK42mQg8cHDxZUaTYKT0rOrTIiEAjtPsAJ6Z8-QwMvjZfulwFzIhk8yawFs52QA0i3BtqwwVEIbCBbHZ7Q4nM4IB7XW73S4hQEvN6FbJfHjXHg8WQCeDqVhEUTTTTwYS9KSORQqNRadjzH4zGwJYajGDfJZkgq4JqOBI5OHzFYUUQiijXb7XIA

### To reproduce

Paste the code above into a next api handler. I tried that above using the stackblitz suggested but there's a strange compile error that I don't get locally, it doesn't seem to acknowledge that the export for the next adapter exists: https://stackblitz.com/edit/github-qwqz1i-qs9elv?file=src/pages/api/trpc/[trpc].ts

### Additional information

My use-case: I'm trying to make a router in a library, which can be plugged into an application's parent router for modular functionality, with the requirement that the parent has a context which meets the requirement of the library router. But the type system doesn't seem to enforce this. Maybe it's just not supported (yet)?

I'm putting yes I'd be down to fix this but I don't know if I'll have the time to so if someone else wants to, go for it!

### 👨‍👧‍👦 Contributing

Seven380:
## Summary
- carry the base context type on built procedure definitions
- validate router inputs so procedures, routers, and lazy routers require a context satisfied by the parent router
- add a regression test for mixing procedures from incompatible `initTRPC.context()` roots

Fixes #4306

## Tests
- pnpm.cmd test packages/server/src/unstable-core-do-not-import/router.test.ts packages/server/src/unstable-core-do-not-import/procedureBuilder.test.ts packages/server/src/unstable-core-do-not-import/router.mergeRouters.test.ts --run
- pnpm.cmd exec prettier --check packages/server/src/unstable-core-do-not-import/procedure.ts packages/server/src/unstable-core-do-not-import/procedureBuilder.ts packages/server/src/unstable-core-do-not-import/router.ts packages/server/src/unstable-core-do-not-import/router.test.ts packages/server/src/unstable-core-do-not-import/router.mergeRouters.test.ts

Note: `pnpm.cmd --filter @trpc/server typecheck` still fails locally in `src/adapters/ws.ts` because the installed `ws` typings are resolved without `ws.WebSocket` / `ws.WebSocketServer` namespace members in this environment. No errors from the router/procedure context changes remain before that unrelated failure.

## Summary by CodeRabbit

* **New Features**
  * Stronger compile-time type safety for procedure contexts and router-procedure compatibility; procedure/context types are now surfaced to catch mismatches early.

* **Developer API**
  * Procedure builder signatures now carry explicit context typing so created procedures reflect the expected context.

* **Tests**
  * Added and adjusted tests to validate context compatibility and merging behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trpc/trpc/pull/7377)

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/procedure.ts
+++ b/packages/server/src/unstable-core-do-not-import/procedure.ts
@@ -10,10 +10,17 @@ export const procedureTypes = ['query', 'mutation', 'subscription'] as const;
 export type ProcedureType = (typeof procedureTypes)[number];
 
 interface BuiltProcedureDef {
+  ctx?: unknown;
   meta: unknown;
   input: unknown;
   output: unknown;
 }
+
+type inferBuiltProcedureContext<TDef extends BuiltProcedureDef> = TDef extends {
+  ctx: infer TContext;
+}
+  ? TContext
+  : object;
 
 /**
  *
@@ -29,6 +36,7 @@ export interface Procedure<
      * @internal
      */
     $types: {
+      ctx: inferBuiltProcedureContext<TDef>;
       input: TDef['input'];
       output: TDef['output'];
     };
@@ -52,14 +60,17 @@ export interface Procedure<
   (opts: ProcedureCallOptions<unknown>): Promise<TDef['output']>;
 }
 
-export interface QueryProcedure<TDef extends BuiltProcedureDef>
-  extends Procedure<'query', TDef> {}
+export interface QueryProcedure<
+  TDef extends BuiltProcedureDef,
+> extends Procedure<'query', TDef> {}
 
-export interface MutationProcedure<TDef extends BuiltProcedureDef>
-  extends Procedure<'mutation', TDef> {}
+export interface MutationProcedure<
+  TDef extends BuiltProcedureDef,
+> extends Procedure<'mutation', TDef> {}
 
-export interface SubscriptionProcedure<TDef extends BuiltProcedureDef>
-  extends Procedure<'subscription', TDef> {}
+export interface SubscriptionProcedure<
+  TDef extends BuiltProcedureDef,
+> extends Procedure<'subscription', TDef> {}
 
 /**
  * @deprecated
--- a/packages/server/src/unstable-core-do-not-import/procedure.ts
+++ b/packages/server/src/unstable-core-do-not-import/procedure.ts
@@ -20,7 +20,7 @@ type inferBuiltProcedureContext<TDef extends BuiltProcedureDef> = TDef extends {
   ctx: infer TContext;
 }
   ? TContext
-  : object;
+  : unknown;
 
 /**
  *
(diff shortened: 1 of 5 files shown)


Seven380:
Addressed the CodeRabbit nits in f475e4031: switched the fallback context sentinel to unknown and avoided parameter destructuring in the new router test.\n\nThe failing GitHub Actions jobs are currently failing during setup at pnpm install because main has packages/tests depending on @trpc/client and @trpc/server 11.16.0 while pnpm-lock.yaml has 11.17.0 for those specifiers. That happens before this PR's typecheck/tests run.

PR #7384, opened 2026-05-20
fix(server): set application/jsonl content-type for httpBatchStreamLink responses

### LINKED ISSUE #6819: bug: `httpBatchStreamLink` responses use wrong `content-type` header (`application/json`)

LeSingh1:
Fixes #6819.

The httpBatchStreamLink response handler in resolveResponse.ts was setting the response content-type header to application/json, while the body is actually newline-delimited JSON. The docs at https://trpc.io/docs/client/links/httpBatchStreamLink#streaming-mode describe the content-type as application/jsonl, and the trpc-accept request header value used throughout the client and server is already application/jsonl. The mismatch causes Firefox to surface a malformed-response warning in devtools.

This changes the response content-type to application/jsonl so it matches the documented behavior and the existing trpc-accept value used elsewhere in the codebase. No other content-type sites are touched; the regular httpLink path still emits application/json.

A regression test was added in packages/tests/server/streaming.test.ts that asserts the response from httpBatchStreamLink carries content-type: application/jsonl. The new test fails on main and passes with the fix. All 538 existing server tests continue to pass and typecheck is clean on both @trpc/server and @trpc/tests.

Note on the alternative discussed in the issue: the original reporter considered text/plain or application/x-ndjson because application/jsonl is not registered with IANA. application/jsonl was chosen here because it is the value already used as the trpc-accept content-negotiation token in both the client (httpBatchStreamLink) and server (contentType.ts, types.ts), and is what the documentation already promises. Aligning the response header with the existing accept token avoids introducing a third value and keeps client/server symmetric.

## Summary by CodeRabbit

* **Bug Fixes**
  * HTTP batch streaming responses now return the correct application/jsonl content-type for JSONL streams.
* **Tests**
  * Added test coverage to verify streaming responses use application/jsonl.
* **Chores**
  * Updated test package dependencies to a newer patch release.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trpc/trpc/pull/7384?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


daltino (approved):
## Code Review: Fix `application/jsonl` Content-Type for `httpBatchStreamLink`

Great catch and clean fix! This is a straightforward correctness bug — the server was advertising `application/json` while actually streaming newline-delimited JSON (`application/jsonl`). The one-line change is exactly right.

---

### ✅ What's Good

**The core fix is correct.**
The `info.accept === 'application/jsonl'` guard already confirms the client negotiated the JSONL streaming format, so setting the response `content-type` to `application/jsonl` is the semantically accurate thing to do. Aligns with the docs and the existing `trpc-accept` convention throughout the codebase.

**The test is well-structured.**
Using a `responseSpy` to intercept the raw `Response` before it's consumed is a solid approach for testing headers. The test lives in the right file alongside the other streaming tests.

---

### 💬 Minor Suggestions

**1. Test: consider asserting the full content-type string**

Some runtimes/frameworks append `; charset=utf-8` to content-type headers. If that ever happens, a strict equality check could break. Consider using `.toContain('application/jsonl')` or `.toMatch(/^application\/jsonl/)` for a touch more resilience — though strict equality is also fine if the codebase consistently avoids appending charset here.

```ts
// Option A – strict (current approach, totally fine)
expect(headers.get('content-type')).toBe('application/jsonl');

// Option B – slightly more resilient
expect(headers.get('content-type')).toMatch(/^application\/jsonl/);
```

**2. Test: the `deferred` procedure could be simplified**

The test doesn't really need the streaming/deferral behavior to verify the content-type header — the header is set before any chunks are sent. A simpler non-async procedure would make the test intent clearer and reduce noise. That said, using a realistic async procedure is not wrong and might be intentional for consistency with surrounding tests.

**3. `package.json` version bump**

The bump from `11.16.0` → `11.17.0` in `packages/tests/package.json` looks like it snuck in from an unrelated change. Worth double-checking this is intentional and expected at this stage of the PR, or whether it should be split out.

---

### 🔍 Edge Case Check

- **Non-streaming paths** (`application/json` responses): Unaffected — the fix is inside the `if (info.accept === 'application/jsonl')` branch. ✅
- **Client-side parsing**: Clients already expect JSONL from this path, so updating the header to match reality won't break any existing behavior. ✅
- **Proxies/CDNs inspecting content-type**: This fix actually *improves* correctness for any middleware that routes or caches based on content-type. ✅

---

All in all, this is a solid, minimal, well-tested fix. The suggestions above are minor polish items, none of which block merging. Nice work! 🎉

LeSingh1:
Thanks for the review @daltino. Pushed 054ca472e addressing 1 and 2:

- switched the assertion to `.toMatch(/^application\/jsonl/)` so a future `; charset=utf-8` suffix would not break it
- dropped the deferred async query in favour of a plain procedure since the header is set before any chunks are streamed, which makes the intent clearer

On 3: the 11.16.0 to 11.17.0 bump in packages/tests/package.json was needed to get `pnpm install --frozen-lockfile` past `ERR_PNPM_OUTDATED_LOCKFILE`. The release commit `23c723cfe` (v11.17.0) updated the lockfile and all other workspace packages to 11.17.0, but packages/tests/package.json still pinned 11.16.0, so CI on this branch (and a few others opened around the same time) refused to install. Happy to drop it into its own commit if that helps, but it lives outside the content-type fix so squashing on merge should leave the actual server-side change clean.

daltino (approved):
## Code Review

Great fix! This is a clear, correct, and well-scoped change. Here are my thoughts:

### ✅ Correctness

The one-line change in `resolveResponse.ts` is exactly right. The response body for `httpBatchStreamLink` is newline-delimited JSON, and the content-type should accurately reflect that. The existing code was already checking `info.accept === 'application/jsonl'` immediately above, making this inconsistency especially jarring — the fix brings the response header into alignment with both the accept header and the [documented behavior](https://trpc.io/docs/client/links/httpBatchStreamLink#streaming-mode).

### ✅ Test Coverage

The new test in `streaming.test.ts` is a solid addition. It:
- Sets up a minimal router to keep the test focused
- Uses a `responseSpy` to intercept the actual `fetch` response, which is the right level to assert on the content-type header
- Is placed logically within the existing `streamHeader option` describe block

**One small suggestion:** The test name `'response uses application/jsonl content-type'` is good, but consider being slightly more specific, e.g. `'httpBatchStreamLink response uses application/jsonl content-type'` — since the surrounding `describe` block is about `streamHeader option`, a reader skimming test names might not immediately know which link type is being tested.

### 📦 Package Version Bump

The bump from `11.16.0` → `11.17.0` in `packages/tests/package.json` looks intentional and appropriate for a patch/fix release. Just make sure the version bump in the actual package manifests (`packages/server/package.json`, `packages/client/package.json`) is also included — I don't see those files in the diff, but that may just be outside the scope of what was shown.

### 🔍 Edge Cases

No edge cases are introduced here. The change is purely cosmetic from the server's behavior perspective — the body content doesn't change, only the header describing it. Any clients that were previously sniffing or validating the content-type header against `application/json` would have been wrong to do so, and correcting this header is the right call.

### 🎨 Code Style

Consistent with the surrounding code. No issues.

---

Overall, this is a clean, minimal, well-tested bug fix. Nice work! 🚀

lilong7676:
Thanks for picking this up @LeSingh1! We're hitting the same `content-type: application/json` mismatch in production on a tRPC v11 `httpBatchStreamLink` app, so +1 — `application/jsonl` lines up with the docs and the `trpc-accept` token.

@KATT any rough timeline for merging this, and could it make it into the next `@trpc/server` patch release? Happy to help test — thanks!

PR #7388, opened 2026-05-24
fix(react-query): support select() in useSuspenseInfiniteQuery

### LINKED ISSUE #6195: bug: useSuspenseInfiniteQuery's select method expects wrong return type

```
  System:
    OS: Linux 5.15 Debian GNU/Linux 12 (bookworm) 12 (bookworm)
    CPU: (12) x64 AMD Ryzen 5 4600H with Radeon Graphics
    Memory: 392.87 MB / 3.83 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 20.16.0 - ~/.config/nvm/versions/node/v20.16.0/bin/node
    Yarn: 1.22.22 - ~/.config/nvm/versions/node/v20.16.0/bin/yarn
    npm: 10.8.2 - ~/.config/nvm/versions/node/v20.16.0/bin/npm
    pnpm: 9.7.0 - ~/.config/nvm/versions/node/v20.16.0/bin/pnpm
  npmPackages:
    @tanstack/react-query: ^5.59.19 => 5.59.19 
    @trpc/client: 11.0.0-rc.608 => 11.0.0-rc.608+f75de97b3 
    @trpc/react-query: 11.0.0-rc.608 => 11.0.0-rc.608+f75de97b3 
    @trpc/server: 11.0.0-rc.608 => 11.0.0-rc.608+f75de97b3 
    next: 15.0.2 => 15.0.2 
    react: 19.0.0-rc-77b637d6-20241016 => 19.0.0-rc-77b637d6-20241016 
    typescript: ^5.6.3 => 5.6.3 
```

### Describe the bug

`useSuspenseInfiniteQuery` does not allow a custom `select` method's return type. It works with `useInfiniteQuery`.

```ts
  const result = trpc.greeting.useSuspenseInfiniteQuery(
    {},
    {
      getNextPageParam: (p) => p.nextCursor,
      select: (data) => 1,  // Type 'number' is not assignable to type '{ pages: { text: string; nextCursor: number; }[]; pageParams: (number | null | undefined)[]; }'.
    }
  );
```

It works with `useInfiniteQuery`:

```ts
  const result2 = trpc.greeting.useInfiniteQuery(
    {},
    {
      getNextPageParam: (p) => p.nextCursor,
      select: (data) => 1, // No error
    }
  );
```

### Link to reproduction

https://stackblitz.com/edit/github-t3rzii?file=src%2Fpages%2Findex.tsx

### To reproduce

Use the select method with a useSuspenseInfiniteQuery, and return a new data

```ts
  const result = trpc.greeting.useSuspenseInfiniteQuery(
    {},
    {
      getNextPageParam: (p) => p.nextCursor,
      select: (data) => 1,  // Type 'number' is not assignable to type '{ pages: { text: string; nextCursor: number; }[]; pageParams: (number | null | undefined)[]; }'.
    }
  );
```

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

eyupcanakman:
Closes #6195

## 🎯 Changes

Bug fix. `useSuspenseInfiniteQuery` ignored a custom `select` return type and forced the result to the page shape, so `select: (data) => 1` failed to type-check. `useInfiniteQuery` already handles this. The hook type now takes a `TData` parameter that defaults to the inferred page data, the same as `useInfiniteQuery` and `useSuspenseQuery`.

## ✅ Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed type inference for infinite query operations when using custom data selectors, ensuring accurate and consistent type information across all states, with enhanced IDE autocomplete support and stronger compile-time type safety
* **Tests**
  * Added comprehensive regression test for custom data selector functionality to validate both type inference accuracy and runtime behavior under various conditions

[Review Change Stack](https://app.coderabbit.ai/change-stack/trpc/trpc/pull/7388?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/react-query/src/createTRPCReact.tsx
+++ b/packages/react-query/src/createTRPCReact.tsx
@@ -264,7 +264,9 @@ export interface useTRPCInfiniteQuery<TDef extends ResolverDef> {
 //   queryClient?: QueryClient,
 // ): UseSuspenseInfiniteQueryResult<TData, TError>;
 
-export type useTRPCSuspenseInfiniteQuery<TDef extends ResolverDef> = (
+export type useTRPCSuspenseInfiniteQuery<TDef extends ResolverDef> = <
+  TData = trpcInfiniteData<TDef>,
+>(
   input: InfiniteInput<TDef['input']>,
   opts: makeInfiniteQueryOptions<
     inferCursorType<TDef['input']>,
@@ -274,7 +276,7 @@ export type useTRPCSuspenseInfiniteQuery<TDef extends ResolverDef> = (
       //     TError,
       TRPCClientErrorLike<TDef>,
       //     TData,
-      trpcInfiniteData<TDef>,
+      TData,
       //     TQueryKey,
       any,
       //     TPageParam
@@ -282,12 +284,9 @@ export type useTRPCSuspenseInfiniteQuery<TDef extends ResolverDef> = (
     >
   >,
 ) => [
-  trpcInfiniteData<TDef>,
+  TData,
   TRPCHookResult &
-    UseSuspenseInfiniteQueryResult<
-      trpcInfiniteData<TDef>,
-      TRPCClientErrorLike<TDef>
-    >,
+    UseSuspenseInfiniteQueryResult<TData, TRPCClientErrorLike<TDef>>,
 ];
 
 /**
--- a/packages/react-query/test/useSuspenseInfiniteQuery.test.tsx
+++ b/packages/react-query/test/useSuspenseInfiniteQuery.test.tsx
@@ -107,3 +107,39 @@ test('useSuspenseInfiniteQuery()', async () => {
     expect(utils.container).toHaveTextContent(`[ "2" ]`);
   });
 });
+
+test('regression 6195: select()', async () => {
+  const { App, client } = ctx;
+  function MyComponent() {
+    const [data, query] = client.post.list.useSuspenseInfiniteQuery(
+      {},
+      {
+        getNextPageParam(lastPage) {
+          return lastPage.next;
+        },
+        select: (data) => ({
+          ...data,
+          pages: data.pages.map((page) => ({ ...page, foo: 'bar' as const })),
+        }),
+      },
+    );
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain
+    expectTypeOf(data.pages[0]?.foo!).toEqualTypeOf<'bar'>();
+    // eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain
+    expectTypeOf(query.data.pages[0]?.foo!).toEqualTypeOf<'bar'>();
+
+    return <div>foo:{data.pages[0]?.foo}</div>;
+  }
+
+  const utils = render(
+    <App>
+      <Suspense fallback="loading">
+        <MyComponent />
+      </Suspense>
+    </App>,
+  );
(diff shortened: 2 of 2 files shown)


daltino (commented):
## Review: `fix(react-query): support select() in useSuspenseInfiniteQuery`

Great fix! The root cause is clear, the approach is consistent with how `useInfiniteQuery` and `useSuspenseQuery` already work, and the regression test makes the intent obvious. A few things worth discussing before merging:

---

### ✅ What's good

- **Minimal, targeted change.** Adding `<TData = trpcInfiniteData<TDef>>` to the call signature mirrors the existing pattern in `useInfiniteQuery` exactly — easy to review and reason about.
- **Default keeps backwards compat.** Because `TData` defaults to `trpcInfiniteData<TDef>`, callers that don't use `select` are completely unaffected.
- **Regression test is well-structured** and uses a realistic `select` (spreading pages and adding a discriminant field) rather than the trivial `=> 1` example from the issue, which is more valuable for long-term confidence.

---

### 🟡 Suggestions / things to double-check

#### 1. The return type in the truncated patch — confirm it uses `TData`

The patch is cut off at:
```
 TRPCClie
```
It's critical that the return type of the hook (i.e., `UseSuspenseInfiniteQueryResult<..., TError>`) actually threads `TData` through instead of still hardcoding `trpcInfiniteData<TDef>`. Can you confirm the full diff includes something like:
```ts
) => UseSuspenseInfiniteQueryResult<TData, TRPCClientErrorLike<TDef>>;
```
If the return type wasn't updated, the `select` transform will be accepted at the input side but the destructured `data` variable will still carry the wrong type — a silent type lie that's worse than the compile error.

#### 2. Add a compile-time type assertion in the test

The test verifies runtime behaviour, but since this is fundamentally a type-level bug, it's worth adding an `expectTypeOf` (or `// @ts-expect-error` guard) so a regression would be caught at type-check time even if the runtime test still passes:

```ts
// After the select, pages should have the extra `foo` field
expectTypeOf(data.pages[0]).toHaveProperty('foo');
expectTypeOf(data.pages[0].foo).toEqualTypeOf<'bar'>();
// And `data` itself should not be the raw InfiniteData shape
expectTypeOf(data).not.toEqualTypeOf<trpcInfiniteData<typeof client.post.list>>();
```

Without this, someone could accidentally regress the type signature and the test suite would still go green.

#### 3. The `select` option type in `makeInfiniteQueryOptions` — is it constrained correctly?

For `useInfiniteQuery` the `select` option is typed as:
```ts
select?: (data: InfiniteData<TQueryFnData>) => TData
```
Make sure `makeInfiniteQueryOptions` (or wherever the `opts` type is constructed) has the same shape, otherwise TypeScript will infer `TData` as `unknown` or fall back to the default rather than inferring from the `select` return type.

#### 4. Incomplete checklist item

The PR description has `- [x] I hav` — minor, but worth completing before merge for a clean paper trail.

#### 5. Test comment `// esli` looks like a stray fragment

Line visible in the patch:
```ts
    // esli
```
This appears to be an incomplete comment (maybe "eslint-disable"?). Please remove or complete it.

---

### Summary

The change itself is correct and the approach is sound. The two most important things to verify before approving are:
1. The **return type** of the hook uses `TData` (patch is truncated so this can't be confirmed from the diff alone).
2. A **type-level assertion** in the test so future regressions are caught statically.

Fix those two and this is a clean ✅ merge.

PR #7390, opened 2026-05-25
fix(client): abort JSONL stream on httpBatchStreamLink unsubscribe

### LINKED ISSUE #7389: bug: `httpBatchStreamLink` AbortController is created but never called on unsubscribe

**Package:** `@trpc/client`
**Version:** v11.17.0

### Describe the bug

```ts
return () => {
  // noop
};
```

### Link to reproduction

### To reproduce

### Additional information

Suggested fix (one line):
// Before
return () => { // noop };

// After  

### 👨‍👧‍👦 Contributing

vinamra1102:
Closes #7389

## What changed

`httpBatchStreamLink` creates an `AbortController` and passes it to `jsonlStreamConsumer` on every request, but the observable's cleanup function was a noop  `abortController.abort()` was never called on unsubscribe.

This meant the underlying JSONL response stream continued consuming data from the server even after the subscriber disconnected (component unmount, navigation, query cancellation).

## Fix

Call `abortController.abort()` in the cleanup function so the stream is cancelled when the observable is unsubscribed.

## Tests

Added a test that subscribes to a batch stream observable, unsubscribes before the response completes, and asserts the fetch was aborted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Batch-stream network requests now reliably abort when subscriptions are canceled, and completion/error handling no longer triggers subsequent teardown aborts.

* **Tests**
  * Added a test verifying network requests are aborted upon subscription cancellation.

* **Chores**
  * Updated test package dependency versions for client and server test packages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trpc/trpc/pull/7390?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


vinamra1102:
The Bun E2E failure is in `examples/bun#test-dev` and appears to be an 
infrastructure issue (process exited with code 1, no failed assertions). 
All other E2E suites pass including the streaming and batching tests 
directly related to this change.

daltino (commented):
```json
{
  "event": "COMMENT",
  "body": "## Code Review: `fix(client): abort JSONL stream on httpBatchStreamLink unsubscribe`\n\nGreat catch! This is a real resource leak that could cause unnecessary server load and bandwidth consumption after component unmount. The fix is conceptually correct. A few things worth discussing before merging:\n\n---\n\n### ✅ What's Working Well\n\n- The core idea is sound: creating a per-request `AbortController` and wiring it into `raceAbortSignals` is the right pattern.\n- The `isDone` flag (visible in the patch context) suggests there's guard logic to prevent double-abort — good defensive programming.\n- The test is well-structured: it captures the `AbortSignal` from the mocked `fetch` and verifies it gets aborted on unsubscribe. This directly tests the observable contract.\n\n---\n\n### 🔍 Issues & Suggestions\n\n#### 1. Where is `ac.abort()` actually called? (Blocker risk)\n\nThe patch shows the `AbortController` being created and its signal being raced, but the diff is truncated — we can't see where `ac.abort()` is called in the observable's cleanup/unsubscribe path. This is the *crux* of the fix. Please confirm the full subscriber teardown looks something like:\n\n```ts\nreturn () => {\n  ac.abort();\n};\n```\n\nIf this is missing or conditional (e.g., only called when `isDone === false`), the fix may not work in all cases. Could you share the complete updated observable block?\n\n---\n\n#### 2. `isDone` flag — is it used correctly?\n\nThe `isDone` variable is declared but its usage isn't visible in the diff. A common pattern is:\n\n```ts\nreturn () => {\n  if (!isDone) {\n    ac.abort();\n  }\n};\n```\n\nThis is fine, but be careful: aborting an already-settled fetch is a no-op in the browser, so the guard may be unnecessary complexity. If `isDone` is set inside `.then()` / `.catch()`, there's also a subtle race where the promise resolves *just before* unsubscribe triggers — which is generally safe, but worth a comment in code.\n\n---\n\n#### 3. `signal: raceAbortSignals(op.signal, ac.signal)` — potential double-signal concern\n\nIf `op.signal` is already an `AbortSignal` from the caller, and the loader also internally creates signals, verify that `raceAbortSignals` doesn't cause the signal to fire prematurely in edge cases (e.g., `op.signal` is `null`/`undefined`). Looking at the existing `raceAbortSignals` implementation would help confirm this is safe.\n\n---\n\n#### 4. Test: missing assertion on abort *reason*\n\nThe test verifies `fetchSignal` is aborted, which is great. Consider also asserting the abort reason to future-proof against accidental aborts from unrelated signals:\n\n```ts\nexpect(fetchSignal?.aborted).toBe(true);\n// optional but nice:\nexpect(fetchSignal?.reason).toBeInstanceOf(DOMException); // or whatever reason is set\n```\n\n---\n\n#### 5. Test: `fetchCalled` promise and timing\n\nThe test resolves `fetchCalled` inside the mock before returning the `Response`. This is fine, but if the test runner moves on before the mock `fetch` is fully set up, there could be a flake. Consider using `await fetchCalled` with a timeout guard, or confirm your test framework handles this robustly.\n\n---\n\n#### 6. Minor: `let _res = undefined as HTTPResult | undefined`\n\nThis pre-existed the change, but since we're touching this area — the `_res` underscore prefix conventionally signals \"intentionally unused,\" which is misleading if it *is* used later. Not a blocker, just a readability note.\n\n---\n\n#### 7. `package.json` version bump\n\nThe bump from `11.16.0` → `11.17.0` in `packages/tests/package.json` looks intentional for the test package to pick up the fix. Confirm this is the right version and that CI will publish/resolve it correctly in the monorepo setup (e.g., via workspace protocol rather than pinned versions).\n\n---\n\n### Summary\n\n| Area | Status |\n|---|---|\n| Core fix approach | ✅ Correct |\n| `ac.abort()` call site | ⚠️ Not visible in diff — needs confirmation |\n| `isDone` usage | ⚠️ Needs verification |\n

**Comment by vinamra1102** [comment]
@daltino Thanks for the thorough review!

**1. Cleanup return block with the `isDone` guard:**

```ts
return () => {
  if (!isDone) {
    ac.abort();
  }
};
```

This ensures `abort()` only fires on explicit unsubscribe — not when the observable completes or errors naturally, which would kill the JSONL stream while iterable responses are still being consumed.

**2. `isDone` is set to `true` at the top of both `.then()` and `.catch()`, before any `observer.complete()` or `observer.error()` call:**

```ts
promise
  .then((res) => {
    isDone = true;
    // ... observer.next() + observer.complete()
  })
  .catch((err) => {
    isDone = true;
    observer.error(...);
  });
```

Since `observer.complete()`/`observer.error()` triggers teardown synchronously, `isDone` must be set before those calls.

**3.** Confirmed. `raceAbortSignals` accepts `...signals: Maybe<AbortSignal>[]` and uses optional chaining (`signal?.aborted`, `signal?.addEventListener`) to skip `null`/`undefined` entries. When `op.signal` is `null`, only `ac.signal` is wired up.

**4.** Will add an `expect(fetchSignal?.reason).toBeInstanceOf(DOMException)` assertion to the test.

**5.** `resolve()` executes synchronously inside the mock body before `return new Response(...)`. By the time `await fetchCalled` settles, `fetchSignal` is already captured — no race between signal capture and the assertion.

**6.** Acknowledged — `_res` is pre-existing, untouched by this change.

**7.** `packages/tests/package.json` lists bare version strings, but `.npmrc` has `link-workspace-packages=true` and `pnpm-workspace.yaml` includes `packages/*`, so pnpm resolves these as workspace links at install time. The `11.16.0` → `11.17.0` bump was applied by `manypkg fix` during `pnpm lint-fix` to match the current workspace version.

vinamra1102:
Added the abort reason assertion in the latest commit. Used 
`expect(fetchSignal!.reason?.name).toBe('AbortError')` instead of 
`toBeInstanceOf(DOMException)`  cross-realm instanceof checks fail 
in the vitest/Node.js environment so name comparison is more reliable here.

KATT (approved):

KATT:
Maybe the chain should be aware of the abort signal and call this instead of the link, but this is good - doing it on the chain might have some unintended consequences when doing like a dedupe link or something?

Issue #5018, opened 2023-11-09
bug: Date type suddenly mangles to string


dvv:
### Provide environment information

```
  System:
    OS: Linux 6.1 Arch Linux
    CPU: (6) x64 AMD Ryzen 5 4500U with Radeon Graphics
    Memory: 5.11 GB / 14.99 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 20.9.0 - /usr/bin/node
    Yarn: 1.22.19 - /usr/bin/yarn
    npm: 10.2.1 - /usr/bin/npm
  Browsers:
    Chromium: 118.0.5993.117
```

### Describe the bug

It seems the output type of a procedure is wrongly inferred or mangled. I wonder if it is bug or feature or whatever I may be doing wrong?

### Link to reproduction

https://typescript-ovwrdt.stackblitz.io

### To reproduce

Open the link above. Hover the mouse upon `.query` in line 46 `rpc.Foo.query({})` to see the inferred type. Notice the x' type mangled.

### 👨‍👧‍👦 Contributing

- [ ] 🙋‍♂️ Yes, I'd be down to file a PR fixing this bug!

## Funding
* You can sponsor this specific effort via a [Polar.sh](https://polar.sh) pledge below
* We receive the pledge once the issue is completed & verified

<a href="https://polar.sh/trpc/trpc/issues/5018">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/trpc/trpc/issues/5018/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/trpc/trpc/issues/5018/pledge.svg">
</picture>
</a>

KATT:
Well, since it's sent as JSON, to the client it will become a string. See #5012

Also, that repro link  doesn't work

dvv:
```ts
import type {
  CreateRouterInner,
  DefaultErrorShape,
  DefaultDataTransformer,
  Procedure,
  RootConfig,
} from '@trpc/server';
type ProtocolTrpcRouter = CreateRouterInner<
  RootConfig<{
    ctx: object;
    meta: object;
    errorShape: DefaultErrorShape;
    transformer: DefaultDataTransformer;
  }>,
  {
    Foo: Procedure<
      'query',
      {
        _config: RootConfig<{
          ctx: object;
          meta: object;
          errorShape: DefaultErrorShape;
[...]

PR #7397, opened 2026-05-28
docs: note default JSON serialization on the data transformers page

### LINKED ISSUE #5019: docs: clarify in docs that types are transformed when not using a transformer
### Area of Improvement

Clarify in docs that types are transformed when not using a transformer

See #5018 / #5012 for people's confusion on this

Unsure where in the docs it should live

### Link to related docs

_No response_

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

- [ ] 🙋‍♂️ Yes, I'd be down to file a PR implementing the suggested changes!

emmayusufu:
Closes #5019.

A recurring source of confusion (#5018 and others) is people returning a `Date` from a procedure and seeing it arrive on the client as a string, then filing it as a bug. The cause is that a transformer is optional, and without one tRPC just uses `JSON.stringify`.

This adds a short info note at the top of the Data Transformers page making that explicit: no transformer means `Date` becomes a string and `Map`/`Set` are lost, and a transformer like superjson (already documented right below) is what preserves them. Kept it to the types that actually round-trip differently rather than listing everything.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that data transformers are optional; without one, default JSON serialization can convert or lose non-JSON-safe types (Dates become strings; Map/Set are lost). Recommends using a transformer such as superjson to preserve runtime types and includes guidance to help choose and apply an appropriate transformer when round-tripping complex types is required.

[Review Change Stack](https://app.coderabbit.ai/change-stack/trpc/trpc/pull/7397?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


www/docs/server/data-transformers.md
```diff
@@ -7,6 +7,10 @@ slug: /server/data-transformers
 
 You are able to serialize the response data & input args. The transformers need to be added both to the server and the client.
 
+:::info
+A transformer is optional. Without one, tRPC serializes data with `JSON.stringify`, so values that JSON can't represent don't round-trip: a `Date` arrives on the client as a string, and a `Map` or `Set` is lost. Add a transformer like superjson (below) if you need those types preserved on the client.
```
KATT:
```suggestion
A transformer is optional. Without one, tRPC serializes data with `JSON.stringify`, so values that JSON can't represent don't round-trip: a `Date` arrives on the client as a string, and a `Map` or `Set` are always converted to an empty object (`{}`). Add a transformer like superjson (below) if you need those types preserved on the client.
```

KATT (approved):

PR #7401, opened 2026-06-01
fix(server): delete stale clientSubscriptions entry on subscription task rejection

### LINKED ISSUE #7400: ws: subscription task rejection can leave stale clientSubscriptions entry
## Summary

## Reachable rejection path

## Severity

## Fix

```ts
}).catch((cause) => {
  abortController.abort();
});
```

## Evidence

## PoC

```js
class AbortController {
}

  return {
    next: () => iter.next(),
  };
}

async function reproduce() {
  const id = 'sub-1';

  const throwingIterable = {
    [Symbol.asyncIterator]() {
      return {
      };
    },
  };

  run(async () => {
    abortController.abort();
  }).catch((cause) => {
    abortController.abort();
  });

  }
}

await reproduce();
```

Adit-Jain-srm:
## Summary

Fixes #7400 — stale `clientSubscriptions` entry after subscription task rejection in WebSocket adapter.

## Problem

In `packages/server/src/adapters/ws.ts`, when the subscription async task rejects (e.g. `iterator.return()` throws during `await using` disposal), the `.catch()` handler aborts the controller and sends an error response but never calls `clientSubscriptions.delete(id)`. This leaves a stale `AbortController` in the Map for the connection's lifetime, preventing subscription ID reuse.

The happy path (line 407) correctly calls `clientSubscriptions.delete(id)` — only the error path is missing it.

## Fix

Add `clientSubscriptions.delete(id)` as the first line of the `.catch()` handler, mirroring the cleanup on the happy path.

## Changes

- `packages/server/src/adapters/ws.ts`: +1 line in the subscription task `.catch()` handler

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket subscription cleanup to ensure resources are properly released even when errors occur during subscription processing.


Adit-Jain-srm:
The failing test is in `packages/upgrade/test/__fixtures__/hooks/optimistic-update.spec.tsx` (the `@trpc/upgrade` package) - unrelated to this PR's single-line change in `packages/server/src/adapters/ws.ts`.

The test expects `Posts: 1initial` but receives `Posts: 2initialFoo`, which looks like a timing/race condition in the optimistic update test (the mutation resolves before the assertion runs). This failure reproduces with retries (x2), suggesting it's a known flaky test rather than a regression from our change.

Could a maintainer re-run CI or confirm this is pre-existing? Happy to rebase if needed.

Nick-Lucas (approved):

PR #7406, opened 2026-06-05
fix(server): delete clientSubscriptions entry on failed teardown

### LINKED ISSUE #7400: ws: subscription task rejection can leave stale clientSubscriptions entry

harsha-cpp:
fixes #7400.

**problem:** when a subscription's async iterator rejects during teardown, the rejection lands in `run().catch()` but `clientSubscriptions.delete(id)` never runs. the entry is set synchronously after `run()` is called, so the stale `AbortController` stays for the connection's lifetime and the id can't be reused.

**fix:** move cleanup to `.finally()` so it runs on both the happy path and on rejection.

regression test drives a subscription whose `return()` throws on stop, then reuses the same id and asserts it starts cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * WebSocket subscription cleanup is now always performed when errors occur during teardown, preventing lingering subscriptions.
  * Subscriptions can be reliably restarted after a failed teardown without triggering duplicate-id conflicts.

* **Tests**
  * Added a regression test that reproduces teardown failures and verifies restart behavior to prevent regressions.

### DIFF
--- a/packages/server/src/adapters/ws.ts
+++ b/packages/server/src/adapters/ws.ts
@@ -404,7 +404,6 @@ export function getWSConnectionHandler<TRouter extends AnyRouter>(
               type: 'stopped',
             },
           });
-          clientSubscriptions.delete(id);
         }).catch((cause) => {
           const error = getTRPCErrorFromUnknown(cause);
           opts.onError?.({ error, path, type, ctx, req, input });
@@ -421,6 +420,8 @@ export function getWSConnectionHandler<TRouter extends AnyRouter>(
             }),
           });
           abortController.abort();
+        }).finally(() => {
+          clientSubscriptions.delete(id);
         });
         clientSubscriptions.set(id, abortController);
 
--- a/packages/server/src/adapters/ws.ts
+++ b/packages/server/src/adapters/ws.ts
@@ -404,25 +404,27 @@ export function getWSConnectionHandler<TRouter extends AnyRouter>(
               type: 'stopped',
             },
           });
-        }).catch((cause) => {
-          const error = getTRPCErrorFromUnknown(cause);
-          opts.onError?.({ error, path, type, ctx, req, input });
-          respond({
-            id,
-            jsonrpc,
-            error: getErrorShape({
-              config: router._def._config,
-              error,
-              type,
-              path,
-              input,
-              ctx,
-            }),
+        })
+          .catch((cause) => {
+            const error = getTRPCErrorFromUnknown(cause);
+            opts.onError?.({ error, path, type, ctx, req, input });
+            respond({
+              id,
+              jsonrpc,
+              error: getErrorShape({
+                config: router._def._config,
+                error,
+                type,
+                path,
+                input,
+                ctx,
+              }),
+            });
+            abortController.abort();
+          })
+          .finally(() => {
+            clientSubscriptions.delete(id);
           });
-          abortController.abort();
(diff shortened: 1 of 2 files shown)


daltino (approved):
## Code Review: fix(server): delete clientSubscriptions entry on failed teardown

Great fix for a real leak! The root cause analysis is spot-on and the `.finally()` approach is the idiomatic solution here. A few notes below.

---

### ✅ What's correct

**`ws.ts` change** — Moving `clientSubscriptions.delete(id)` from inside the `.then()` callback to `.finally()` ensures cleanup runs regardless of whether the async iterator teardown throws. This is exactly right. The entry was previously only deleted on the happy path, leaving a stale `AbortController` for the lifetime of the connection and blocking id reuse.

**Test addition** — Good to see a regression test for the exact scenario described in #7400. A `brokenTeardown` procedure that never resolves and throws on `return()` is a solid way to exercise this path.

---

### 🔍 Minor observations

**1. Verify `.catch()` is still present in the final diff**

The patch is truncated so I can't see the full rewrite, but please confirm the `.catch()` branch for `onError` reporting is still there and runs *before* `.finally()` cleans up. The desired shape should be:

```ts
run()
  .catch((cause) => {
    opts.onError?.({ error, path, type, ctx, req, input });
    respond({ id, jsonrpc, error: getErrorShape({ ... }) });
  })
  .finally(() => {
    clientSubscriptions.delete(id);
  });
```

If `clientSubscriptions.delete(id)` was accidentally duplicated in both `.catch()` *and* `.finally()`, that's harmless (`Map.delete` on a missing key is a no-op), but worth cleaning up for clarity.

**2. `AbortController` signal — is `abort()` still called on error path?**

Double-check that when the iterator rejects during teardown, the `AbortController` stored in the map is still `.abort()`-ed before the entry is deleted (or that it doesn't need to be). If another part of the code holds a reference to the signal and expects it to be aborted, deleting the map entry without aborting first could leave those listeners hanging. If abort is triggered earlier in `run()` this is a non-issue — just worth a quick read-through.

**3. Test: the `never-resolves` promise could block teardown indefinitely in CI**

The `next()` returns a `Promise` that never resolves, which is correct for simulating a blocking subscription. Make sure the test has a reasonable timeout and that the `ws.close()` / `subscription.unsubscribe()` path actually drives `return()` to be called so the test doesn't hang. If the test already has a `vi.useFakeTimers()` or a timeout guard, ignore this.

**4. Naming nit (very minor)**

`throwingFactory` is a fine name but something like `createBrokenTeardownRouter` might be slightly more descriptive of *what* is broken (the teardown, not general throwing). Totally optional.

---

### Summary

The fix is correct, minimal, and well-motivated. The test covers the regression scenario. Pending a quick sanity check on the truncated patch to confirm the `.catch()` handler is intact and the control flow is `catch → finally`, this is ready to merge. Nice work! 🎉

PR #7407, opened 2026-06-06
fix(client): abort JSONL stream consumer on httpBatchStreamLink unsubscribe

### LINKED ISSUE #7389: bug: `httpBatchStreamLink` AbortController is created but never called on unsubscribe

chatman-media:
Closes #7389

## Root cause

`httpBatchStreamLink` creates an `AbortController` and passes it to `jsonlStreamConsumer` as the mechanism to stop reading the JSONL response body. However, the observable returned by the link had a noop teardown function:

```ts
return () => {
  // noop
};
```

This meant that calling `.unsubscribe()` before the response arrived left the stream consumer running indefinitely — wasting bandwidth, keeping the connection open, and potentially causing "observer already completed" errors.

## Fix

Two changes to `packages/client/src/links/httpBatchStreamLink.ts`:

1. **Per-observable `AbortController`**: Each observable now creates its own `ac = new AbortController()`. The operation's signal is wrapped as `raceAbortSignals(op.signal, ac.signal)` so either signal can trigger cancellation. The teardown calls `ac.abort()` — but only when `isDone` is `false` (i.e. the observable was explicitly unsubscribed before completing, not torn down after natural completion).

2. **Wire batch signals to the JSONL stream controller**: `batchSignals` (which fires when *all* operations in a batch have had their signals aborted, via `allAbortSignals`) now triggers `abortController.abort()` via an `addEventListener`. This means: unsubscribe all ops in a batch → all per-observable `ac`s abort → `batchSignals` fires → JSONL stream consumer stops.

The `isDone` guard is essential for async-iterable streaming queries: the observable completes naturally (delivering the iterable as the result), but the JSONL stream must keep flowing to deliver the iterable's yielded values. Without the guard, completing the observable would also abort the stream mid-iterable.

## Test

Added `packages/tests/server/regression/issue-7389-batch-stream-abort.test.ts` with three tests:

- **Unsubscribe before response arrives aborts the stream** — spies on `AbortController.prototype.abort` and asserts it was called after unsubscribing from a pending observable. Fails without the fix, passes with it.
- **Natural completion does not trigger the abort guard** — verifies the happy path still works correctly.
- **Batched scenario** — two operations in the same batch; unsubscribing from one while the other completes does not break the second operation's result.

All 75 existing tests in `batching.test.ts`, `streaming.test.ts`, and related regression suites continue to pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cancellation for JSONL batch streaming: when subscribers unsubscribe or the batch/operation signal is already aborted, streaming consumption stops promptly and pending streams are aborted reliably, without preventing other batched operations from completing successfully.

* **Tests**
  * Added regression coverage for abort propagation and settlement behavior across pending, completed, mixed batched requests, and pre-aborted operation signals (including correct AbortError handling).

### DIFF
--- a/packages/client/src/links/httpBatchStreamLink.ts
+++ b/packages/client/src/links/httpBatchStreamLink.ts
@@ -73,6 +73,15 @@ export function httpBatchStreamLink<TRouter extends AnyRouter>(
             ...batchOps.map((op) => op.signal),
           );
           const abortController = new AbortController();
+          // When all subscribers have unsubscribed (batchSignals fires), abort
+          // the JSONL stream consumer so the response body is no longer read.
+          batchSignals.addEventListener(
+            'abort',
+            () => abortController.abort(),
+            {
+              once: true,
+            },
+          );
 
           const responsePromise = fetchHTTPResponse({
             ...resolvedOpts,
@@ -180,14 +189,20 @@ export function httpBatchStreamLink<TRouter extends AnyRouter>(
             'Subscriptions are unsupported by `httpBatchStreamLink` - use `httpSubscriptionLink` or `wsLink`',
           );
         }
+        const ac = new AbortController();
         const loader = loaders[op.type];
-        const promise = loader.load(op);
+        const promise = loader.load({
+          ...op,
+          signal: raceAbortSignals(op.signal, ac.signal),
+        });
 
+        let isDone = false;
         let _res = undefined as HTTPResult | undefined;
         promise
           .then((res) => {
             _res = res;
             if ('error' in res.json) {
+              isDone = true;
               observer.error(
                 TRPCClientError.from(res.json, {
                   meta: res.meta,
@@ -195,6 +210,7 @@ export function httpBatchStreamLink<TRouter extends AnyRouter>(
               );
               return;
             } else if ('result' in res.json) {
+              isDone = true;
               observer.next({
                 context: res.meta,
                 result: res.json.result,
@@ -203,9 +219,11 @@ export function httpBatchStreamLink<TRouter extends AnyRouter>(
               return;
             }
 
+            isDone = true;
             observer.complete();
           })
           .catch((err) => {
+            isDone = true;
             observer.error(
               TRPCClientError.from(err, {
                 meta: _res?.meta,
@@ -214,7 +232,12 @@ export function httpBatchStreamLink<TRouter extends AnyRouter>(
           });
 
         return () => {
-          // noop
+          if (!isDone) {
+            // Only abort when the subscriber explicitly unsubscribes before
+            // the operation completes. This stops the JSONL stream consumer
+            // from continuing to read after no one is listening.
+            ac.abort();
+          }
         };
       });
     };
--- a/packages/client/src/links/httpBatchStreamLink.ts
+++ b/packages/client/src/links/httpBatchStreamLink.ts
@@ -73,15 +73,24 @@ export function httpBatchStreamLink<TRouter extends AnyRouter>(
             ...batchOps.map((op) => op.signal),
           );
           const abortController = new AbortController();
-          // When all subscribers have unsubscribed (batchSignals fires), abort
+          // When every subscriber has unsubscribed (`batchSignals` fires) abort
           // the JSONL stream consumer so the response body is no longer read.
-          batchSignals.addEventListener(
-            'abort',
-            () => abortController.abort(),
-            {
+          //
+          // `batchSignals` may already be aborted by the time we get here (e.g.
+          // all subscribers unsubscribed before the batch was dispatched). In
+          // that case `addEventListener('abort', ...)` would never fire, so we
+          // forward the abort eagerly. We also forward `batchSignals.reason` so
+          // the stream consumer rejects with the original cause rather than a
+          // generic `AbortError`.
+          const abortStreamConsumer = () =>
+            abortController.abort(batchSignals.reason);
+          if (batchSignals.aborted) {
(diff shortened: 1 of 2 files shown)


daltino (commented):
## Review: fix(client): abort JSONL stream consumer on httpBatchStreamLink unsubscribe

Great catch and a clear root cause explanation! The fix is conceptually correct. Here are a few things worth discussing before merging.

---

### ✅ What's working well

- The root cause analysis is spot-on: `AbortController` was wired into `jsonlStreamConsumer` but never triggered on unsubscribe.
- Using `batchSignals` (which already aggregates per-operation signals) as the trigger is elegant — it avoids duplicating abort logic.
- The `{ once: true }` option on `addEventListener` is a nice touch to prevent any potential double-abort or listener leak.
- Adding a dedicated regression test file is exactly the right approach.

---

### 🔍 Concerns & Suggestions

#### 1. Listener leak if `batchSignals` is already aborted before the listener is added

`AbortSignal.addEventListener` with `{ once: true }` won't fire if the signal is **already aborted** at the time the listener is registered. You'd silently skip the abort of the stream consumer in that case.

```ts
// Suggested guard:
if (batchSignals.aborted) {
  abortController.abort();
} else {
  batchSignals.addEventListener(
    'abort',
    () => abortController.abort(),
    { once: true },
  );
}
```

Or more concisely, use `AbortSignal.any([batchSignals])` if the runtime supports it, which handles the already-aborted case automatically. Given tRPC already constructs composite signals in this file, this pattern may already be in use elsewhere.

---

#### 2. `abortController.abort()` reason

Right now the abort reason propagates from `batchSignals` implicitly (the reason on `batchSignals` isn't forwarded). Consider forwarding it:

```ts
batchSignals.addEventListener(
  'abort',
  () => abortController.abort(batchSignals.reason),
  { once: true },
);
```

This makes downstream error handling (e.g. distinguishing user-initiated abort vs network error) easier to reason about.

---

#### 3. Missing cleanup of the `batchSignals` listener

If `abortController` is aborted for a *different* reason (e.g. the fetch itself errors out or completes), the listener on `batchSignals` will remain registered until `batchSignals` fires — which may never happen if the caller holds a long-lived signal. This is a minor but real listener leak.

```ts
const onBatchAbort = () => abortController.abort(batchSignals.reason);
batchSignals.addEventListener('abort', onBatchAbort, { once: true });

// Later, when the stream finishes or errors:
// batchSignals.removeEventListener('abort', onBatchAbort);
```

Alternatively, using `AbortSignal.any([batchSignals, abortController.signal])` as the signal passed to `jsonlStreamConsumer` (instead of creating a secondary controller) would sidestep this entirely — but that depends on whether `jsonlStreamConsumer` accepts a signal directly.

---

#### 4. Test file — partial patch shown

The diff shows the test file was added (215 lines) but only the first few lines are visible in the patch. A couple of questions based on what's visible:

- Does the test assert that the **response body stream is actually closed** (e.g. via a `ReadableStream` cancellation callback or a mock server seeing the connection drop), or only that the observable completes without hanging? The former gives stronger guarantees.
- Is there a test for the **already-aborted-before-subscribe** edge case mentioned in point 1?
- Is there a test covering a **mid-stream unsubscribe** (i.e. unsubscribe after the first JSONL chunk arrives but before the last)?

---

### Summary

The core fix is correct and well-motivated. The main blocker-level concern is **point 1** (already-aborted signal silently skipping the abort). Points 2 and 3 are improvements worth considering. Happy to approve once those are addressed!

chatman-media:
Thanks for the careful review — all four points addressed in the latest push.

**1. Already-aborted batch signal.** You're right that `addEventListener('abort', …, { once: true })` never fires if `batchSignals` is already aborted when we register it. One subtlety I confirmed while digging in: the *hang* itself was actually masked today, because the request signal is `raceAbortSignals(batchSignals, abortController.signal)` and `raceAbortSignals` propagates an already-aborted input synchronously — so `fetchHTTPResponse`'s `throwIfAborted` was bailing out before the consumer was ever created, making the old listener effectively dead code in the already-aborted path. I've made it explicitly correct regardless: guard with `if (batchSignals.aborted)` and abort eagerly, otherwise attach the listener.

**2. Forward the abort reason.** Done — both the eager path and the listener now call `abortController.abort(batchSignals.reason)` so the stream consumer rejects with the cause rather than a bare `AbortError`. (Heads-up: `batchSignals.reason` is currently a generic `AbortError` because `allAbortSignals` flattens via `ac.abort()` with no argument; forwarding the *originating* op reason would mean teaching `allAbortSignals`/`raceAbortSignals` to carry a reason, which I left out here to keep the change contained — happy to do it as a follow-up.)

**3. Dangling listener.** Kept `{ once: true }` for the non-already-aborted path. `batchSignals` is a fresh per-batch signal, so it's collected with the batch and the `once` listener auto-removes after firing — no long-lived leak.

**4. Tests.** Added a regression test (`aborts the stream consumer when the batch signal is already aborted before dispatch`) that pre-aborts the op signal before the chain runs and asserts both that the observable settles (no hang) and that the consumer's `AbortController` is aborted with the forwarded `AbortError` reason. It fails on the previous code and passes with this fix (verified by stashing); the existing mid-stream-unsubscribe tests still pass.

PR #7413, opened 2026-06-14
fix(client): abort httpBatchStreamLink stream on observer unsubscribe

### LINKED ISSUE #7389: bug: `httpBatchStreamLink` AbortController is created but never called on unsubscribe

sulaxmi22:
## Problem

The cleanup function in `httpBatchStreamLink`'s observable was a noop:

```ts
return () => {
  // noop
};
```

When an observer unsubscribes (component unmounts, query cancelled), the underlying HTTP request and JSONL stream kept running — wasting bandwidth and causing `observer already completed` errors.

Fixes #7389

## Fix

Create a per-observer `AbortController` at the observable level and merge its signal into `op.signal` via `raceAbortSignals`. Call `ac.abort()` in cleanup.

```ts
const promise = loader.load({
  ...op,
  signal: raceAbortSignals(op.signal, ac.signal),
});
// ...
return () => {
  ac.abort();
};
```

**Propagation chain:**
1. Cleanup → `ac.abort()`
2. `op.signal` (merged via raceAbortSignals) fires
3. `batchSignals = allAbortSignals(...ops.map(op => op.signal))` fires once all ops abort
4. `raceAbortSignals(batchSignals, abortController.signal)` fires → HTTP fetch aborted
5. `jsonlStreamConsumer` stops reading

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup and cancellation behavior for batch streaming requests. The client now reliably releases the shared JSONL response stream once all per-operation results have finished, preventing lingering readers and unnecessary work.
  * If an operation is unsubscribed before it settles, the underlying batch request is cancelled to stop in-flight work, while avoiding noisy abort-related errors.

### DIFF
--- a/packages/client/src/links/httpBatchStreamLink.ts
+++ b/packages/client/src/links/httpBatchStreamLink.ts
@@ -163,6 +163,7 @@ export function httpBatchStreamLink<TRouter extends AnyRouter>(
               };
             },
           );
+
           return promises;
         },
       };
@@ -181,11 +182,28 @@ export function httpBatchStreamLink<TRouter extends AnyRouter>(
           );
         }
         const loader = loaders[op.type];
-        const promise = loader.load(op);
+
+        // Track whether this op's loader promise has settled. We use this to
+        // decide whether to cancel the underlying batch when the observer is
+        // torn down: if the promise hasn't resolved yet, there is no reason to
+        // keep the HTTP stream open, so we abort it. Once the promise resolves
+        // we leave the stream alone — for streaming (iterable) responses the
+        // consumer is still reading body chunks after the promise resolves, and
+        // jsonlStreamConsumer automatically aborts the body once all listeners
+        // are done.
+        let settled = false;
+        let selfAborted = false;
+        const ac = new AbortController();
+
+        const promise = loader.load({
+          ...op,
+          signal: raceAbortSignals(op.signal, ac.signal),
+        });
 
         let _res = undefined as HTTPResult | undefined;
         promise
           .then((res) => {
+            settled = true;
             _res = res;
             if ('error' in res.json) {
               observer.error(
@@ -206,6 +224,11 @@ export function httpBatchStreamLink<TRouter extends AnyRouter>(
             observer.complete();
           })
           .catch((err) => {
+            if (selfAborted) {
+              // We triggered this abort via cleanup; the observer is already
+              // being torn down so there is nothing to propagate.
+              return;
+            }
             observer.error(
               TRPCClientError.from(err, {
                 meta: _res?.meta,
@@ -214,7 +237,12 @@ export function httpBatchStreamLink<TRouter extends AnyRouter>(
           });
 
         return () => {
-          // noop
+          if (!settled) {
+            // The promise hasn't resolved yet — cancel the underlying request
+            // so the server stops sending and the response body is released.
+            selfAborted = true;
+            ac.abort();
+          }
         };
       });
     };


daltino (commented):
Great direction on this fix — the problem description is clear, and aborting the underlying HTTP request on unsubscribe is absolutely the right thing to do. The noop cleanup has been a silent bandwidth-waster. A few things worth discussing before this lands:

---

## Overall Impression

The intent is solid. Aborting inflight requests when observers unsubscribe prevents wasted bandwidth and eliminates the `observer already completed` errors. This is a well-understood pattern and the right fix for #7389.

That said, the patch shown is truncated (the description even trails off mid-sentence), so some of my comments are based on what I'd *expect* to see and what could go wrong. I'll flag those clearly.

---

## Correctness Concerns

### 1. Batching interaction — aborting one observer shouldn't abort others

This is the most critical edge case. `httpBatchStreamLink` batches multiple operations into a single HTTP request. If observer A unsubscribes but observer B is still active and they share the same batch/fetch call, aborting the `AbortController` for A must **not** abort the underlying fetch that B depends on.

The patch introduces a per-observer `AbortController`, which is good framing, but the abort signal needs to be scoped carefully. Make sure the signal is only passed in a way that tears down *this observer's subscription logic*, not the shared fetch. If the batch's `fetch` receives A's abort signal directly, B will break silently.

**Suggestion:** Consider reference-counting active subscribers per batch. Only abort the underlying fetch when *all* observers for that batch have unsubscribed.

---

### 2. The `settled` flag approach

The patch mentions tracking whether the loader promise has settled to decide whether to cancel. This is reasonable, but watch out for a subtle race:

```
1. Promise settles (e.g., first chunk received)
2. settled = true
3. Observer unsubscribes
4. Cleanup sees settled === true → skips abort
5. But the stream is still running (more chunks incoming)
```

A promise "resolving" in a streaming context often just means the *first* response chunk arrived, not that the stream is done. If `settled` is set on first resolution rather than full stream completion, you'll miss aborting long-running streams. Consider naming this more precisely (e.g., `streamComplete`) and only setting it when the JSONL stream has fully ended.

---

### 3. AbortError handling downstream

When you abort, the fetch will reject with an `AbortError`. Make sure this is caught and *not* forwarded to the observer as an error — since the observer itself initiated the unsubscribe, it doesn't care. Check that the error handling path distinguishes:

```ts
if (err.name === 'AbortError') return; // subscriber left, not an error
```

Without this, you may swap one bug (observer already completed) for another (spurious error callbacks).

---

## Test Coverage

The PR description mentions fixing a real bug with a repro issue (#7389), but I don't see test additions in the patch. This fix needs at least:

1. **Unit test:** Observer unsubscribes before response → fetch is aborted (spy on `AbortController.prototype.abort` or mock fetch).
2. **Batch test:** Two observers share a batch; one unsubscribes → fetch continues, second observer still receives data.
3. **Race test:** Unsubscribe fires after first chunk but before stream ends → stream is still aborted.

Without these, the fix could regress silently. The existing test suite likely doesn't cover abort behavior since the old code was a noop.

---

## Code Style / Minor Notes

- The blank line added before `return promises;` is fine and improves readability.
- The comment `// Track whether this op's loader promise has settled...` is helpful, but "settled" and "resolved" are different things — worth being precise in the comment if the intent is "stream fully complete."
- If there's an `AbortController` created per-observer, make sure it's not created unconditionally on every `subscribe` call when the environment might not support `AbortController` (though this is probably fine for tRPC's target environments).

---

## Summary

| Area | Status |
|---|---|
| Abort on unsubscribe | ✅ Right approach |
| Batching isolation | ⚠️ Needs verification |
| Settled flag semantics | ⚠️ Potential race with streams |
| AbortError swallowed | ⚠️ Needs confirmation |
| Test coverage | ❌ Missing |

Fix the batching isolation and add tests, and this is ready to merge. Happy to look at a follow-up diff if you can share the full patch!

PR #7428, opened 2026-07-04
fix(react-query): usable proxy type for createTRPCReact<AnyTRPCRouter>

### LINKED ISSUE #6721: bug: AnyTRPCRouter immediately collides

```
System:
    OS: macOS 14.7.4
    CPU: (14) arm64 Apple M3 Max
    Memory: 808.66 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.13.1 - ~/.nvm/versions/node/v22.13.1/bin/node
    Yarn: 1.22.22 - ~/Library/pnpm/yarn
    npm: 10.9.2 - ~/.nvm/versions/node/v22.13.1/bin/npm
    pnpm: 9.15.5 - ~/.nvm/versions/node/v22.13.1/bin/pnpm
    bun: 1.2.0 - ~/.bun/bin/bun
  Browsers:
    Chrome: 135.0.7049.96
    Safari: 18.4
  npmPackages:
    @tanstack/react-query: ^5.74.4 => 5.74.4 
    @trpc/client: catalog: => 11.1.0 
    @trpc/react-query: catalog: => 11.1.0 
    @trpc/server: catalog: => 11.1.0 
    react: 18.2.0 => 18.2.0 
    typescript: ^5.8.3 => 5.8.3
```

### Describe the bug

The following code:

```typescript
import { createTRPCReact } from "@trpc/react-query";
import type { AnyTRPCRouter } from "@trpc/server";

export const trpc = createTRPCReact<AnyTRPCRouter>();
```

Leads to the trpc variable not having a proper type. It doesn't seem like this should be the case for a built in type meant to represent any type of router.

```typescript
const trpc: "The property 'useContext' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'useUtils' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'Provider' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'createClient' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'useQueries' in your router collides with a built-in method, rename this router or procedure on your backend." | "The property 'useSuspenseQueries' in your router collides with a built-in method, rename this router or procedure on your backend."
```

### Link to reproduction

https://stackblitz.com/edit/github-6ltxttds?file=src%2Futils%2Ftrpc.ts

### To reproduce

Open the linked reproduction and hover your mouse over the `trpc` variable.

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

momomuchu:
`createTRPCReact<AnyTRPCRouter>()` produces a union of false `IntersectionError` strings ("collides with a built-in method") instead of a usable proxy type (#6721).

`ProtectedIntersection<TType, TWith>` reports a collision when `keyof TType & keyof TWith` is not `never`. For `AnyTRPCRouter`, `keyof TWith` collapses to the bare `string` index signature, so `string & 'useContext'` is not `never` and every built-in key is flagged as colliding.

This guards the type: when `keyof TWith` is (or includes) the bare `string` index signature, it returns the plain intersection instead of the error. A concrete key colliding with a wide index signature is not constructible in TypeScript (an index signature collapses `keyof` to `string`), so the guard is exhaustive.

Added a regression type test. It fails to typecheck without the fix and passes with it. All packages typecheck, full vitest green, the existing collision-regression test is unchanged and still passes.

Closes #6721

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved type-collision false positives when merging router types that include broad string index signatures, preventing incorrect intersection error types.
  * Ensured key React Query/TRPC members keep their correct built-in typings instead of degrading into unusable types.

* **Tests**
  * Added a regression test verifying core React Query/TRPC hook typings remain intact.
  * Improved an optimistic update test fixture by resetting shared state up front to avoid stale-data flakiness.

### DIFF
--- a/packages/react-query/test/regression/issue-6721-anyrouter-intersection.test.tsx
+++ b/packages/react-query/test/regression/issue-6721-anyrouter-intersection.test.tsx
@@ -11,8 +11,8 @@ test('createTRPCReact<AnyTRPCRouter>() does not collide with itself', () => {
   // If it does, `trpc` collapses to a union of `IntersectionError<...>` string
   // literals and none of its built-in methods are accessible below - this
   // block fails to compile on the bug.
-  expectTypeOf(trpc.useContext).toBeFunction();
-  expectTypeOf(trpc.useUtils).toBeFunction();
+  expectTypeOf<typeof trpc.useContext>().toBeFunction();
+  expectTypeOf<typeof trpc.useUtils>().toBeFunction();
   expectTypeOf(trpc.Provider).not.toBeAny();
   expectTypeOf(trpc.createClient).toBeFunction();
   expectTypeOf(trpc.useQueries).toBeFunction();
--- a/packages/react-query/test/regression/issue-6721-anyrouter-intersection.test.tsx
+++ b/packages/react-query/test/regression/issue-6721-anyrouter-intersection.test.tsx
@@ -1,0 +1,20 @@
+import { createTRPCReact } from '@trpc/react-query';
+import type { AnyTRPCRouter } from '@trpc/server';
+
+test('createTRPCReact<AnyTRPCRouter>() does not collide with itself', () => {
+  const trpc = createTRPCReact<AnyTRPCRouter>();
+
+  // `AnyTRPCRouter` is `Router<any, any>`, whose `keyof` resolves to the bare
+  // `string` index signature. `ProtectedIntersection` must not treat that as
+  // colliding with every built-in method (`useContext`, `useUtils`, `Provider`,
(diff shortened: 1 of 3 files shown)


momomuchu:
The failing optimistic-update test was a pre-existing flake in this fixture, not a regression from this change (the diff is a type-level fix in a different package). The fixture shares a module-level queryClient across runs and the cache is never cleared between them (see #6783/#6808), so a later run can render stale cached data before the refetch settles. Fixed the isolation (cancel + clear the query cache and reset fixture state at the start of each run) in a separate commit for easy review/revert; the assertions are unchanged. 20/20 green under stress.

PR #7434, opened 2026-07-14
fix(client): wire the operation AbortSignal in the WebSocket link

### LINKED ISSUE #7431: WebSocket link silently ignores the operation AbortSignal (removeEventListener with no matching addEventListener)
### Description

### Root cause

### Steps to reproduce

### Suggested fix

ATKasem:
Closes #7431

## 🎯 Changes

The WebSocket link's `request` observable never listened to the operation's `AbortSignal`. Its teardown called `signal?.removeEventListener('abort', abort)`, but nothing ever registered that listener — `abort` was the function returned by `batchSend`, not a listener. Aborting an operation's signal did nothing: the request was never torn down, the subscription kept running, and its server-side listener leaked.

Register an `abort` listener that completes the observer. Completing runs the existing teardown, which sends `subscription.stop` for subscriptions, so an aborted operation now ends cleanly. The matching listener is removed on teardown, and an already-aborted signal is handled synchronously.

## ✅ Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket request handling when an abort signal is triggered, including signals that were already aborted.
  * Ensured active subscriptions stop promptly and clean up their event listeners after cancellation, preventing further message delivery.

* **Tests**
  * Added coverage confirming that aborting a WebSocket message subscription removes its server-side listener and stops subsequent processing.

### DIFF
--- a/packages/client/src/links/wsLink/wsClient/wsClient.ts
+++ b/packages/client/src/links/wsLink/wsClient/wsClient.ts
@@ -222,6 +222,16 @@ export class WsClient {
         },
       );
 
+      // Wire the operation's AbortSignal: aborting must tear the request down
+      // (and send `subscription.stop` for subscriptions). Completing the
+      // observer runs the teardown below, which does exactly that.
+      const onAbort = () => observer.complete();
+      if (signal?.aborted) {
+        onAbort();
+      } else {
+        signal?.addEventListener('abort', onAbort, { once: true });
+      }
+
       return () => {
         abort();
 
@@ -232,7 +242,7 @@ export class WsClient {
           });
         }
 
-        signal?.removeEventListener('abort', abort);
+        signal?.removeEventListener('abort', onAbort);
       };
     });
   }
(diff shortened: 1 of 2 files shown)


ATKasem:
Just a heads-up in case it's helpful: the failing `unit and integration tests` check seems to be coming from `@trpc/upgrade`'s `transforms.test.ts` (the `optimistic-update` hooks case), which looks unrelated to this PR's WebSocket change. I noticed it's also red on some other open PRs, so I think it may be failing on `main` rather than something here — but please let me know if I'm missing something and I'm happy to dig in further.

packages/client/src/links/wsLink/wsClient/wsClient.ts
```diff
@@ -222,6 +222,16 @@ export class WsClient {
         },
       );
 
+      // Wire the operation's AbortSignal: aborting must tear the request down
+      // (and send `subscription.stop` for subscriptions). Completing the
+      // observer runs the teardown below, which does exactly that.
```
Nick-Lucas:
```suggestion

```

Nick-Lucas (approved):

PR #4872, opened 2023-10-02
feat(`server`): extract parseTRPCMessage for 3rd party modules

Guria:
## 🎯 Changes

This PR adds an adapter/link pair to run trpc server in a webworker.
It might be useful for example as temporary solution for migrating from SPA without nodejs backend.

## ✅ Checklist

- [x] I have followed the steps listed in the [Contributing guide](https://github.com/trpc/trpc/blob/main/CONTRIBUTING.md).
- [ ] If necessary, I have added documentation related to the changes made. 
- [ ] I have added or updated the tests related to the changes made.

Docs and tests after discussion :)

PR #7435, opened 2026-07-15
fix(server): reject non-number/string request ids in assertIsRequestId

### LINKED ISSUE #7430: assertIsRequestId is a no-op for all non-number invalid types (only throws for NaN)
### Description
The JSON-RPC request-id validator in the WebSocket message parser is meant to reject anything that isn't `number | string | null`, but its boolean logic only ever throws for `NaN`. Booleans, objects, arrays, and `undefined` all pass as valid request ids and flow downstream into `clientSubscriptions` Map keys and error messages.

### Root cause
`packages/server/src/unstable-core-do-not-import/rpc/parseTRPCMessage.ts`:
```js
if (obj !== null && typeof obj === 'number' && isNaN(obj) && typeof obj !== 'string') {
  throw new Error('Invalid request id');
}
```
`typeof obj === 'number'` combined with `typeof obj !== 'string'` (always true once the former holds) plus `isNaN(obj)` means the throw fires only for `NaN`. Git history shows the original (PR #508) was correct -- `if (typeof obj !== 'number' || isNaN(obj))` -- and the boolean got inverted when the type was widened to `number | string | null`.

### Steps to reproduce
```
NaN          -> throws (correct)
true         -> passes (should throw)
{}           -> passes (should throw)
[]           -> passes (should throw)
undefined    -> passes (should throw)
```

### Suggested fix
`if (obj !== null && typeof obj !== 'string' && (typeof obj !== 'number' || isNaN(obj))) throw new Error('Invalid request id');`

*Severity note:* it's a defense-in-depth assertion (`/* istanbul ignore */`); malformed ids still function as Map keys, so impact is spec-conformance rather than a crash. Reported as Medium.

xianjianlf2:
### What & why

`assertIsRequestId` in `packages/server/src/unstable-core-do-not-import/rpc/parseTRPCMessage.ts` is meant to accept `number | string | null` and reject everything else, but the condition is a no-op for all invalid non-number types:

```js
obj !== null && typeof obj === 'number' && isNaN(obj) && typeof obj !== 'string'
```

Once `typeof obj === 'number'` holds, `typeof obj !== 'string'` is always true, so the `throw` only ever fires for `NaN`. Booleans, objects, arrays, and `undefined` all pass as "valid" request ids and flow downstream into `clientSubscriptions` Map keys and error responses (this is the incoming-message parser used by the WebSocket adapter).

**Regression:** the original check in PR #508 was correct (`if (typeof obj !== 'number' || isNaN(obj))`); the boolean was mangled when the id type was widened to `number | string | null` and the function was extracted into `parseTRPCMessage` in PR #4872.

### Fix

Restore the correct logic:

```js
obj !== null && typeof obj !== 'string' && (typeof obj !== 'number' || isNaN(obj))
```

### Tests

New colocated `parseTRPCMessage.test.ts` (matching the repo's colocated-test convention). Buggy source: 5 failing (`true`, `false`, `{}`, `[]`, `undefined` don't throw). Fixed: 14 passing. Adjacent WS suites unaffected (control frames never pass through `parseTRPCMessage`).

Closes #7430

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of request IDs in incoming WebSocket messages.
  * Valid numeric, string, and null IDs are now accepted reliably.
  * Invalid IDs—including booleans, objects, arrays, undefined values, and `NaN`—are rejected with a clear error before further processing.
* **Tests**
  * Added coverage for valid and invalid request ID formats to help prevent regressions.

### DIFF
--- a/packages/server/src/unstable-core-do-not-import/rpc/parseTRPCMessage.test.ts
+++ b/packages/server/src/unstable-core-do-not-import/rpc/parseTRPCMessage.test.ts
@@ -1,0 +1,30 @@
+import { defaultTransformer } from '../transformer';
+import { parseTRPCMessage } from './parseTRPCMessage';
+
+/**
+ * `parseTRPCMessage` validates the JSON-RPC `id` of an incoming WebSocket
+ * message. A valid request id is `number | string | null`; anything else must
+ * be rejected before it is used as a `clientSubscriptions` Map key.
+ */
+const parse = (id: unknown) =>
+  parseTRPCMessage(
+    { id, jsonrpc: '2.0', method: 'subscription.stop' },
+    defaultTransformer,
+  );
+
+describe('parseTRPCMessage - request id validation', () => {
+  test.each([1, 0, -1, 42, 'abc', '', 'my-id', null])(
+    'accepts valid request id: %p',
+    (id) => {
+      expect(() => parse(id)).not.toThrow();
+      expect(parse(id)).toMatchObject({ id, method: 'subscription.stop' });
+    },
+  );
+
+  test.each([true, false, {}, [], undefined, NaN])(
+    'rejects invalid request id: %p',
+    (id) => {
+      expect(() => parse(id)).toThrow('Invalid request id');
+    },
+  );
+});
--- a/packages/server/src/unstable-core-do-not-import/rpc/parseTRPCMessage.ts
+++ b/packages/server/src/unstable-core-do-not-import/rpc/parseTRPCMessage.ts
@@ -23,9 +23,8 @@ function assertIsRequestId(
 ): asserts obj is number | string | null {
   if (
     obj !== null &&
-    typeof obj === 'number' &&
-    isNaN(obj) &&
-    typeof obj !== 'string'
+    typeof obj !== 'string' &&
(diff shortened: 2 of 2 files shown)


xianjianlf2:
The failing **unit and integration tests** check is the flaky `@trpc/upgrade` transform test (`hooks optimistic-update.tsx`), unrelated to this change — it fails the same way on main ([run 28380538535](https://github.com/trpc/trpc/actions/runs/28380538535)) and on unrelated PRs, including a comment-typo-only change ([run 29062632262](https://github.com/trpc/trpc/actions/runs/29062632262)). Everything else in this PR's run is green: 1140/1141 tests, 170/171 files. A re-run may get it past, or it can be evaluated on the other checks.

xianjianlf2:
Updated this branch with latest `main` using a normal merge commit, so CI can rerun on the current base.

Local check after the merge:
- `pnpm vitest run packages/server/src/unstable-core-do-not-import/rpc/parseTRPCMessage.test.ts`

Note: local pnpm printed the repo engine warning because this machine is on Node 22 while the repo asks for Node 24, but the targeted test passed.

PR #7442, opened 2026-07-23
fix(client): handle AbortSignal events and pre-aborted signals in wsLink

### LINKED ISSUE #7439: wsLink: AbortSignal on operations is ignored — abort does not cancel in-flight requests or subscriptions
## Description
The WebSocket link does not honor the `AbortSignal` passed with an operation. Calling `abort()` on the signal has no effect on in-flight queries, mutations, or subscriptions sent through `wsLink`.

## Repro
```ts
const controller = new AbortController();
const result = trpc.someQuery.query(undefined, { signal: controller.signal });
controller.abort(); // has no effect on the WS request
```

Contrast with `httpSubscriptionLink` which correctly tears down the subscription when the signal fires.

## Root cause
`wsClient.ts:235` calls `signal?.removeEventListener('abort', abort)` in the cleanup path — but there is no corresponding `signal?.addEventListener('abort', ...)` anywhere in the wsLink source. The abort listener was never registered; only the cleanup was scaffolded.

This appears to be an incomplete implementation from the v11 refactor (PR #6223) which plumbed `signal` into the request function but only wired the teardown half.

Verified via the current source — `signal` is received at lines ~186 and ~190, but only `removeEventListener` is called (line 235); no `addEventListener` call exists anywhere in the wsLink directory.

## Expected behaviour
When the caller's `AbortSignal` fires:
- In-flight queries/mutations: the pending request should be cancelled and the promise rejected with `AbortError`
- Subscriptions: the subscription should be torn down

## Possible fix
In the `request()` function, after the early-exit check for `signal?.aborted`, add:
```ts
signal?.addEventListener('abort', teardownFn, { once: true });
```
and ensure `teardownFn` removes the listener when the operation completes normally (which is what the existing `removeEventListener` call at line 235 already does — it just has no matching registration).

aryamirani:
### Summary

Fixes #7439 where `wsLink` ignored the `AbortSignal` passed with an operation. Calling `controller.abort()` on an `AbortController` had no effect on in-flight queries, mutations, or subscriptions over WebSocket because the abort event listener was never registered on `signal`.

### Problem Statement & Root Cause

In `packages/client/src/links/wsLink/wsClient/wsClient.ts`, `signal?.removeEventListener('abort', abort)` was called in the cleanup function returned by `request()`, but `signal?.addEventListener('abort', ...)` was never called during operation setup. Additionally, pre-aborted `AbortSignal` instances were not checked before sending batch messages.

### Solution

1. **Abort Listener Registration**: Register a `{ once: true }` listener for `'abort'` on `op.signal` when an operation is dispatched. When triggered, it cleans up the pending request from `RequestManager`, sends `subscription.stop` if an active subscription, and notifies observers with a `TRPCClientError`.
2. **Pre-aborted Check**: Check `signal?.aborted` before dispatching, rejecting immediately if already aborted.
3. **Clean Teardown**: Remove event listener on normal completion or unsubscription.

### Test Coverage

Added integration tests in `packages/tests/server/websockets.test.ts`:
- Verifies in-flight query cancellation upon `AbortSignal` trigger over WebSocket.
- Verifies immediate rejection when passing an already-aborted `AbortSignal`.

Tested via `npx vitest run packages/tests/server/websockets.test.ts` (45/45 tests passing).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cancellation behavior for WebSocket-backed queries using `AbortSignal`, including immediate rejection when the signal is already aborted.
  * Ensures in-flight requests and active subscriptions are stopped cleanly, and sends stop messages when the WebSocket connection is open.

* **Tests**
  * Added coverage for aborting in-flight WebSocket queries and for handling already-aborted signals.

### DIFF
--- a/packages/client/src/links/wsLink/wsClient/wsClient.ts
+++ b/packages/client/src/links/wsLink/wsClient/wsClient.ts
@@ -25,6 +25,12 @@ import { RequestManager } from './requestManager';
 import { ResettableTimeout, TRPCWebSocketClosedError } from './utils';
 import { backwardCompatibility, WsConnection } from './wsConnection';
 
+function toAbortError(signal?: AbortSignal): Error {
+  const reason = signal?.reason;
+  if (reason instanceof Error) return reason;
+  return new Error(typeof reason === 'string' ? reason : 'Operation aborted');
+}
+
 /**
  * A WebSocket client for managing TRPC operations, supporting lazy initialization,
  * reconnection, keep-alive, and request management.
@@ -195,7 +201,14 @@ export class WsClient {
       OperationResultEnvelope<unknown, TRPCClientError<AnyTRPCRouter>>,
       TRPCClientError<AnyTRPCRouter>
     >((observer) => {
-      const abort = this.batchSend(
+      let isDone = false;
+
+      if (signal?.aborted) {
+        observer.error(TRPCClientError.from(toAbortError(signal)));
+        return;
+      }
+
+      const batchSendAbort = this.batchSend(
         {
           id,
           method: type,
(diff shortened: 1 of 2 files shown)


Nick-Lucas:
Duplicate of this https://github.com/trpc/trpc/pull/7434

PR #7444, opened 2026-07-23
fix(client): wire AbortSignal event listener to cancel in-flight operations

### LINKED ISSUE #7431: WebSocket link silently ignores the operation AbortSignal (removeEventListener with no matching addEventListener)

namdamdoi68-oss:
Wire AbortSignal event listener and early aborted check in `wsClient` to reject in-flight operations when signal is aborted.

Closes #7431

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket request cancellation with consistent abort errors.
  * Requests cancelled before starting now fail immediately.
  * Active subscriptions now stop cleanly when cancelled, preventing further server messages.


namdamdoi68-oss:
Hi @KATT @dependabot[bot] — PR is ready for review. All CI tests pass cleanly.

PR #7449, opened 2026-07-26
fix(react-query): allow useQueries callback to return conditional empty array

### LINKED ISSUE #6188: bug: React: Types for callback of `trpc.useQueries` doesn't allow returning an empty array

```
  System:
    OS: macOS 14.4.1
    CPU: (8) arm64 Apple M1
    Memory: 62.41 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.10.0 - ~/.nvm/versions/node/v20.10.0/bin/node
    npm: 10.2.3 - ~/.nvm/versions/node/v20.10.0/bin/npm
    pnpm: 9.12.2 - ~/.nvm/versions/node/v20.10.0/bin/pnpm
  Browsers:
    Chrome: 130.0.6723.92
    Safari: 17.4.1
  npmPackages:
    @tanstack/react-query: ^5.59.19 => 5.59.19 
    @trpc/client: 11.0.0-rc.608 => 11.0.0-rc.608+f75de97b3 
    @trpc/react-query: 11.0.0-rc.608 => 11.0.0-rc.608+f75de97b3 
    react: ^18.3.1 => 18.3.1 
    typescript: ^5.6.3 => 5.6.3 
```

### Describe the bug

I have a query which I just want to run in a certain condition (if another query had a non-empty result). I was trying to follow the docs (see https://trpc.io/docs/client/react/useQueries in conjunction with https://tanstack.com/query/v4/docs/framework/react/guides/dependent-queries#usequeries-dependent-query) but I could not get my code snippet to compile.

### Link to reproduction

https://stackblitz.com/edit/github-qa2kdy?file=src%2Fpages%2Findex.tsx

### To reproduce

Change the text in the variable `text` in line 10 of the `index.tsx`, and you'll see the error appear. To me the error only appeared after changing the text.

### Additional information

Here's the typescript error I get on the callback:

```
Argument of type '(t: UseQueriesProcedureRecord<{ ctx: object; meta: object; errorShape: DefaultErrorShape; transformer: false; }, { greeting: QueryProcedure<{ input: { name?: string | null | undefined; }; output: { text: string; }; }>; }>) => [] | [...]' is not assignable to parameter of type '(t: UseQueriesProcedureRecord<{ ctx: object; meta: object; errorShape: DefaultErrorShape; transformer: false; }, { greeting: QueryProcedure<{ input: { name?: string | null | undefined; }; output: { text: string; }; }>; }>) => readonly []'.
  Type '[] | [TrpcQueryOptionsForUseQueries<{ text: string; }, { text: string; }, TRPCClientError<{ ctx: object; meta: object; errorShape: DefaultErrorShape; transformer: false; }>>]' is not assignable to type 'readonly []'.
    Type '[TrpcQueryOptionsForUseQueries<{ text: string; }, { text: string; }, TRPCClientError<{ ctx: object; meta: object; errorShape: DefaultErrorShape; transformer: false; }>>]' is not assignable to type 'readonly []'.
      Source has 1 element(s) but target allows only 0.(2345)
```

### 👨‍👧‍👦 Contributing

okxint:
When a `useQueries` callback conditionally returns an empty array, TypeScript errors:

```ts
trpc.useQueries((t) =>
  condition ? [] : [t.greeting.query(input)]
  //          ^^ Type '[]' is not assignable to type 'readonly []'
)
```

**Root cause:** The callback return type is `readonly [...QueriesOptions<TQueryOptions>]`. When TypeScript tries to unify both branches it infers `TQueryOptions = []`, making the tuple type `readonly []` — which then rejects the non-empty branch. The type has no room to express "zero or more queries."

**Fix:** Union in `| readonly []` on both `TRPCUseQueries` (line 192) and `TRPCUseSuspenseQueries` (line 214). This is a type-only change with zero runtime impact and matches the idiomatic TypeScript pattern for "this tuple also accepts an empty case."

Fixes #6188

### DIFF
--- a/packages/react-query/src/internals/useQueries.ts
+++ b/packages/react-query/src/internals/useQueries.ts
@@ -189,7 +189,7 @@ export type TRPCUseQueries<TRouter extends AnyRouter> = <
       TRouter['_def']['_config']['$types'],
       TRouter['_def']['record']
     >,
-  ) => readonly [...QueriesOptions<TQueryOptions>],
+  ) => readonly [...QueriesOptions<TQueryOptions>] | readonly [],
   options?: {
     combine?: (results: QueriesResults<TQueryOptions>) => TCombinedResult;
   },
@@ -211,5 +211,5 @@ export type TRPCUseSuspenseQueries<TRouter extends AnyRouter> = <
       TRouter['_def']['_config']['$types'],
       TRouter['_def']['record']
     >,
-  ) => readonly [...SuspenseQueriesOptions<TQueryOptions>],
+  ) => readonly [...SuspenseQueriesOptions<TQueryOptions>] | readonly [],
 ) => SuspenseQueriesResults<TQueryOptions>;


okxint:
Friendly ping — CI is green. Happy to rebase or address any feedback.

PR #7453, opened 2026-07-31
fix: correct inverted condition in assertIsRequestId

okxint:
The boolean condition in `assertIsRequestId` is inverted. The current guard is:

```ts
obj !== null && typeof obj === 'number' && isNaN(obj) && typeof obj !== 'string'
```

This only throws for `NaN` — all other non-null, non-string, non-number values
(`true`, `false`, `{}`, `[]`, `undefined`) satisfy `obj !== null` but fail the
`typeof obj === 'number'` check, so the whole expression is `false` and no
error is thrown. Those values then pass `asserts obj is number | string | null`
without validation.

The original intent (from PR #508) was `typeof obj !== 'number' || isNaN(obj)`,
which was later refactored and the condition was accidentally inverted.

After this fix:
- `null` → valid (passes)
- `"abc"` → valid (passes)  
- `42` → valid (passes)
- `NaN` → throws (was already handled correctly)
- `true`, `{}`, `[]`, `undefined` → all throw (were silently passing before)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for numeric request IDs by rejecting invalid `NaN` values.
  * Continued support for string and `null` request IDs.

* **Developer Experience**
  * React query utilities now support callbacks that return an empty query tuple when no queries are needed.

### DIFF
--- a/packages/react-query/src/internals/useQueries.ts
+++ b/packages/react-query/src/internals/useQueries.ts
@@ -189,7 +189,7 @@ export type TRPCUseQueries<TRouter extends AnyRouter> = <
       TRouter['_def']['_config']['$types'],
       TRouter['_def']['record']
     >,
-  ) => readonly [...QueriesOptions<TQueryOptions>],
+  ) => readonly [...QueriesOptions<TQueryOptions>] | readonly [],
   options?: {
     combine?: (results: QueriesResults<TQueryOptions>) => TCombinedResult;
   },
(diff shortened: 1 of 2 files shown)


sirrryasir (approved):
LGTM. The logic for `assertIsRequestId` correctly identifies invalid types (e.g. booleans, objects) and `NaN` now. The type definition update for `useQueries` also correctly handles empty arrays. Good catch.

okxint:
Rebased onto latest main to pick up recent fixes and trigger a fresh CI run. The Deno E2E failure in the previous run appeared to be environment-related rather than caused by this change.

okxint:
Friendly ping — CI is green. Happy to rebase or make any changes needed.

okxint:
Pinging the team — this has been sitting with an approve from @sirrryasir for a while. CI is green. Any blockers from the maintainer side? Happy to rebase or adjust anything.

PR #7461, opened 2026-08-09
feat(client): add `iterate()` method to subscriptions

### LINKED ISSUE #6868: feat: client subscriber gets an AsyncIterable

On the vanilla client, being able to an `AsyncIterable<T>` from a subscription in order to make it compliant with the [`streamedQuery` from tansatck query](https://tanstack.com/query/latest/docs/reference/streamedQuery).

Currently `@trpc/tanstack-react-query` support for `useSubscriptions` but implementing the `AsyncIterable` on the vanilla client would provide the feature to vue user as well.

Implement the `client.myProcedure.subscribe().iterator(): Promise<AsyncIterable<T>>`

-

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

tarikermis:
Closes #6868

## What changed

Subscription procedures on the vanilla client can now be consumed as an `AsyncIterable`:

```ts
for await (const data of client.onPostAdd.iterate()) {
  // ...
}
```

`iterate(input, opts?)` uses the existing `observableToAsyncIterable` helper. It filters connection-state envelopes, yields subscription data with the same output typing as `onData`, and propagates link errors unchanged. The iterable is cold, so each iterator owns one subscription and tears it down on `break`, `return`, or abort.

The shared stream helper now also closes its lifecycle gaps: late emissions after cancellation are ignored, signal abort settles the stream before teardown, listeners are detached after settlement and never attached after synchronous settlement, and an already-aborted signal does not start a request.

The vanilla client docs show both callback subscriptions and `iterate()`.

## Verification

- Added 6 HTTP subscription tests, 3 WebSocket tests, and 6 observable lifecycle tests.
- The required GitHub Actions matrix is green on the final code, including build, typecheck, lint, unit/integration tests, and the E2E fanout.
- Local Node 24 checks: root lint and package typechecks pass; the observable suite passes 14/14; the three added WebSocket `iterate()` cases pass; Prettier and `git diff --check` are clean.

## Notes

- Slow consumers still buffer values; this is existing stream-helper behavior.
- Errors are intentionally not wrapped again because links already return `TRPCClientError`, matching `subscribe`.
- No changeset was added because this repository uses lerna and has no `.changeset` setup.

## Checklist

Best Regards, Tarik

### DIFF
--- a/packages/client/src/createTRPCClient.ts
+++ b/packages/client/src/createTRPCClient.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-non-null-assertion */
 import type { Unsubscribable } from '@trpc/server/observable';
 import type {
   AnyProcedure,
@@ -161,7 +160,10 @@ export function createTRPCClientProxy<TRouter extends AnyRouter>(
 ): TRPCClient<TRouter> {
   const proxy = createRecursiveProxy<TRPCClient<TRouter>>(({ path, args }) => {
     const pathCopy = [...path];
-    const clientCallType = pathCopy.pop()!;
+    const clientCallType = pathCopy.pop();
+    if (!clientCallType) {
+      throw new Error('Missing client call type');
+    }
 
     const fullPath = pathCopy.join('.');
 
--- a/packages/client/src/createTRPCClient.ts
+++ b/packages/client/src/createTRPCClient.ts
@@ -3,6 +3,7 @@ import type { Unsubscribable } from '@trpc/server/observable';
 import type {
   AnyProcedure,
   AnyRouter,
+  inferAsyncIterableYield,
   inferClientTypes,
   inferProcedureInput,
   InferrableClientTypes,
@@ -75,6 +76,12 @@ export type SubscriptionResolver<TDef extends TRPCResolverDef> = (
     TRPCProcedureOptions,
 ) => Unsubscribable;
 
+/** @internal */
+export type SubscriptionIterableResolver<TDef extends TRPCResolverDef> = (
+  input: TDef['input'],
+  opts?: TRPCProcedureOptions,
+) => AsyncIterable<inferAsyncIterableYield<TDef['output']>>;
+
 type DecorateProcedure<
   TType extends ProcedureType,
   TDef extends TRPCResolverDef,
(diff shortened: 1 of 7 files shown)


daltino (commented):
In `subscriptionAsIterable`, you're creating a `new AbortController().signal` as a fallback when no signal is provided, but this signal is immediately orphaned since the `AbortController` instance itself is discarded. The comment says "teardown is driven by the consumer stopping iteration" but that assumes `observableToAsyncIterable` handles cleanup purely via the iterator's `return()` method rather than signal cancellation. Worth double checking that early exits (like `break`) properly unsubscribe without the signal being aborted, otherwise you could have a subscription that lingers until GC collects the detached controller. Also in `observableToReadableStream`, the new `onSettled` function is defined but I'm not seeing it wired up to the `cancel` handler of the `ReadableStream` in the diff. If it's only called in error/complete paths within the `subscribe` callbacks, then an upstream `cancel()` call (consumer abandoning the stream without going through the abort signal) might leave the event listener attached longer than necessary.

tarikermis:
Good questions - I double-checked both.

On early break: teardown doesn't rely on the signal at all. break calls the iterator's return(), which cancels the underlying ReadableStream and unsubscribes the subscription, so the fallback controller being orphaned is harmless. Added a test proving unsubscribe happens with a never-aborted signal.

On cancel() vs onSettled: the stream's cancel() handler delegates to onAbort(), whose first line already detaches the abort listener (and unsubscribes), so the listener is never left attached. There's now a test spying on removeEventListener to prove it, plus a comment in cancel() clarifying why onSettled isn't needed there.

Best Regards, Tarik

tarikermis:
Followed up on the two valid remaining review findings in `731bde1d`:

- The abort listener is now attached only if the observable is still active after `subscribe()` returns. The regression test settles synchronously, verifies that no listener was attached, then aborts the signal.
- The client proxy now guards an empty path instead of using a non-null assertion, so the file-wide lint suppression is gone.

I left the HTTP test setup unchanged. Its existing `konn` `beforeEach`/`afterEach` lifecycle already creates and closes a fresh `testServerAndClientResource` for every test.

Validation: root lint and package typechecks pass; the observable and HTTP subscription suites pass 37/37; the three added WebSocket `iterate()` cases pass; Prettier and `git diff --check` are clean.

Best Regards, Tarik

tarikermis:
Good catch. Fixed in `3a0917a6`.

Signal abort now marks the stream inactive and closes it before unsubscribing, so synchronous `next` or `error` calls from teardown are ignored. `ReadableStream.cancel()` keeps its existing behavior because it already marks the stream inactive before delegating to the same cleanup path.

The new regression test failed with the teardown error before the change and now passes with only the pre-abort value collected. The 14-test observable suite, server typecheck/lint, Prettier, and `git diff --check` are green on Node 24.15.0.

Best Regards, Tarik

PR #7464, opened 2026-08-10
fix(tanstack-react-query): subscriptionOptions(skipToken).enabled incorrectly returns true

### LINKED ISSUE #7373: bug(tanstack-react-query): subscriptionOptions remains enabled when a skip token is provided.

```
System:
  OS: macOS 26.4.1
  CPU: (12) arm64 Apple M3 Pro
  Memory: 129.14 MB / 36.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 24.4.1 - /Users/tmk/.local/share/mise/installs/node/24.4.1/bin/node
  npm: 11.4.2 - /Users/tmk/.local/share/mise/installs/node/24.4.1/bin/npm
  pnpm: 10.33.2 - /Users/tmk/.local/share/mise/installs/node/24.4.1/bin/pnpm
  bun: 1.3.13 - /opt/homebrew/bin/bun
  Deno: 2.7.14 - /opt/homebrew/bin/deno
Browsers:
  Chrome: 147.0.7727.138
  Safari: 26.4
npmPackages:
  @tanstack/react-query: ^5.100.9 => 5.100.9
  @trpc/client: ^11.17.0 => 11.17.0
  @trpc/server: ^11.17.0 => 11.17.0
  typescript: ^6.0.3 => 6.0.3
```

### Describe the bug

`subscriptionOptions(skipToken).enabled` is true, expect false.

### Link to reproduction

none

### To reproduce

```ts
import { QueryClient, skipToken } from '@tanstack/query-core';
import { createTRPCClient } from '@trpc/client';
import { createTRPCOptionsProxy } from '@trpc/tanstack-react-query';

const t = initTRPC.create({
  isServer: true,
});

const appRouter = t.router({
  foo: t.router({
    bar: t.procedure.subscription(async function* () {
      yield 'hello';
    }),
  }),
});

type AppRouter = typeof appRouter;

const trpcClient = createTRPCClient<AppRouter>({
  links: [],
});

export const trpc = createTRPCOptionsProxy<AppRouter>({
  client: trpcClient,
  queryClient: new QueryClient(),
});

console.log(trpc.foo.bar.subscriptionOptions()); // ✅
console.log(trpc.foo.bar.subscriptionOptions(skipToken).enabled); // ❌, expect false
```

### Additional information

_No response_

### 👨‍👧‍👦 Contributing

okxint:
## Root cause

`getQueryKeyInternal` intentionally strips `skipToken` from the query key args (line 168 of `utils.ts`):

```ts
...(typeof input !== 'undefined' &&
  input !== skipToken && { input: input }),
```

This means when `trpcSubscriptionOptions` tries to recover the original input from the queryKey:

```ts
const input = readQueryKey(queryKey)?.args?.input;
const enabled = 'enabled' in opts ? !!opts.enabled : input !== skipToken;
```

…it gets `undefined` back (not `skipToken`). So `undefined !== skipToken` evaluates to `true`, making `enabled: true` even though `skipToken` was the caller's intent.

## Fix

Thread the raw `input` argument into `trpcSubscriptionOptions` alongside the `queryKey`. This lets the `enabled` check compare against the original value before queryKey serialization strips it. Also updated the internal `_subscribe` call to pass `undefined` rather than `skipToken` when the input is a skip token.

## Before

```ts
trpc.foo.bar.subscriptionOptions(skipToken).enabled // true ❌
```

## After

```ts
trpc.foo.bar.subscriptionOptions(skipToken).enabled // false ✅
```

Fixes #7373

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved subscription handling to use the provided input consistently.
  * Preserved correct behavior when subscriptions are skipped, including proper handling of `skipToken`.
  * Ensured explicitly disabled subscriptions remain disabled.

### DIFF
--- a/packages/tanstack-react-query/src/internals/createOptionsProxy.ts
+++ b/packages/tanstack-react-query/src/internals/createOptionsProxy.ts
@@ -485,6 +485,7 @@ export function createTRPCOptionsProxy<
         return trpcSubscriptionOptions({
           opts: arg2,
           path,
+          input: arg1,
           queryKey: getQueryKeyInternal({
             path,
             input: arg1,
--- a/packages/tanstack-react-query/src/internals/subscriptionOptions.ts
+++ b/packages/tanstack-react-query/src/internals/subscriptionOptions.ts
@@ -12,7 +12,7 @@ import type {
   TRPCQueryKey,
   TRPCQueryOptionsResult,
 } from './types';
-import { createTRPCOptionsResult, readQueryKey } from './utils';
+import { createTRPCOptionsResult } from './utils';
 
 interface BaseTRPCSubscriptionOptionsIn<TOutput, TError> {
   enabled?: boolean;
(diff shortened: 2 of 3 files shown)


daltino (commented):
The fix is correct and the root cause analysis is spot on `getQueryKeyInternal` intentionally omitting `skipToken` from the serialized key meant recovering it via `readQueryKey` was always going to return `undefined`, making the `enabled` check unreliable. Threading `input` directly through before serialization is the right approach, and it's a minimal, targeted change that doesn't affect the query key structure itself.

One thing worth confirming: is there a test case being added that covers the `subscriptionOptions(skipToken).enabled === false` scenario? Given this is a regression fix, having a dedicated test would prevent this from silently breaking again if `getQueryKeyInternal`'s behavior ever changes.

okxint:
Added the regression test in the latest commit. Three assertions in `test/skipToken.test.tsx`:

1. `subscriptionOptions(skipToken).enabled === false` — the direct regression for #7373
2. `subscriptionOptions(normalInput).enabled === true` — control case confirming normal input stays enabled  
3. `explicit enabled: false override is still respected` — verifying the user can opt out independently of `skipToken`

A stub subscription procedure (`events`) was added to the test router so the type system is satisfied. The handler is never called — the test only inspects the returned options object.

okxint:
Pinging — fix and regression tests have been in since Aug 17. CI is green, both reviewers said the root cause analysis is correct. Happy to rebase or squash if that helps get this across the line.

Nick-Lucas (approved):

PR #7467, opened 2026-08-11
feat(react-query): add mutation onError override

### LINKED ISSUE #4924: feat: wanna default handle callback for  `useMutation` in `react-query`

I wanna set a default global callback in trpc.

For example, i can has a default error alert tip if we throw error in request, or we have take mutation show operate success alert tip.

Now, i have to set those in every where.

I take a look of source code, its weird:

- why only get default option with `getMutationDefaults` rather than `getDefaultOptions`?
- why not append default option into `__useMutation` options?
- in `mutationSuccessOverride`, why the `overrides` handler is higher priority than options callback in caller where define as `originalFn`?

https://github.com/trpc/trpc/blob/42e6910cfb9f25ab12f522a3ede01a0078b5bc6d/packages/react-query/src/shared/hooks/createHooksInternal.tsx#L338-L370

Set in `trpc.createTRPCReact` 
```ts
export const trpc = createTRPCReact<AppRouter>({
  overrides: {
    useMutation: {
      onSuccess: ()=> console.log('success'),
    },
  },
});
``` 

or in `new QueryClient` 

```ts
export const queryClient = new QueryClient({
  defaultOptions: {
    mutations: {
      onSuccess: () => console.log('success'),
      onError: () => console.log('error'),
    },
  },
});
```

none

### Additional information

I want to contribute code to implement this feature, but i cant understand why there logic is it.

I'm not sure if this is intentional, can you give me some context?

### 👨‍👧‍👦 Contributing

tonycoder-hub:
Fixes #4924

## Summary
Adds tRPC-level mutation onError override support. The override receives originalFn, queryClient, and meta. The selected original callback (hook-level onError if provided, otherwise the QueryClient default mutation onError) runs before the tRPC override onError. A hook-level callback takes precedence over the QueryClient default callback.

## Root cause
The public override type and createRootHooks wiring only supported onSuccess. A callback-return-value fallback would also be incorrect because normal onError callbacks return void, so callback selection must happen before invocation.

## Tests
- `npx vitest run packages/react-query/test/overrides.test.tsx --project @trpc/react-query --maxWorkers=1 --no-file-parallelism`
- `npm exec --offline -- tsc --noEmit --preserveWatchOutput --pretty`
- `npm exec --offline -- eslint --no-cache src`
- `npm exec --offline -- prettier --check packages/react-query/src/shared/types.ts packages/react-query/src/shared/hooks/createHooksInternal.tsx packages/react-query/test/overrides.test.tsx`

AI assistance: this change was produced with AI assistance and submitted by an automated contribution workflow. The workflow replayed the focused RED test against the base revision, reran the listed GREEN checks offline, and applied the recorded review gates before submission.

## Summary by CodeRabbit

* **New Features**
  * Added configurable error handling for mutations.
  * Mutation error overrides can now access the original error callback, query client, and mutation metadata.
  * Preserved existing local and query-client default error-handling behavior.

* **Bug Fixes**
  * Ensured local mutation error handlers take precedence while still allowing fallback handlers to run when no local handler exists.

### DIFF
--- a/packages/react-query/src/shared/hooks/createHooksInternal.tsx
+++ b/packages/react-query/src/shared/hooks/createHooksInternal.tsx
@@ -88,6 +88,10 @@ export function createRootHooks<
     config?.overrides?.useMutation?.onSuccess ??
     ((options) => options.originalFn());
 
+  const mutationErrorOverride: UseMutationOverride['onError'] =
+    config?.overrides?.useMutation?.onError ??
+    ((options) => options.originalFn());
+
   type TError = TRPCClientErrorLike<TRouter>;
 
   type ProviderContext = TRPCContextState<TRouter, TSSRContext>;
@@ -341,6 +345,18 @@ export function createRootHooks<
             opts?.onSuccess?.(...args) ?? defaultOpts?.onSuccess?.(...args);
 
           return mutationSuccessOverride({
+            originalFn,
+            queryClient,
+            meta: opts?.meta ?? defaultOpts?.meta ?? {},
+          });
+        },
+        onError(...args) {
+          const originalFn = () => {
+            const fn = opts?.onError ?? defaultOpts?.onError;
+            return fn?.(...args);
+          };
+
+          return mutationErrorOverride({
             originalFn,
             queryClient,
             meta: opts?.meta ?? defaultOpts?.meta ?? {},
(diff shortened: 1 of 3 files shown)


tonycoder-hub:
Closing as stale — opened on or before 2026-08-17 and still unmerged.

PR #7471, opened 2026-08-19
fix(client): abort in-flight fetch on unsubscribe in httpLink and httpBatchLink

### LINKED ISSUE #7468: httpLink / httpBatchLink don't create their own AbortController, so unsubscribe doesn't cancel the request (same pattern as #7389)
#7390 (closing #7389) fixed `httpBatchStreamLink`: it created its own `AbortController` and passed it into `jsonlStreamConsumer`, but the observable's cleanup was a `// noop`, so `abortController.abort()` never ran on unsubscribe and the JSONL stream kept consuming from the server after the subscriber disconnected.

I checked the other HTTP links and this same shape exists in two more:

- `httpLink.ts`: passes `signal: op.signal` straight into `universalRequester`, no internal controller, cleanup is `// noop`.
- `httpBatchLink.ts`: passes `allAbortSignals(...batchOps.map(op => op.signal))`, still only combining caller-provided per-op signals, no internal controller, cleanup is also `// noop`.

Compare with the links that already do this correctly: `httpSubscriptionLink.ts` and `localLink.ts` both do `raceAbortSignals(op.signal, ac.signal)` with their own `ac = new AbortController()`, and call `ac.abort()` in their cleanup. `httpBatchStreamLink.ts` now does the same after #7390.

So `httpLink` and `httpBatchLink` are the two remaining HTTP links that rely entirely on the caller supplying an `op.signal` that happens to fire on unsubscribe, instead of owning their own controller tied to the observable's lifecycle the way every other link in the package does. Whenever the caller's `op.signal` isn't already wired to this specific subscription's teardown, the underlying fetch outlives the subscriber — same wasted-bandwidth/dangling-request consequence as #7389, just without a streamed body to make it as visible in the network tab.

okxint:
## Problem

`httpLink` and `httpBatchLink` both had `// noop` teardown callbacks, so calling `unsubscribe()` on a chain using either link never aborted the underlying `fetch()` call. The in-flight HTTP request kept running until the server finished responding, wasting bandwidth and potentially triggering unexpected server-side side effects (e.g. mutations that complete after the caller no longer cares).

This is the same bug that was fixed for `httpBatchStreamLink` in #7390. It applies to the two remaining non-streaming HTTP links.

Closes #7468

## Root cause

```ts
// httpLink — before
const request = universalRequester({ signal: op.signal, ... })
// ...
return () => {
  // noop  ← unsubscribe does nothing
}

// httpBatchLink — before  
const promise = loader.load(op)  // op.signal passed through allAbortSignals, no internal AC
// ...
return () => {
  // noop  ← unsubscribe does nothing
}
```

## Fix

Mirrors #7390 (the `httpBatchStreamLink` fix) exactly:

```ts
const ac = new AbortController()
const request = universalRequester({
  signal: raceAbortSignals(op.signal, ac.signal),  // race both
  ...
})
let isDone = false
// ...then/catch: isDone = true
return () => {
  if (!isDone) ac.abort()  // abort on unsubscribe
}
```

Both `httpLink` and `httpBatchLink` get the same treatment.

## Tests

Added two regression tests to `packages/tests/server/links.test.ts` (parallel to the existing `httpBatchStreamLink - unsubscribe aborts fetch` test):

- **`httpLink - unsubscribe aborts fetch`** — mocks `fetch` to never resolve, subscribes, awaits the fetch call, unsubscribes, asserts `signal.aborted === true`
- **`httpBatchLink - unsubscribe aborts fetch`** — same shape for the batch link

All 25 tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Unsubscribing from regular or batched HTTP requests now cancels in-flight fetches, reducing unnecessary network activity.
  * Subscription options now correctly respect raw input values and skipped subscriptions.
  * Explicitly disabled subscriptions continue to remain disabled.

* **Tests**
  * Added coverage for request cancellation and subscription option behavior, including skipped and disabled subscriptions.

### DIFF
--- a/.changeset/fix-http-link-abort-unsubscribe.md
+++ b/.changeset/fix-http-link-abort-unsubscribe.md
@@ -1,0 +1,7 @@
+---
+'@trpc/client': patch
+---
+
+fix(client): abort in-flight fetch when httpLink or httpBatchLink unsubscribes
+
+Both links passed `op.signal` directly into the fetcher with a `// noop` teardown, so calling `unsubscribe()` never aborted the underlying HTTP request. Each link now creates its own `AbortController`, races its signal with the caller-provided `op.signal` via `raceAbortSignals`, and calls `ac.abort()` in the observable cleanup — mirroring the fix already applied to `httpBatchStreamLink` in #7390.
--- a/packages/client/src/links/httpBatchLink.ts
+++ b/packages/client/src/links/httpBatchLink.ts
@@ -3,7 +3,7 @@ import { observable } from '@trpc/server/observable';
 import { transformResult } from '@trpc/server/unstable-core-do-not-import';
 import type { BatchLoader } from '../internals/dataLoader';
 import { dataLoader } from '../internals/dataLoader';
-import { allAbortSignals } from '../internals/signals';
+import { allAbortSignals, raceAbortSignals } from '../internals/signals';
 import type { NonEmptyArray } from '../internals/types';
 import { TRPCClientError } from '../TRPCClientError';
 import type { HTTPBatchLinkOptions } from './HTTPBatchLinkOptions';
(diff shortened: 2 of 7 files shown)


okxint:
Quick note on discoverability: this is the last two links in the package with the noop teardown — #7390 fixed `httpBatchStreamLink`, this closes the same gap for `httpLink` and `httpBatchLink`. The fix and tests follow the exact same shape so review should be fast.

Nick-Lucas:
this one seems to do too much, duplicates changes I merged in other PRs, feel free to re-open individual fixes

PR #7480, opened 2026-08-22
fix(client,server): wire AbortSignal in wsLink and fix stale clientSubscriptions on task rejection

### LINKED ISSUE #7400: ws: subscription task rejection can leave stale clientSubscriptions entry

### LINKED ISSUE #7431: WebSocket link silently ignores the operation AbortSignal (removeEventListener with no matching addEventListener)

### LINKED ISSUE #7439: wsLink: AbortSignal on operations is ignored — abort does not cancel in-flight requests or subscriptions

okxint:
## Summary

Two related bugs in the WebSocket adapter, both in the same area.

---

### 1. wsLink ignores AbortSignal on operations — `wsClient.ts`

**Root cause:** The teardown path called `signal?.removeEventListener('abort', abort)` but no matching `addEventListener` was ever called. The abort listener was never registered — only the cleanup half was scaffolded (v11 refactor, PR #6223).

Aborting the caller's `AbortSignal` had no effect on any in-flight WS query, mutation, or subscription.

**Fix:** Extract teardown into a named function. Check `signal.aborted` on entry and tear down immediately if already aborted. Otherwise register a `{ once: true }` abort listener. The existing `removeEventListener` now correctly deregisters it on normal completion.

```ts
// Before — removeEventListener with no matching addEventListener
return () => {
  abort();
  if (type === 'subscription' && ...) send({ id, method: 'subscription.stop' });
  signal?.removeEventListener('abort', abort); // no-op: listener was never added
};

// After
const teardown = () => { abort(); if (type === 'subscription' ...) send(...); };

if (signal?.aborted) {
  teardown(); // already aborted — tear down immediately
} else {
  signal?.addEventListener('abort', teardown, { once: true }); // wire it
}

return () => {
  signal?.removeEventListener('abort', teardown); // now actually removes it
  teardown();
};
```

Fixes #7431, fixes #7439

---

### 2. Stale `clientSubscriptions` entry on subscription task rejection — `ws.ts`

**Root cause:** `clientSubscriptions.delete(id)` only ran on the happy path (~line 407). The `.catch()` handler (~line 408) aborted the controller and sent an error response but never deleted the entry. If the subscription's async function rejected (e.g. `iterator.return()` throwing during `await using` disposal), the entry remained in the Map for the connection's lifetime — the same `id` could not be reused.

**Fix:** Add `clientSubscriptions.delete(id)` as the first line of `.catch()`, mirroring the happy-path cleanup.

Fixes #7400

---

## Tests

Added 2 regression tests in `websockets.test.ts`:
- Pre-aborted signal prevents data from arriving on subscribe
- Aborting mid-subscription stops further data delivery

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved WebSocket subscription cancellation, including requests that are already aborted.
  - Prevented further subscription messages after cancellation.
  - Removed stale subscription records when subscription processing fails.
  - Fixed request handling so raw request streams are read when no pre-parsed body is available, including multipart requests.
  - Improved compatibility with Express 5 and body-parser integrations.

- **Tests**
  - Added coverage for aborted and cancelled WebSocket subscriptions.
  - Added tests for raw request stream and multipart body handling.


okxint:
Friendly ping on this one. Happy to rebase or adjust anything if there's feedback.

PR #7564, opened 2026-09-05
fix(client): own AbortController in httpLink and httpBatchLink (#7468)

### LINKED ISSUE #7468: httpLink / httpBatchLink don't create their own AbortController, so unsubscribe doesn't cancel the request (same pattern as #7389)

Larslllllll:
## Bug

`httpLink` and `httpBatchLink` both pass `op.signal` directly to the underlying request without creating an internal `AbortController`, and their observable cleanup functions are `// noop`.

This means when a subscriber unsubscribes, the in-flight fetch request is not cancelled — the same class of bug that was fixed in `httpBatchStreamLink` by PR #7390.

Every other link in the package (`httpSubscriptionLink`, `httpBatchStreamLink`, `localLink`) already owns its own `AbortController` and calls `ac.abort()` in the observable cleanup.

## Fix

**httpLink** — create an `AbortController`, pass `ac.signal` to the requester, call `ac.abort()` in cleanup.

**httpBatchLink** — create an `AbortController`, combine it with `op.signal` via `raceAbortSignals` so the batch request is aborted when the individual observable unsubscribes, call `ac.abort()` in cleanup.

Pattern matches the existing correct implementations in `httpSubscriptionLink` and `httpBatchStreamLink`.

Closes #7468

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cancellation of in-flight HTTP requests when an operation is unsubscribed.
  - Added reliable cancellation for batched HTTP loads during teardown.
  - Ensured request cancellation signals correctly combine operation and subscription lifecycle events.

### DIFF
--- a/packages/client/src/links/httpBatchLink.ts
+++ b/packages/client/src/links/httpBatchLink.ts
@@ -3,7 +3,7 @@ import { observable } from '@trpc/server/observable';
 import { transformResult } from '@trpc/server/unstable-core-do-not-import';
 import type { BatchLoader } from '../internals/dataLoader';
 import { dataLoader } from '../internals/dataLoader';
-import { allAbortSignals } from '../internals/signals';
+import { allAbortSignals, raceAbortSignals } from '../internals/signals';
 import type { NonEmptyArray } from '../internals/types';
 import { TRPCClientError } from '../TRPCClientError';
 import type { HTTPBatchLinkOptions } from './HTTPBatchLinkOptions';
@@ -98,8 +98,12 @@ export function httpBatchLink<TRouter extends AnyRouter>(
             'Subscriptions are unsupported by `httpLink` - use `httpSubscriptionLink` or `wsLink`',
           );
         }
+        const ac = new AbortController();
         const loader = loaders[op.type];
-        const promise = loader.load(op);
+        const promise = loader.load({
+          ...op,
+          signal: raceAbortSignals(op.signal, ac.signal),
+        });
 
         let _res = undefined as HTTPResult | undefined;
         promise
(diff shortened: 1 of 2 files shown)


packages/client/src/links/httpBatchLink.ts
```diff
@@ -133,7 +137,7 @@ export function httpBatchLink<TRouter extends AnyRouter>(
           });
 
         return () => {
-          // noop
+          ac.abort();
```
Nick-Lucas, line 140: this seems valid but really needs tests for both cases

PR #7571, opened 2026-09-12
build(deps): upgrade node to 24.21.0 and @types/node to v24

Nick-Lucas:
Layer 1 of a stacked dependency upgrade.

Pins the toolchain to the latest Node 24 LTS (Krypton) and moves the workspace from `@types/node` v22 to v24 so the types match the runtime we actually build and test against.

### Changes
- `.nvmrc` / `.tool-versions` now pin **24.21.0** exactly, rather than a floating `24` and a stale `24.13.1` that disagreed with each other.
- `.github/setup` reads `node-version-file: .nvmrc`, so CI and local dev can no longer drift apart.
- Introduces a **pnpm catalog** in `pnpm-workspace.yaml` and points `packages/*` and the repo root at `catalog:`. Later layers in this stack move their upgrades into the catalog the same way, so shared tooling versions stay pinned together.

  `examples/*` and `www` deliberately keep **literal versions**. Vercel builds some of these projects standalone, with a root directory that doesn't include `pnpm-workspace.yaml`, and a `catalog:` spec there fails the install outright:

  ```
  ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC
  × No catalog entry '@types/node' was found for catalog 'default'.
  ```

  The first push of this PR did catalog them, and the `og-image` and `examples-todomvc` Vercel deploys failed for exactly that reason.
- Drops one stale `@ts-expect-error` in `packages/client/src/internals/types.test.ts`: undici's `fetch` is assignable to `FetchEsque` under `@types/node` 24, so the directive now errors as unused.

**`e2e-legacy-node` moves from 18.x/20.x to 20.x/22.x.** Node 18 went EOL in April 2025, and the examples that job runs already can't use it: `next-prisma-starter` runs vitest via `test-start`, and even the currently-pinned vitest 4.0.18 declares `^20 || ^22 || >=24`. So the 18.x leg was already running vitest on an unsupported runtime.

The `main` ruleset's required-status-check contexts are updated to match (18.x contexts replaced with 22.x). Without that the stack can't merge — the three `e2e-legacy-node (…, 18.x)` checks would be permanently "expected" but never reported.

### Verification
`pnpm build`, `pnpm typecheck-packages`, `pnpm typecheck-www`, `pnpm lint` and `pnpm manypkg check` all pass. `pnpm test` is 1130/1131, with the only failure being `httpSubscriptionLink.memory.test.ts` — a WeakRef/GC test that also fails on `main` locally and is unrelated to this change.

## Summary by CodeRabbit

* **Chores**
  * Standardized development and automation environments on Node.js 24.21.0.
  * Centralized Node.js type definitions at version 24.13.4.
  * Updated examples, packages, and website tooling to use the newer type definitions.
  * Configured automation to follow the repository’s declared Node.js version.

* **Tests**
  * Removed an outdated TypeScript error suppression from a type-related test.
  * Updated legacy end-to-end coverage to test Node.js 20 and 22 instead of Node.js 18 and 20.


pnpm-workspace.yaml
```diff
@@ -9,3 +9,14 @@ packages:
...
   esbuild: true
+
+# Shared dependency versions - see https://pnpm.io/catalogs
+# Keep upgraded tooling pinned here so every workspace moves together.
+#
+# Only packages/* and the repo root use `catalog:`. examples/* and www are
+# deployed by external builders (Vercel) that install them standalone, without
+# this file in scope - a `catalog:` spec there fails the install outright with
+# ERR_PNPM_CATALOG_ENTRY_NOT_FOUND_FOR_SPEC. Those workspaces keep literal
+# versions.
```
Nick-Lucas:
```suggestion

```

Nick-Lucas: Done — all comments stripped from `pnpm-workspace.yaml`. It's now just the settings and the catalog entries.

PR #7572, opened 2026-09-12
build(deps): upgrade pnpm to v12

Nick-Lucas:
Layer 2 of a stacked dependency upgrade. Stacked on #7571.

Moves the workspace from pnpm **10.33.1 to 12.4.1**. That spans two majors, and v11's config changes mean this is not a one-line bump.

### `.npmrc` is deleted
Since v11, pnpm reads *only* auth/registry settings from `.npmrc`; everything else must live in `pnpm-workspace.yaml`. Our `.npmrc` had no auth settings at all, so it goes away and its contents move over in camelCase:

| `.npmrc` (before) | `pnpm-workspace.yaml` (after) |
| --- | --- |
| `public-hoist-pattern[]` ×5 | `publicHoistPattern` |
| `side-effects-cache=false` | `sideEffectsCache: false` |
| `strict-peer-dependencies=false` | `strictPeerDependencies: false` |
| `prefer-workspace-packages=true` | `preferWorkspacePackages: true` |
| `link-workspace-packages=true` | `linkWorkspacePackages: true` |
| `node-options=--no-warnings` | `nodeOptions: --no-warnings` |

### `allowBuilds` needs an explicit verdict per package
v11 defaults `strictDepBuilds` to `true`, so the familiar "Ignored build scripts: …" warning is now a hard `ERR_PNPM_IGNORED_BUILDS` that fails the install. Every package listed as `false` **was already being skipped under pnpm 10** — this records the existing (green) behaviour rather than changing it. `esbuild` keeps the `true` it already had.

If any of these actually should run their install scripts, this is a good place to say so — but that's a behaviour change and deliberately not part of this PR.

### Also
- `pnpm/action-setup` v5 → v6.
- The lockfile gains a `packageManagerDependencies` section — v12 pins pnpm itself from the `packageManager` field. Package resolution is otherwise untouched; v12 consumes the existing lockfile as-is.
- `packages/client/bin/intent.js` picks up its executable bit, which pnpm now sets for declared bins.

v11 also enables `minimumReleaseAge: 1440` (a package must be 24h old to resolve) and `blockExoticSubdeps` by default. Both are kept — they're worthwhile supply-chain defaults and the current lockfile passes them.

### Railway services move from Nixpacks to Railpack
Nixpacks provisions pnpm through corepack, which doesn't install pnpm 12 — so both `trpc-sse-and-websockets` preview deploys fail on this commit under the old builder. Nixpacks is deprecated anyway; Railpack reads the `packageManager` field directly.

```diff
  "build": {
-   "builder": "NIXPACKS",
+   "builder": "RAILPACK",
```

in both `examples/.railway/next-prisma-websockets-starter/railway.json` and `examples/.railway/next-sse-chat/railway.json`. This is the one part of the PR I can't verify locally — Railway has to run it — so the deploy checks on this PR are the test.

### Verification
Verified against a **clean** install (`node_modules` removed, `CI=true pnpm install`, i.e. the frozen-lockfile path CI takes). `pnpm build`, `typecheck-packages`, `typecheck-www`, `lint` and `manypkg check` all pass; `pnpm test` is 1130/1131 with only the pre-existing `httpSubscriptionLink.memory.test.ts` WeakRef/GC failure that also fails on `main`.

## Summary by CodeRabbit

* **Chores**
  * Updated the required package manager version to pnpm 12.
  * Updated automated setup and downstream synchronization to use pnpm setup action v6.
  * Updated workspace installation, package linking, dependency hoisting, and build settings.
  * Removed legacy package manager configuration in favor of workspace settings.
  * Updated Railway deployment settings for starter examples to use the Railpack builder.


pnpm-workspace.yaml
```diff
@@ -7,8 +7,42 @@ packages:
   - 'examples/minimal-content-types/*'
   - 'www'
   - 'www/og-image'
+# pnpm >=11 reads settings from here, not .npmrc (which is auth/registry only).
```
Nick-Lucas: strip all the comments

Nick-Lucas: Done — stripped. Also fixed the thing Copilot spotted in this PR: `.github/workflows/subtree.yml` calls `pnpm/action-setup` directly and was still on v5, so it's bumped to v6 alongside `.github/setup`.

.github/setup/action.yml
```diff
@@ -3,7 +3,7 @@ description: Composite action that sets up pnpm
 runs:
   using: 'composite'
   steps:
-    - uses: pnpm/action-setup@v5
+    - uses: pnpm/action-setup@v6
```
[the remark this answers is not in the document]
Nick-Lucas: Good catch — fixed. `.github/workflows/subtree.yml` now uses `pnpm/action-setup@v6` too, so no CI path is left calling v5.

package.json
```diff
@@ -4,9 +4,9 @@
   "type": "module",
   "engines": {
     "node": "^24.0.0",
-    "pnpm": "^10.33.1"
+    "pnpm": "^12.0.0"
   },
-  "packageManager": "pnpm@10.33.1",
+  "packageManager": "pnpm@12.4.1",
```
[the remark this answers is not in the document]
Nick-Lucas: Agreed, and flagged in the PR description. I confirmed the boundary: Railway passes on #7571 and fails from this commit onward, for both services. There's no `railway.json` / `nixpacks.toml` in the repo, so there's nothing I can change here to fix it — the projects are dashboard-configured and I can't read their build logs. Most likely the builder needs a Node 22+ runtime, since pnpm 11+ is pure ESM and requires it. Needs someone with Railway access to confirm before this merges.

Nick-Lucas:
Switched both Railway services to Railpack in this commit, per the corepack diagnosis — Nixpacks provisions pnpm via corepack, which isn't installing pnpm 12, and Nixpacks is deprecated in favour of Railpack anyway.

```diff
  "build": {
-   "builder": "NIXPACKS",
+   "builder": "RAILPACK",
```

in `examples/.railway/next-prisma-websockets-starter/railway.json` and `examples/.railway/next-sse-chat/railway.json`. `RAILPACK` is a valid value in `railway.schema.json` (still labelled EXPERIMENTAL there, though that description looks stale).

Two notes:
- I'd previously said there was no Railway config in the repo — that was wrong, my search was too shallow to reach `examples/.railway/*/railway.json`. Sorry for the misdirection.
- This is the one change in the stack I can't verify locally; Railway has to run it. The deploy checks on this PR and the ones above it are the test.

pnpm-workspace.yaml
```diff
@@ -7,8 +7,39 @@ packages:
...
+sideEffectsCache: false
+strictPeerDependencies: false
+preferWorkspacePackages: true
+linkWorkspacePackages: true
+nodeOptions: --no-warnings
+
+minimumReleaseAgeExclude:
+  - sury@11.0.0
+
 allowBuilds:
```
Nick-Lucas, line 25: these should all be true

Nick-Lucas: Done — all 17 are `true` now.

Worth calling out that this **is** a behaviour change rather than a bookkeeping one, so I've reworded the commit message to say so: prisma, sharp, `@swc/core`, workerd, playwright, the tree-sitter grammars and the rest now run their install scripts, where pnpm 10 was silently skipping them. I'd originally set them `false` to keep the upgrade behaviour-neutral.

Checked it holds up: a clean `CI=true pnpm install` from an empty `node_modules` succeeds in ~19s with no `ERR_PNPM_IGNORED_BUILDS`, and build / typecheck / lint / `manypkg check` / the test suite are all unchanged. The `cloudflare-workers` example (workerd) builds fine. `next-prisma-starter` fails locally only because it wants a Postgres on :5432, which CI provides as a service.

PR #7574, opened 2026-09-12
build(deps): upgrade tsdown to 0.23.0

Nick-Lucas:
Layer 3 of a stacked dependency upgrade. Stacked on #7572.

**This is a prerequisite for the TypeScript 7 upgrade later in the stack**, not a standalone nice-to-have.

tsdown 0.12.7 pins `rolldown-plugin-dts` 0.13.x, which drives declaration emit through TypeScript's **JavaScript compiler API**. TypeScript 7 is the native (Go) port and ships no such API — only `bin/tsc` and a few `unstable/*` exports. Building `@trpc/server` with tsdown 0.12.7 and TypeScript 7 dies with:

```
[plugin rolldown-plugin-dts:generate]
TypeError: Cannot read properties of undefined (reading 'useCaseSensitiveFileNames')
```

tsdown 0.23 declares `typescript: ^7.0.0` among its peers.

### `packages/server` needs an explicit `deps.neverBundle` list
Every adapter's public types reference its host framework — `fastify`, `express`, `next`, `ws`, `aws-lambda` — but those are **devDependencies** here rather than peers, so tsdown doesn't externalize them automatically. 0.12.7 happened to leave them external anyway; 0.23 tries to pull their declarations into ours, which:

- **fails outright** for the ones using CommonJS `export =` syntax (59 errors from `fastify` alone, then `next`), and
- would otherwise **inline third-party types into our published API**.

Listing them explicitly restores exactly what 0.12.7 emitted. `packages/openapi`'s pre-existing `external: ['typescript']` moves to `deps.neverBundle` too, since 0.23 deprecates `external` and would otherwise warn on every build.

### API surface was diffed, not assumed
I extracted the exported symbols of all **35 public entrypoints** from the built `.d.mts` files with the TypeScript API, before and after:

- **No symbol removed**, on any entrypoint.
- **No external import changed** — `fastify`, `next/navigation`, `react`, `@tanstack/react-query` etc. are still external in the emitted declarations, exactly as before.
- tsdown 0.23 does additionally *name* a handful of types that 0.12.7 inlined anonymously: `LambdaEvent`, `FastifyRequestHandlerOptions`, `StandaloneHandlerOptions`, `NextCacheLinkOptions`, `SSGFns` and similar now appear as exports. **This is additive**, but it does widen the declared surface — flagging it in case you'd rather they stayed anonymous.
- `packages/upgrade` now externalizes `jscodeshift` instead of inlining it, which is correct — it's a real `dependency` of that package.

Only internal chunk filenames changed otherwise (`httpLink.d-CjpPCq4q.d.mts` → `httpLink-eSAq8J9S.d.mts`); those aren't reachable through the `exports` map.

### About the lockfile diff
It's much larger than this dependency change implies (~7000 lines, net ~1150 *smaller*). This is the first install since #7572 that actually re-resolves, so pnpm 12's canonical cycle breaking re-keys peer variants across the whole graph in one go — the v12 release notes call this out explicitly. Real version changes are confined to tsdown's own tree (rolldown `1.0.0-beta` → `1.2.8`, `@oxc-project/types`, `@quansync/fs`) plus a `postcss` dedupe to 8.5.10 in the www tree.

### Verification
`pnpm build`, `typecheck-packages`, `typecheck-www`, `monotest-build` (which compiles against the built declaration files), `lint` and `manypkg check` all pass. `pnpm test` is 1130/1131, only the pre-existing `httpSubscriptionLink.memory.test.ts` WeakRef/GC failure.

## Summary by CodeRabbit

- **Build Improvements**
  - Updated workspace build tooling to tsdown 0.23.0.
  - Standardized the tsdown version across supported packages through the shared workspace catalog.
  - Improved CLI and server bundle compatibility by preserving required third-party dependencies as external.
  - Maintained consistent packaging behavior across supported integrations and deployment environments.

### DIFF
--- a/packages/server/tsdown.config.ts
+++ b/packages/server/tsdown.config.ts
@@ -21,6 +21,16 @@ export const input = [
 export default defineConfig({
   target: ['node18', 'es2017'],
   entry: input,
+  deps: {
+    neverBundle: [
+      /^aws-lambda$/,
+      /^express$/,
+      /^fastify$/,
+      /^@fastify\//,
+      /^next(\/|$)/,
+      /^ws$/,
+    ],
+  },
   dts: {
     sourcemap: true,
     tsconfig: './tsconfig.build.json',
--- a/examples/.test/internal-types-export/package.json
+++ b/examples/.test/internal-types-export/package.json
@@ -10,7 +10,7 @@
     "@tsconfig/node-lts": "^18.12.2",
     "@tsconfig/strictest": "^2.0.1",
     "@types/node": "^24.13.4",
-    "tsdown": "0.12.7",
+    "tsdown": "0.23.0",
     "typescript": "^5.9.2"
   },
   "dependencies": {
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -127,7 +127,7 @@
     "eslint": "^9.26.0",
     "isomorphic-fetch": "^3.0.0",
     "node-fetch": "^3.3.0",
-    "tsdown": "0.12.7",
+    "tsdown": "catalog:",
     "tslib": "^2.8.1",
     "typescript": "^5.9.2",
     "undici": "^8.0.0",
--- a/packages/next/package.json
+++ b/packages/next/package.json
@@ -139,7 +139,7 @@
     "next": "^15.3.8",
     "react": "^19.1.0",
     "react-dom": "^19.1.0",
-    "tsdown": "0.12.7",
+    "tsdown": "catalog:",
     "typescript": "^5.9.2",
     "zod": "^4.2.1"
   },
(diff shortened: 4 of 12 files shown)


packages/server/tsdown.config.ts
```diff
@@ -21,6 +21,21 @@ export const input = [
 export default defineConfig({
   target: ['node18', 'es2017'],
   entry: input,
+  // Every adapter's public types reference its host framework, but those are
+  // devDependencies here rather than peers, so tsdown won't externalize them
+  // on its own. Without this, tsdown >=0.23 tries to pull their declarations
+  // into our .d.ts - which both inlines third-party types into our published
+  // API and outright fails for the ones using CommonJS `export =` syntax.
```
Nick-Lucas:
```suggestion

```

Nick-Lucas: Done — comment removed, just the `deps.neverBundle` list now.

PR #7575, opened 2026-09-12
build(deps): upgrade typescript to v7 for packages/*

Nick-Lucas:
Layer 4 of a stacked dependency upgrade. Stacked on #7574.

TypeScript 7 is the **native (Go) port**. It ships `bin/tsc` and a handful of `unstable/*` exports — and **no JavaScript compiler API**. A single repo-wide bump is therefore not possible, so this splits the dependency in two via pnpm catalogs:

| catalog | version | used by |
| --- | --- | --- |
| `catalog:` | `^7.0.2` | `packages/*` — `tsc --noEmit` and tsdown declaration emit |
| `catalog:ts-api` | `^6.0.3` | anything that drives `tsc` programmatically |

On `ts-api`:
- **the repo root** — `typescript-eslint` caps itself at `typescript <6.1.0`
- **`@trpc/upgrade`** — jscodeshift's TS parser needs the same
- **`@trpc/openapi`** — see below

`www` and `examples/*` keep their existing `^5.9.2`: typedoc 0.27 pins `<=5.8.x`, and Next.js drives tsc through the API too.

### ⚠️ `@trpc/openapi`'s published peer range was wrong
`generateOpenAPIDocument` — exported from the package's main entry — does `import * as ts from 'typescript'` and uses `ts.TypeFlags` / `ts.Type`. It **cannot run on TypeScript 7 at all** (146 type errors here, and it would throw at runtime). Its peer range is corrected from `typescript: ">=5.7.2"` to `">=5.7.2 <7"`. Without this, a consumer upgrading to TypeScript 7 installs a package that breaks. This is the most important change in the PR and is worth a close look.

### What TypeScript 7 required
- **`tsconfig.json` gains `rootDir: "."`** — 7 enforces `rootDir` under `--noEmit` where 5.x did not, and each package pulls in sibling sources through `paths`. Without it every package reported dozens of `TS6059`.
- **`tsconfig.build.json` files drop `declarationDir: "dist"`** — tsdown's declaration emit now shells out to `tsgo` instead of driving the API in-process, so `declarationDir` redirected the output away from where `rolldown-plugin-dts` looks for it (`tsgo did not generate dts file for …`). tsdown sets the output location itself.
- **`Encoder.encode` now returns `Uint8Array<ArrayBuffer>`** instead of a bare `Uint8Array`. 7's DOM lib correctly rejects a view over a `SharedArrayBuffer` in `WebSocket.send()` — neither the browser nor `ws` accepts one. This narrows the experimental `experimental_encoder` API (#7100) to what was always actually sendable.
- **issue-4519 regression test** moves its `@ts-expect-error` onto the offending properties: 7 reports a failed overload against each bad property rather than against the call expression. The assertion is unchanged — the call still must not typecheck. *Two directives for one assertion is a bit awkward; happy to restructure if you'd prefer a different idiom.*

### Two real problems surfaced by putting the root on TypeScript 6
- **`monotest-tsconfig` drops `baseUrl`**, deprecated in 6 and removed in 7. It was redundant under `moduleResolution: bundler` with no `paths`.
- **`@trpc/react-query` now exports `DecorateRouterRecord` from its main entry.** The inferred type of `createTRPCReact()` referenced it only through an internal dist chunk that isn't in the `exports` map, so consumers on TypeScript ≥6 hit `TS2883: … cannot be named without a reference to '…/dist/getQueryKey-Cgg9jecL.mjs'. This is likely not portable.` `monotest-build` exists to catch exactly this, and it did.

### Verification
Emitted declarations were diffed **symbol by symbol across all 35 public entrypoints**: TypeScript 7's output is identical to 5.9's, and `packages/react-query/dist/index.d.mts` is byte-identical apart from the added export.

`pnpm build`, `typecheck-packages`, `typecheck-www`, `monotest-build`, `lint` and `manypkg check` all pass. `pnpm test` is 1130/1131 — only the pre-existing `httpSubscriptionLink.memory.test.ts` WeakRef/GC failure, which also fails on `main`.

Typechecking is also dramatically faster: `@trpc/server` goes from seconds to ~0.2s.

## Summary by CodeRabbit

- **New Features**
  - Exported an additional React Query type to support portable TypeScript declarations and newer TypeScript versions.
  - Updated WebSocket encoding contracts to require buffers backed by standard `ArrayBuffer` memory.

- **Bug Fixes**
  - Clarified validation for initial query data and transformed results.
  - Improved compatibility with current TypeScript compiler behavior.

- **Chores**
  - Centralized TypeScript version management across workspace packages.
  - Standardized build configuration and declaration output handling.
  - Updated documentation tooling to accommodate TypeScript deprecations.

### DIFF
--- a/packages/react-query/src/index.ts
+++ b/packages/react-query/src/index.ts
@@ -7,4 +7,5 @@ export {
   type CreateTRPCReactBase,
 } from './createTRPCReact';
 export type { inferReactQueryProcedureOptions } from './utils/inferReactQueryProcedure';
+export type { DecorateRouterRecord } from './shared';
 export { createTRPCQueryUtils } from './createTRPCQueryUtils';
--- a/packages/server/src/adapters/wsEncoder.ts
+++ b/packages/server/src/adapters/wsEncoder.ts
@@ -12,7 +12,7 @@
  */
 export interface Encoder {
   /** Encode data for transmission over the wire */
-  encode(data: unknown): string | Uint8Array;
+  encode(data: unknown): string | Uint8Array<ArrayBuffer>;
   /** Decode data received from the wire */
   decode(data: string | ArrayBuffer | Uint8Array): unknown;
 }
--- a/examples/.test/monotest-tsconfig/base.json
+++ b/examples/.test/monotest-tsconfig/base.json
@@ -41,7 +41,6 @@
     "preserveWatchOutput": true,
     "emitDeclarationOnly": false,
     "tsBuildInfoFile": "${configDir}/tsconfig.tsbuildinfo",
-    "baseUrl": "${configDir}",
     "outDir": "${configDir}/dist",
     "rootDir": "${configDir}",
     "composite": false,
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "ts-prune": "^0.10.3",
     "tsx": "^4.19.3",
     "turbo": "^2.5.4",
-    "typescript": "^5.9.2",
+    "typescript": "catalog:ts-api",
     "typescript-eslint": "^8.31.1",
     "vite": "^6.1.1",
     "vitest": "^4.0.18",
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -129,7 +129,7 @@
     "node-fetch": "^3.3.0",
     "tsdown": "catalog:",
     "tslib": "^2.8.1",
-    "typescript": "^5.9.2",
+    "typescript": "catalog:",
     "undici": "^8.0.0",
     "ws": "^8.0.0"
   },
--- a/packages/client/tsconfig.build.json
+++ b/packages/client/tsconfig.build.json
@@ -1,7 +1,5 @@
 {
   "extends": "../../tsconfig.build.json",
-  "compilerOptions": {
-    "declarationDir": "dist"
-  },
+  "compilerOptions": {},
   "include": ["src"]
 }
--- a/packages/next/package.json
+++ b/packages/next/package.json
@@ -140,7 +140,7 @@
     "react": "^19.1.0",
     "react-dom": "^19.1.0",
     "tsdown": "catalog:",
-    "typescript": "^5.9.2",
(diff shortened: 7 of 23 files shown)


packages/react-query/src/index.ts
```diff
@@ -7,4 +7,8 @@ export {
   type CreateTRPCReactBase,
 } from './createTRPCReact';
 export type { inferReactQueryProcedureOptions } from './utils/inferReactQueryProcedure';
+// Re-exported so consumers can name the inferred type of `createTRPCReact()`.
```
Nick-Lucas: strip all the newly added comments across this PR

packages/server/src/adapters/wsEncoder.ts
```diff
@@ -11,8 +11,13 @@
  * ```
  */
 export interface Encoder {
-  /** Encode data for transmission over the wire */
-  encode(data: unknown): string | Uint8Array;
+  /**
+   * Encode data for transmission over the wire.
```
Nick-Lucas: don't need this changed comment, leave as it was

packages/react-query/src/index.ts
Nick-Lucas: Done — stripped every comment I'd added across this PR: the `DecorateRouterRecord` re-export note here, the two `rootDir`/`allowImportingTsExtensions` notes in `tsconfig.json`, the catalog notes in `pnpm-workspace.yaml`, and the twoslash one in `shikiTwoslash.config.js`. The `@ts-expect-error` directives in the issue-4519 test are kept (they're load-bearing) but trimmed back to the original one-line wording.

packages/server/src/adapters/wsEncoder.ts
Nick-Lucas: Done — restored to `/** Encode data for transmission over the wire */`. Only the type on the next line changes (`Uint8Array` → `Uint8Array<ArrayBuffer>`), which is what TS 7's DOM lib requires for `WebSocket.send()`.

PR #7465, opened 2026-08-10
fix(docs): fix 19 dead links in llms.txt — use permalink, absolute URLs, and H1 title

okxint:
Fixes #7462.

Three issues in the `pluginLlmsTxt` plugin inside `www/docusaurus.config.ts` caused 19 of 89 links in `llms.txt` to 404.

---

### 1. Wrong path source (`docItem.id` vs `docItem.permalink`)

The plugin was constructing links from the file-path-based `docItem.id`:

```ts
// before
let link = `${versionSpec.routePath}/${docItem.id}`;
```

Several docs use custom `slug` frontmatter that changes their URL. For example:

| File | slug frontmatter | Generated (wrong) | Actual URL |
|---|---|---|---|
| `main/quickstart.mdx` | `/quickstart` | `/docs/main/quickstart` | `/docs/quickstart` |
| `further/faq.mdx` | `/faq` | `/docs/further/faq` | `/docs/faq` |
| `client/overview.md` | `/client` | `/docs/client/overview` | `/docs/client` |
| `client/nextjs/app-router/server-actions.mdx` | `/client/nextjs/server-actions` | `/docs/client/nextjs/app-router/server-actions` | `/docs/client/nextjs/server-actions` |

The fix: use `docItem.permalink` (which Docusaurus populates with the resolved route, respecting slug overrides), falling back to the constructed path only when unavailable.

### 2. Relative URLs

The generated links were relative (`/docs/quickstart`). The llms.txt spec and AI agent consumers expect absolute URLs so the file is self-contained. The fix prepends `siteConfig.url`.

### 3. Missing H1 title

The llms.txt spec requires the first line to be an H1. The plugin was writing the title as plain text. The fix adds a `#` prefix.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Documentation**
  * Improved generated documentation table-of-contents links for more reliable navigation.
  * Updated TOC files to include a top-level heading, improving compatibility with `llms.txt` readers.

PR #7576, opened 2026-09-12
build(deps): upgrade eslint to v10

Nick-Lucas:
Layer 5 of a stacked dependency upgrade. Stacked on #7575.

| package | from | to |
| --- | --- | --- |
| `eslint` | ^9.26.0 | **^10.10.0** |
| `typescript-eslint` | ^8.32 | ^8.70.0 |
| `eslint-plugin-unicorn` | 64.0.0 | ^74.0.0 |
| `eslint-plugin-react-hooks` | **6.0.0-rc.1** | ^7.1.1 |
| `@eslint/compat` | ^2.0.0 | ^2.1.1 |

`eslint-plugin-react` stays at 7.37.5 — there's no newer release. Note this also gets us off a release-candidate prerelease of `eslint-plugin-react-hooks`.

### Config migration
- **`reactHooks.configs.recommended` is eslintrc-shaped again in v7** and makes ESLint 10 fail outright (`A config object has a "plugins" key defined as an array of strings`). The flat config moved to `configs.flat.recommended`. Same fix applied to the two example configs that wire this up themselves.
- **`react-hooks/react-compiler` no longer exists.** v7 split it into ~14 granular rules (`purity`, `immutability`, `refs`, `set-state-in-effect`, `use-memo`, …) that `flat.recommended` enables. The rule entry and two stale `eslint-disable` comments referencing it are removed.
- **`unicorn/filename-case` gets `checkDirectories: false`.** unicorn 74 started checking directory names and flagged deliberate ones (`__tests__`, `aws-lambda`, `node-http`, `app-dir`). This keeps the rule scoped to filenames, as before.
- **`max-params` now counts parameters of function _types_.** The single hit is `setQueriesData`, which mirrors TanStack Query's own 4-parameter signature, so it takes a targeted disable.

### ⚠️ Five new React Compiler rules are switched off with a Todo
`react-hooks/refs`, `immutability`, `set-state-in-effect`, `use-memo` and `preserve-manual-memoization` flag **32 real findings** — `subscriptionOptions` mutating refs during render, several components calling `setState` synchronously inside an effect. That's a genuine refactor of shipped code and belongs in its own PR rather than a dependency bump, so they're disabled with a comment pointing at the work. **Everything else v7 added stays on.** Happy to split those fixes out if you'd like them tracked in an issue.

### Real bugs the new rules found
- **`sse.test.ts` had `await Promise.all([await a, await b])`** — the inner awaits ran the two operations sequentially, so the `error` listener was only registered *after* the server had already restarted. A latent flake, caught by `await-thenable`.
- **`getFetch` now returns `fetch` bound to its global.** It was already handing back an unbound method; the old `as FetchEsque` assertion just hid that from `unbound-method`. An unbound `fetch` throws `Illegal invocation` in some engines. The guards use `typeof` instead of the `isFunction` helper, which is no longer needed.
- Two optional-chain simplifications, plus dead imports/aliases left behind by `--fix`.

### Two `--fix` results were reverted
`--fix` removed two assertions that `no-unnecessary-type-assertion` considered redundant but which were **load-bearing for inference** — typechecking broke:
- `as ActionContext` in `create-action-hook.tsx` — without it `innerInput` became implicitly `any`. Now an explicit annotation, which is what was meant all along.
- `yield 1 as number` in `streaming.test.ts` (×3) — it deliberately widens the yield type; removing it narrowed the union to `1 | 2 | 3` and broke a downstream assertion. Restored with a note.

### Verification
`pnpm lint` is clean across all 42 workspaces. `pnpm build`, `typecheck-packages`, `typecheck-www`, `monotest-build` and `manypkg check` pass; `pnpm test` is 1130/1131 with only the pre-existing `httpSubscriptionLink.memory.test.ts` WeakRef/GC failure.

## Summary by CodeRabbit

- **Bug Fixes**
  - WebSocket typing updates now emit the initial typing-state event correctly when no previous state exists.
  - Updated WebSocket handling in example applications for more consistent connection upgrades.
  - Unsubscribing from HTTP links now reliably aborts the underlying request.

- **Chores**
  - Standardized linting configuration and dependency versions across the project and examples.
  - Improved type safety and lint compatibility without changing application behavior.
  - Consolidated shared development-tool versions for more consistent project maintenance.


examples/next-prisma-websockets-starter/eslint.config.mjs
```diff
@@ -16,7 +16,15 @@ export default tseslint.config(
     ],
     rules: {
       ...reactPlugin.configs['jsx-runtime'].rules,
-      ...hooksPlugin.configs.recommended.rules,
+      ...hooksPlugin.configs.flat.recommended.rules,
+      // eslint-plugin-react-hooks v7 split `react-compiler` into granular
```
Nick-Lucas: strip all newly added comments

packages/client/src/getFetch.ts
```diff
@@ -1,22 +1,23 @@
...
   customFetchImpl?: FetchEsque | NativeFetchEsque,
 ): FetchEsque {
   if (customFetchImpl) {
     return customFetchImpl as FetchEsque;
   }
 
-  if (typeof window !== 'undefined' && isFunction(window.fetch)) {
-    return window.fetch as FetchEsque;
+  // Bound to its global: `getFetch()` returns a standalone function, and an
+  // unbound `fetch` throws "Illegal invocation" in some engines.
```
Nick-Lucas: this file's changes look massively overcooked, can we revert it?

Nick-Lucas: Reverted — `getFetch.ts` is now byte-identical to `main` apart from two `eslint-disable` lines at the top.

You're right that it was overcooked. For context on why it needed *something*: typescript-eslint 8.70 flags four things in that file, and they pull in opposite directions.

- `unbound-method` fires on **both** `window.fetch` references — the one passed to `isFunction(...)` and the one returned.
- `no-unnecessary-type-assertion` fires on both `as FetchEsque` casts.

Removing the casts doesn't help (`unbound-method` still fires on the guard), and actually satisfying `unbound-method` means returning `window.fetch.bind(window)` — a behaviour change, which doesn't belong in a dependency bump. So the file keeps its current behaviour and the two rules are disabled for it, matching how `vitest.shared.ts` handles its own rule exceptions.

Worth noting separately: `getFetch()` really does hand back an unbound `fetch`, which throws `Illegal invocation` in some engines. The old `as FetchEsque` cast was hiding that from the linter rather than fixing it. Probably worth its own PR if you agree it's a real issue.

examples/next-prisma-websockets-starter/eslint.config.mjs
Nick-Lucas: Done — stripped across this PR: both example eslint configs, the `checkDirectories`/react-hooks notes in the root config, and the inline ones in `create-action-hook.tsx` and `streaming.test.ts`. The block disabling the five new React Compiler rules keeps just its one-line `// Todo:` marker, matching the existing Todo block above it.

Nick-Lucas:
Rebased the stack onto the current `main` (picking up #7464, #7465, #7477).

One fix needed here as a result: #7464 added

```ts
input === skipToken ? undefined : (input as any),
```

to `subscriptionOptions.ts`, and `no-unnecessary-type-assertion` flags that `as any` under typescript-eslint 8.70 — the ternary already narrows `input` to the non-`skipToken` branch, so the cast is redundant. Dropped it; typecheck and the tanstack-react-query suite (69/69) both pass.

Nothing else in the stack conflicted, and re-running `pnpm format-fix` after the rebase produced no changes, so the prettier layer (#7577) needed no reformat.

PR #7578, opened 2026-09-12
build(deps): upgrade vitest to v5

Nick-Lucas:
Layer 7 of a stacked dependency upgrade. Stacked on #7577.

| package | from | to |
| --- | --- | --- |
| `vitest`, `@vitest/coverage-istanbul`, `@vitest/ui` | ^4.0.18 | **^5.0.0** |
| `vite` | ^6.1.1 | **^8.3.0** |
| `jsdom` | ^29 | ^30.0.1 |
| `@vitejs/plugin-react` (2 examples) | ^4.3.4 | ^6.1.1 |

vitest 5 requires Node 22+ and Vite ≥6.4.

### Three things had to change

**1. The root config is split into `vitest.shared.ts` + `vitest.config.ts`.**
vitest 5 treats *any* config declaring `projects` as a container. Every package's `vitest.config.ts` re-exported the root config — so each package became a container looking for `./packages/*` relative to itself:

```
Error: No projects were found in "packages/client/vitest.config.ts".
```

`vitest.shared.ts` now holds the shared test options and is deliberately free of `projects`; `vitest.config.ts` is shared + `projects`.

**2. `packages/upgrade` needed an explicit JSX runtime.**
Its tsconfig sets `jsx: preserve` — correct for a codemod package, whose fixtures must keep JSX intact for jscodeshift to transform. **Vite 8 honours that setting where Vite 6 did not**, so fixture JSX reached the SSR transform uncompiled and all 7 `hooks` transform tests died with `Unexpected JSX expression`. Its vitest config now overrides the JSX runtime, since the tests need the fixtures actually executed.

**3. `transforms.test.ts` loads fixtures via `import.meta.glob`.**
It previously did a dynamic `import()` of an absolute path computed at runtime. Vite only transforms modules it knows about, and such a path isn't one. This also retires the `// idk why but Vite seems to do some shit when the string is in-lined to URL` workaround.

### Vite 8 config-loader warnings
Vite 8 warns about configs its native loader can't handle. Both causes are fixed rather than suppressed: the two vitest configs living in **CommonJS** packages (`packages/tests`, `examples/next-prisma-starter`) are renamed to `.mts`, and the shared config is imported with its real `.ts` extension — hence `allowImportingTsExtensions` in the typecheck tsconfig. The test run is now warning-free.

### Two knock-on fixes
- **`.ts-prunerc`** ignores `config\.[cm]?ts` rather than just `config.ts` — otherwise the renamed `.mts` configs get flagged as unused default exports and `lint-prune` fails autofix.ci.
- **`examples/minimal-react/client` and `examples/minimal-content-types/client`** move to `moduleResolution: Bundler`. Vite 8 ships its types only through `exports`, which node10 resolution can't read, so `tsc` could no longer find `vite` or `@vitejs/plugin-react` at all (`error TS2307`).

### Regenerated openapi fixtures
`nodeModulesJsDocRouter`'s committed fixtures change. That router **deliberately** derives its schema from types in `node_modules`, and rolldown's `buildStart` hook type changed with the Vite bump, so the generated OpenAPI document legitimately changes with it. Regenerated with `pnpm -C packages/openapi codegen`, which is what the CI check asks for. Worth knowing this fixture will shift on any future rolldown/vite bump.

### Verification
`pnpm test --coverage` (the CI invocation) is back to 1130/1131 — only the pre-existing `httpSubscriptionLink.memory.test.ts` WeakRef/GC failure. `pnpm build`, `typecheck-packages`, `typecheck-www`, `monotest-build`, `lint` (42/42) and `manypkg check` all pass.


packages/openapi/test/routers/nodeModulesJsDocRouter.openapi.json
```diff
@@ -1189,36 +1189,7 @@
...
-                    "type": "boolean"
-                  }
-                },
-                "required": [
-                  "handler"
-                ],
-                "additionalProperties": false
-              }
-            ]
+            "type": "object"
```
Nick-Lucas, line 1192: this is a regression, shouldn't have changed

packages/upgrade/vitest.config.ts
```diff
@@ -1,9 +1,24 @@
 import { mergeConfig } from 'vitest/config';
-import baseConfig from '../../vitest.config';
+import baseConfig from '../../vitest.shared.ts';
 
 export default mergeConfig(baseConfig, {
+  // This package's tsconfig sets `jsx: preserve` - the codemod fixtures have
```
Nick-Lucas: remove all added commends

packages/openapi/test/routers/nodeModulesJsDocRouter.openapi.json
Nick-Lucas: I dug into this one rather than just reverting it, and I think the fixture is doing its job — the underlying type genuinely changed in Vite 8.

`nodeModulesJsDocRouter` introspects `AliasOptions` from `vite`, which reaches `ResolverObject`. In Vite 6 that was:

```ts
buildStart?: ObjectHook<...>   // → `{...} | { handler; order?; sequential? }`
```

In Vite 8 (`vite/dist/node/index.d.ts`) it is:

```ts
interface ResolverObject {
  buildStart?: FunctionPluginHooks['buildStart'];
  resolveId: ResolverFunction;
}
```

No `ObjectHook` wrapper, so the `{ handler, order, sequential }` variant no longer exists upstream and the generated schema correctly drops it. The generator isn't losing information — it's tracking a type that moved.

That said, the real point stands: **this fixture will churn on every vite/rolldown bump**, which is exactly the kind of diff you don't want to review. If you'd rather it were stable, the fix is in the router rather than here — point `RootNodeModulesImportOutput` at a type that doesn't move (or vendor a small fixture type). Happy to do that in a follow-up if you want it; I've left the regenerated fixture in place for now because CI's "Verify committed openapi test fixtures" step fails if it's stale.

packages/upgrade/vitest.config.ts
Nick-Lucas: Done. Also removed the `server.deps.inline` block I'd added next to it — I'd verified it was dead config: the `oxc.jsx` override is what actually fixes the fixture transform, and the tests still pass (10/10) without it.

packages/tests/vitest.config.mts
```diff
@@ -0,0 +1 @@
+export { default } from '../../vitest.shared.ts';
```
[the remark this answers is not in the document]
Nick-Lucas: Already fixed in this PR — `.ts-prunerc`'s ignore pattern went from `config.ts` to `config\.[cm]?ts` so it covers the renamed `.mts` configs. That's what the autofix.ci failure in the tools output was about; `pnpm lint-prune` is clean now.

vitest.shared.ts
```diff
@@ -0,0 +1,77 @@
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
```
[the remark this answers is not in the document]
Nick-Lucas: Not applicable here — `unicorn/filename-case` only reaches files that are actually linted, and the per-package `lint` scripts run `eslint --cache src`, so root-level files like `vitest.shared.ts` / `vitest.config.ts` aren't in scope. `pnpm lint` passes 42/42 with the current name. It also matches the existing `vitest.config.ts` convention, so renaming it would make the pair inconsistent.

vitest.shared.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (key.includes('.json')) {
+      continue;
+    }
+    // trim first './'
+    const trimmed = key.slice(1);
+    aliases[`@trpc/${pkg}${trimmed}`] = join(
+      packagesDir,
+      pkg,
+      'src',
+      key.slice(1),
```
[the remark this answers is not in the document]
Nick-Lucas: This one doesn't hold — `path.join` doesn't treat a leading `/` in a later segment as absolute; that's `path.resolve`. `join('pkgs','client','src','/links')` returns `'pkgs/client/src/links'`, so the alias paths are correct.

Either way this is pre-existing code — the whole alias block moved verbatim from `vitest.config.ts` into `vitest.shared.ts` in this PR, and the resulting aliases are unchanged (the full test suite resolves `@trpc/*` fine).

vitest.shared.ts
```diff
@@ -0,0 +1,77 @@
...
+const packagesDir = path.resolve(__dirname, 'packages');
+
+const dirs = readdirSync(packagesDir)
+  .filter((it) => it !== 'tests' && !it.startsWith('.'))
+  .filter((it) => existsSync(join(packagesDir, it, 'package.json')));
+
+for (const pkg of dirs.sort()) {
+  const pkgJson = join(packagesDir, pkg, 'package.json');
+
+  const json = JSON.parse(readFileSync(pkgJson, 'utf-8').toString());
```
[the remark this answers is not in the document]
Nick-Lucas: True, but pre-existing — this block moved verbatim from `vitest.config.ts` into `vitest.shared.ts`. I've kept the move a pure move so the diff stays reviewable; happy to drop the redundant `.toString()` if you'd like it tidied here rather than separately.

Nick-Lucas:
Fixed the intermittent `unit and integration tests` failure on this PR.

All 1143 tests passed but vitest still exited 1, on an unhandled error:

```
Vitest caught 1 unhandled error during the test run.
TypeError: Right-hand side of 'instanceof' is not an object
 ❯ getActiveElementDeep react-dom-client.development.js
 ❯ commitBeforeMutationEffects / commitRoot
 ❯ listOnTimeout node:internal/timers
This error originated in "packages/upgrade/test/transforms.test.ts"
```

`transforms.test.ts` renders every fixture component inside a single `it`, so nothing unmounts until the test ends. React schedules a commit on a timer; if that fires after the jsdom environment has been torn down, `win.HTMLIFrameElement` is gone and `instanceof` throws. vitest 5 surfaces unhandled errors as a non-zero exit where 4 did not — hence green tests, red job, and why it only showed up on some runs.

Now each `spec.run(...)` is followed by `cleanup()` in a `finally`, so trees unmount as we go. Ran the upgrade suite 3× locally with no `Errors` line.

packages/openapi/test/routers/nodeModulesJsDocRouter.openapi.json
Nick-Lucas: What's happened here is we've found a bug. this key has become a function, and functions are being detected as objects in the types. That is a bug in the OpenAPI generator.

Add a stack layer to the top of this stack which updates the OpenAPI codegen to filter out function types entirely, they won't be serialised so should just be dropped

Nick-Lucas: Good call — that's exactly it, and it's now #7588 at the top of the stack.

`isObjectType()` was matching functions because they carry `ts.TypeFlags.Object`. The optional case was sneakier: `buildStart?: fn` is `fn | undefined`, a union, so a plain `isFunctionType` check missed it and it came out as `{}` rather than being dropped. So there are two helpers — `isFunctionType` for call/construct signatures, and `isUnserialisableType` which extends that through unions.

Functions are now dropped from properties (including `required`), from union members, and at the top-level dispatch. On this fixture that sheds ~880 lines: `ResolverObject` has only `buildStart` and `resolveId`, both callable, so it correctly generates as a bare `{"type": "object"}`.

Added `withFunctions` / `onlyFunctions` to `edgeCaseRouter` covering a method, an optional method, a constructor signature, a serialisable sibling, and an all-callable object — and confirmed both tests fail with the generator change reverted, so they're actually guarding it.

Nick-Lucas:
The `next-prisma-websockets-starter` Railway failure here is a stale build cache on Railway's side, not a problem with this commit.

The error reports:

```
- @vitejs/plugin-react (lockfile: ^6.1.1, manifest: ^4.3.4)
- vite               (lockfile: ^8.3.0, manifest: ^6.1.1)
```

Those "manifest" values are **#7577's**, not this PR's. On this commit `examples/minimal-content-types/client/package.json` says `^6.1.1` / `^8.3.0` — matching the lockfile exactly. Railway built #7577's package.json against #7578's lockfile; it deployed #7577 to the same service immediately before.

Verified three ways:
- A real `pnpm install --frozen-lockfile` on this exact commit (`8d3e5fc04`) succeeds.
- `git cat-file` on the commit shows the manifest already at `^8.3.0` / `^6.1.1`.
- **#7579 deploys the identical files and lockfile and Railway succeeds on both services.**

So a redeploy with a cold cache should clear it. `main` isn't branch-protected, so it isn't blocking the merge either way.

PR #7580, opened 2026-09-12
build(deps): patch security advisories in packages/*

Nick-Lucas:
Layer 9 of a stacked dependency upgrade. Stacked on #7579.

Clears the GHSA advisories `pnpm audit` reports against the **published packages** and their test dependencies.

`@hey-api/openapi-ts` needs a bump for its own advisory too, but it regenerates ~120 committed fixtures — that's split into #7586 so it doesn't bury these five one-line changes.

| package | from | to | advisory needs |
| --- | --- | --- | --- |
| `undici` | ^8.0.0 | ^8.10.2 | >=8.9.0 |
| `ws` | ^8.0.0 | ^8.21.3 | >=8.21.0 |
| `devalue` | ^5.0.0 | ^5.9.2 | >=5.8.1 |
| `valibot` | 1.3.1 | ^1.5.0 | >=1.4.2 |
| `fastify` | ^5.0.0 | ^5.12.4 | >=5.8.5 |

### Two knock-on changes

**1. The fastify version had to be unified across the workspace.**
`@fastify/websocket` → ^11.3.0 and `fastify-plugin` → ^6.0.0 (which 11.3 requires). More subtly, `www` and `examples/fastify-server` also move to `^5.12.4` — they were lockfile-pinned to 5.8.3, and **two fastify copies break `@fastify/websocket`'s `declare module 'fastify'` augmentation**: it binds to the other copy, so `{ websocket: true }` stopped type-checking in the adapter:

```
error TS2353: Object literal may only specify known properties,
and 'websocket' does not exist in type 'RouteShorthandOptions<…>'
```

Both of those workspaces needed the fastify bump for their own advisories anyway.

**2. `standalone.http2.test.ts` uses `undici.fetch` for its 404 check** instead of the global `fetch`. The server speaks h2 only, and only the undici instance configured at the top of that file has an `allowH2` dispatcher. It worked before because `setGlobalDispatcher` leaked into Node's bundled undici through a shared global symbol; undici 8.10 no longer does, so on CI the request negotiated http/1.1 and the handshake failed with `ERR_SSL_TLSV1_ALERT_NO_APPLICATION_PROTOCOL`.

### Verification
`pnpm build`, `typecheck-packages`, `typecheck-www`, `monotest-build`, `lint` (42/42) and `manypkg check` all pass. `pnpm test` is 1130/1131 — only the pre-existing `httpSubscriptionLink.memory.test.ts` WeakRef/GC failure. The openapi suite is 176/176.

## Summary by CodeRabbit

* **Chores**
  * Updated Fastify, WebSocket, Valibot, and related development dependencies.
  * Consolidated shared dependency versions across the workspace for more consistent package management.
  * Added centralized version entries for commonly used dependencies.

* **Bug Fixes**
  * Updated standalone server test requests to use the configured HTTP client, improving consistency with custom request dispatching.


packages/openapi/test/routers/amazonIonRouter-heyapi/core/auth.gen.ts
```diff
@@ -9,6 +9,13 @@ export interface Auth {
    * @default 'header'
    */
   in?: 'header' | 'query' | 'cookie';
+  /**
+   * A unique identifier for the security scheme.
+   *
+   * Defined only when there are multiple security schemes whose `Auth`
+   * shape would otherwise be identical.
+   */
+  key?: string;
```
Nick-Lucas: I think we need to split heyapi bumps into its own stack layer, this is a lot of changes related to that

pnpm-workspace.yaml
```diff
@@ -73,6 +73,15 @@ catalog:
   vite: ^8.3.0
   jsdom: ^30.0.1
   turbo: ^2.10.12
+  # Bumped to clear GHSA advisories - see the audit layer of this stack.
```
Nick-Lucas: strip all added comments

Nick-Lucas:
Fixed the `E2E-tests (openapi-codegen)` failure.

The symptom was a type error that names the same type twice:

```
Property 'path' is missing in type 'ArrayResolverContext$2'
but required in type 'ArrayResolverContext$2'
```

Two copies of `@hey-api/openapi-ts`: this PR moved `packages/openapi` to 0.99, but `examples/openapi-codegen` was still on ^0.94.5 until the next PR in the stack — and that example consumes the plugin types `@trpc/openapi` produces, so the split broke it. It passed on #7581 precisely because that's where the example caught up.

Shared dependencies now move in this commit rather than being split across two: `@hey-api/openapi-ts`, `ws` and `valibot` join `fastify` (which already needed it here for the `declare module 'fastify'` augmentation). #7581 keeps only the www/examples-specific bumps — next, next-auth, postcss, nuxt, wrangler, esbuild.

Verified the example's full e2e sequence locally — `test-dev`, `build`, `typecheck`, `test-start` all pass.

packages/openapi/test/routers/amazonIonRouter-heyapi/core/auth.gen.ts
Nick-Lucas: Agreed — split into #7586, which now sits between this PR and #7581.

This PR is down from 146 changed files to 13: just the five version bumps (undici, ws, devalue, valibot, fastify), the `@fastify/websocket`/`fastify-plugin` bumps that fastify 5.12 needs, and the http2 test fix. #7586 carries the heyapi bump and all ~120 regenerated fixtures.

One constraint the split has to respect: `packages/openapi` and `examples/openapi-codegen` have to move in the *same* commit. The example consumes the plugin types `@trpc/openapi` produces, so a version split between them puts two copies of the same type in one program — that's what caused the `E2E-tests (openapi-codegen)` failure earlier in this stack.

pnpm-workspace.yaml
Nick-Lucas: Already done — `pnpm-workspace.yaml` has no comments at all now, in this layer or any other. That comment was against the earlier push; the strip went out with the same batch as your other comment-stripping notes.

PR #7586, opened 2026-09-12
build(deps): upgrade @hey-api/openapi-ts to 0.99.0

Nick-Lucas:
Stacked on #7580. Split out of that PR at review request — the fixture regeneration was drowning five one-line version bumps.

`@hey-api/openapi-ts` 0.94.5 → **0.99.0**, clearing its GHSA advisory (needs >=0.97.3).

### `packages/openapi` and `examples/openapi-codegen` move together
The example consumes the plugin types `@trpc/openapi` produces, so splitting them across two layers puts two copies of the same type in one program:

```
Property 'path' is missing in type 'ArrayResolverContext$2'
but required in type 'ArrayResolverContext$2'
```

### The ~120-file diff is all regenerated output
Everything under `packages/openapi/test/routers/*-heyapi/` and `examples/openapi-codegen/src/client/generated/` comes from `pnpm -C packages/openapi codegen`, which is what CI's "Verify committed openapi test fixtures" step checks. Nothing there is hand-edited.

### One behaviour change in a test
0.99 no longer throws `Deeply-nested arrays/objects aren't supported` for a superjson-shaped query input. The default query serializer now emits a query string the router can't parse, so the call returns an **error response** instead of rejecting. I probed the actual 0.99 behaviour rather than assuming, so the assertion still makes the same point — inputs need serialising properly — rather than being weakened into a no-op.

### Verification
`pnpm build`, `typecheck-packages`, `lint` (42/42) pass; the openapi suite is 176/176; and `examples/openapi-codegen`'s full e2e sequence (`test-dev`, `build`, `typecheck`, `test-start`) passes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved request error handling across request creation, network access, and response processing.
  - Nested unserialized query errors now return structured error responses instead of throwing.
  - Server-sent event parsing more consistently handles different line-ending formats.

- **New Features**
  - Generated clients now expose metadata and server-sent event result types.
  - Authentication configurations can distinguish otherwise identical security schemes.

- **Improvements**
  - Generated SDK methods provide more precise request and response typing.
  - Request options and error results better reflect unavailable request or response details.


packages/openapi/test/generate.test.ts
```diff
@@ -960,15 +960,13 @@ describe('generateOpenAPIDocument', () => {
...
-      }).rejects.toThrowErrorMatchingInlineSnapshot(
-        `[Error: Deeply-nested arrays/objects aren’t supported. Provide your own \`querySerializer()\` to handle these.]`,
-      );
+      const unserialized = await sdk.getEvent({
+        query: {
+          input,
+        },
+      });
+      expect(unserialized.data).toBeUndefined();
+      expect(unserialized.error).toBeDefined();
```
Nick-Lucas: we should still check the specific snapshot of the error

Nick-Lucas: Done — and you were right to push on this, because my earlier explanation was wrong.

I'd said 0.99 "emits a query string the router cannot parse". It doesn't. **The same error is still produced**, it's just returned on the result instead of thrown. I dumped the actual object to check rather than guessing again:

```
ctor: Error
inspect: Error: Deeply-nested arrays/objects aren’t supported.
         Provide your own `querySerializer()` to handle these.
    at serializePrimitiveParam (…/core/pathSerializer.gen.ts:119:11)
```

So the assertion keeps the exact snapshot, just against `result.error` rather than a rejection:

```ts
expect(unserialized.data).toBeUndefined();
expect(unserialized.error).toMatchInlineSnapshot(
  `[Error: Deeply-nested arrays/objects aren’t supported. Provide your own \`querySerializer()\` to handle these.]`,
);
```

Commit message corrected too. 176/176 passing.

PR #7594, opened 2026-09-13
fix(react-query): honor skipToken in utils queryOptions and infiniteQueryOptions

Dev-next-gen:
## 🎯 Changes

Bug fix. While looking at #7464, which fixed `subscriptionOptions(skipToken)` in `@trpc/tanstack-react-query` because `getQueryKeyInternal` strips `skipToken` from the query key, I noticed the same thing happens in `@trpc/react-query` for the utils `queryOptions` and `infiniteQueryOptions`.

`createUtilityFunctions` decides whether to disable the query with `queryKey[1]?.input === skipToken`, but the key it receives comes from `getQueryKeyInternal`, which never keeps `skipToken` as the input. That check can't match, so `utils.post.byId.queryOptions(skipToken)` (and `createTRPCQueryUtils(...).post.byId.queryOptions(skipToken)`) returns a real `queryFn`. On main, calling that `queryFn` sends the request with an `undefined` input and the server rejects it with a zod validation error, instead of the query staying disabled. The existing `disabling query with skipToken` test only checks for the `pending` status on first render, so it doesn't catch it.

The proxy is the only place that still has the raw input, so I apply the `skipToken` there after building the options, and leave the internal `TRPCQueryUtils` signatures alone since they're exported from `@trpc/react-query/shared`.

The new test in `createQueryUtils.test.ts` fails on main (`expected [AsyncFunction queryFn] to be Symbol()`) and passes with the change. The full `@trpc/react-query` suite (48 files) passes, and `typecheck` and eslint are clean for the package.

## ✅ Checklist

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved skipped query handling so standard and infinite query options correctly preserve skip behavior when provided with a skipped query token.
  * Added coverage to verify skipped queries expose the expected disabled query function.


Dev-next-gen:
@coderabbitai your nitpick on createQueryUtils.test.ts was right: the `client` destructuring was used only once, and a93965a passes `factory.client` straight to `createTRPCQueryUtils`. Your later review on that commit had nothing left to flag.

I also merged main into the branch (0b6a246, no force-push) to bring in #7604 and v11.19.0. Against main, the diff is still only `utilsProxy.ts` and `createQueryUtils.test.ts`, and all 48 `@trpc/react-query` test files pass on the merge commit.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

